PR #8, opened 2020-07-12
feat: redesigning top-n component

anandtiwary:


projects/observability/src/shared/components/top-n/top-n-chart.component.test.ts
```diff
@@ -1,12 +1,18 @@
...
-describe('TopN Chart Component', () => {
+describe('Top N Chart Component', () => {
   let spectator: Spectator<TopNChartComponent>;
 
   const createHost = createHostFactory({
     component: TopNChartComponent,
-    shallow: true
+    shallow: true,
+    imports: [CommonModule, TooltipModule, FormattingModule],
+    providers: [mockProvider(NavigationService), ...getMockFlexLayoutProviders()]
```
skjindal93, line 15: I think we can get rid of `CommonModule`, `TooltipModule` imports and `getMockFlexLayoutProviders` provider here

anandtiwary: I doesn't work without it. May be i will have to mock the directive. But I am not sure I want to do a shallow test here. Testing the display value post formatting seems an important test case for this component. So I think i must test it against the actual formatting pipe. I am not testing the internals of display formatter, only using it as a black box because the correct output is relevant to my top-n-view. @aaron-steinfeld What are your thoughts?

projects/observability/src/shared/components/top-n/top-n-chart.component.ts
```diff
@@ -1,18 +1,31 @@
...
-import { sortBy } from 'lodash';
-import { HistogramBarData } from '../histogram/histogram-chart.component';
+import { maxBy } from 'lodash';
 
 @Component({
   selector: 'ht-top-n-chart',
+  styleUrls: ['./top-n-chart.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <ht-histogram-chart
```
aaron-steinfeld, line 10: Do we want the histogram and top-n diverging, or should this change be to the underlying histogram component?

projects/observability/src/shared/components/top-n/top-n-chart.component.test.ts
aaron-steinfeld: Generally, the job of this test ends once it goes into the next component/pipe - the formatting it produces should be tested by that pipe. There is a place for deeper tests especially for complex components working together, but I don't think a display component like this is one of them.

anandtiwary: True, but this is a very basic component. I don't think I should skip checking whether the correct text (value with percentage) is being displayed or not.

anandtiwary: Removing the imports and mock providers breaks the test. Should i ng-mock the tooltip and display pipe? Or is there a better way? I want this change in soon.

skjindal93: @anandtiwary It passes for me. That's why I didn't ask you to remove `FormattingModule` import and get rid of everything else

```
    component: TopNChartComponent,
    shallow: true,
    imports: [FormattingModule],
});
```

anandtiwary:
Screen Shot 2020-07-12 at 9 49 32 PM

This repo needs to show latest commit's build status here. Github bug or some setting that we need to enable?

projects/observability/src/shared/components/top-n/top-n-chart.component.ts
anandtiwary: Yes. IMO it should diverge since now top-n is significantly different from the histogram/bar chart. Histogram may still have usage is some cases. So let's roll with this and probably we can come back to this later after few months.

aaron-steinfeld:
> Screen Shot 2020-07-12 at 9 49 32 PM
> 
> This repo needs to show latest commit's build status here. Github bug or some setting that we need to enable?

Sometimes GH caches things on browser back I've noticed, generally a browser refresh fixes it.

projects/observability/src/shared/components/top-n/top-n-chart.component.test.ts
aaron-steinfeld: pipes have to be mocked or provided, but a shallow test means you don't have to mock components or directives unless you're interacting with them in the test. @skjindal93 's suggestion seems right

PR #11, opened 2020-07-13
feat: adding link support to top-n model

anandtiwary:


projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget.model.ts
```diff
@@ -10,10 +18,14 @@ import { TopNDataSourceModel, TopNWidgetDataFetcher } from './data/top-n-data-so
...
-    type: STRING_PROPERTY.type
+    // tslint:disable-next-line: no-object-literal-type-assertion
+    type: {
+      key: ModelPropertyType.TYPE,
+      defaultModelClass: WidgetHeaderModel
+    } as ModelModelPropertyTypeInstance,
+    key: 'header'
   })
-  public title: string = '';
+  public header!: WidgetHeaderModel;
```
skjindal93, line 28: I don't think the header is a required field. Can we make it optional? and have `required:false` in the ModelProperty?

anandtiwary: Good catch. It should be optional. The renderer does make that assumption but this should be updated too.

PR #14, opened 2020-07-13
feat: add service name col

sanjay-nagaraj:


projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -227,10 +239,22 @@ export class ExplorerDashboardBuilder {
...
+              attribute: 'serviceName'
+            },
+            'click-handler': {
+              type: 'api-trace-navigation-handler'
+            }
+          },
           {
             type: 'table-widget-column',
             title: 'Name',
-            width: '30%',
```
aaron-steinfeld, line 233: For spans, I think it makes more sense to take the width from status - way over provisioned right now since it doesn't have a special renderer. Names can be fairly long.

PR #22, opened 2020-07-14
feat: add support for multiple group by keys

aaron-steinfeld:


projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -65,16 +64,19 @@ export class ExploreCartesianDataSourceModel extends GraphQlDataSourceModel<Metr
         (selection): selection is RequireBy<ExploreSpecification, 'aggregation'> => selection.aggregation !== undefined
       );
 
-    if (request.interval && !request.groupBy) {
+    const groupByKeys = request.groupBy?.keys || [];
```
anandtiwary, line 67: `??` ?

aaron-steinfeld: Old habits - although they're identical in this case, will update.

PR #30, opened 2020-07-17
chore: update to ng10

aaron-steinfeld:


package.json
```diff
@@ -7,7 +7,7 @@
   },
   "license": "LicenseRef-LICENSE",
   "scripts": {
-    "postinstall": "ngcc --properties es2015 browser module main",
+    "postinstall": "ngcc --properties browser main --tsconfig './tsconfig.spec.json'",
```
anandtiwary, line 10: Sorry i think i missed to ask this in the other PR. Why is this using spec tsconfig?

aaron-steinfeld: So I dug into a bit why this is needed, since the CLI already does this as part of its build (and only once). The reason is that our tests use jest so don't go through the same CLI preprocessing - there's an open ticket on upgrading the jest-angular support for that. Since this is only used by jest tests, that indicated that we didn't need to run the module transform here, only the main umd one (cli will still run module transform if needed). The second part, the tsconfig is similar reasoning, however additionally the change to the solution style tsconfig is flagged by ngcc as invalid (and that's the default), so we need to explicitly provide a tsconfig here.

anandtiwary: okay

PR #34, opened 2020-07-20
fix: use object rather than label in top n selection

aaron-steinfeld:


projects/observability/src/shared/components/top-n/top-n-chart.component.ts
```diff
@@ -63,20 +63,20 @@ export class TopNChartComponent implements OnChanges {
...
 }
 export interface TopNData {
   label: string;
   value: number;
 }
 
-interface TopNItemOptions extends TopNData {
+interface TopNItemOption<T extends TopNData> extends TopNData {
   width: string;
+  original: T;
```
anandtiwary, line 81: What's the reason for this change?

aaron-steinfeld: The previous API had no way of disambiguating two objects with identical labels.

anandtiwary: I see. Thanks!

PR #37, opened 2020-07-21
feat(api detail breadcrumbs): Make endpoint list crumb clickable

arjunlalb:
Make endpoint list crumb clickable

### DIFF
--- a/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
+++ b/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
@@ -6,11 +6,11 @@ import { GraphQlTimeRange, SpecificationBuilder } from '@hypertrace/distributed-
 import { GraphQlRequestService } from '@hypertrace/graphql-client';
 import { Observable } from 'rxjs';
 import { map, switchMap, take } from 'rxjs/operators';
-import { ENTITY_METADATA, EntityMetadata, EntityMetadataMap } from '../../../shared/constants/entity-metadata';
+import { EntityMetadata, EntityMetadataMap, ENTITY_METADATA } from '../../../shared/constants/entity-metadata';
 import { Entity, ObservabilityEntityType } from '../../../shared/graphql/model/schema/entity';
 import {
-  ENTITY_GQL_REQUEST,
-  EntityGraphQlQueryHandlerService
+  EntityGraphQlQueryHandlerService,
+  ENTITY_GQL_REQUEST
 } from '../../../shared/graphql/request/handlers/entities/query/entity/entity-graphql-query-handler.service';
 
 @Injectable({ providedIn: 'root' })
--- a/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
+++ b/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
@@ -58,7 +58,7 @@ export class ApiDetailBreadcrumbResolver implements Resolve<Observable<Breadcrum
           {
             label: 'Endpoints',
             icon: this.apiEntityMetadata?.icon,
-            url: parentEntityMetadata?.endpointsListPath?.(api.parentId)
+            url: parentEntityMetadata?.apisListPath?.(api.parentId)
           }
         ]
       : [];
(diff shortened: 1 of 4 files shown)


projects/observability/src/shared/constants/entity-metadata.ts
```diff
@@ -5,6 +5,7 @@ export interface EntityMetadata {
   icon: string;
   detailPath(id: string, sourceRoute?: string): string[];
   listPath?: string[];
+  endpointsListPath?(id: string): string[];
```
skjindal93, line 8: Do we want to name this as "endpoints"? That's just a name for APIs right now. It might change later

May be a more generic name?

arjunlalb: this is the list path for endpoints. Generic name would be a misfit here. If we rename endpoints to something later, we'll have to refactor these as well.

skjindal93: May be `apisListPath` (_if we don't want to keep it generic_) or `childListPath` (_if it makes more sense and is generic_)

arjunlalb: entityMetadata map could be extended to be used with other pages as well. So, childListPath would become ambiguous, if we want to add another child list path here in the future. I'm good with apisListPath, but isn't that the same thing as apisListPath?

We should lean towards using one name. I was assuming we are calling it endpoints hereafter everywhere.

skjindal93: We are actually calling APIs as "APIs" everywhere. All the component classes, EntityTypes etc. Names will keep changing IMO

Probably want to be consistent everywhere

arjunlalb: In that case, I prefer `apisListPath` as well. Updating.

arjunlalb: @skjindal93 Done.

projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
```diff
@@ -51,13 +47,18 @@ export class ApiDetailBreadcrumbResolver implements Resolve<Observable<Breadcrum
...
+  private getParentBreadcrumbs(api: ApiBreadcrumbDetails, parentEntityMetadata?: EntityMetadata): Breadcrumb[] {
     return parentEntityMetadata !== undefined
       ? [
           {
             label: api.parentName,
             icon: parentEntityMetadata?.icon,
             url: parentEntityMetadata?.detailPath(api.parentId)
+          },
+          {
+            label: 'Endpoints',
```
aaron-steinfeld, line 59: Should we maintain the unclickable endpoints breadcrumb when there is no parent? That was removed here, unsure if that was intentional.

src/app/entity-metadata.ts
```diff
@@ -17,6 +17,7 @@ export const entityMetadata: EntityMetadataMap = new Map([
       icon: ObservabilityIconType.Service,
       detailPath: (id: string) => ['services', 'service', id],
       listPath: ['services'],
+      apisListPath: (id: string) => ['services', 'service', id, 'endpoints'],
```
aaron-steinfeld, line 20: Hmm - rather than introducing this which is only relevant to a small subset of entities, could we concatenate the detail path of one entity with the list path of the api entity? It would make the assumption that the list path is always at the same subroute, but I think that's the behavior we want anyway.

projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
arjunlalb: That was intentional. It did not make sense to have a breadcrumb of the format `Services / Endpoints / <Endpoint Name>`. Without the parent name, the 'endpoints' section is a non existent page. @aaron-steinfeld

src/app/entity-metadata.ts
arjunlalb: I did it this way so that we'd have flexibility on where the list path is. Moreover, it would keep everything in _configs_ without having to add a subroute information to the resolver code.

arjunlalb:
@aaron-steinfeld codecov seems to be failing. no reason for a 10% drop in coverage. override merge please :)

aaron-steinfeld:
> @aaron-steinfeld codecov seems to be failing. no reason for a 10% drop in coverage. override merge please :)

The first run it only uploaded from one of the test containers. Reran it, and both uploaded and now it's queued for processing. It looks like codecov is catching up from their backlog earlier, so it should come through shortly. Let me know if it's time sensitive.

arjunlalb:
@aaron-steinfeld @skjindal93 merge please :)

PR #38, opened 2020-07-21
feat(card-list-widget): click handlers on card

skjindal93:


projects/observability/src/shared/dashboard/widgets/card-list/card-list-widget-renderer.component.ts
```diff
@@ -21,15 +21,20 @@ import { CardListWidgetModel } from './card-list-widget-model';
         <ht-card-list>
           <ht-card-container *ngFor="let card of cards" [ngSwitch]="this.model.cardType">
             <div class="summary-card" *ngSwitchCase="'${CardType.Summary}'">
-              <htc-summary-card [name]="card.name" [color]="card.color" [summaries]="card.summaries"></htc-summary-card>
+              <htc-summary-card
+                [name]="card.name"
+                [color]="card.color"
+                [summaries]="card.summaries"
+                (click)="this.model.clickHandler?.execute(card.original)"
```
aaron-steinfeld, line 28: Is original supposed to be populated by the data source? That doesn't make a whole lot of sense. A pattern like `original` holding the provided reference is meant to be internal to a class so it can give back what it took in. It should never traverse class boundaries - give the model back what it gave you, the card.

skjindal93: `original` is a bad name then here. The motive was to pass the data to the card, so that it can be executed by the click handlers

It can be Entity, or API Trace or anything else

How about renaming `original` to `context`?

projects/observability/src/shared/dashboard/widgets/card-list/card-list-widget-renderer.component.ts
```diff
@@ -21,15 +21,23 @@ import { CardListWidgetModel } from './card-list-widget-model';
...
+              ></htc-summary-card>
             </div>
           </ht-card-container>
         </ht-card-list>
       </htc-titled-content>
     </div>
   `
 })
-export class CardListWidgetRendererComponent<T extends Card[]> extends WidgetRenderer<CardListWidgetModel<T>, T> {
+export class CardListWidgetRendererComponent<P, T extends Card<P>[]> extends WidgetRenderer<
```
anandtiwary, line 37: Sorry. This typing is confusing to me.

anandtiwary: Looks like they are only useful in tests. Do we need to include a new generics P? What if in the card we use `unknown` instead of generic T ?

skjindal93: I can do that

PR #40, opened 2020-07-21
Top n styling

anandtiwary:


projects/observability/src/shared/components/top-n/top-n-chart.component.scss
```diff
@@ -29,6 +28,7 @@
     border-radius: 3px;
     height: 6px;
     background-color: $gray-2;
+    cursor: pointer;
```
aaron-steinfeld, line 31: Progress bar isn't clickable, why is it a pointer?

skjindal93: I thought it's going to show some tooltip on it, since we removed the %age

aaron-steinfeld: Don't think so, but tooltip shouldn't be pointer regardless since it's still not clickable

anandtiwary: Yeah, I missed to remove this. I had added it for a reason. nvm. Will update this

PR #41, opened 2020-07-21
fix: updating the back icon

anandtiwary:


projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts
```diff
@@ -17,7 +17,7 @@ import { ApiTraceDetails, ApiTraceDetailService } from './api-trace-detail.servi
         <div class="back">
           <htc-icon
             (click)="this.onClickBack()"
-            icon="${IconType.KeyboardBackspace}"
+            icon="${IconType.ArrowLeft}"
```
skjindal93, line 20: Are we planning to use back as ArrowLeft everywhere? Then, we should get rid of `KeyboardBackspace` icon, once we have replaced the usages everywhere

PR #43, opened 2020-07-22
feat(donut-widget): color palette

skjindal93:


projects/observability/src/shared/dashboard/widgets/donut/donut-widget.model.ts
```diff
@@ -65,13 +69,25 @@ export class DonutWidgetModel {
...
+  public colorPalette: string[] = [];
+
   @ModelInject(MODEL_API)
   private readonly api!: ModelApi;
 
+  @ModelInject(ColorService)
+  private readonly colorService!: ColorService;
+
   public getData(): Observable<DonutResults> {
+    this.colorService.registerColorPalette(DonutWidgetModel.DONUT_COLORS, this.colorPalette);
```
aaron-steinfeld, line 86: Don't need to register this and it doesn't really make sense since it's local to this instance of this mode.

projects/observability/src/shared/dashboard/widgets/donut/donut-widget.model.ts
```diff
@@ -82,4 +98,13 @@ export class DonutWidgetModel {
       }))
     );
   }
+
+  private buildSeriesWithColors(series: DonutSeries[]): DonutSeries[] {
+    const colors = this.colorService.getColorPalette(DonutWidgetModel.DONUT_COLORS).forNColors(series.length);
```
aaron-steinfeld, line 103: `this.colorService.getColorPalette(this.colorPalette)` - that assumes we change the default for this to be undefined rather than an empty array, which is an invalid palette.

projects/observability/src/shared/dashboard/widgets/donut/donut-widget.model.ts
```diff
@@ -65,13 +69,25 @@ export class DonutWidgetModel {
   })
   public header?: WidgetHeaderModel;
 
+  @ModelProperty({
+    key: 'color-palette',
+    type: ARRAY_PROPERTY.type,
+    required: false
+  })
+  public colorPalette: string[] = [];
```
aaron-steinfeld, line 77: A color palette should also accept a string as key - the color service reflects this scenario. Since we don't have unions in our crude type system, can just leave that as `UNKNOWN_PROPERTY.type` - at some point, we'll create a color property.

PR #46, opened 2020-07-22
feat: put selected explorer tab in url ENG-4538

aaron-steinfeld:


projects/observability/src/pages/explorer/explorer.component.test.ts
```diff
@@ -110,14 +118,15 @@ describe('Explorer component', () => {
     discardPeriodicTasks(); // Some of the newly instantiated components also uses async, need to wait for them to settle
   };
 
-  beforeEach(fakeAsync(() => {
-    spectator = createComponent();
+  const init = (...params: Parameters<typeof createComponent>) => {
```
anandtiwary, line 121: Why changing this to init function?

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -79,25 +86,34 @@ import {
...
+    @Inject(EXPLORER_DASHBOARD_BUILDER_FACTORY) explorerDasboardBuilderFactory: ExplorerDashboardBuilderFactory,
+    activatedRoute: ActivatedRoute
   ) {
     this.explorerDashboardBuilder = explorerDasboardBuilderFactory.build();
     this.resultsDashboard$ = this.explorerDashboardBuilder.resultsDashboard$;
     this.vizDashboard$ = this.explorerDashboardBuilder.visualizationDashboard$;
+    this.contextLabel$ = activatedRoute.queryParamMap.pipe(
+      map(paramMap => paramMap.get(ExplorerComponent.SCOPE_QUERY_PARAM)),
+      map(queryParam => this.queryParamToContextLabel(queryParam)),
+      tap(label => this.onContextUpdated(label))
```
anandtiwary, line 115: Keeping this here wouldn't update URL again? Since the change is coming from URL, why call this function here?

projects/observability/src/pages/explorer/explorer.component.test.ts
aaron-steinfeld: This is generally why I prefer to keep the component creation outside of beforeEach and in each test instead - I had a use case in one of the tests where I needed to override the injects to test its initialization behavior (activated route in this case, I believe)

projects/observability/src/pages/explorer/explorer.component.ts
aaron-steinfeld: Technically it would, but it's idempotent. The first time through, we do want it to update the URL to apply the default, it's just on any later url changes where it becomes unnecessary. I fixed it in one of the iterations, but felt like the extra complexity to avoid the dupe calls wasn't worth it.

PR #48, opened 2020-07-23
fix: checking if touch event exist on the browser

anandtiwary:
Fixes zoom drag on firefox browser

### DIFF
--- a/projects/observability/src/shared/components/topology/d3/interactions/zoom/topology-zoom.ts
+++ b/projects/observability/src/shared/components/topology/d3/interactions/zoom/topology-zoom.ts
@@ -258,7 +258,7 @@ export class TopologyZoom<TContainer extends Element = Element, TTarget extends 
 
   private isPrimaryMouseOrTouchEvent(receivedEvent: ZoomSourceEvent): receivedEvent is TouchEvent | MouseEvent {
     return (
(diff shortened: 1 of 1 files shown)


projects/observability/src/shared/components/topology/d3/interactions/zoom/topology-zoom.ts
```diff
@@ -258,7 +258,7 @@ export class TopologyZoom<TContainer extends Element = Element, TTarget extends
 
   private isPrimaryMouseOrTouchEvent(receivedEvent: ZoomSourceEvent): receivedEvent is TouchEvent | MouseEvent {
     return (
-      receivedEvent instanceof TouchEvent ||
+      receivedEvent instanceof window.TouchEvent ||
```
aaron-steinfeld, line 261: Does FF have another global TouchEvent? Why does this need qualification?

anandtiwary: This interface is not defined by FF. The qualification avoids the reference error

skjindal93: I have seen Firefox complain many times that TouchEvent is not defined. And, that used to come on topology page

PR #63, opened 2020-07-27
refactor: simplify nav dash api

aaron-steinfeld:


projects/observability/src/pages/apis/api-detail/metrics/api-metrics.component.ts
```diff
@@ -22,9 +21,7 @@ export class ApiMetricsComponent {
   public constructor(apiDetailService: ApiDetailService) {
     this.filterConfig$ = apiDetailService.entityFilter$.pipe(
       map(filter => ({
-        scope: ObservabilityTraceType.Api,
-        implicitFilters: [filter],
-        hideFilterBar: true
+        implicitFilters: [filter]
```
anandtiwary, line 24: If we don't set scope what attributes does the filter bar shows?

aaron-steinfeld: It's not shown - the scope is only for the filter bar, where it's required. So rather than having a boolean and a scope as siblings, where the scope is meaningless if the boolean is true and required if it's false, I changed it to be an optional object in https://github.com/hypertrace/hypertrace-core-ui/pull/34

PR #76, opened 2020-08-03
feat(top-n): Convert top-n widget to work with explore query

arjunlalb:
Use explore query for top n widget usecase.
Make select options part of data source.
Provide name/id and context override support in options.


projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -46,41 +34,62 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
...
+            [entityTypeKey]: this.getEntityTypeForContext(this.context)
           }
         }))
       )
     );
   }
+
+  private getEntityTypeForContext(context: string): string {
+    switch (context) {
+      case 'API':
```
arjunlalb, line 80: currently we do not have an enum for explore context values. Should I add one or is there a reason why we haven't added it yet? @skjindal93 @aaron-steinfeld

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -46,41 +34,62 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
...
+          direction: 'DESC',
+          key: metricSpecification
+        }
+      ]
     })).pipe(
       map(response =>
         response.results.map(entity => ({
-          label: entity[this.attributeSpecification.resultAlias()] as string,
-          value: (entity[metricSpecification.resultAlias()] as MetricAggregation).value,
+          label: (entity[this.nameAttributeSpec.resultAlias()] as GraphQlExploreResultValue).value as string,
```
skjindal93, line 67: We can use the generics provided by `GraphQlExploreResultValue` as `GraphQlExploreResultValue<string>` here

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -46,41 +34,62 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
...
+          value: (entity[metricSpecification.resultAlias()] as GraphQlExploreResultValue).value as number,
```
skjindal93, line 68: similarly here too. `GraphQlExploreResultValue<number>`

projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget.model.ts
```diff
@@ -7,10 +7,10 @@ import {
   ModelProperty,
   ModelPropertyType
 } from '@hypertrace/hyperdash';
-import { ModelInject, MODEL_API } from '@hypertrace/hyperdash-angular';
+import { MODEL_API, ModelInject } from '@hypertrace/hyperdash-angular';
```
skjindal93, line 10: something is wrong with our lint checks? How come this was not caught before? 🤔

projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget-renderer.component.ts
```diff
@@ -4,14 +4,14 @@ import { SelectOption, SelectSize } from '@hypertrace/components';
 import { InteractiveDataWidgetRenderer } from '@hypertrace/dashboards';
 import { MetadataService } from '@hypertrace/distributed-tracing';
 import { Renderer } from '@hypertrace/hyperdash';
-import { RendererApi, RENDERER_API } from '@hypertrace/hyperdash-angular';
+import { RENDERER_API, RendererApi } from '@hypertrace/hyperdash-angular';
```
skjindal93, line 7: this too! strange!

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
arjunlalb: we actually don't have to do casting here. updated it.

projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget.model.ts
arjunlalb: might be my IDE. updated

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
aaron-steinfeld: Shouldn't need a new one - explorer context is just a superset of other enums - trace type, entity type etc. Because of that, the translation here isn't needed - these should be identical.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -46,41 +33,62 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
...
+          direction: 'DESC',
+          key: metricSpecification
+        }
+      ]
     })).pipe(
       map(response =>
         response.results.map(entity => ({
-          label: entity[this.attributeSpecification.resultAlias()] as string,
-          value: (entity[metricSpecification.resultAlias()] as MetricAggregation).value,
+          label: entity[this.nameAttributeSpec.resultAlias()].value as string,
```
aaron-steinfeld, line 66: Similar to above comment, we're assuming this will be configured with an entity which isn't necessarily true in an explorer context. Need to either make it generic, or rename + scope down to entities.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -46,41 +33,62 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
...
     });
   }
 
-  private fetchDataWithMetric(metricSpecification: MetricAggregationSpecification): Observable<TopNWidgetValueData[]> {
-    return this.queryWithNextBatch<EntitiesGraphQlQueryHandlerService, EntitiesResponse>(filters => ({
-      requestType: ENTITIES_GQL_REQUEST,
-      entityType: this.entityType,
+  private readonly specBuilder: ExploreSpecificationBuilder = new ExploreSpecificationBuilder();
+
+  private readonly nameAttributeSpec: ExploreSpecification = this.specBuilder.exploreSpecificationForKey('name');
```
aaron-steinfeld, line 43: These values are different in different contexts - which is why they were part of the definition before. If we want to limit this data source to a subset of contexts where this holds, then this is fine but we should change up the naming + typing to indicate that.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
aaron-steinfeld: My comment was a little premature - this falls under the same family as the other comments - doesn't make sense except for entities.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
arjunlalb: I am taking that up as the next step. Will add support for overriding/individually specifying name/id and context. First, I wanted everything to be converted to be explorer queries that works fine as is.

Will use appropriate names for now.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
arjunlalb: I'm working with the assumption that top-n will want to fetch entity information ultimately. I don't think its possible / a worthwhile pursuit to write something thats super generic without complicating it.

Here, the return type is TopnData with entity information.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
aaron-steinfeld: It makes sense to make it work as is, but since this is already changing the api, I don't think it makes sense to change it once to something explorer-centric, then back to something more entity centric. Ultimately, the explorer implementation isn't super relevant to the json, so I'd push for changing the naming here. If the name/id are always the same for the supported entity types, we can leave them hardcoded.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
aaron-steinfeld: Agree it's not worth doing right now - let's just adjust this one to have a name + api that make sense specifically for entities.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -1,41 +1,29 @@
...
+  GraphQlExploreResponse
+} from '../../../../graphql/request/handlers/explore/explore-graphql-query-handler.service';
 
 @Model({
   type: 'top-n-data-source'
 })
 export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFetcher> {
   @ModelProperty({
-    key: 'entity',
+    key: 'context',
```
anandtiwary, line 21: Is there any benefit of keeping the entity based data source? If yes, I think we should create an explore-top-n-data-source instead?

aaron-steinfeld, line 21: If we switch this back to `entity`, we can remove `getEntityTypeForContext`

projects/observability/src/shared/graphql/request/handlers/explore/explore-graphql-query-handler.service.ts
```diff
@@ -149,6 +158,11 @@ export interface GraphQlExploreResult {
   [key: string]: GraphQlExploreResultValue;
 }
 
+export enum ExploreQueryContextType {
+  Api = 'API',
```
aaron-steinfeld, line 162: Don't think this is needed - it is required to be identical to the entity type, and isn't used in here.

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
arjunlalb: I don't know if there's any benefit to it. All the existing usages have to be moved over to an explore query to support some of the upcoming requirements.

We can always bring it back if we want I guess. No strong preference there.

projects/observability/src/shared/graphql/request/handlers/explore/explore-graphql-query-handler.service.ts
arjunlalb: There are so many places where we use this as a string. I've started using this instead.

arjunlalb:
@aaron-steinfeld @skjindal93 @anandtiwary  Made changes as per our discussion. I have marked previous comments as resolved since some significant rewrites were done. Please review again :)

projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget.model.ts
aaron-steinfeld: This was correct as is - make sure your IDE is using the same tslint as the project (this changed in newer versions of tslint, but all files should be up to date0

projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget-renderer.component.ts
```diff
@@ -63,11 +63,15 @@ export class TopNWidgetRendererComponent extends InteractiveDataWidgetRenderer<T
     private readonly entityNavService: EntityNavigationService
   ) {
     super(api, changeDetector);
-    this.setInitialMetricSpecification();
-    this.setOptionsObservables();
+
+    this.model.getData().subscribe(fetcher => {
```
aaron-steinfeld, line 67: This should move into the tap that was removed in the `fetchData` function. Otherwise, the getData call there isn't ever being used (and we'd be calling getData indiscriminately, regardless of whether it's ultimately used).

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-explore-selection-specification.model.ts
```diff
@@ -0,0 +1,52 @@
...
+  @ModelProperty({
+    key: 'metric',
+    // tslint:disable-next-line: no-object-literal-type-assertion
+    type: {
+      key: ModelPropertyType.TYPE,
+      defaultModelClass: ExploreSelectionSpecificationModel
+    } as ModelModelPropertyTypeInstance,
+    required: true
+  })
+  public exploreSpec!: ExploreSpecification;
```
aaron-steinfeld, line 51: Is there a reason not to match the model prop name? If no, we should be consistent

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -44,32 +35,50 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
   })
   public resultLimit: number = 10;
 
+  @ModelProperty({
+    type: ARRAY_PROPERTY.type,
+    key: 'select-option-metrics',
```
aaron-steinfeld, line 40: Since we've moved this and it's a breaking change to the json anyway, what are your thoughts about adding in a rename to just `options`?

projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.test.ts
```diff
@@ -31,32 +32,40 @@ describe('Top N Data Source Model', () => {
...
     model.api = mockApi as ModelApi;
     model.resultLimit = 3;
-    model.attributeSpecification = attributeSpecificationModel;
 
-    model.query$.subscribe(query => (emittedQuery = query.buildRequest([]) as GraphQlEntitiesQueryRequest));
+    model.query$.subscribe(query => (emittedQuery = query.buildRequest([]) as GraphQlExploreRequest));
   });
 
-  test('builds expected Entity requests for Last Hour', fakeAsync(() => {
+  test('builds expected explore requests for Last Hour', fakeAsync(() => {
```
aaron-steinfeld, line 39: last hour not really relevant to test

projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget.model.ts
arjunlalb: I think this is a side effect of working in the parent project in the IDE.

In a related topic, do we really have to care about how the imports are ordered? IMO its just another headache that we can avoid.

aaron-steinfeld: That could be, depending on the IDE. For my VS code, linting + testing only works for the main project, not submodules (that's one of the main reasons I work directly against the individual repos).

The import ordering should be more or less a no-op, since it's autofixed by both prettier and tslint, both of which should be running. It's part of pretty much every style guide I've ever read to keep things consistent make it easier to read, and lower the diff sizes of changes. If it were a burdensome rule, I could see the arg for removing it - but here, I think the issue is just that the linter isn't running, which means more critical rules would be missed too.

arjunlalb: Its not really a burdensome rule. But in this case, the import optimization doesn't work properly with tslint in the IDE. Spending time to then fix it to order it when it gets messed up is burdensome IMO. Hence the question about if its really necessary.

aaron-steinfeld: I think we should focus on getting tslint working in the IDE (or documenting the expected dev flow), since I think linting in general becomes quite burdensome and not practical without IDE integration.

PR #78, opened 2020-08-03
Interaction mixins

jake-bassett:


projects/observability/src/shared/components/card-list/card-list.component.scss
```diff
@@ -1,5 +1,4 @@
-@import 'font';
-@import 'layout';
+@import 'mixins';
```
aaron-steinfeld, line 1: Tons of files presumably use the imports - are we updating this one specifically because something changed? Should we update all of them?

jake-bassett: I updated this one specifically because it required adding the interaction.scss import. I have just been doing them one by one as needed. Not sure I see a benefit to updating them until required.

aaron-steinfeld: consistency, mainly. We discussed on the HTC PR the merits of a shared import vs explicit, but I think having a hybrid approach is kind of the worst of both worlds. They'll never be updated unless it's done explicitly

aaron-steinfeld: That said, I'm ok doing this separately to get the breaking change dealt with.

PR #80, opened 2020-08-03
feat: update explorer toggle group

jake-bassett:


projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -93,7 +91,18 @@ export class ExplorerComponent {
   private readonly explorerDashboardBuilder: ExplorerDashboardBuilder;
   public readonly resultsDashboard$: Observable<ExplorerGeneratedDashboard>;
   public readonly vizDashboard$: Observable<ExplorerGeneratedDashboard>;
-  public readonly contextLabel$: Observable<ContextLabel>;
+
+  public contextItems: ToggleItem<ContextLabel>[] = [
+    {
+      label: ExplorerComponent.API_TRACES,
```
aaron-steinfeld, line 97: We convert back aand forth between theses labels and the associated values because the toggle-button didn't support a distinct value - we should fix that here (Removing `ContextLabel` the type should give a decent how-to guide on this)

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -124,12 +134,16 @@ export class ExplorerComponent {
...
   }
 
-  public onContextUpdated(label: ContextLabel): void {
-    this.context = this.contextLabelToDashboardContext(label);
-    // Set query param async to allow any initating route change to complete
+  private getContextItem(label: string): ToggleItem<ContextLabel> | undefined {
+    return this.contextItems.find(item => item.label === label);
+  }
+
+  public onContextUpdated(item: ToggleItem<ContextLabel> = this.contextItems[0]): void {
```
aaron-steinfeld, line 141: This should just take a context. It would remove the `value!`s we see below, and the label isn't relevant here.

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -92,17 +92,23 @@ export class ExplorerComponent {
   public readonly resultsDashboard$: Observable<ExplorerGeneratedDashboard>;
   public readonly vizDashboard$: Observable<ExplorerGeneratedDashboard>;
 
-  public contextItems: ToggleItem<ContextLabel>[] = [
+  public contextItems: ToggleItem<ExplorerContextScope>[] = [
     {
       label: ExplorerComponent.API_TRACES,
-      value: ExplorerComponent.API_TRACES
+      value: {
```
aaron-steinfeld, line 98: ooh that's nice

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -92,17 +92,23 @@ export class ExplorerComponent {
...
+        context: ObservabilityTraceType.Api,
+        scope: ScopeQueryParam.EndpointTraces
```
aaron-steinfeld, line 100: scope is a bit of an overloaded term (context actually replaces scope for explorer, which is lovely) - what we're putting in this scope field is actually a query parameter that's local to this component. All the usage is right, just naming question. Maybe just match the enum with `scopeQueryParam` ?

PR #82, opened 2020-08-04
Isolate top-n query

arjunlalb:
Isolate top-n-data query

### DIFF
--- a/projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
@@ -57,7 +57,7 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
     const labelAttributeSpec: ExploreSpecification = this.specBuilder.exploreSpecificationForKey(metricSpec.nameKey);
     const idAttributeSpec: ExploreSpecification = this.specBuilder.exploreSpecificationForKey(metricSpec.idKey);
 
-    return this.queryWithNextBatch<ExploreGraphQlQueryHandlerService, GraphQlExploreResponse>(filters => ({
+    return this.queryIsolated<ExploreGraphQlQueryHandlerService, GraphQlExploreResponse>(filters => ({
       requestType: EXPLORE_GQL_REQUEST,
       context: metricSpec.context,
       limit: this.resultLimit,


projects/observability/src/shared/dashboard/widgets/top-n/data/top-n-data-source.model.ts
```diff
@@ -57,7 +57,7 @@ export class TopNDataSourceModel extends GraphQlDataSourceModel<TopNWidgetDataFe
     const labelAttributeSpec: ExploreSpecification = this.specBuilder.exploreSpecificationForKey(metricSpec.nameKey);
     const idAttributeSpec: ExploreSpecification = this.specBuilder.exploreSpecificationForKey(metricSpec.idKey);
 
-    return this.queryWithNextBatch<ExploreGraphQlQueryHandlerService, GraphQlExploreResponse>(filters => ({
+    return this.queryIsolated<ExploreGraphQlQueryHandlerService, GraphQlExploreResponse>(filters => ({
```
anandtiwary, line 60: Curious to know the reason for this.

arjunlalb: Top-n-queries are apparently higher cost queries and since they are at the bottom of the page everywhere, we can isolate them and have rest of the page load faster.

anandtiwary: I see. And once the query call is completed, browser would spend more time in scripting and rendering all the data that could freeze the UI for a few seconds. This is something we saw recently on service/api detail with large time range. 

Are we really getting a perf improvement by batching the queries? Or is it slowing down the overall experience? What if we update all queries to be sent as separate http call? I feel it would improve overall user experience 

cc @jake-bassett @aaron-steinfeld @sj thoughts?

aaron-steinfeld: top n should be trivial work client side. I think there's an argument to cap the batch size, and we can test out not batching them, but there's no real reason batching should be slower - that's more indicative of other issues (backend, frontend or both).

arjunlalb: @anandtiwary We fetch very little data for top-n, so I would assume client side work is trivial like Aaron mentioned. The queries with multiple group-bys are apparently heavier on the backend. We can evaluate the performance with them being separate and later take a decision based on what we see.
I agree with @aaron-steinfeld 's point that there could be an argument to cap the batch size. Some queries are really big.

skjindal93: Top N was a heavy query when we were using entities query handler, since it was touching multiple Pinot tables (_I now forgot why it was touching multiple Pinot tables_)

Since, we moved it to explore, the query should be fast enough, since Explore works off a single Pinot table

Anyways, that's a backend issue

The problem should be solved generically by capping the batch size, instead making the queries isolated

PR #83, opened 2020-08-04
feat: update explorer for optional ToggleItem value

jake-bassett:


projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -140,10 +140,10 @@ export class ExplorerComponent {
   }
 
   private getContextItemFromValue(value: ScopeQueryParam): ToggleItem<ExplorerContextScope> | undefined {
-    return this.contextItems.find(item => value === item.value.scopeQueryParam);
+    return this.contextItems.find(item => value === item.value!.scopeQueryParam);
```
anandtiwary, line 143: But this can be false right? why not use '?'  or handle the case when value is not present?

jake-bassett: It can't be undefined. We set the value ourselves on line 95.

anandtiwary: Since the value field is optional, i think its more correct to use '?' here. Someone else can add a new context item with optional value. If that's not gonna happen, then why did we mark value as optional?

jake-bassett: Because the field is optional. :-)

Many times just a label is fine, so requiring users to always provide a value seemed heavy.

I agree that `?` is more correct if we don't own the `value` properties, but since we own and assign them ourselves, `!` assertion seems right to me.

anandtiwary: But you just made the field optional didn't you? https://github.com/hypertrace/hypertrace-core-ui/pull/60/files

aaron-steinfeld: This gets back to the question of the revert (sorry was out so didn't get to comment on it). It makes sense to allow omitting it, but it also makes sense to require it to be defined - different use cases. My preference for the api would be to omit the value field entirely - leave it up to the consumer to dictate the shape of the object as long as it satisfies what the component needs, which is label. It can omit value, have value be required, optional, or call value something else, whatever. This would change the component to be generic though, so something like `<T extends ToggleItem>`, where `ToggleItem = { label: string}`.

If we don't want to do that right now, then I'd argue for declaring our own type here, say `type ContextToggleItem = Required<ToggleItem>;` It's more or less the same as what I suggested above, the only difference is the types on the toggle component aren't quite right (but they're the same in the consumer).

PR #90, opened 2020-08-06
Update navigation

anandtiwary:


src/app/shared/navigation/navigation.component.scss
```diff
@@ -0,0 +1,7 @@
+@import 'mixins';
```
jake-bassett, line 1: Oooo. Look how easy that is to import!

anandtiwary: :D

PR #117, opened 2020-08-17
Enable conditonal service annotation

samarth-gupta-traceable:
Enable conditonal service annotation


helm/templates/service.yaml
```diff
@@ -4,6 +4,10 @@ metadata:
   name: {{ .Chart.Name }}
   labels:
     release: {{ .Release.Name }}
+  {{- if .Values.service.annotations }}
+  annotations:
```
aaron-steinfeld, line 8: we avoid the conditional with `with` in most places don't we? For example in this chart,
```
    {{- with .Values.imagePullSecrets }}
      imagePullSecrets:
        {{- toYaml . | nindent 8 }}
    {{- end }}
```
Seems slightly cleaner and more consistent with the rest of the chart

samarth-gupta-traceable: done

PR #119, opened 2020-08-18
feat: update metric widgets on home dash

arjunlalb:
Update metric widgets on home dash.

1. Change order of errors/second & calls/second widgets.
2. Add average latency and p99 latency widgets.

### DIFF
--- a/src/app/home/home.dashboard.ts
+++ b/src/app/home/home.dashboard.ts
@@ -153,6 +153,7 @@ export const homeDashboard: DashboardDefaultConfiguration = {
           {
             type: 'metric-display-widget',
             title: 'p99 Latency',
+            subscript: 'ms',
             'title-position': TitlePosition.Footer,
             data: {
               type: 'trace-metric-aggregation-data-source',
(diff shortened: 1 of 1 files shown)


src/app/home/home.dashboard.ts
```diff
@@ -152,52 +152,52 @@ export const homeDashboard: DashboardDefaultConfiguration = {
...
-                aggregation: MetricAggregationType.Sum
+                metric: 'duration',
+                aggregation: MetricAggregationType.P99
               }
             }
           },
           {
             type: 'metric-display-widget',
-            title: 'Calls/Second',
+            title: 'Avg. Latency',
```
skjindal93, line 168: P50 is Median, right?

src/app/home/home.dashboard.ts
```diff
@@ -152,52 +152,54 @@ export const homeDashboard: DashboardDefaultConfiguration = {
         children: [
           {
             type: 'metric-display-widget',
-            title: 'Total Calls',
+            title: 'p99 Latency',
+            subscript: 'ms',
```
skjindal93, line 156: The metric display widget does not support units out of the box? 🤔

arjunlalb: trace-metric data source doesn't do that yet.

PR #120, opened 2020-08-19
fix: customizable context in error percent data source

arjunlalb:
Rename 'api-error-percentage-data-source' to 'explore-error-percentage-data-source'.
Add support for custom context.

### DIFF
--- a/projects/observability/src/shared/dashboard/data/graphql/api-error-percentage/api-error-percentage-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/api-error-percentage/api-error-percentage-data-source.model.ts
@@ -5,7 +5,7 @@ import {
   MetricAggregationType,
   MetricHealth
 } from '@hypertrace/distributed-tracing';
-import { Model } from '@hypertrace/hyperdash';
+import { Model, ModelProperty, STRING_PROPERTY } from '@hypertrace/hyperdash';
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { ExploreSpecification } from '../../../../graphql/model/schema/specifications/explore-specification';
@@ -17,7 +17,7 @@ import {
 } from '../../../../graphql/request/handlers/explore/explore-graphql-query-handler.service';
 
 @Model({
-  type: 'api-error-percentage-data-source'
+  type: 'explore-error-percentage-data-source'
 })
 export class ApiErrorPercentageDataSourceModel extends GraphQlDataSourceModel<MetricAggregation> {
   public getData(): Observable<MetricAggregation> {
@@ -29,6 +29,12 @@ export class ApiErrorPercentageDataSourceModel extends GraphQlDataSourceModel<Me
       }))
     );
   }
+
+  @ModelProperty({
+    key: 'context',
+    type: STRING_PROPERTY.type
+  })
+  public context!: string;
 
   private readonly errorCountSpec: ExploreSpecification = new ExploreSpecificationBuilder().exploreSpecificationForKey(
     'errorCount',
@@ -42,7 +48,7 @@ export class ApiErrorPercentageDataSourceModel extends GraphQlDataSourceModel<Me
   private fetchErrorCountData(): Observable<number> {
     return this.query<ExploreGraphQlQueryHandlerService, GraphQlExploreResponse>({
       requestType: EXPLORE_GQL_REQUEST,
-      context: 'API',
+      context: this.context,
       selections: [this.errorCountSpec, this.numCallsSpec],
       timeRange: this.getTimeRangeOrThrow(),
       limit: 1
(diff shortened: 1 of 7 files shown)


projects/observability/src/shared/dashboard/data/graphql/api-error-percentage/api-error-percentage-data-source.model.ts
```diff
@@ -30,6 +30,12 @@ export class ApiErrorPercentageDataSourceModel extends GraphQlDataSourceModel<Me
     );
   }
 
+  @ModelProperty({
```
skjindal93, line 33: Is `required: true` by default in ModelProperty?

arjunlalb: nope. added it.

projects/observability/src/shared/dashboard/data/graphql/explore-error-percentage/explore-error-percentage-data-source.model.test.ts
```diff
@@ -8,11 +8,11 @@ import {
   EXPLORE_GQL_REQUEST,
   GraphQlExploreResponse
 } from '../../../../graphql/request/handlers/explore/explore-graphql-query-handler.service';
-import { ApiErrorPercentageDataSourceModel } from './api-error-percentage-data-source.model';
+import { ExploreErrorPercentageDataSourceModel } from './explore-error-percentage-data-source.model';
 
 describe('API error percentage data source model', () => {
```
skjindal93, line 13: nit: test name

skjindal93 (approved):

projects/observability/src/shared/dashboard/data/graphql/explore-error-percentage/explore-error-percentage-data-source.model.ts
```diff
@@ -17,9 +17,9 @@ import {
 } from '../../../../graphql/request/handlers/explore/explore-graphql-query-handler.service';
 
 @Model({
-  type: 'api-error-percentage-data-source'
+  type: 'explore-error-percentage-data-source'
```
aaron-steinfeld, line 20: Going forward, we should try to keep the implementation out of the class/model names - the fact that this uses the explore API isn't relevant, it's just an `error-percentage-data-source`

arjunlalb: this is intended as a customizable error percentage data source that works on explore query, hence the name. so that we can avoid defining these again for similar use cases.

aaron-steinfeld: but if we're customizing the numerator and the denominator, isn't it just a percentage data source? there's nothing tied to errors in it, and how it resolves the data isn't relevant to someone selecting a model

arjunlalb: this one can be just a percentage data source model for explore queries that need no filters to be applied. The name is just left over from the unused model that we revamped. will fix it.

arjunlalb: On revisiting, I think for our current use cases, we can use the other data source from the above linked PR for this purpose as well. 
That also _percentages_ values against the total call count, which is what we are trying to do here as well.

aaron-steinfeld: I'm imagining something of the pseudo form below. It's more verbose, but I think handles all the cases we need. The only gotcha right now is an aggregation data source doesn't support context, but we could add that if need be (not sure, do we have a use case there?)
```
{
  type: percentage-data-source,
  context: "API"
  numerator: {
    type: metric-aggregation-data-source, // These can already apply filters
    metric: 'foo',
    aggregation:  SUM
  },
  denominator: {
    type: metric-aggregation-data-source,
    metric: 'bar',
    aggregation:  SUM
  }
}
```

arjunlalb: Yes, this is the obvious solution. I thought the reason why we didn't do this in UI already is because we had plans to move percentage calculation to backend. Any ideas about that?

aaron-steinfeld: I'd guess they're not on the short term road map, and likely wouldn't support the type of flexibility we'd need here anyway - things like apply filters to just the numerator/denominator, or using different contexts for both.

arjunlalb: I'll create this data source for explore queries next.

PR #121, opened 2020-08-19
fix: support custom metric keys in explore error % data source

arjunlalb:


skjindal93 (approved):

projects/observability/src/shared/dashboard/data/graphql/explore-error-percentage/explore-error-percentage-data-source.model.ts
```diff
@@ -20,6 +20,27 @@ import {
...
+    required: true
+  })
+  public context!: string;
+
+  @ModelProperty({
+    key: 'error-count-metric-key',
+    type: STRING_PROPERTY.type,
+    required: true
+  })
+  public errorCountMetricKey!: string;
```
aaron-steinfeld, line 35: going a step further than my last comment, this actually feels unnecessarily specific - what in here has anything to do with errors other than the names? This looks like it's purely a division data source (the parts that are still hardcoded being that they a. share a context and b. assume a sum aggregation)

PR #122, opened 2020-08-19
feat: define and use trace-calls-percentage data source

arjunlalb:
Screenshot 2020-08-19 at 2 05 21 PM

Define trace calls percentage data source.
Add new widgets to home dash.

### DIFF
--- a/projects/observability/src/shared/dashboard/data/graphql/calls-percentage/calls-percentage-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/calls-percentage/calls-percentage-data-source.model.ts
@@ -5,7 +5,7 @@ import {
   MetricAggregationType,
   MetricHealth
 } from '@hypertrace/distributed-tracing';
-import { Model } from '@hypertrace/hyperdash';
+import { Model, ModelProperty, STRING_PROPERTY } from '@hypertrace/hyperdash';
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { ExploreSpecification } from '../../../../graphql/model/schema/specifications/explore-specification';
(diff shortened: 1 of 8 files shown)


projects/observability/src/shared/dashboard/data/graphql/trace-calls-percentage/trace-calls-percentage-data-source.model.ts
```diff
@@ -0,0 +1,71 @@
...
+    return forkJoinSafeEmpty([this.fetchCallCountWithFilters(), this.fetchNumCallsData()]).pipe(
+      map(results => ({
+        value: getPercentage(results[0], results[1]),
+        health: MetricHealth.NotSpecified,
+        units: '%'
+      }))
+    );
+  }
+
+  private readonly context: string = 'API_TRACE';
```
skjindal93, line 33: Can we also probably make this as ErrorPercentageDataSourceModel? Like this one can take context as a ModelProperty and call count key?

arjunlalb: sure. I didn't do that today because I couldn't find a usecase. thats all

arjunlalb: @skjindal93 done

projects/observability/src/shared/dashboard/data/graphql/calls-percentage/calls-percentage-data-source.model.ts
```diff
@@ -0,0 +1,83 @@
...
+})
+export class CallsPercentageDataSourceModel extends GraphQlDataSourceModel<MetricAggregation> {
+  @ModelProperty({
+    key: 'context',
+    type: STRING_PROPERTY.type,
+    required: true
+  })
+  public context!: string;
+
+  @ModelProperty({
```
aaron-steinfeld, line 30: Same comment as the earlier PR - why do we need specific data sources for each use case - they're all the same, taking in two metrics and dividing them.

arjunlalb: difference is that this data source needs filters to be applied on the numerator, where as the error-percentage calculation do not.
I tried to keep this generic enough within the context of calls. This data source can be reused for various cases in the numerator like you see in the home dashboard in this PR.

arjunlalb: I did think of defining a generic percentage data source model, but there are too many possibilities around that to cover. Some percentage calculations make calls in the same context for both numerator and denominator, some others don't. Some need filters on the numerator, some would need on denominator, and some might need at both places.

arjunlalb: @aaron-steinfeld if there are no other comments apart from the specificity of data sources, shall I merge this and take up that refactor after?

aaron-steinfeld: sure

PR #124, opened 2020-08-19
feat: filter button component in explorer

jake-bassett:


projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
```diff
@@ -38,6 +38,10 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
   })
   public childEntityType?: ObservabilityEntityType;
 
+  public getScope(): string {
```
aaron-steinfeld, line 41: can this take in a row ? hacky, but don't see another way.

jake-bassett: A row? Do you mean array?

aaron-steinfeld: For entities (and other trees), we may have rows in the table that have a different scope from other rows. I don't remember where this is called, but the data source itself doesn't have enough information to give a scope without being told what row it's giving the scope for.

anandtiwary: Service list?

anandtiwary: Can we address the review comments in a separate PR? I kind of pulled on latest htc and now need those changes to continue t-ui changes.

jake-bassett: I created a story for the backlog.

PR #125, opened 2020-08-20
feat: adding selection styles to cards

anandtiwary:


projects/observability/src/shared/components/card-list/card-list.component.ts
```diff
@@ -19,6 +24,22 @@ export class CardListComponent {
...
   public mode?: CardListMode = CardListMode.List;
+
+  public selectedCard?: CardContainerComponent;
+
+  public onCardClicked(card: CardContainerComponent): void {
+    this.selectedCard = this.selectedCard === card ? undefined : card;
+  }
+
+  public getCardStyle(card: CardContainerComponent): string[] {
+    const classes = [(this.mode ?? CardListMode.List).toString()];
```
aaron-steinfeld, line 35: why is `toString()` needed here - this is a const enum which is a subclass of string

anandtiwary: Ts complains. Don't know. may be because I am adding a string obj below.

aaron-steinfeld: that would do it. instead of calling a function, can just add the type: 
```suggestion
    const classes: string[] = [this.mode ?? CardListMode.List];
```

PR #142, opened 2020-08-24
feat: define and use a generic percentage data source

arjunlalb:
Define a metric aggregation data source with custom context support.
Use the above data source to define a generic % data source.

Tests are WIP.


src/app/home/home.dashboard.ts
```diff
@@ -215,35 +215,67 @@ export const homeDashboard: DashboardDefaultConfiguration = {
...
-                  type: 'calls-percentage-data-source',
-                  context: 'API_TRACE',
-                  'call-count-metric-key': 'calls',
-                  filters: [
-                    {
-                      type: 'graphql-key-value-filter',
-                      key: 'duration',
-                      operator: GraphQlOperatorType.GreaterThan,
-                      value: 1000
+                  type: 'percentage-data-source',
```
anandtiwary, line 218: data source that depends on other data sources. Should we rename it differently? `percentage-composite-data-source` may be?

arjunlalb: Makes sense. Will merge this and send a follow up PR modifying the name.

PR #151, opened 2020-08-28
feat: adding grouping to timeline cards

anandtiwary:


projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.test.ts
```diff
@@ -10,7 +11,7 @@ describe('Timeline Card List component', () => {
     declareComponent: false,
     component: TimelineCardListComponent,
     imports: [TimelineCardListModule],
-    providers: []
+    providers: [mockProvider(PopoverService)]
```
anandtiwary, line 14: Don't want to make this a shallow test since I need container component to project contents.

projects/observability/src/shared/components/timeline-card-list/container/timeline-card-container.component.ts
```diff
@@ -9,4 +9,7 @@ import { ContentHolder, CONTENT_HOLDER_TEMPLATE } from '@hypertrace/components';
 export class TimelineCardContainerComponent extends ContentHolder {
   @Input()
   public timestamp!: number | Date;
+
+  @Input()
+  public similarToPrevious?: boolean;
```
aaron-steinfeld, line 14: This API strikes me as pretty strange - components don't generally know about their siblings.

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.ts
```diff
@@ -30,4 +55,40 @@ export class TimelineCardListComponent {
...
+  public ngAfterContentInit(): void {
+    queryListAndChanges$(this.cards)
+      .pipe(map(list => list.toArray()))
+      .subscribe(allCards => {
+        this.allCards = allCards;
+        this.buildItems(allCards);
+      });
+  }
+
+  public readonly isSelectedCard = (card: TimelineCardContainerComponent): boolean =>
```
aaron-steinfeld, line 71: why make this an instance rather than prototype function? Intentional?

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.ts
```diff
@@ -30,4 +55,40 @@ export class TimelineCardListComponent {
...
+    this.selectedIndex !== undefined && this.selectedIndex >= 0 && this.allCards[this.selectedIndex] === card;
+
+  private buildItems(allCards: TimelineCardContainerComponent[]): void {
```
aaron-steinfeld, line 74: What if we changed this to create groups rather than a boolean referring to a sibling. Like:
```html
    <ht-timeline-card-list>
      <ht-timeline-card-group [collapsed]="true"> <!-- could add things like summary, etc in the future to this -->
        <ht-timeline-card-container [timestamp]="cardData.timestamp">
          <div class="custom-card">
            <div class="title">{{cardData.name}}</div>
          </div>
        </ht-timeline-card-container>
      </ht-timeline-card-group>
    </ht-timeline-card-list>
```

That API seems clearer to me, but I'm not sure on how we're using it.

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.ts
anandtiwary: Oh. I was trying to use htcMemoize for this, but i think it only takes one argument, which is card here. It doesn't trigger the function when selectedIndex change. If there is a way to pass more than one argument, then I can use it for with both card and selectedIndex and have it trigger only when one of them change.

Ofcourse after trying, I forgot to remove it as instance.

anandtiwary: Do you think it is easy to have memoize support multiple arguments?

aaron-steinfeld: already does - example from memoize tests:
```typescript
    const mockFn = jest.fn((...args: string[]) => args.join(','));
    const spectator = createHost(
      `
    <pipe-holder>
      {{ hostFn | htcMemoize : arg1 : arg2 }}
    </pipe-holder>`,
      {
        hostProps: {
          hostFn: mockFn,
          arg1: 'foo',
          arg2: 'alpha'
        }
      }
    );
    expect(mockFn).toHaveBeenCalledTimes(1);
    expect(spectator.query(byText('foo,alpha'))).toExist();
```

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.ts
anandtiwary: The reason I didn't chose this is that it allows random grouping of cards. The grouping we want has to be strictly adjacent in time. Otherwise, **we are hiding important information from the user** and the timeline view will also be broken.

> This API strikes me as pretty strange - components don't generally know about their siblings.

Yes, it is strange. But it is also on a container component and not on a generalized component. The `<ht-timeline-card-list>` is responsible for rendering the cards, the containers are just supposed to provide meta info for the cards. With this understanding, i think it makes sense. And, if `ht-timeline-card-container` is a structural directive then the same would look much more decent :)

projects/observability/src/shared/components/timeline-card-list/container/timeline-card-container.component.ts
anandtiwary: explained below

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.ts
anandtiwary: It does seem to support it. I think I was using a wrong syntax.

anandtiwary: I was passing wrong syntax. Fixed and pushing now

anandtiwary: thanks

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.ts
aaron-steinfeld: So what are the criteria we'd generally group by? Timestamps are within some threshold of each other and card data is also matching some implementation-specific equality (e.g. same title) ?

I still like the idea of telling the timeline list how to group with some form of metadata, vs a boolean back reference which forces logic on either side. For example, passing a grouping function, or a grouping key + time threshold.

anandtiwary: So exactly preceding in time and same type. 

The grouping function approach still has the potential to break the timeline imo. I want to avoid a scenario where we are showing cards out of order of timestamps. With a grouping function that can happen if the function itself is written incorrectly. And we will also have to maintain unique grouping keys for correctness.

In the current implementation, the adjacent grouping is strictly enforced. If EventA occurs again after EventB, then they are treated differently. Current api is ensuring this criteria.

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.scss
```diff
@@ -7,7 +7,13 @@
   padding-left: 1px;
   padding-right: 1px;
 
-  .record {
+  .button {
+    margin-top: 8px;
+    margin-left: 88px;
```
jake-bassett, line 12: Strange margin.

anandtiwary: Had to put this so that it is aligned with the card boundary

anandtiwary:
Hah.. With that comment and approval, I am put on spot. Of course I want to merge this change, but don't want to reject Aaron's comments. Will wait now :)

projects/observability/src/shared/components/timeline-card-list/timeline-card-list.component.ts
aaron-steinfeld: I guess I'd think that functiin would be used only to check boundaries, while the timie is stsill enforced. Anywayy, I'm ok with htis prroceediinig. can revissitt if needed. (sorry mac keyboard)

PR #152, opened 2020-08-28
feat: added gauge list component

skjindal93:
A gauge chart component which does not have a sort order on the data and can take in a color key for custom colors for key names and gauge

TopN chart passes in sorted data to Gauge Chart component

With the colors keys and custom sorting, the chart can now look like this
Screen Shot 2020-08-28 at 1 51 52 PM


aaron-steinfeld:
High level comment - I would think a gauge would be a single row here, vs a list of them

projects/observability/src/shared/components/top-n/top-n-chart.component.ts
```diff
@@ -1,30 +1,18 @@
...
-          {{ item.label }}
-        </div>
-        <div class="progress">
-          <div class="progress-value" [ngStyle]="{ width: item.width }"></div>
-        </div>
-        <div class="value">
-          <span>{{ item.value | htcDisplayNumber }}</span>
-        </div>
-      </ng-container>
+      <ht-gauge-chart
```
aaron-steinfeld, line 11: If all this is doing is sorting, my vote would be to get rid of this component. Use the underlying, unsorted chart directly from the renderer and have the renderer do the sorting. It seems excessive to have an extra component layer to go `widget -> sorted -> unsorted`

projects/observability/src/shared/components/gauge/gauge-chart.component.ts
```diff
@@ -0,0 +1,99 @@
...
+      </ng-container>
+    </div>
+  `
+})
+export class GaugeChartComponent<T extends GaugeData = GaugeData> implements OnChanges {
+  @Input()
+  public data: T[] = [];
+
+  @Input()
+  public labelClickable: boolean = false;
```
aaron-steinfeld, line 38: If we're refactoring this anyway, I think it'd be a good opportunity to clean up the api. We take in `data`, use `labelClickable` to determine if it can be clicked and `itemClick` to ouput that click - all three are really referring to the same unit.

skjindal93: What do you mean by clean the api though?

aaron-steinfeld: data, label and item all refer to the same thing. So perhaps `data` ->`items` `labelClickable` -> `itemClickable` for example

projects/observability/src/shared/components/gauge/gauge-chart.component.ts
```diff
@@ -0,0 +1,99 @@
...
+          <div class="progress-value" [ngStyle]="{ width: item.width, backgroundColor: item.color }"></div>
+        </div>
+        <div class="value">
+          <span>{{ item.value | htcDisplayNumber }}</span>
+        </div>
+      </ng-container>
+    </div>
+  `
+})
+export class GaugeChartComponent<T extends GaugeData = GaugeData> implements OnChanges {
```
anandtiwary, line 33: This design is now similar to old top-n and histogram. :)
`https://github.com/hypertrace/hypertrace-ui/blob/main/projects/observability/src/shared/components/histogram/histogram-chart.component.ts`

anandtiwary: We should name it differently for sure. Gauge chart has a different meaning.

projects/observability/src/shared/components/gauge/gauge-chart.component.ts
skjindal93: ah, just renames. makes sense

projects/observability/src/shared/components/sort-n/sort-n-chart.component.ts
```diff
@@ -0,0 +1,98 @@
...
+      original: option
+    }));
+  }
+
+  private buildColorLookupForData(barData: SortNData[]): Map<string, string> | undefined {
+    const uniqueDataValues = new Set(barData.map(data => data.colorKey ?? data.label));
+
+    return this.determineColor
+      ? new Map(Array.from(uniqueDataValues).map(value => [value, this.determineColor!(value)]))
+      : undefined;
```
arjunlalb, line 84: would an empty map be a safer choice here?

projects/observability/src/shared/components/sort-n/sort-n-chart.component.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+export interface SortNData {
+  label: string;
+  value: number;
+  colorKey?: string;
```
arjunlalb, line 91: simplify to `color`?

skjindal93: We actually follow this concept of passing in a determineColor function as an Input to the component, which takes in a colorKey and outputs a color string.

Hence, the data will come with a colorKey in it, and then use the determineColor function passed to determine the color corresponding to the colorKey

projects/observability/src/shared/components/sort-n/sort-n-chart.component.ts
skjindal93: I can do empty map

projects/observability/src/public-api.ts
```diff
@@ -184,9 +184,9 @@ export * from './shared/components/timeline-card-list/timeline-card-list.module'
 export * from './shared/components/interval-select/interval-select.component';
 export * from './shared/components/interval-select/interval-select.module';
 
-// Top N Chart
-export * from './shared/components/top-n/top-n-chart.module';
-export * from './shared/components/top-n/top-n-chart.component';
+// Sort N Chart
+export * from './shared/components/sort-n/sort-n-chart.module';
```
aaron-steinfeld, line 188: To save some effort, we should probably discuss and agree on names before you go through the effort of implementing the change - I didn't like top N and I'm not sure I understand the change to sort-n.

projects/observability/src/shared/dashboard/widgets/top-n/top-n-widget-renderer.component.ts
```diff
@@ -74,18 +74,19 @@ export class TopNWidgetRendererComponent extends InteractiveDataWidgetRenderer<T
...
+  protected fetchData(): Observable<SortNData[]> {
     return this.model.getData().pipe(
       tap(fetcher => {
         this.fetcher = fetcher;
         this.setInitialMetricSpecification(fetcher);
         this.setOptionsObservables(fetcher);
       }),
-      switchMap(() => this.buildDataObservable())
+      switchMap(() => this.buildDataObservable()),
+      map(data => data.sort((first, second) => second.value - first.value))
```
aaron-steinfeld, line 85: So sort n doesn't sort? I think this pattern is right, it's just a naming thing. What do we think of `gauge-list`?

skjindal93: renamed it to gauge-list

projects/observability/src/public-api.ts
skjindal93: yeah. learnt it the hard way

PR #162, opened 2020-09-03
refactor(gauge-list): rename sort n to gauge item

skjindal93:


projects/observability/src/shared/components/gauge-list/gauge-list.component.test.ts
```diff
@@ -1,5 +1,5 @@
 import { FormattingModule } from '@hypertrace/common';
-import { GaugeListComponent, SortNData } from '@hypertrace/observability';
+import { GaugeItem, GaugeListComponent } from '@hypertrace/observability';
```
aaron-steinfeld, line 2: bad import, should be relative.

skjindal93: Interesting. I guess I will have to fix WebStorm

skjindal93: The auto suggested import always suggests the aliased imports

aaron-steinfeld (approved):

PR #164, opened 2020-09-03
feat: separated cell parsers

jake-bassett:


projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
```diff
@@ -28,16 +28,16 @@ import { EntityNavigationService } from '../../services/navigation/entity/entity
...
 
   @Input()
   public navigable: boolean = true;
 
   @Input()
   public showIcon: boolean = false;
 
-  public name!: unknown;
-  public entityIconType!: string;
+  public name?: unknown;
```
anandtiwary, line 39: why not string?

jake-bassett: Not sure. I just left it as it was, but I can change in my local to string.

anandtiwary (approved):

PR #165, opened 2020-09-04
Add htc

anandtiwary:


jake-bassett (approved):

README.md
```diff
@@ -13,7 +13,7 @@ Install Node + NPM
 
   `npm install`
 
-- Initialize Submodule (https://github.com/hypertrace/hypertrace-core-ui/)
+- Initialize Submodule (https://github.com/hypertrace/)
```
aaron-steinfeld, line 16: There no longer is a submodule right?

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "hypertrace-core-ui"]
```
aaron-steinfeld, line 1: Shouldn't this whole file be deleted?

package.json
```diff
@@ -22,7 +22,7 @@
     "prettier:check": "prettier --check '**'",
     "test:ci": "ng test hypertrace-ui -i --ci --coverage",
     "docs": "compodoc -p tsconfig.json",
-    "core-ui:clean": "rm -rf hypertrace-core-ui/node_modules"
+    "core-ui:clean": "rm -rf node_modules"
```
aaron-steinfeld, line 25: no longer needed, and can be removed from commands above.

adriancole:
like the action and reduced submodules, but may have missed the license aspect possibly the source history part, too.

If this is a carry-on from a gdoc I wrote about macrorepo, one thing maybe missed was that having a "core" directory allows you to know which files are licensed as what. This is moot if the license is unified now, but I'm not sure if that was intentional. The source history is less important with a recent project, but something to consider in PR description to say explicitly we are ok to reset the history (instead of using a technique to allow it to be retained pinned to a merge commit)

Meanwhile, it is early enough to adjust however we choose. Only major thing to address now is what the license of the code is.

aaron-steinfeld:
Good point. I thought I mentioned this in one of the conversations but it was likely lost. We need to address the licensing here - the easiest way to do that would be to put a license in each project folder, but I'm not sure if that's sufficient. If we need to organize the projects into core/non-core folders, we can do that instead with minimal impact.

anandtiwary:
Yeah we were planning to work on the licensing part, but other high priority items came up in security. Will get to this soon.

PR #179, opened 2020-09-09
Move units addition to meta data

anandtiwary:


projects/observability/src/shared/graphql/request/handlers/entities/query/entities-graphql-query-builder.service.ts
```diff
@@ -92,44 +88,20 @@ export class EntitiesGraphqlQueryBuilderService {
   }
 
   public normalizeEntity(rawResult: Dictionary<unknown>, request: EntityDescriptor): Observable<Entity> {
-    return forkJoinSafeEmpty(
```
anandtiwary, line 95: This logic was repeated in entities handler, interaction handler and is supposed to be used in the new actor and actors handler. So I decided to move it to meta data service so that the code can be shared. 
@aaron-steinfeld what do you think? Will add some test after initial feedback.

projects/distributed-tracing/src/shared/graphql/model/schema/specifier/specification.ts
```diff
@@ -10,3 +10,8 @@ export interface Specification {
   asGraphQlOrderByFragment(): GraphQlSortWithoutDirection;
   extractFromServerData(resultContainer: Dictionary<unknown>): unknown;
 }
+
+export interface SpecificationResult {
+  alias: string;
```
aaron-steinfeld, line 15: Can we get a type on the data? at a minimum, `T`? It seems like it should reflect the potential presence of a units field though.

projects/distributed-tracing/src/shared/services/metadata/metadata.service.ts
```diff
@@ -116,6 +92,75 @@ export class MetadataService {
...
+          return this.resultUnits(scope, spec).pipe(
+            map(units => ({
+              alias: alias,
+              data: { units: units, ...(data as object) }
+            }))
+          );
+        }
+
+        return of({
+          alias: alias,
```
aaron-steinfeld, line 121: So having alias in the result ass far as I can see is just a way of mapping it back to the specific specification that was the source of the mapped data - so can we provide a Map<Specification, SpecficationResult> instead to be more explicit?

anandtiwary: Although it would make it more readable, I feel I would be complicating the code.  Here is how it would look

```
  public buildSpecificationResultWithUnits(
    rawResult: Dictionary<unknown>,
    specifications: Specification[],
    scope: string
  ): Observable<Map<Specification, SpecificationResult<unknown>>> {
    return forkJoinSafeEmpty(
      specifications.map(spec => {
        const alias = spec.resultAlias();
        const data = spec.extractFromServerData(rawResult);

        if (isMetricSpecification(spec) && isMetricAggregation(data)) {
          return this.resultUnits(scope, spec).pipe(
            map(
              units =>
                [
                  spec,
                  {
                    alias: alias,
                    data: { units: units, ...(data as object) }
                  } as SpecificationResult<unknown>
                ] as [Specification, SpecificationResult<unknown>]
            )
          );
        }

        return of([
          spec,
          {
            alias: alias,
            data: data as SpecificationResult<unknown>
          }
        ] as [Specification, SpecificationResult<unknown>]);
      })
    ).pipe(map(results => new Map<Specification, SpecificationResult<unknown>>(results)));
  }
```

projects/distributed-tracing/src/shared/graphql/model/schema/specifier/specification.ts
anandtiwary: Every specification result can be different. We may be able to type it here, but in `buildSpecificationResultWithUnits` I don't think we can pass all the types, so i may have to pass unknown only. So I am not sure what we will achieve.

projects/distributed-tracing/src/shared/services/metadata/metadata.service.ts
anandtiwary: On the caller side, the specification in the map would be used. We just iterate over key value, only use value to add it in the final result object. So I am not convinced this would be super helpful. Thoughts?

projects/distributed-tracing/src/shared/graphql/model/schema/specifier/specification.ts
aaron-steinfeld: the value might be different (although the specific shape depends on the caller), but what is under the purview of the new code is the presents of a unit field. For example, to match the metadata logic we could do:
```
export interface SpecificationResult<T> {
  alias: string;
  data: T extends MetricAggregation ? T & { units : string} : T;
}
```

projects/distributed-tracing/src/shared/services/metadata/metadata.service.ts
aaron-steinfeld: My mistake on the example type - I meant to replace `SpecificationResult` entirely since the map removes need for an alias - so something like `Map<Specification, unknown>`. We do iterate the final value, but maps are still iterable and it's keeping the result matching the input, rather than this implementation being tied to how the caller's using it.

anandtiwary: Yeah. This makes sense

anandtiwary: It conflicts with the other comment though :). I would prefer this over the new types.

aaron-steinfeld: Kind of conflicts, I just got lazy with the unknown - the new type comment could still be relevant here though;
```
export type MaybeWithUnits<T> = T extends MetricAggregation ? T & { units : string} : T;
```
then return `Map<Specification, MaybeWithUnits<T>>`

anandtiwary: I think for this to work, Specification should itself by typed with T. Since `extractFromServerData` returns unknown, I don't think I can map it to a generic T/MaybeWithUnits<T>. 
```
export interface Specification {
  resultAlias(): string;
  name: string;
  displayName?: string;
  asGraphQlSelections(): GraphQlSelection | GraphQlSelection[];
  asGraphQlOrderByFragment(): GraphQlSortWithoutDirection;
  extractFromServerData(resultContainer: Dictionary<unknown>): unknown;
}
```

aaron-steinfeld: Spec probably should be typed, but I think the two changes are independent - it's still on the caller to decide what the type of the result is since an array can contain specs with different shapes (hence my comment on the other PR about getting a standard data wrapper object) - this change is more about the returned type giving the caller an indication of what's happening. It's really in the same vane as that comment, I think of it as more of a proto-wrapper object, that we could update in the future. For example, if an entity/span/trace became a dictionary of `MaybeWithUnits` it could still be used as is today, and when we eventually started putting units on all things not just metric aggregations, that would be reflected in the type system vs be transparent in an unknown type. We can hold off on that now and address that as a stand alone effort though.

projects/distributed-tracing/src/shared/services/metadata/metadata.service.ts
```diff
@@ -138,3 +177,5 @@ export class MetadataService {
     return displayName;
   }
 }
+
+export type MaybeWithUnits<T> = T extends MetricAggregation ? T & { units: string } : T;
```
aaron-steinfeld, line 181: Don't think we ended up using this

aaron-steinfeld (approved):

projects/distributed-tracing/src/shared/services/metadata/metadata.service.ts
anandtiwary: okay. Will address this separately. Will open a tech debt today.

projects/distributed-tracing/src/shared/services/metadata/metadata.service.ts
anandtiwary: Will remove it in a follow up.

PR #180, opened 2020-09-10
fix: stop syncing explorer page data to url

aaron-steinfeld:


skjindal93 (approved):

anandtiwary:
Wouldn't this break the filtering ? cc @jake-bassett

aaron-steinfeld:
> Wouldn't this break the filtering ? cc @jake-bassett 

Filters are still part of the url, this takes out paging and sort data

anandtiwary:
And why are we taking them out?

aaron-steinfeld:
> And why are we taking them out?

This isn't actually a full screen dashboard for one. As a consequence there's a bug where, if changing the dashboard filters when on the second page of data, the new data is fetched but the page is not reset. We still need to fix this for full page tables I suspect (depends on how they react to a new data source) but this change gets the most problematic spot and seems right regardless

anandtiwary:
okay

PR #182, opened 2020-09-10
feat: label tag component

arjunlalb:
Creating a label tag component with support for custom colors and text.
Updating bet tag component use the newly defined label tag component.
Defining a `color.ts` file to define color values.


projects/common/src/color/colors.ts
```diff
@@ -0,0 +1,6 @@
+export enum Colors {
+  Purple2_15pOpacity = '#f2d0f5',
```
skjindal93, line 2: Can we just call it Purple2, since we call it the same in _color-palette.scss?

skjindal93: An enum for Colors sounds awesome!

projects/components/src/label-tag/label-tag.component.ts
```diff
@@ -0,0 +1,28 @@
...
+  public label?: string;
+
+  @Input()
+  public labelColor?: string;
+
+  @Input()
+  public backgroundColor?: string;
+
+  public ngOnInit() {
+    this.backgroundColor = this.backgroundColor ?? Colors.Gray1;
```
skjindal93, line 25: I think the `ngStyle` for that property does not get applied, if the value is undefined, so probably we can define these defaults directly as CSS

projects/common/src/color/colors.ts
arjunlalb: aah. my bad. I was testing something and forgot to change this back.

projects/components/src/label-tag/label-tag.component.ts
arjunlalb: makes sense.

projects/components/src/label-tag/label-tag.component.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+export class LabelTagComponent {
+  @Input()
+  public label?: string;
+
+  @Input()
+  public labelColor?: Color;
+
+  @Input()
+  public backgroundColor?: Color;
```
anandtiwary, line 22: No sure about making these styles as inputs. Can we do it by introducing an input for roles or style enum  and hiding all the required css changes inside the component? An example of this is the button component. 

It would improve readability as we would not have to go into the details of choosing the right color and background. We can just choose a role/style and all corresponding css would come. 

It would also allow us to not maintain a color js object.

arjunlalb: Any particular reason why we don't want to maintain a color js object? I feel its good given we use so many hex strings in the code already.

Also, for style inputs, I think this works best for the tag use case because we might want to use it for a varied set of usecases. Example, `Beta` / `New` or even other usecases in the future. I feel that the role/style using CSS will be cumbersome to maintain and won't give flexibility for a very generic component like this.

anandtiwary: Regarding color object, there was some discussion before about doing this with a service that would read directly from css variables. The advantages are
1. we wouldn't have to maintain colors at two places. With the current color object approach, if Michael changes any of the color (as part of rebranding), we would have to change at both places. 
2. As this color object grows, it would become difficult to maintain. 
3. Implementing Themeing would be a big problem. For dark theme, we wouldn't be applying the same colors. 

The service approach would work best as it will only read from one source of truth and can quickly switch from one theme to another.

Similarly, by exposing label colors and background colors as input properties, we are :
1. In a way sacrificing on the component abstraction for more flexibility. 
2. If Michael choose a different style, we will have to modify all its usages to pass new colors, rather than modifying the css from within this component.
3. Themeing becomes way more harder as every component usage is passing colors explicitly. If this pattern grows, it will become much more difficult. 


If we just introduce roles, we can achieve the same thing with all the flexibility, just that we are pushing it one layer deep.

projects/components/src/header/page/page-header.component.ts
```diff
@@ -29,7 +29,13 @@ import { NavigableTab } from '../../tabs/navigable/navigable-tab';
           ></ht-icon>
 
           <ht-label [label]="titlecrumb.label"></ht-label>
-          <ht-beta-tag class="beta" *ngIf="this.isBeta"></ht-beta-tag>
+          <ht-label-tag
+            *ngIf="this.isBeta"
+            class="beta"
+            label="Beta"
```
aaron-steinfeld, line 35: It seems like there's still an argument for a beta tag component even if it's noow built on the lower level label-tag component - better a tiny component than having to pass the same 3 params in every place we need a beta tag and updating them all together.

projects/common/src/color/color.ts
```diff
@@ -0,0 +1,7 @@
+export enum Color {
```
aaron-steinfeld, line 1: const

projects/components/src/label-tag/label-tag.component.ts
aaron-steinfeld: I don't love the idea of maintaining two separate color declarations, but I think it's strictly better than the hardcoding we're doing now - maybe we should go with it and revisit when we have more time to think on color in general. The idea of roles is independent vs explicit values is independent of this, and I generally do agree there - the problem being that we don't want to have the generic component understand where it's being used (if it's supplying the roles, which in a tag component can be pretty varied and specific).

So I guess what I'm saying is - without knowing what we're doing with theming (and now's not the time for that) it's hard to say which solution will be easier to adapt in the future, and I'd personally be fine with either way right now.

projects/components/src/header/page/page-header.component.ts
arjunlalb: That makes sense. Will add a `beta` tag component.

projects/common/src/color/color.ts
arjunlalb: I made this an enum so that this can also be used as a type for the input params. Please let me know if there's a strong preference to avoid enum here.

projects/components/src/label-tag/label-tag.component.ts
arjunlalb: That was what I was aiming for also. 2 sources is better than all the string values we have lying around the code right now. I was planning on cleaning those up as well. When we revisit theming and define a color service, this work isn't going to hurt us. If anything, it'll make life easier compared to moving from what we have now.

And yes, my reasoning for specific colors also was the same. Tag component can be very varied, I don't think we should push that use case specific styling down to the component.

Like @aaron-steinfeld mentioned in another comment. If we have a use case like beta-tag, which is going to be used at different places, we can actually create another small component on top of this one to make life easier. I feel thats a cleaner solution than pushing styling as roles into this generic component.

@anandtiwary thoughts?

projects/common/src/color/color.ts
skjindal93: I think @aaron-steinfeld meant `export const enum`

projects/components/src/label-tag/label-tag.component.ts
anandtiwary: Isn't a beta tag component over label tag same as combining them into one component and selecting a role?. That also works, but we are adding another component which is not doing much. 

Do we want to use this pattern elsewhere? Like, what are your thoughts on applying the same pattern on button component? We would have a single button component that would take different css as inputs and separate specific button components like `button-primary`, `button-additive`, `button-destructive` etc. Would you prefer it over a single component?

That being said, it is still a small component, so I am not too worried. We can surely iterate over it. :)

projects/common/src/color/color.ts
arjunlalb: oh. right. ok. now I feel stupid.

projects/components/src/label-tag/label-tag.component.ts
arjunlalb: @anandtiwary technically, its kinda the same thing. We customize a component for the usecase. If we have just a one off use case, the label component can be directly used with colors as inputs. If there are multiple use cases, we can write a wrapper component (like `beta-tag`) to avoid duplication.

IMO, button usecase is slightly different from this because our design system choices would dictate the kind of buttons we use. We won't use many different styles of buttons nor would we want one off customizations for it. And the styles of a component like button is dictated by the functionality of it. Whereas a tag is just a text with a style (style being the differentiator) and theoretically, it could have any style applied to it.

If this tag ever becomes an actionable tag (ex: click to open something, show a popover etc), then I'd see more merit in defining role based styles. Because then we would want the user to relate a particular style with a particular action also. For a display only tag, I feel it should be left customizable, in control of the consumer.

projects/common/src/color/color.ts
aaron-steinfeld: 🦜 (no party parrot to be found)

projects/components/src/label-tag/label-tag.component.ts
aaron-steinfeld: It's the same idea but an inversion of concerns - a button with multiple roles means the base component has knowledge of the whole design system. For a button where we have a limited number based on an abstract concept like primary/additive/destructive, that makes sense. New concepts shouldn't be added with any regularity, they should be driven by the design system, and we want to prevent any stray from the specific predefined roles. On the other hand, the tag's categories aren't really look and feel driven, but use case - such as a beta tag. So building use cases in to the base component there seems like the wrong approach, as these additions may be driven by specific features rather than the design system, and we don't want feature changes to require changes to the underlying component system.

projects/components/src/beta-tag/beta-tag.component.test.ts
```diff
@@ -1,16 +1,21 @@
 import { createHostFactory, Spectator } from '@ngneat/spectator/jest';
+import { LabelTagComponent } from '../label-tag/label-tag.component';
 import { BetaTagComponent } from './beta-tag.component';
 
 describe('Beta Tag Component', () => {
   let spectator: Spectator<BetaTagComponent>;
 
   const createHost = createHostFactory({
     component: BetaTagComponent,
+    entryComponents: [LabelTagComponent],
```
aaron-steinfeld, line 10: shouldn't need an entry component here - maybe a declaration if we want to go one level deep on the test (current approach), or to keep it strictly shallow, a mocked label tag component.

projects/components/src/beta-tag/beta-tag.module.ts
```diff
@@ -1,8 +1,11 @@
+import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
+import { LabelTagModule } from '../label-tag/label-tag.module';
 import { BetaTagComponent } from './beta-tag.component';
 
 @NgModule({
   declarations: [BetaTagComponent],
+  imports: [CommonModule, LabelTagModule],
```
aaron-steinfeld, line 8: Don't see anything used from common module here

arjunlalb: CommonModule is for ngStyle

aaron-steinfeld: ngStyle is used by label-tag rather than beta directly so is a concern of `LabelTagModule`

projects/components/src/beta-tag/beta-tag.component.test.ts
arjunlalb: tried doing that. the Color enum within LabelTag component doesn't get resolved when I use a declaration. Any alternate ideas? @aaron-steinfeld

projects/components/src/beta-tag/beta-tag.module.ts
arjunlalb: oh. I looked at the wrong place. Sorry. Yes. Its not required here. Will rmeove.

projects/components/src/beta-tag/beta-tag.component.test.ts
aaron-steinfeld: That's strange - I'd suspect that would be an artifact of test caching, as an entry vs declaration shouldn't be relevant there. They're sometimes a bit squirrelly around the enums I've found. If it's still happening after switching back to declaration now, try clearing the jest cache (`--clear-cache` arg on whatever test command you're running) and rerunning - and if that doesn't work, we can leave it as an entry component.

arjunlalb: it was the jest cache. Works after clearing. Updated.

projects/components/src/label-tag/label-tag.component.ts
arjunlalb: Seems like we are on the same page on this.

aaron-steinfeld (approved):

projects/components/src/label-tag/label-tag.component.ts
arjunlalb: Merged this change for now. We can iterate on this.

PR #196, opened 2020-09-15
feat: menu dropdown component

arjunlalb:
A generic menu dropdown component


projects/components/src/menu-dropdown/menu-dropdown.component.ts
```diff
@@ -0,0 +1,46 @@
...
+      <ht-popover [closeOnClick]="true" class="menu-container">
+        <ht-popover-trigger>
+          <div class="trigger-content">
+            <ht-label *ngIf="this.label" class="trigger-label" [label]="this.label"> </ht-label>
+            <ht-icon *ngIf="this.icon" class="trigger-icon" [icon]="this.icon" size="${IconSize.Small}"></ht-icon>
+          </div>
+        </ht-popover-trigger>
+        <ht-popover-content>
+          <div class="dropdown-content">
+            <div *ngFor="let item of items" (click)="item.action()" class="menu-item">
```
jake-bassett, line 21: What happens if action is undefined?

arjunlalb: as of now, nothing happens. the consumer needs to handle the actions. We could log a warning.

projects/components/src/menu-dropdown/menu-item.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  @Input()
+  public label!: string;
+
+  @Input()
+  public icon?: IconType;
+
+  @Input()
+  public iconColor?: string;
+
+  @Input()
```
aaron-steinfeld, line 20: Rather than passing in a function, can this instead be an output?

projects/components/src/menu-dropdown/menu-dropdown.component.ts
jake-bassett: So no console error on click for an undefined method? If no error, this is fine. Not looking to log a warning, just didn't want an undefined method to not be handled.

projects/components/src/menu-dropdown/menu-dropdown.component.test.ts
```diff
@@ -0,0 +1,76 @@
...
+import { IconType } from '@hypertrace/assets-library';
+import { MenuDropdownComponent, MenuItemComponent } from '@hypertrace/components';
+import { createHostFactory, SpectatorHost } from '@ngneat/spectator/jest';
+import { MockComponent } from 'ng-mocks';
+import { LabelComponent } from '../label/label.component';
+
+describe('Menu dropdown Component', () => {
+  const hostFactory = createHostFactory<MenuDropdownComponent>({
+    component: MenuDropdownComponent,
+    entryComponents: [MenuItemComponent],
```
aaron-steinfeld, line 11: not an entry component

projects/components/src/menu-dropdown/menu-dropdown.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  public icon?: IconType;
+
+  @Input()
+  public label?: string;
+
+  @ContentChildren(MenuItemComponent)
+  public items?: QueryList<MenuItemComponent>;
+
+  @Output()
+  public output: EventEmitter<() => Observable<any>> = new EventEmitter();
```
aaron-steinfeld, line 57: The linter beat me to it, but never any!

projects/components/src/menu-dropdown/menu-dropdown.component.ts
```diff
@@ -0,0 +1,62 @@
...
+      <ht-popover [closeOnClick]="true" class="menu-container">
+        <ht-popover-trigger>
+          <div class="trigger-content">
+            <ht-label *ngIf="this.label" class="trigger-label" [label]="this.label"> </ht-label>
+            <ht-icon *ngIf="this.icon" class="trigger-icon" [icon]="this.icon" size="${IconSize.Small}"></ht-icon>
+          </div>
+        </ht-popover-trigger>
+        <ht-popover-content>
+          <div class="dropdown-content">
+            <div *ngFor="let item of items" (click)="this.onMenuItemClick(item)" class="menu-item">
```
aaron-steinfeld, line 30: What we discussed yesterday was instead of explicitly rendering these in the dropdown, move the icon into the item template and here just render the content via `ng-content`. This both makes the DOM more similar to the template but also removes the need for an output at all, since click can be directly bound by the user of menu-dropdown. See toggle button group for an example.

arjunlalb: @aaron-steinfeld I'm working on that. Ignore the last few commits. I had made these changes earlier. Just wanted to pull it in another project.

projects/components/src/menu-dropdown/menu-item/menu-item.component.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  test('should display icon and label as expected', fakeAsync(() => {
+    const spectator = createHost(
+      '<ht-menu-item [icon]="icon" [label]="label" [iconColor]="iconColor"></ht-menu-item>',
+      {
+        hostProps: { icon: IconType.MoreHorizontal, label: 'Item', iconColor: Color.Gray1 }
+      }
+    );
+    expect(spectator.query('.menu-item', { root: true })).toExist();
```
aaron-steinfeld, line 20: why are these root queries? Also should we verify the icon and icon color beyond existence? Mocking the icon component would make that trivial

projects/components/src/menu-dropdown/menu-dropdown.component.test.ts
```diff
@@ -2,14 +2,12 @@ import { fakeAsync } from '@angular/core/testing';
...
 import { createHostFactory, SpectatorHost } from '@ngneat/spectator/jest';
-import { MockComponent } from 'ng-mocks';
 import { LabelComponent } from '../label/label.component';
 
 describe('Menu dropdown Component', () => {
   const hostFactory = createHostFactory<MenuDropdownComponent>({
     component: MenuDropdownComponent,
-    entryComponents: [MenuItemComponent],
-    declarations: [MockComponent(LabelComponent)],
+    declarations: [LabelComponent, MenuItemComponent],
```
aaron-steinfeld, line 10: Using the real components here means this isn't a shallow test - if that's intentional, let's remove line 11. If not, let's mock the menu and label components.

projects/components/src/menu-dropdown/menu-item/menu-item.component.test.ts
arjunlalb: oops. thats left over from a copy paste. removing.

I didnt' want to check the exact icon and color because that is tested in the icon component. Thought it'd be redundant.

aaron-steinfeld: Not checking the resolution of them but the input - so checking that we correctly wire the icon and color inputs from this component to the icon component. In otherwords, something like `expect(spectator.query(IconComponent)?.icon).toBe(IconType.MoreHorizontal);`

projects/components/src/menu-dropdown/menu-dropdown.component.test.ts
arjunlalb: Will remove shallow then. For this test to work, we need the MenuItemComponent. Mocking doesn't work. Or we have to use entryComponent

aaron-steinfeld: That's only because the test is written to assume the internals of the item. The following change removes that assumption and keeps things shallow.
```diff
diff --git a/projects/components/src/menu-dropdown/menu-dropdown.component.test.ts b/projects/components/src/menu-dropdown/menu-dropdown.component.test.ts
index 878d6e9..cabcd7d 100644
--- a/projects/components/src/menu-dropdown/menu-dropdown.component.test.ts
+++ b/projects/components/src/menu-dropdown/menu-dropdown.component.test.ts
@@ -8,7 +8,7 @@ import { LabelComponent } from '../label/label.component';
 describe('Menu dropdown Component', () => {
   const hostFactory = createHostFactory<MenuDropdownComponent>({
     component: MenuDropdownComponent,
-    declarations: [MockComponent(LabelComponent), MenuItemComponent],
+    declarations: [MockComponent(LabelComponent), MockComponent(MenuItemComponent)],
     shallow: true
   });
 
@@ -36,11 +36,11 @@ describe('Menu dropdown Component', () => {
     );
 
     spectator.click('.trigger-content');
-    const optionElements = spectator.queryAll('.menu-item', { root: true });
-    expect(spectator.query('.dropdown-content', { root: true })).toExist();
-    expect(optionElements.length).toBe(2);
+    const optionComponents = spectator.queryAll(MenuItemComponent);
+    expect(spectator.query('.dropdown-content')).toExist();
+    expect(optionComponents.length).toBe(2);
 
-    expect(optionElements[0]).toHaveText('Do X');
-    expect(optionElements[1]).toHaveText('Do Y');
+    expect(optionComponents[0].label).toBe('Do X');
+    expect(optionComponents[1].label).toBe('Do Y');
   }));
 });

```

projects/components/src/menu-dropdown/menu-item/menu-item.component.test.ts
aaron-steinfeld: Importing the module is still using the real one - this is just bringing in all the real icon dependencies too. See below diff, which works with no test changes:
```diff
diff --git a/projects/components/src/menu-dropdown/menu-item/menu-item.component.test.ts b/projects/components/src/menu-dropdown/menu-item/menu-item.component.test.ts
index 33defb3..8681630 100644
--- a/projects/components/src/menu-dropdown/menu-item/menu-item.component.test.ts
+++ b/projects/components/src/menu-dropdown/menu-item/menu-item.component.test.ts
@@ -1,14 +1,15 @@
 import { fakeAsync } from '@angular/core/testing';
-import { IconLibraryTestingModule, IconType } from '@hypertrace/assets-library';
+import { IconType } from '@hypertrace/assets-library';
 import { Color, NavigationService } from '@hypertrace/common';
-import { IconComponent, IconModule, MenuItemComponent } from '@hypertrace/components';
+import { IconComponent, MenuItemComponent } from '@hypertrace/components';
 import { createHostFactory, mockProvider } from '@ngneat/spectator/jest';
+import { MockComponent } from 'ng-mocks';
 
 describe('Menu Item Component', () => {
   const createHost = createHostFactory<MenuItemComponent>({
     component: MenuItemComponent,
-    imports: [IconModule, IconLibraryTestingModule],
     shallow: true,
+    declarations: [MockComponent(IconComponent)],
     providers: [mockProvider(NavigationService)]
   });
```

projects/components/src/menu-dropdown/menu-dropdown.component.test.ts
arjunlalb: aah. interesting. thanks!

aaron-steinfeld (approved):

PR #197, opened 2020-09-15
fix: waterfall indentation

jake-bassett:


projects/components/src/table/table.component.ts
```diff
@@ -496,7 +496,7 @@ export class TableComponent
   }
 
   public isExpanderColumn(columnConfig: TableColumnConfig): boolean {
-    return columnConfig === this.expandableToggleColumnConfig;
+    return columnConfig.display === this.expandableToggleColumnConfig.display;
```
anandtiwary, line 499: So easy. Should there be any test to capture this particular failure?

anandtiwary (approved):

projects/components/src/table/table.component.ts
jake-bassett: Agreed. Added.

aaron-steinfeld (approved):

PR #198, opened 2020-09-15
refactor: remove json string encoding from span detail

anandtiwary:


aaron-steinfeld:
What's an example of a before/after body here?

jake-bassett (approved):

anandtiwary:
Decode the JSON payload --> convert `\"` to `"`

"{\"content\": 2 }" -> {"content": 2}

projects/distributed-tracing/src/shared/components/span-detail/call/body/span-detail-call-body.component.ts
```diff
@@ -30,7 +30,7 @@ import { ToggleViewMode } from '@hypertrace/components';
             </ht-message-display>
           </section>
           <section *ngSwitchCase="this.rawLabel">
-            <pre class="pre-body">{{ this.body | json }}</pre>
+            <pre class="pre-body">{{ this.rawBody }}</pre>
```
aaron-steinfeld, line 33: Are we not pretty printing this anymore? Did this ever work - I thought the json pipe takes objects.

aaron-steinfeld (commented):
> Decode the JSON payload --> convert `\"` to `"`
> 
> `"{\"content\": 2 }" `-> {"content": 2}

I saw the conversion of `\"`, but not the leading or trailing quotes.

projects/distributed-tracing/src/shared/components/span-detail/call/body/span-detail-call-body.component.ts
anandtiwary: Yep. Don't think it worked for raw format.

aaron-steinfeld (approved):

PR #201, opened 2020-09-16
Update cdk dependency

jake-bassett:


aaron-steinfeld (approved):

projects/components/src/input/input.component.test.ts
```diff
@@ -63,6 +63,11 @@ describe('Input Component', () => {
     );
 
     const inputEl = spectator.query('input') as HTMLInputElement;
-    expect(inputEl.placeholder).toEqual('placeholder');
+    /*
+     * Upgrading @angular/material from 10.0.2 to 10.2.1 made this test fail. The HTML5 placeholder attribute is not
```
aaron-steinfeld, line 67: Looks like they explicitly changed it: https://github.com/angular/components/blob/master/src/material/input/input.ts#L80-L82
Wouldn't it make more sense to change to `spectator.query(MatInput)` then we don't have to be concerned with how the placeholder is implemented, we can just check the input.

jake-bassett: Yes. Yes it would. Thanks!

PR #208, opened 2020-09-21
feat: table column resize

jake-bassett:


projects/components/src/table/table.component.ts
```diff
@@ -330,6 +343,85 @@ export class TableComponent
     this.rowStateSubject.complete();
     this.columnStateSubject.complete();
     this.columnConfigsSubject.complete();
+
+    this.mouseMoveFunction && this.mouseMoveFunction();
```
aaron-steinfeld, line 347: Not clear why these are needed on destroy

projects/components/src/table/table.component.ts
```diff
@@ -330,6 +343,85 @@ export class TableComponent
...
+    this.mouseUpFunction && this.mouseUpFunction();
+  }
+
+  public trackItem(_index: number, column: TableColumnConfigExtended): string {
+    return `${column.id}-${column.width}`;
+  }
+
+  public queryHeaderRowElement(): HTMLElement {
+    return this.elementRef.nativeElement.querySelector('cdk-header-row');
```
aaron-steinfeld, line 356: Why can't this use a normal view child selector?

projects/components/src/table/table.component.ts
```diff
@@ -330,6 +343,85 @@ export class TableComponent
...
+  private getVisibleColumnConfig(index: number): TableColumnConfigExtended {
+    return this.columnConfigsSubject.value.filter(cc => cc.visible)[index];
+  }
+
+  public onResizeMouseDown(event: MouseEvent, index: number): void {
+    const leftColumn: ColumnInfo = this.buildColumnInfo(index - 1);
+    const rightColumn: ColumnInfo = this.buildColumnInfo(index);
+
+    this.lastOffsetX = this.calcOffsetX(event, 0, leftColumn.bounds.left, rightColumn.bounds.right);
+    this.mouseMoveFunction = this.renderer.listen('document', 'mousemove', e =>
```
aaron-steinfeld, line 372: Why do we need these on the document rather than on the element? Offsets? It would seem those offsets should be able to be local to the component.

projects/components/src/table/table.component.ts
```diff
@@ -330,6 +343,85 @@ export class TableComponent
...
+
+  public trackItem(_index: number, column: TableColumnConfigExtended): string {
+    return `${column.id}-${column.width}`;
+  }
+
+  public queryHeaderRowElement(): HTMLElement {
+    return this.elementRef.nativeElement.querySelector('cdk-header-row');
+  }
+
+  public queryHeaderCellElement(index: number): HTMLElement {
```
aaron-steinfeld, line 359: Why are these public? Also, could we get this through angular? For example, by iterating through a view children query.

projects/components/src/table/table.component.ts
```diff
@@ -66,15 +68,19 @@ import { TableColumnConfigExtended, TableService } from './table.service';
...
-      <div *ngFor="let columnDef of this.columnConfigsSubject.value; index as index">
+      <div *ngFor="let columnDef of this.columnConfigs$ | async; trackBy: this.trackItem; index as index">
         <ng-container [cdkColumnDef]="columnDef.id">
           <cdk-header-cell
             *cdkHeaderCellDef
             [style.flex-basis]="columnDef.width"
             [style.max-width]="columnDef.width"
             class="header-cell"
           >
+            <div *ngIf="index !== 0" class="header-column-handle" (mousedown)="this.onResizeMouseDown($event, index)">
```
aaron-steinfeld, line 79: Rather than working by index and in the dom, requerying each time - couldn't we pass in the column def, and have the column def hold the changing data? Or at least serve as a reference to it if we're not mutating the column already.

projects/components/src/table/table.component.ts
```diff
@@ -330,6 +343,85 @@ export class TableComponent
...
+
+  public queryHeaderRowElement(): HTMLElement {
+    return this.elementRef.nativeElement.querySelector('cdk-header-row');
+  }
+
+  public queryHeaderCellElement(index: number): HTMLElement {
+    return this.elementRef.nativeElement.querySelectorAll('cdk-header-cell').item(index);
+  }
+
+  private getVisibleColumnConfig(index: number): TableColumnConfigExtended {
```
aaron-steinfeld, line 363: Is this assuming the index is from visible cols rather than all cols? If so, we should rename it, because looks like a bug otherwise.

projects/components/src/table/table.component.ts
aaron-steinfeld: Are these being removed? Also are they being invoked on events outside of table?

projects/components/src/table/table.component.ts
```diff
@@ -330,6 +343,85 @@ export class TableComponent
...
+    return {
+      config: this.getVisibleColumnConfig(index),
+      element: element,
+      bounds: this.calcBounds(element)
+    };
+  }
+
+  private calcBounds(element: HTMLElement): ColumnBounds {
+    return {
+      left: element.offsetLeft + 12,
```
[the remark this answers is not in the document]
jake-bassett: They are additional pixels to prevent us reducing the width too much. I don't know of a way to do that with css.

projects/components/src/table/table.component.ts
jake-bassett: The method name should be clear enough.

projects/components/src/table/table.component.ts
jake-bassett: I tried using ViewChildren, but it doesn't update correctly. Even subscribing to `changes` on the QueryList gave wrong results because of the ngFor with dynamic content, I assume. It would start with the correct columns, then as we modify them they get removed from the QueryList and never return.

projects/components/src/table/table.component.ts
jake-bassett: I'm not sure I understand what the goal is here. What does this give us?

aaron-steinfeld: The goal is preventing multiple DOM queries per mouse move event when those query results are known - they only change if columns are reconfigured.

projects/components/src/table/table.component.ts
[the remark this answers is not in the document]
aaron-steinfeld: OK - sounds like they're not not based on css then - I thought they might be reflective of certain margin or padding values in the stylesheet and needed to stay in sync.

projects/components/src/table/table.component.ts
aaron-steinfeld: Hmm strange. Anyway, can we make this private? Looks to only be called from a private method.

projects/components/src/table/table.component.ts
jake-bassett: Guess I'm still not following. The mousedown event is only triggered when there is mousedown over the column-handle. It doesn't fire off any other time.

projects/components/src/table/table.component.ts
jake-bassett: The only reason I made it public was so I could override the method during test. I don't like modifying code just for a test, but the alternative is importing the CDK into the test and building the full DOM there, which is a bigger undertaking.

aaron-steinfeld: Agree that alternative is worse. Another potential alternative which we've used elsewhere for similar issues is `DomElementMeasurerService` - allowing mocking measurements without mocking the dom itself.

projects/components/src/table/table.component.ts
aaron-steinfeld: Discussed offline - the comment position doesn't make a whole lot of sense here, but the ask is to move the expensive querying and measuring (if possible) to the mouse down handler, rather than in each cycle of the mousemove handler.

aaron-steinfeld (approved):

PR #211, opened 2020-09-23
chore: move filtering components into filtering dir

jake-bassett:


aaron-steinfeld (approved):
You sure love your code with a certain level of organization

PR #214, opened 2020-09-23
refactor: new filter builders

jake-bassett:


projects/components/src/filtering/filter/builder/types/abstract-filter-builder.ts
```diff
@@ -0,0 +1,38 @@
...
+  public static supportedAttributeTypes: FilterAttributeType[];
+  public static supportedOperators: FilterOperator[];
+
+  protected abstract supportedAttributeTypes(): FilterAttributeType[];
+  protected abstract supportedOperators(): FilterOperator[];
+
+  protected abstract convertStringToValue(value: string): TValue | undefined;
+  protected abstract convertValueToString(value: TValue): string;
+
+  public buildFilters(attribute: FilterAttribute, value: TValue): Filter<TValue>[] {
```
aaron-steinfeld, line 16: Naming on this one - presuming this is used to generate potential choices for auto complete/cell filtering, but without looking at the implementation that's not obvious. Maybe `buildPotentialFilters` or `buildFilterForEachOperator`

aaron-steinfeld (approved):

projects/components/src/filtering/filter/builder/types/abstract-filter-builder.ts
jake-bassett: Yep! I'll change locally.

PR #215, opened 2020-09-24
fix: explorer now keeps url filter on back from waterfall

jake-bassett:


projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
```diff
@@ -77,13 +77,13 @@ export class ExploreQueryEditorComponent implements OnChanges, OnInit {
...
-    if (changeObject.filters) {
-      this.visualizationBuilder.filters(this.filters);
-    }
-
     if (changeObject.context) {
       this.visualizationBuilder.context(this.context);
     }
+
+    if (changeObject.filters) {
+      this.visualizationBuilder.filters(this.filters);
```
skjindal93, line 85: Since visualizationBuilder is a builder, how does the preference matter?

aaron-steinfeld: It's a bit confusing api, but it basically exposes an observable which emits on each change in the builder - changing the context resets the whole thing though, so we'd want that to happen first.

aaron-steinfeld (approved):

projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
jake-bassett: 👆

PR #216, opened 2020-09-24
feat: Define modalRef and modalService. Support for data in modal config

arjunlalb:
Define modalref with an output object.
Modal service to work with modalref and inject it.
Add data support to modalconfig.
New method in overlay service to render modals.

New confirmation dialog component. (Leaving it here for easier testing and demonstrating sample usage. Will remove at the end to keep PRs separate)


projects/components/src/overlay/modal/modal-overlay.module.ts
```diff
@@ -3,8 +3,10 @@ import { NgModule } from '@angular/core';
 import { ButtonModule } from '../../button/button.module';
 import { TooltipModule } from '../../tooltip/tooltip.module';
 import { ModalOverlayComponent } from './modal-overlay.component';
+import { ModalService } from './modal-service';
 
 @NgModule({
+  providers: [ModalService],
```
aaron-steinfeld, line 9: providers should either belong to a component or be in root. Declaring them in the module is effectively root, but is more confusing to the reader.

projects/components/src/overlay/modal/modal-ref.ts
```diff
@@ -0,0 +1,14 @@
+import { Subject } from 'rxjs';
+import { PopoverRef } from '../../popover/popover-ref';
+
+export class ModalRef extends PopoverRef {
+  public readonly output$: Subject<unknown> = new Subject<unknown>();
```
aaron-steinfeld, line 5: I would argue for keeping this completely decoupled from popover rather than extending it - it's built on top of popover, but should be completely its own concept.

projects/components/src/confirmation-modal/confirmation-modal.component.ts
```diff
@@ -0,0 +1,64 @@
+import { ChangeDetectionStrategy, Component, Inject } from '@angular/core';
+import { ButtonRole } from '../button/button';
+import { ModalOverlayConfig } from '../overlay/modal/modal';
+import { ModalRef } from '../overlay/modal/modal-ref';
+import { POPOVER_DATA } from '../popover/popover';
+
+@Component({
+  selector: 'lib-confirmation-modal',
```
aaron-steinfeld, line 8: `ht-`

projects/components/src/confirmation-modal/confirmation-modal.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { ConfirmationModalComponent } from './confirmation-modal.component';
+
+// TODO: Add tests here
```
aaron-steinfeld, line 5: Assuming these are still wip since they're not using spectator.

projects/components/src/confirmation-modal/confirmation-modal.component.ts
```diff
@@ -0,0 +1,64 @@
...
+    </div>
+  `
+})
+export class ConfirmationModalComponent {
+  public actionButtonText?: string;
+  public descriptionText?: string;
+
+  public constructor(
+    private readonly modalRef: ModalRef,
+    @Inject(POPOVER_DATA) private readonly popoverData: ModalOverlayConfig<ConfirmationModalData>
```
aaron-steinfeld, line 39: as far as a component using modals, I would expect it to be able to have two special injects available - `ModalRef` (can be a class-based inject, as it is here) and `MODAL_DATA` which is of an unknown type (so requires an inject token, similar to POPOVER_DATA), provided with the value of the data arg from the config.

projects/components/src/confirmation-modal/confirmation-modal.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  ) {
+    this.actionButtonText = this.popoverData?.data?.confirmButtonText;
+    this.descriptionText = this.popoverData?.data?.descriptionText;
+  }
+
+  public onConfirmation(): void {
+    console.log('Confirm clicked');
```
aaron-steinfeld, line 46: Rather than explicitly expose an output like this, I'd take them as an argument to close. If we think it's valuable, we could expose it as two functions like `close` and `dismiss` and then return a `ModalResult` object, but that's likely overkill.

projects/components/src/overlay/modal/modal-service.ts
```diff
@@ -0,0 +1,46 @@
+import { Injectable, Injector } from '@angular/core';
+import { PopoverOptions, POPOVER_DATA } from '../../popover/popover';
+import { PopoverRef } from '../../popover/popover-ref';
+import { PopoverService } from '../../popover/popover.service';
+import { ModalRef } from './modal-ref';
+
+@Injectable()
+export class ModalService extends PopoverService {
```
aaron-steinfeld, line 8: Similar to the ref, I think we should avoid extending popoverservice, and instead use it - similar to how the existing overlay service works.

projects/components/src/overlay/overlay.service.ts
```diff
@@ -46,10 +52,13 @@ export class OverlayService implements OnDestroy {
...
 
-  public createModal(config: ModalOverlayConfig, injector: Injector = this.defaultInjector): PopoverRef {
+  public createModal<TData = unknown>(
+    config: ModalOverlayConfig<TData>,
+    injector: Injector = this.defaultInjector
+  ): ModalRef {
     this.activeModalPopover?.close();
 
-    const popover = this.popoverService.drawPopover({
+    const modal = this.modalService.drawPopover({
```
aaron-steinfeld, line 61: If we want to put this in its own service, let's remove the method here and have callers use modal service directly.

projects/components/src/overlay/overlay.ts
```diff
@@ -1,7 +1,12 @@
 import { TemplateRef, Type } from '@angular/core';
+import { Observable } from 'rxjs';
 
 export interface OverlayConfig {
   showHeader?: boolean;
   title?: string;
   content: TemplateRef<unknown> | Type<unknown>;
 }
+
+export interface ConfirmationModalConfig extends OverlayConfig {
```
aaron-steinfeld, line 10: Confirmation is just another modal in the library, it shouldn't be part of the overlay api itself.

projects/components/src/popover/popover-ref.ts
```diff
@@ -25,9 +25,9 @@ export class PopoverRef {
   private _visible: boolean = false;
 
   public constructor(
-    private readonly overlayRef: OverlayRef,
-    private readonly positionBuilder: PopoverPositionBuilder,
-    private readonly navigationService: NavigationService
+    public readonly overlayRef: OverlayRef,
+    public readonly positionBuilder: PopoverPositionBuilder,
```
aaron-steinfeld, line 29: Why do these need to be exposed?

projects/components/src/confirmation-modal/confirmation-modal.component.spec.ts
arjunlalb: Yep. PR is still a work in progress. Tests, cleanup, and reorganizing pending.

projects/components/src/overlay/modal/modal-ref.ts
arjunlalb: What does that gain us here when we just want one extra property?

aaron-steinfeld: It's not just one extra property though - most of the methods on popoverRef are not relevant to modalRef. In other words, inheritance is a bad fit here. What methods do we actually want to expose on ModalRef? I'd think something along the lines of below to start.
```typescript
interface ModalRef<TResult> {
   closed$: Observable<TResult>;
   close(result: TResult): void;
}
```
As far as implementation, we can easily take in a popoverref to the ModalRef implementation to support this. A good rule of thumb on inheritance - are you actually using it polymorphically? If yes, and it satisfies an `is a` relationship, then it might be a good fit. If it's being used to cut down on code written, it's likely not. It tightly couples things, making the child much more susceptible to changes in the parent (vs changes in the parent's API in the suggested wrapping approach) and makes it harder to test in isolation as you can't separate the parent logic out to mock it.

projects/components/src/popover/popover-ref.ts
arjunlalb: this was done for the constructor in ModalRef. If we are moving to an independent ModalRef, this won't be required.

projects/components/src/overlay/modal/modal-ref.ts
anandtiwary: I would agree with the api aaron suggested. If we use has-a relationship, it would be cleaner and it would avoid unnecessary changes to popover to support the inheritance.

projects/components/src/overlay/overlay.service.ts
```diff
@@ -46,10 +52,13 @@ export class OverlayService implements OnDestroy {
     return popover;
   }
 
-  public createModal(config: ModalOverlayConfig, injector: Injector = this.defaultInjector): PopoverRef {
+  public createModal<TData = unknown>(
+    config: ModalOverlayConfig<TData>,
```
anandtiwary, line 56: I would think that we should maintain a different config for confirmation model. `ConfirmationModelConfig` may extend from ModalOverlayConfig with the addition of certain properties like `confirmButtonLabel`, `confirmClickHandler`, `cancelButtonLabel` and `cancelClickHandler`.  Based on the handler user clicks on, we can emit true/false on the `closed` observable.

projects/components/src/confirmation-modal/confirmation-modal.component.ts
anandtiwary: +1. The action should be controlled from within the component itself.

projects/components/src/overlay/overlay.ts
anandtiwary: +1

projects/components/src/overlay/overlay.service.ts
aaron-steinfeld: I agree with the properties, but rather than extending, I believe a ModalOverlayConfig should have a data object, and the data object is what would be injected into the modal. That would work the same way as popover. Bringing it all together:

```typescript
export interface ModalConfig <TData = never> {
  componentOrTemplate: Type<unknown> | TemplateRef<TData>;
  data?: TData; // available as MODAL_DATA
  parentInjector?: Injector;
}
```

Used by some modal, such as a confirmation modal like:
```typescript
class ConfirmationModalComponent {
  public constructor(
     private readonly modalRef: ModalRef,
     @Inject(MODAL_DATA) private readonly data: ConfirmationModalConfig) {}

   public onClose(): void {
    this.modalRef.close(true);
  }

   public onCancel(): void {
    this.modalRef.close(false);
  }
}
```

arjunlalb:
@aaron-steinfeld @anandtiwary This PR has been updated. Please review again. Will add tests after that.

projects/components/src/overlay/modal/modal-ref.ts
```diff
@@ -0,0 +1,70 @@
+import { OverlayRef } from '@angular/cdk/overlay';
+import { Portal } from '@angular/cdk/portal';
+import { NavigationService } from '@hypertrace/common';
+import { Observable, Observer, ReplaySubject } from 'rxjs';
+import { filter, takeUntil } from 'rxjs/operators';
+
```
aaron-steinfeld, line 6: This class doesn't need to mirror popoverRef - it can greatly simplify. To start, I'd probably strip it all the way down to something like:
```typescript
export class ModalRef<TResult> {
  public readonly closed$: Observable<TResult>;

  private readonly closedObserver: Observer<TResult>;

  public constructor(private readonly popoverRef: PopoverRef) {
    const closedSubject = new ReplaySubject<TResult>(1);
    this.closedObserver = closedSubject;
    this.closed$ = closedSubject.asObservable();
    this.popoverRef.closed$.subscribe({
      complete: () => this.closedObserver.complete(),
      error: err => this.closedObserver.error(err)
    });
  }

  public close(result?: TResult): void {
    if (result !== undefined) {
      this.closedObserver.next(result);
    }
    this.popoverRef.close();
  }
```

projects/components/src/overlay/modal/modal-service.ts
```diff
@@ -0,0 +1,146 @@
...
+import { ModalOverlayComponent } from './modal-overlay.component';
+import { ModalRef } from './modal-ref';
+
+@Injectable({ providedIn: 'root' })
+export class ModalService {
+  private activeModalPopover?: ModalRef;
+  private modalCloseSubscription?: Subscription;
+
+  public constructor(
+    public readonly overlay: Overlay,
```
aaron-steinfeld, line 23: like this modalref, we don't need to completely reimplement popover here. Instead, inject popoverservice - a modal is just a specialized popover.

projects/components/src/overlay/modal/modal.ts
```diff
@@ -1,9 +1,31 @@
...
 export const enum ModalSize {
   Small = 'small'
 }
+
+export const MODAL_DATA = new InjectionToken<unknown>('MODAL_DATA');
+
+export interface ModalOptions<TData = never> {
+  componentOrTemplate: Type<unknown> | TemplateRef<TData>;
+  data?: TData;
+  parentInjector?: Injector;
```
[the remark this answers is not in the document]
arjunlalb: Agreed. I left it in there just in case we had a usecase.

projects/components/src/overlay/modal/modal-service.ts
arjunlalb: We need to override the ref and data injection in this class hence the choice to reimplement it here instead of injecting popoverservice.

aaron-steinfeld: We don't need to override it though - we just need to provide our own injector including those things.

aaron-steinfeld: I was thinking something along the lines of:
```typescript
@Injectable({
  providedIn: 'root'
})
export class ModalService {
  private activeModal?: ModalRef<unknown>;

  public constructor(private readonly popoverService: PopoverService, private readonly defaultInjector: Injector) {}

  public createModal<T>(config: ModalOverlayConfig, injector: Injector = this.defaultInjector): ModalRef<T> {
    this.activeModal?.close();

    const popover = this.popoverService.drawPopover({
      componentOrTemplate: ModalOverlayComponent,
      parentInjector: injector,
      position: {
        type: PopoverPositionType.Fixed,
        location: PopoverFixedPositionLocation.Centered
      },
      data: {
        config: config,
        parentInjector: injector
      },
      backdrop: PopoverBackdrop.Opaque
    });

    popover.closeOnNavigation();
    this.activeModal = new ModalRef(popover);

    return this.activeModal;
  }
}
```

aaron-steinfeld: we could then use this data passed into the popover to create an injector for the modal that doesn't expose the underlying popover usage:

```diff
diff --git a/projects/components/src/overlay/modal/modal-overlay.component.ts b/projects/components/src/overlay/modal/modal-overlay.component.ts
index b4e8ffb..27f4dd7 100644
--- a/projects/components/src/overlay/modal/modal-overlay.component.ts
+++ b/projects/components/src/overlay/modal/modal-overlay.component.ts
@@ -1,9 +1,10 @@
-import { ChangeDetectionStrategy, Component, Inject, TemplateRef, Type } from '@angular/core';
+import { ChangeDetectionStrategy, Component, Inject, Injector, TemplateRef, Type } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
 import { ButtonSize, ButtonStyle } from '../../button/button';
 import { POPOVER_DATA } from '../../popover/popover';
 import { PopoverRef } from '../../popover/popover-ref';
-import { ModalOverlayConfig, ModalSize } from './modal';
+import { ModalOverlayConfig, ModalSize, MODAL_DATA } from './modal';
+import { ModalRef } from './modal-ref';
 
 @Component({
   selector: 'ht-modal-overlay',
@@ -26,7 +27,7 @@ import { ModalOverlayConfig, ModalSize } from './modal';
       <div class="content-wrapper">
         <div class="content">
           <ng-container *ngIf="this.isComponentModal; else templateRenderer">
-            <ng-container *ngComponentOutlet="this.renderer"></ng-container>
+            <ng-container *ngComponentOutlet="this.renderer; injector: this.componentInjector"></ng-container>
           </ng-container>
           <ng-template #templateRenderer>
             <ng-container *ngTemplateOutlet="this.renderer"></ng-container>
@@ -42,19 +43,42 @@ export class ModalOverlayComponent {
   public readonly size: ModalSize;
   public readonly isComponentModal: boolean;
   public readonly renderer: TemplateRef<unknown> | Type<unknown>;
+  public readonly componentInjector: Injector;
 
   public visible: boolean = true;
 
-  public constructor(private readonly popoverRef: PopoverRef, @Inject(POPOVER_DATA) config: ModalOverlayConfig) {
+  public constructor(private readonly popoverRef: PopoverRef, @Inject(POPOVER_DATA) metadata: ModalMetadata) {
+    const config = metadata.config;
     this.showHeader = config.showHeader === true;
     this.modalTitle = config.title === undefined ? '' : config.title;
     this.size = config.size;
     this.isComponentModal = !(config.content instanceof TemplateRef);
     this.renderer = config.content;
+    this.componentInjector = this.buildComponentInjector(metadata, popoverRef);
   }
 
   public close(): void {
     this.visible = false;
     this.popoverRef.close();
   }
+
+  private buildComponentInjector(metadata: ModalMetadata, popoverRef: PopoverRef): Injector {
+    return Injector.create({
+      providers: [
+        {
+          provide: MODAL_DATA,
+          useValue: metadata.config.data
+        },
+        {
+          provide: ModalRef,
+          useValue: new ModalRef(popoverRef)
+        }
+      ],
+      parent:   metadata.parentInjector});
+  }
 }
+
+export interface ModalMetadata {
+  config: ModalOverlayConfig;
+  parentInjector: Injector;
+}
\ No newline at end of file

```

arjunlalb: I'm working with these suggestions, but I'm kind of confused where to inject the modalRef. In the changes you suggested, modalRef is created twice. One in modal service (that is returned to consumer) and the other in modalOverlay. Any recommendations? @aaron-steinfeld

aaron-steinfeld: Sorry about, confused me too looking back at it, and it may need some tweaks, but here's how I got there:

The pattern we used for popover itself can be seen in popover service - first create the ref, then create the underlying component and use the result of that to initialize the ref before returning it. The one gotcha in this case is that since we're sending the ref both down to the underlying component and up to the caller, we'd want to make sure it's initialized before either receives it. Some ways we could do that:
- Have the service initialize it before returning it. In this scenario, the modaloverlay needs to know when it's "ready for use". That could be done for example by giving the modaloverlay an observable of the modalref (or even of the whole modal injector) and waiting for that before instantiating the modal. The popover would provide that once it had the popover ref back in hand.
- Give it back in an observable. This is basically the role reversal of the above, with the modaloverlay doing the initialization and at that point the service emits the ref. This would change the API of the modal service though, since it can't block waiting for the overlay to emit.
- Have two instances of modalRef. I can't really think of any cases where object equality would matter as long as there's no issues that could lead to them becoming out of sync (which I don't think there are, but worth testing - the observer invocations would likely need to be updated from my example and wrapped with checks of the closed flag to ensure complete isn't called by both ModalRefs. This is basically what I did above, saving us from having to do a back and forth between the service and overlay at the price of having two copies of the ref.

Doing a bit more in depth thinking, I think the first option is probably the best bet, but up to you. If you have questions on it, feel free to ping me.

PR #217, opened 2020-09-24
refactor: add parsers to new filtering code

jake-bassett:


projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
```diff
@@ -0,0 +1,66 @@
+import { FilterAttribute } from '../../filter-attribute';
+import { FilterAttributeType } from '../../filter-attribute-type';
+import { FilterOperator } from '../../filter-operators';
+import { ParsedFilter } from '../parsed-filter';
+
+export abstract class AbstractFilterParser<TValue> {
+  public static supportedAttributeTypes: FilterAttributeType[];
+  public static supportedOperators: FilterOperator[];
+
+  protected abstract supportedAttributeTypes(): FilterAttributeType[];
```
aaron-steinfeld, line 10: Seems strange to have this both as an instance method and static property.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
```diff
@@ -0,0 +1,66 @@
...
+      .find(op => filterString.includes(op));
+
+    if (matchingOperator === undefined) {
+      return undefined;
+    }
+
+    const parts = filterString
+      .split(matchingOperator)
+      .map(str => str.trim())
+      .filter((part: string | undefined) => (part ?? '').length > 0);
```
aaron-steinfeld, line 54: how could this be undefined? we invoke trim the line above, so we'd have already errored out

projects/components/src/filtering/filter/parser/types/comparison-filter-parser.ts
```diff
@@ -0,0 +1,62 @@
+import { assertUnreachable } from '@hypertrace/common';
+import { isBoolean } from 'lodash-es';
+import { FilterAttribute } from '../../filter-attribute';
+import { FilterAttributeType } from '../../filter-attribute-type';
+import { FilterOperator } from '../../filter-operators';
+import { FilterParser } from '../filter-parser.decorator';
+import { AbstractFilterParser } from './abstract-filter-parser';
+
+@FilterParser({
+  supportedAttributeTypes: [FilterAttributeType.Boolean, FilterAttributeType.Number, FilterAttributeType.String],
```
aaron-steinfeld, line 10: How do we match parsers? Not all of these attribute types work with all of these operators.

jake-bassett: Actually going through that part of the code now as I build the lookup service. My thought at the moment is to do it by operator.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
jake-bassett: This is something I'm not thrilled about either. Its kind of confusing to the reader. The statics in the abstract class are never set. They are strictly there for the decorator to populate for the subclasses.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
jake-bassett: Ah, true. I'll fix locally for now unless you want this addressed before merge.

aaron-steinfeld: No need

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
aaron-steinfeld: I see in this change how they're used at an instance level - are they also being used at a static level? What's being done with the decorater-set values?

projects/components/src/filtering/filter/parser/types/comparison-filter-parser.ts
aaron-steinfeld: hmm - we can merge this as is, but may have questions once we get to the lookup part. Curious how the static values are used.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
jake-bassett: You might be on to something here. This whole statics model with the builders (and now parsers) may not be needed anymore with the way I'm doing the lookups this time. I carried over the statics from the old code, but I don't think I'm going to need them.

Can we keep them for now, and once I've done the full end to end and determined I don't need them anymore I'll remove them?

jake-bassett: Take that back. They weren't even in the old code. I'm 99% sure I won't need them. I'll remove them next PR.

jake-bassett: Now that I'm looking at how I'm doing the lookup service, I won't actually need those. I'll remove them. Is it okay to do so in the next PR?

PR #223, opened 2020-09-28
Update batching and merging

anandtiwary:


projects/graphql-client/src/graphql.module.ts
```diff
@@ -1,29 +1,27 @@
...
-}
-
 @NgModule({
-  imports: [HttpLinkModule, ApolloModule],
   providers: [
     {
       provide: APOLLO_OPTIONS,
-      useFactory: createApollo,
-      deps: [HttpLink, GRAPHQL_URI]
+      useFactory: (httpLink: HttpBatchLink, uri: string, batchSize: number) => ({
```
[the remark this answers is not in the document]
anandtiwary: Yep it does support now

projects/graphql-client/src/graphql.module.ts
```diff
@@ -1,29 +1,27 @@
...
+        cache: new InMemoryCache(),
```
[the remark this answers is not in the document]
anandtiwary: `dataIdFromObject` is deprecated. InMemoryCache now supports Type policies for defining cache strategies. By default, it would use TypeName + data id to store the object and would take into account the request params to determine whether it needs to refetch or it can serve from cache. We can set different policies for each Gql Request. I plan to add that, but not as part of this PR.

For an array, it will use request and type name and  `[]` to serve from cache. So if request changes, previous cache would not be used.

projects/graphql-client/src/utils/builders/request/graphql-request-builder.ts
```diff
@@ -24,23 +27,35 @@ export class GraphQlRequestBuilder {
...
-        const valueAsString = isEmpty(value.childFragment)
-          ? ''
-          : ` ${this.translateMergedRequestToGql(value.childFragment)}`;
-        const keyAsString = this.convertKeyToFieldExpression(key, value);
-        const argString = this.convertArgumentMapToString(value.argMap);
-
-        return `${keyAsString}${argString}${valueAsString}`;
+  public buildSelectionToGqlRequestStringMap(): Map<GraphQlSelection, string> {
+    return new Map(
+      this.originalSelections.map(selection => {
```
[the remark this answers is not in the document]
anandtiwary: Yes. So to avoid complexity (and to use Apollo's batching), I am sending each selection in its own gql request. The HttpBatchLink batches these selections into individual http request according to our config and returns the results seamlessly.

projects/graphql-client/src/graphql.module.ts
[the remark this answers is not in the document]
aaron-steinfeld: We can optimize caching as a later PR as long as things are still working here. Type name and data id is what it originally used which is not sufficient because we request different attributes for the same entity in different requests. I'm certainly ok with this as long as it's working but please pay extra careful attention to making sure requests with different arguments, different selections etc. are all working for the same entity.

projects/graphql-client/src/utils/builders/request/graphql-request-builder.ts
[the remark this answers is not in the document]
aaron-steinfeld: It makes sense to send each query as a separate request and use apollo's batching, but by keying off the original request aren't we negating all merging? I would think we could use the previous logic almost identically, but rather than returning a string return an array of strings (since we know all the top level keys of the merged result in the request build, we have the info we need to split these into appropriate queries)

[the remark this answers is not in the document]
aaron-steinfeld: Or is this returning identical queries for many different selections, and apollo is collapsing them?

projects/graphql-client/src/graphql.module.ts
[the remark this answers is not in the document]
anandtiwary: Yes. So If we request a new attribute and apollo can't find it in the cache for the given object, it would fire a new request, update the cache and then server the response. I have tested few scenarios like this. Didn't find any issues yet.

projects/graphql-client/src/utils/builders/request/graphql-request-builder.ts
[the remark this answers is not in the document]
anandtiwary: The original selection is only used for response parsing. For example, The TraceHandler requests both trace and span query using a selection map. So this given request has two selection. By maintaining original selection to merged request fragment mapping, we ensure that responses for both merged selection are passed on to the gql extractor for further processing.

In the Graphql Request service, we extract values array from this map, and we only intend to pass unique values to apollo. I just realized that I missed the `unique` call there. I will add it.

projects/graphql-client/src/graphql-config.ts
```diff
@@ -47,3 +46,9 @@ export const enum GraphQlHandlerType {
   Mutation = 'mutation',
   Query = 'query'
 }
+
+export interface GraphQlOptions {
+  uri: string;
+  batchSize?: number;
+  batchDuration: number;
```
aaron-steinfeld, line 53: misleading name - maybe `batchDebounceTimeMs` ? Also doesn't look to be required since we're providing a default.

src/app/config.module.ts
```diff
@@ -10,12 +10,11 @@ import { FeatureResolverModule } from './shared/feature-resolver/feature-resolve
...
-      provide: GRAPHQL_URI,
-      useValue: environment.graphqlUri
-    },
-    {
-      provide: GRAPHQL_BATCH_SIZE,
-      useValue: environment.graphqlUri
+      provide: GRAPHQL_OPTIONS,
+      useValue: {
+        uri: environment.graphqlUri,
+        batchSize: environment.graphqlBatchSize
```
aaron-steinfeld, line 16: bath size can probably just be hardcoded here - I don't expect it to be something that diverges prod vs dev.

projects/distributed-tracing/src/shared/dashboard/data/graphql/graphql-data-source.model.ts
```diff
@@ -53,7 +53,6 @@ export abstract class GraphQlDataSourceModel<TData> implements DataSource<TData>
 
     this.querySubject.next({
       buildRequest: this.convertToBuilder(requestOrBuilder),
-      isolated: true,
```
aaron-steinfeld, line 56: This does mean we'll be batching certain things that we hadn't batched before we removed batching the first time - such as topology, which has always been isolated.

projects/observability/src/pages/explorer/explorer.component.test.ts
```diff
@@ -115,11 +115,17 @@ describe('Explorer component', () => {
...
     patchRouterNavigateForTest(spectator);
     detectQueryChange();
-    querySpy = spectator.inject(GraphQlRequestService).queryImmediately;
+    querySpy = spectator.inject(GraphQlRequestService).query;
   };
 
   test('fires query on init for traces', fakeAsync(() => {
-    init();
+    init({
+      providers: [
```
aaron-steinfeld, line 123: If we're already doing on 75 why do we need to override in each test?

projects/graphql-client/src/graphql-request.service.test.ts
```diff
@@ -90,159 +154,143 @@ describe('GraphQl Request Service', () => {
...
+    spectator.service.registerHandler(buildMutationHandlerOne());
     const mutateFnMock = mockMutateMethod();
-    spectator.service.mutate(buildRequest());
+    spectator.service.mutate(buildRequestOne());
     expect(mutateFnMock).toHaveBeenCalled();
   });
 
-  test('collects and fires debounced quries together', fakeAsync(() => {
-    spectator.service.registerHandler(buildQueryHandler());
+  test('collects and fires different requests separately. Apollo bataches them', fakeAsync(() => {
```
aaron-steinfeld, line 171: typo
```suggestion
  test('collects and fires different requests separately. Apollo batches them', fakeAsync(() => {
```

projects/graphql-client/src/graphql-request.service.test.ts
```diff
@@ -90,159 +154,143 @@ describe('GraphQl Request Service', () => {
...
+    spectator.service.registerHandler(buildQueryHandlerOne());
+    spectator.service.registerHandler(buildQueryHandlerTwo());
+    const queryFnMock = mockQueryMethod();
+    spectator.service.query(buildRequestOne()).subscribe();
+    spectator.service.query(buildRequestTwo()).subscribe();
+    tick(debounceTimeMs);
```
aaron-steinfeld, line 177: should verify `queryFnMock` not called before tick, otherwise a non-batched implementation would pass here.

projects/graphql-client/src/graphql-request.service.test.ts
```diff
@@ -90,159 +154,143 @@ describe('GraphQl Request Service', () => {
...
+    spectator.service.registerHandler(buildQueryHandlerOne());
     const queryFnMock = mockQueryMethod();
-    spectator.service.queryDebounced(buildRequest()).subscribe();
-    spectator.service.queryDebounced(buildRequest()).subscribe();
+    spectator.service.query(buildRequestOne()).subscribe();
+    spectator.service.query(buildRequestOne()).subscribe();
     expect(queryFnMock).not.toHaveBeenCalled();
     tick(debounceTimeMs);
     expect(queryFnMock).toHaveBeenCalledTimes(1);
+    expect(queryFnMock).toHaveBeenCalledWith({
```
aaron-steinfeld, line 208: this assertion isn't wrong but not relevant to this particular test (request uniquness)

projects/graphql-client/src/graphql-request.service.test.ts
```diff
@@ -90,159 +154,143 @@ describe('GraphQl Request Service', () => {
...
+    const queryFnMock = mockQueryMethod();
+    tick(debounceTimeMs);
+    expect(queryFnMock).toHaveBeenCalledWith({
+      query: gql(
+        new GraphQlRequestBuilder()
+          .withSelects({ path: 'testone', children: [{ path: 'id' }, { path: 'value' }] })
+          .build()
+      ),
+      errorPolicy: 'all',
+      fetchPolicy: 'no-cache'
```
aaron-steinfeld, line 251: In general, we want to assert what we're testing for in this specific test. the rest is just noise that needs to all be updated on each change. Here for example, we're testing that the request option is being propagated so we could just say
```typescript
    expect(queryFnMock).toHaveBeenCalledWith(expect.objectContaining({fetchPolicy: 'no-cache'}));`
```

projects/graphql-client/src/graphql-request.service.test.ts
```diff
@@ -90,159 +154,143 @@ describe('GraphQl Request Service', () => {
...
+      mutation: gql(
+        `mutation ${new GraphQlRequestBuilder()
+          .withSelects({ path: 'testone', children: [{ path: 'id' }, { path: 'value' }] })
+          .build()}`
+      )
+    });
+    expect(handler.getRequestOptions).toHaveBeenCalledTimes(1);
   }));
 
-  test('resubscribing to an immediate query triggers an immediate new request', () => {
```
aaron-steinfeld, line 193: Where did the new request on new subscription tests go? These should still be verified with a mocked apollo.

projects/graphql-client/src/graphql-config.ts
anandtiwary: cool

src/app/config.module.ts
anandtiwary: bath size. Yeah!

projects/distributed-tracing/src/shared/dashboard/data/graphql/graphql-data-source.model.ts
anandtiwary: and this would be fine wouldn't it?

projects/observability/src/pages/explorer/explorer.component.test.ts
anandtiwary: I have seen sometimes they keep record of calls made by other tests. Providing in the test fixes the issue. It is weird and not sure if i am missing something.

projects/graphql-client/src/graphql-request.service.test.ts
anandtiwary: Pushing another commit with these tests. Almost done

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -0,0 +1,243 @@
...
+  //   spectator.expectOne(graphQlUri, HttpMethod.POST).flush(modifiedResponse);
+  //   // Now make the first request again - it should be the original result
+  //   spectator.service.queryImmediately(buildRequest()).subscribe(result => {
+  //     expect((result as { value: string }).value).toEqual('bar');
+  //   });
+  //   // Tick, our responses are always async
+  //   tick();
+  // }));
+
+  // test('late subscribers to a debounced query can access results', fakeAsync(() => {
```
anandtiwary, line 193: This test fails now. Resubscribing do not fire a new query for observable returned by apollo. Do we need this behavior?

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -0,0 +1,243 @@
...
+  //   expect(spectator.expectOne(graphQlUri, HttpMethod.POST).request.body).toEqual([
+  //     {
+  //       variables: {},
+  //       query: '{\n  test {\n    id\n    value\n    __typename\n  }\n}\n'
+  //     }
+  //   ]);
+  //   flush();
+  // }));
+
+  // test('resubscribing to a query triggers an immediate new request', fakeAsync(() => {
```
anandtiwary, line 216: This test fails now for the apollo observable. Do we really need this?

projects/graphql-client/src/graphql-request.service.test.ts
```diff
@@ -90,159 +154,208 @@ describe('GraphQl Request Service', () => {
...
     tick(debounceTimeMs);
-    spectator.expectOne(graphQlUri, HttpMethod.POST);
+    expect(queryFnMock).toHaveBeenNthCalledWith(3, {
+      query: gql(requestString),
+      errorPolicy: 'all',
+      fetchPolicy: 'no-cache'
+    });
   }));
+
+  // test('resubscribing to a mutation triggers an immediate new request', fakeAsync(() => {
```
anandtiwary, line 340: This test fails for mutation. I am suspecting some changes in the way apollo creates response observable. For query it still works probably because of the `defer`. Do we need this behavior?

anandtiwary:
kept some failing tests as commented code so that i can get feedback on it.

projects/graphql-client/src/graphql-request.service.test.ts
aaron-steinfeld: the general family of resubscribe tests are to ensure the observable characteristics (in this case, being a cold unicast observable) of our service match those of a "standard" observable like those returned by httpclient. If we're mocking out apollo, since that's where the observable originates, I agree this test doesn't make a whole lot of sense (and you're correct in the wrapping `defer` being the reason this still works in the debounced query case, as we need to change the underlying characteristics to share the results there).

So we do want this behavior, but we don't need to test it here - that belongs in the int test.

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -0,0 +1,243 @@
...
+  //   spectator.expectOne(graphQlUri, HttpMethod.POST).flush(modifiedResponse);
+  //   // Now make the first request again - it should be the original result
+  //   spectator.service.queryImmediately(buildRequest()).subscribe(result => {
+  //     expect((result as { value: string }).value).toEqual('bar');
+  //   });
+  //   // Tick, our responses are always async
+  //   tick();
+  // }));
+
+  test('late subscribers to a debounced query can access results', fakeAsync(() => {
```
aaron-steinfeld, line 193: this is a feature of query service not the integration with apollo. Right now, this test is effectively the same thing as `only fires queries that are subscribed` above

projects/graphql-client/src/graphql-request.service.test.ts
anandtiwary: Okay.. I noticed it failed in the int test as well. I think apollo is not triggering it on resubscribe. Do you think I am missing something in the test?

projects/graphql-client/src/graphql-apollo-integration.test.ts
aaron-steinfeld: I'd like to get it working but we can drop if it becomes problematic. From looking at the commented out test, I'd suspect the issue is that we're never completing the first request. See from other tests:
```
    // Flush first request out, otherwise apollo deduping will always reuse an identical in flight query
    spectator.expectOne(graphQlUri, HttpMethod.POST).flush(buildServerResponse());
```

projects/graphql-client/src/graphql-request.service.test.ts
aaron-steinfeld: See comment on the int test - believe the issue is a missing response flush.

projects/graphql-client/src/graphql-request.service.test.ts
anandtiwary: Yeah. But i wanted to test if errorpolicy is also present.

projects/graphql-client/src/graphql-apollo-integration.test.ts
anandtiwary: thanks!

projects/graphql-client/src/graphql-request.service.test.ts
anandtiwary: So my test was wrong. Resubscribing wouldn't trigger mutate call again. The underlying observable returned from apollo would call http. But I can't test that here. Removing the test now

projects/distributed-tracing/src/shared/dashboard/data/graphql/graphql-data-source.model.ts
aaron-steinfeld: Those were initially meant for queries known to be expensive, which is why we unrolled the batching in the first place. The flow for example requires multiple lower level queries so is expected to be slower.

projects/graphql-client/src/graphql-request.service.test.ts
aaron-steinfeld: agreed - the test in integration test should be sufficient.

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -24,14 +24,23 @@ describe('GraphQl Apollo Integration Service', () => {
...
   });
 
-  const buildRequestString = (rootPath: string = 'test'): string =>
-    new GraphQlRequestBuilder().withSelects({ path: rootPath, children: [{ path: 'id' }, { path: 'value' }] }).build();
+  const buildRequestString = (rootPath: string = 'test', params: string[] = []): string =>
+    new GraphQlRequestBuilder()
+      .withSelects({
+        path: rootPath,
+        children: [{ path: 'id' }, { path: 'value' }],
+        ...params.map(param => ({ path: param }))
```
aaron-steinfeld, line 32: should these be inside the children array? or am I misreading it

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -169,28 +178,7 @@ describe('GraphQl Apollo Integration Service', () => {
     flush();
   }));
 
-  // test('does not cache entities, even if they share an id', fakeAsync(() => {
```
aaron-steinfeld: what happened to this test? were we unable to make it work?

projects/observability/src/pages/explorer/explorer.component.test.ts
aaron-steinfeld: hmm - will try to pull and see if I can figure out why that is, because we shouldn't need to do this. don't hold for this though, since I might forget...

projects/distributed-tracing/src/shared/dashboard/data/graphql/graphql-data-source.model.ts
anandtiwary: We can have query function accept the `isolated` property or may be the entire request option that includes cacheability. Then root data source would add it to RequestEvent and emit. Query event service would then pass it on to graphql request service. This would give more configuration options for us at data source level.

anandtiwary: how does this sound?

aaron-steinfeld: Sounds right I think.

projects/graphql-client/src/graphql-apollo-integration.test.ts
anandtiwary: This is not a valid test. Why would we want to disable caching for this fetch by id? The default behavior would merge the two response if we ask for new fields. Then only time it will refetch is when the request param like time, filters etc changes.

anandtiwary: That seems correct behavior to me. So why do we not want to cache them?

projects/graphql-client/src/graphql-apollo-integration.test.ts
anandtiwary: aah.. I added this for the other test which i removed https://github.com/hypertrace/hypertrace-ui/pull/223#discussion_r499072296

I can remove this depending up your comments on the other comment.

projects/graphql-client/src/graphql-apollo-integration.test.ts
aaron-steinfeld: if they're requested at different times (hence the queryImmediately behavior in existing test) they wouldn't merge - so it's testing what you're describing, that it refetches on a param change.

anandtiwary: Yes. But in the test we also fire the old request again and expect to get the previously cached data for the same id. Since Apollo got newer data for the id, old info will be overwritten. Only this part of the test doesn't seem valid to me. 

```
  modifiedResponse.data.test.value = 'baz';
     spectator.expectOne(graphQlUri, HttpMethod.POST).flush(modifiedResponse);
     // Now make the first request again - it should be the original result
     spectator.service.queryImmediately(buildRequest()).subscribe(result => {
      expect((result as { value: string }).value).toEqual('bar');
    });
    // Tick, our responses are always async//   modifiedResponse.data.test.value = 'baz';
     spectator.expectOne(graphQlUri, HttpMethod.POST).flush(modifiedResponse);
    // Now make the first request again - it should be the original result
    spectator.service.queryImmediately(buildRequest()).subscribe(result => {
      expect((result as { value: string }).value).toEqual('bar');
     });
    // Tick, our responses are always async
```

anandtiwary: Okay hmm.. I misunderstood the test.

anandtiwary: Yes. We need to change the response for this test though.  Apollo looks at result set and caches it according to the object type and defined type policies. For an array response, default behavior is to keep request-response pair as unique. So any change in request will trigger new http call. It achieves this by setting `[]` as the id for array response. Then it goes inside the array, and caches each individual object according to defined type policies or by simply looking for presence of `id`(default). 

When we make a fetchById query, we will only pass its id in the argument and nothing else. E.g `actor(id){...fields}`.
Apollo would use this id to look for cached data, compare and merge and use it for future request.

In the current test, we are returning an object with id for a query that has not id in argument. 

Fixed this by returning an array to test first scenario. Added another test for second scenario.

projects/distributed-tracing/src/shared/dashboard/data/graphql/graphql-data-source.model.ts
anandtiwary: Done. We can parameterize cacheability and isolated properties on root ds. For now, I am only keeping it as function param.

projects/graphql-client/src/graphql-apollo-integration.test.ts
anandtiwary: should be fixed.

projects/graphql-client/src/graphql-apollo-integration.test.ts
anandtiwary: Done

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -24,28 +24,40 @@ describe('GraphQl Apollo Integration Service', () => {
...
   });
 
-  const buildRequestString = (rootPath: string = 'test', params: string[] = []): string =>
+  const buildRequestString = (rootPath: string = 'test', params: string[] = [], fields: string[] = []): string =>
     new GraphQlRequestBuilder()
       .withSelects({
-        path: rootPath,
-        children: [{ path: 'id' }, { path: 'value' }],
-        ...params.map(param => ({ path: param }))
+        path: `${rootPath}${params.length > 0 ? `(${params.map(param => `${param}: ${param}`).join(',')})` : ''}`,
```
aaron-steinfeld, line 30: By params I assume we mean arguments? Shouldn't they use the arguments portion of the selection?

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -24,28 +24,40 @@ describe('GraphQl Apollo Integration Service', () => {
...
+
+  const buildServerResponse = (rootPath: string = 'test', fields: string[] = []) => ({
     data: {
       [`${rootPath}`]: {
         id: 'foo',
         value: 'bar',
-        ...params
-          .map(param => ({ [`${param}`]: param }))
-          .reduce((previousVal, current) => ({ ...previousVal, ...current }), {}),
+        ...fields.map(field => ({ [`${field}`]: field })),
```
aaron-steinfeld, line 43: This appears to be spreading an array into an object.

projects/graphql-client/src/graphql-apollo-integration.test.ts
```diff
@@ -24,28 +24,40 @@ describe('GraphQl Apollo Integration Service', () => {
...
         __typename: 'test-type'
       }
     }
   });
 
+  const buildServerArrayResponse = (rootPath: string = 'test') => ({
+    data: [
+      {
+        [`${rootPath}`]: {
```
aaron-steinfeld, line 52:
```suggestion
        [rootPath]: {
```

projects/graphql-client/src/graphql-apollo-integration.test.ts
anandtiwary: oh hmm.. I want them to be added as properties. I need to reduce it into an object before spreading

projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
```diff
@@ -83,23 +84,30 @@ export class TopologyDataSourceModel extends GraphQlDataSourceModel<TopologyData
   public edgeMetricSpecifications: MetricAggregationSpecification[] = [];
 
   private readonly specBuilder: SpecificationBuilder = new SpecificationBuilder();
+  private readonly requestOptions: GraphQlRequestOptions = {
+    cacheability: GraphQlRequestCacheability.Cacheable,
```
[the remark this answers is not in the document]
anandtiwary: So, with a future pr, I think we can expose cacheability and isolated as properties on root data source model. Then we can control it through json. If you agree I would prefer to do it separately.

aaron-steinfeld (approved):

PR #225, opened 2020-10-01
feat: updating link to use routerlink

anandtiwary:
- this will convert anchor tag into true links


anandtiwary:
I will be using this on Security/Threat card links. But we need to decide if we want to replace all current click -> navigate calls in the product with this link component or not.

anandtiwary:
I have to make one more change. Marked it ready for review a bit too early.

anandtiwary:
> I have to make one more change. Marked it ready for review a bit too early.

Actually, take a look at this usage https://github.com/Traceableai/traceable-ui/pull/1374/files#r497911929 

Let me know if this api is good or not.

projects/components/src/link/link.component.ts
```diff
@@ -1,45 +1,41 @@
...
   selector: 'ht-link',
   styleUrls: ['./link.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <a *ngIf="this.shouldShowLink" class="ht-link" (click)="this.navigateToUrl()">
+    <a
+      *ngIf="this.navigationParams"
+      class="ht-link"
+      [routerLink]="this.navigationParams[0]"
+      [queryParams]="this.navigationParams[1].queryParams"
```
arjunlalb, line 15: how do we want to handle query params?

Currently, the only input the component takes is the URL. if we expect the URL to be inclusive of query param string, then we have to extract them for this to function. Else, we can just use the URL string and this is redundant.

The downstream method `buildInternalNavigationParams` consumes queryparams only from navigationExtras argument which we are not supplying in L38.

projects/common/src/navigation/navigation.service.ts
```diff
@@ -60,6 +61,53 @@ export class NavigationService {
     return this.currentParamMap.getAll(parameterName);
   }
 
+  public navigateToUrl(url: string): Observable<boolean> {
```
aaron-steinfeld, line 64: I feel like adding these to our current nav is redundant - do we think it would be feasible to condense them while adding this support in a very similar way to what's currently being done?

Current list of public methods (that would be changed)
```
public navigateWithinApp(path: NavigationPath, navigationExtras?: WithinAppNavigationExtras): Observable<boolean>
public navigateExternal(url: string, navigationType: NavigationType = NavigationType.SameWindow): Observable<boolean>
public buildParamsForNavigation(preserveParameters: string[] = []): QueryParamObject;
```
proposed:
```
public navigate(url: string): Observable<boolean>; // Adding overload, would just call convertUrlToNavParams first
public navigate(params: NavigationParams): Observable<boolean>;
public convertUrlToNavParams(url: string): NavigationParams;
private buildQueryParams(preserveParameters: string[] = []): QueryParamObject; // Renamed and private

type NavigationParams = InAppNavigationParams | ExternalNavigationParams;
interface InAppNavigationParams {
  navType: 'in-app'
  path: NavigationPath;
  replaceCurrentHistory?: boolean;
  queryParams?: QueryParamObject;
  queryParamsHandling?: 'merge' | 'preserve';
  relativeTo?: ActivatedRoute;
}

interface ExternalNavigationParams {
  navType: 'external'
  url: string;
  windowHandling: ExternalNavigationWindowHandling; // Currently an enum called NavigationType
  queryParams?: QueryParamObject;
}

```

projects/common/src/navigation/navigation.service.ts
```diff
@@ -60,6 +61,53 @@ export class NavigationService {
...
+    const navigationParams = this.buildNavigationParams(url)!;
+
+    return from(this.router.navigate([navigationParams[0]], navigationParams[1]));
+  }
+
+  public buildNavigationParams(url: string): [NavigationPath, NavigationExtras] | undefined {
```
aaron-steinfeld, line 70: should never return undefined IMO - I'd call this an error situation.

aaron-steinfeld: Also, detailed above, I think returning a tuple publicly is a confusing design - would love to condense down to an object.

projects/components/src/link/link.component.ts
aaron-steinfeld: Given my proposal above, this input could accept a `string | NavigationParams` since that's what the navigate api would accept.

projects/common/src/navigation/navigation.service.ts
anandtiwary: okay looking

projects/components/src/link/link.component.ts
anandtiwary: When I tested this, url had the path and query params. See an example here. https://github.com/Traceableai/traceable-ui/pull/1374/files#r498031200

So, lets just have two inputs in this component, path: NavigationPath (which can be string or array) and params: NavigationParams.

aaron-steinfeld: the params as I described above would already self contain both the path and query params

projects/common/src/navigation/navigation.service.ts
anandtiwary: So i feel like I want to take this up separately. The `navigateToUrl` is not needed for my current change but I thought I might as well add it. It was poorly designed. The purpose of current pr is only to build angular router param and use it with routerlink.

I am going to remove this method for now and will create a separate story/pr with your suggested cleanup.

projects/common/src/navigation/navigation.service.ts
anandtiwary: okay.

projects/components/src/link/link.component.test.ts
```diff
@@ -1,77 +1,77 @@
...
+          })
+        })
+      ]
     });
 
-    expect(hostSpectator.query('.ht-link')).toExist();
-    hostSpectator.triggerEventHandler('.ht-link', 'click', {});
-    expect(hostSpectator.inject(NavigationService).navigateWithinApp).toHaveBeenCalledWith(props.url);
-  }));
+    expect(spectator.query('.ht-link')).toExist();
```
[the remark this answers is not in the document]
anandtiwary: updated the test. The other PR for new api is also almost ready. I can open it up once this is in.

aaron-steinfeld (approved):

PR #226, opened 2020-10-01
refactor: clean up filtering in prep for new string map syntax

jake-bassett:


projects/components/src/filtering/filter/parser/parsed-filter.ts
```diff
@@ -0,0 +1,13 @@
+import { FilterOperator } from '../filter-operators';
+
+export interface ParsedFilter<TValue> {
+  field: string;
+  operator: FilterOperator;
+  value: TValue;
+}
+
+export interface SplitFilter<TOperator extends string> {
+  lhs: string;
```
aaron-steinfeld, line 10: What's the thinking on how this will support a lhs with a dotted path for CONTAINS_KEY_VALUE?

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
```diff
@@ -0,0 +1,94 @@
...
+    // Successfully parsed URL filter
+
+    return {
+      field: attribute.name,
+      operator: splitFilter.operator,
+      value: parsedValue
+    };
+  }
+
+  public static splitFilterStringByOperator<TOperator extends string>(
```
aaron-steinfeld, line 72: Why is this public static? Is it because we can't inject here but want to share? Can we either switch to injection to share it, or move to a stand alone function?

projects/components/src/filtering/filter/filter.service.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(navigationService.addQueryParametersToUrl).toHaveBeenCalledWith({
+      filter: undefined
+    });
+  });
+
+  test('correctly sets filters in url', () => {
+    spectator.service.applyUrlFilter(attributes, filters[0]);
+    spectator.service.applyUrlFilter(attributes, filters[1]);
+
+    expect(navigationService.addQueryParametersToUrl).toHaveBeenCalledWith({
```
aaron-steinfeld, line 85: shouldn't this be two calls? Is this array value because this is the original params anyway?

projects/components/src/filtering/filter/filter.service.ts
```diff
@@ -0,0 +1,205 @@
...
+      [FilterService.FILTER_QUERY_PARAM]: filters.length === 0 ? undefined : filters.map(f => f.urlString)
+    });
+  }
+
+  public applyUrlFilter(attributes: FilterAttribute[], filter: Filter): void {
+    const filters = this.getUrlFilters(attributes);
+
+    const foundIndex = filters.findIndex(f => filter.field === f.field);
+
+    if (foundIndex !== -1) {
```
aaron-steinfeld, line 52: Isn't this an explicit implementation of the behavior we currently have an outstanding bug for? we're replacing rather than adding if it's an existing field.

projects/components/src/filtering/filter/filter.service.ts
```diff
@@ -0,0 +1,205 @@
...
+      filters[foundIndex] = filter;
+    } else {
+      filters.push(filter);
+    }
+
+    this.setUrlFilters([...filters]);
+  }
+
+  public removeUrlFilter(attributes: FilterAttribute[], filter: Filter): void {
+    this.setUrlFilters([...this.getUrlFilters(attributes).filter(f => filter.field !== f.field)]);
```
aaron-steinfeld, line 62: this also should check the RHS to support multiple filters for the same field.

projects/components/src/filtering/filter/filter.service.ts
```diff
@@ -0,0 +1,205 @@
...
+        const supportedUrlOperators = filterBuilder.supportedOperators().map(toUrlFilterOperator);
+
+        const splitUrlFilter = AbstractFilterParser.splitFilterStringByOperator(
+          supportedUrlOperators,
+          filterString,
+          false
+        );
+
+        if (splitUrlFilter === undefined) {
+          return undefined;
```
aaron-steinfeld, line 79: does flatmap handle undefined return? would expect empty array

projects/components/src/filtering/filter/filter.service.ts
```diff
@@ -0,0 +1,205 @@
...
+
+        const filterParser = this.filterParserLookupService.lookup(fromUrlFilterOperator(splitUrlFilter.operator));
+
+        const parsedFilter = filterParser.parseUrlFilterString(attribute, filterString);
+
+        if (parsedFilter === undefined) {
+          return undefined;
+        }
+
+        return splitUrlFilter.lhs === parsedFilter.field
```
aaron-steinfeld, line 90: What's the scenario that could lead to this not being true?

projects/components/src/filtering/filter/filter.service.ts
```diff
@@ -0,0 +1,205 @@
...
+  /*
+   * Autocomplete
+   */
+
+  public autocompleteFilters(attributes: FilterAttribute[], text: string = ''): IncompleteFilter[] {
+    return attributes
+      .filter(attribute => this.filterBuilderLookupService.isBuildableType(attribute.type))
+      .flatMap(attribute => this.autocompleteFilter(attribute, text))
+      .filter(incompleteFilter => {
+        const isStringMatch = incompleteFilter.userString.toLowerCase().includes(text.toLowerCase());
```
aaron-steinfeld, line 106: would autocomplete filter return filters that don't satisfy this condition? Assuming yes, otherwise the rest of the function is irrelevant - but based on the function names, would assume that this function is only mapping the singular version over each attribute

projects/components/src/filtering/filter/parser/types/contains-filter-parser.ts
```diff
@@ -0,0 +1,41 @@
...
+      case FilterAttributeType.String: // Unsupported
+      case FilterAttributeType.Timestamp: // Unsupported
+        return undefined;
+      default:
+        assertUnreachable(attribute.type);
+    }
+  }
+
+  private parseStringMapValue(valueString: string): string[] {
+    return valueString.split(ContainsFilterParser.CONTAINS_DELIMITER).map(String);
```
aaron-steinfeld, line 37: doesn't split return an array of strings

projects/components/src/filtering/filter/parser/types/comparison-filter-parser.ts
```diff
@@ -0,0 +1,61 @@
...
+      case FilterAttributeType.StringMap: // Unsupported
+      case FilterAttributeType.Timestamp: // Unsupported
+        return undefined;
+      default:
+        assertUnreachable(attribute.type);
+    }
+  }
+
+  private parseBooleanValue(valueString: string): boolean | undefined {
+    const val = valueString === 'false' ? false : valueString === 'true';
```
aaron-steinfeld, line 45:
```suggestion
    const val = valueString === 'true';
```
Equivalent and less confusing (also could lower case it to be more robust)

projects/components/src/filtering/filter/parser/types/comparison-filter-parser.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    return isBoolean(val) ? val : undefined;
```
aaron-steinfeld, line 47: When could val not be a boolean? 45 only returns a boolean unless above was meant to be a double ternary and go to undefined for no match.

projects/components/src/filtering/filter/parser/filter-parser-lookup.service.test.ts
```diff
@@ -0,0 +1,332 @@
...
+    expect(
+      spectator.service.isParsableOperatorForType(FilterOperator.ContainsKeyValue, FilterAttributeType.Timestamp)
+    ).toEqual(false);
+  });
+
+  test('can provide a parser to parse user filter strings', () => {
+    expect(
+      spectator.service
+        .lookup(FilterOperator.Equals)
+        .parseFilterString(getTestFilterAttribute(FilterAttributeType.Boolean), 'Boolean Attribute = false')
```
aaron-steinfeld, line 179: this isn't really testing the lookup service any more

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
```diff
@@ -0,0 +1,94 @@
...
+  public abstract supportedAttributeTypes(): FilterAttributeType[];
+  public abstract supportedOperators(): FilterOperator[];
+
+  protected abstract parseValueString(
+    attribute: FilterAttribute,
+    splitFilter: SplitFilter<FilterOperator>
+  ): TValue | undefined;
+
+  protected parseKeyString(attribute: FilterAttribute, splitFilter: SplitFilter<FilterOperator>): string | undefined {
+    return attribute.displayName.toLowerCase() !== splitFilter.lhs.toLowerCase() ? undefined : attribute.name;
```
aaron-steinfeld, line 16: we use display for key? could we match either?

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  public static splitFilterStringByOperator<TOperator extends string>(
+    possibleOperators: TOperator[],
+    filterString: string,
+    expectSpaceAroundOperator: boolean = true
+  ): SplitFilter<TOperator> | undefined {
+    const matchingOperator = possibleOperators
+      .sort((o1: string, o2: string) => o2.length - o1.length) // Sort by length to check multichar ops first
+      .map(op => (expectSpaceAroundOperator ? ` ${op as string} ` : op))
```
aaron-steinfeld, line 79: could we make this work without the flag and just collapse any potential whitespace below?

projects/components/src/filtering/filter/parser/types/contains-filter-parser.ts
jake-bassett: Yeah, this was copy paste. Had it fixed in subsequent PR. Fixed here now.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
jake-bassett: We are using the whitespace to help us find the operator and split on it without inadvertently splitting on operators that may be embedded in the value, for instance `=` is often in url values. The user string will always have operators surrounded by spaces, so we use that as our string split token.

The exception is for URL filters, which have our underscore version of the operator, like `_eq_`, thus the reason for needing the arg flag.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
jake-bassett: Yes, this is used in the follow up PR for the new dot syntax. We need to split in order to discover the operator and know which parser to use.

Will change to export const instead of static.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
jake-bassett: Correct, the User filters all work with Display Name. The URL filters work with name. We could probably match on either, but I haven't tested that thoroughly enough to know if it will work in all cases. Rather leave as is for now and expand the check later if we have a reason to.

projects/components/src/filtering/filter/filter.service.ts
jake-bassett: Correct. This is a check on each attribute to see if it matches the current user string. As the user types, the attributes are matched and eliminated here if they don't include the user entered text.

projects/components/src/filtering/filter/filter.service.ts
jake-bassett: The only case right now is the new string map dot syntax. It requires further parsing.

That said, this method gets refactored in the following PR since I choose to keep _URL_ filters the original syntax and this check goes away. It's a handful of changes that work in concert, so not worth pulling them in here, but I can call it out for review in the next PR.

projects/components/src/filtering/filter/filter.service.ts
jake-bassett: It just creates an undefined element. We then filter those all out as the final step of this chain in order to grab what we found to be valid.

projects/components/src/filtering/filter/filter.service.ts
jake-bassett: Thanks. Will follow up with the multi filter changes likely this week.

projects/components/src/filtering/filter/filter.service.ts
jake-bassett: Will follow up with those changes separately. Trying to keep this PR focused. Its big enough as is.

projects/components/src/filtering/filter/parser/parsed-filter.ts
jake-bassett: I have the following PR ready. Its not much beyond what is here, but since its done its probably easier to just show you the code.

aaron-steinfeld (approved):

PR #235, opened 2020-10-06
fix(table): multi select table row selection

skjindal93:


projects/components/src/table/table.component.ts
```diff
@@ -330,7 +330,7 @@ export class TableComponent
     }
 
     if (changes.selections) {
-      this.initializeRows();
+      this.toggleRowSelections(this.selections);
```
aaron-steinfeld, line 333: Could we add a test? Looks like this got broken by the col resize changes

skjindal93: Tried adding a test, but not even sure what to test on here. It broke because, `this.rowStateSubject.next(undefined);` was being called before `toggleRowSelections`, though no clue why that caused the bug

aaron-steinfeld: Wouldn't that undefined clear the rows? Could we test it by marking a selection, tick, then reading the selection?

skjindal93: Tried doing that, but `TableCdkDataSource` comes in between, with it's all sorts of `rowChange$`, `cachedRows` and everything. 

If we just check for the fact that clicking the check box selects (_emits_) the selections, that would pass irrespectively. Because, the check box was getting checked and getting unchecked automatically, due to `changes.selections` being triggered

It's a different question though as to why `changes.selections` is being triggered on checking a checkbox. That is because, the checkbox click changes the `selections` variable, instead of making a copy of it

aaron-steinfeld: > If we just check for the fact that clicking the check box selects (emits) it, that would pass irrespectively. Because, the check box was getting checked and getting unchecked automatically, due to changes.selections being triggered
Agree - I was thinking, unlike a normal test where we verify that after clicking it emits, that we verify after it emits the checkbox is still checked.

Anyway, will approve to get this fix out, don't need to hold it up while we figure this out.

aaron-steinfeld (approved):

PR #236, opened 2020-10-06
Update link api follow up

anandtiwary:


projects/common/src/navigation/navigation.service.ts
```diff
@@ -106,31 +102,20 @@ export class NavigationService {
...
-  }
-
-  /**
-   * Navigate to an external URL.
-   * URL should be of the form 'https://url' or 'http://url'
-   */
-  public navigateExternal(
-    url: string,
-    navigationType: NavigationType = NavigationType.SameWindow
+  public navigateWithinApp(
```
anandtiwary, line 105: @aaron-steinfeld  Should we rip this off as well? We kind of need a method that would build the NavigationParams from these three inputs

anandtiwary:
Will add more test in navigation service and link component after initial feedback

projects/components/src/navigation/navigation-list.component.ts
```diff
@@ -83,7 +83,7 @@ export class NavigationListComponent {
   }
 
   public navigate(item: NavItemLinkConfig): void {
-    this.navigationService.navigateWithinApp(item.matchPaths[0], { relativeTo: this.activatedRoute });
+    this.navigationService.navigateWithinApp(item.matchPaths[0], this.activatedRoute);
```
aaron-steinfeld, line 86: We could probably use ht-link for the main items too right? That way they get the real href behavior

projects/common/src/navigation/navigation.service.ts
```diff
@@ -106,31 +102,20 @@ export class NavigationService {
...
+    return this.navigate({
+      navType: NavigationParamsType.InApp,
+      path: path,
+      queryParams: this.buildQueryParam(preserveParameters ?? []),
+      relativeTo: relativeTo
+    });
   }
 
-  public buildParamsForNavigation(preserveParameters: string[] = []): QueryParamObject {
+  private buildQueryParam(preserveParameters: string[] = []): QueryParamObject {
```
aaron-steinfeld, line 118: nit: buildQueryParams (or buildQueryParamObject)

projects/common/src/navigation/navigation.service.ts
aaron-steinfeld: by rip off do you mean remove it? If so, I would agree - send everything through navigate. `NavigationParams` already supports the relative path. If the preserved params is still used we can add support for it, but I don't think it is from a quick search.

aaron-steinfeld (approved):

projects/common/src/navigation/navigation.service.ts
anandtiwary: Yeah. I wasn't sure about it either. I will put up a separate PR to remove this method and its usages.

projects/components/src/navigation/navigation-list.component.ts
anandtiwary: Yep. Didn't want to club it with this PR. @arjunlalb is also going to help with this.

projects/common/src/navigation/navigation.service.ts
anandtiwary: Yes for buildQueryParamObject but chose the shorter form. I can update it in the next pr.

PR #239, opened 2020-10-07
New tags syntax

jake-bassett:


projects/components/src/filtering/filter/parser/types/in-filter-parser.ts
```diff
@@ -40,7 +40,7 @@ export class InFilterParser extends AbstractFilterParser<PossibleValuesTypes> {
 
   private parseNumberArrayValue(valueString: string): number[] | undefined {
     const array = valueString.split(InFilterParser.IN_DELIMITER).map(str => {
-      const val = Number(str.trim());
+      const val = str.trim() === '' ? NaN : Number(str.trim());
```
aaron-steinfeld, line 43: nit: we already have number coercer (although it may have this same bug, didn't realize javascript was so useful in converting an empty string to zero for us) - should we reuse that here?

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
```diff
@@ -0,0 +1,130 @@
...
+export class FilterChipService {
+  public constructor(
+    private readonly filterBuilderLookupService: FilterBuilderLookupService,
+    private readonly filterParserLookupService: FilterParserLookupService
+  ) {}
+
+  public autocompleteFilters(attributes: FilterAttribute[], text: string = ''): IncompleteFilter[] {
+    return attributes
+      .filter(attribute => this.filterBuilderLookupService.isBuildableType(attribute.type))
+      .flatMap(attribute => this.autocompleteFilter(attribute, text))
```
aaron-steinfeld, line 23: I asked this in the other PR but don't recall the answer and have the same question again - if we're passing `text` into `autocompleteFilter()`, why would it ever return filters that don't match `text`?

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
```diff
@@ -0,0 +1,130 @@
...
+      .filter(incompleteFilter => this.filterMatchesUserText(text, incompleteFilter));
+  }
+
+  private filterMatchesUserText(text: string, incompleteFilter: IncompleteFilter): boolean {
+    const isStringMatch = incompleteFilter.userString.toLowerCase().includes(text.trim().toLowerCase());
```
aaron-steinfeld, line 28: should this be `startsWith` instead of `includes`?

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    if (incompleteFilter.operator === undefined) {
+      return isStringMatch;
```
aaron-steinfeld, line 31: If `isStringMatch` is true and operator is defined, what does that mean? When would a string match ever not be good enough? Or is this just meant to shortcircuit so lookup doesn't fail on the undefined? If that's the case, I'd shortcircuit on a string match too.

projects/components/src/filtering/filter/builder/types/string-map-filter-builder.ts
```diff
@@ -1,9 +1,12 @@
+import { collapseWhitespace } from '@hypertrace/common';
+import { FilterAttribute } from '../../filter-attribute';
 import { FilterAttributeType } from '../../filter-attribute-type';
 import { FilterOperator } from '../../filter-operators';
 import { AbstractFilterBuilder } from './abstract-filter-builder';
 
 export class StringMapFilterBuilder extends AbstractFilterBuilder<string | string[]> {
-  private static readonly DELIMITER: string = ':';
+  private static readonly STRING_MAP_LHS_DELIMITER: string = '.';
```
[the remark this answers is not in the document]
jake-bassett: Yes! I meant to do this. Thanks for calling me on it.

projects/components/src/filtering/filter/builder/filter-builder-lookup.service.test.ts
```diff
@@ -185,7 +185,7 @@ describe('Filter Builder Lookup service', () => {
       field: getTestFilterAttribute(FilterAttributeType.StringMap).name,
       operator: FilterOperator.ContainsKeyValue,
       value: ['myKey', 'myValue'],
-      userString: 'String Map Attribute CONTAINS_KEY_VALUE myKey:myValue',
+      userString: 'String Map Attribute.myKey CONTAINS_KEY_VALUE myValue',
```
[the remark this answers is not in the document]
jake-bassett: This is how a user will use the filter chip autocomplete though. Not sure I understand why this is wrong.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
jake-bassett: `include` works better here. If we have two attributes like `Service Name` and `API Name`, we want a user to be able to type "name" and see both options. They can then tab to autofill from the options.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
jake-bassett: The returned filters should always have a `userString` that _includes_ the user provided `text`. Additionally, once the entire string matches and the user hits `space`, they are also provided an option for each operator available to that attribute.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
jake-bassett: Yeah, this is a bit confusing. Back in simpler times, the `isStringMatch` could have been the only check. This changed with the new syntax. Going to add the following comment here:

```
    /*
     * In most cases, the isStringMatch should be the only check that is needed, however this check fails for
     * STRING_MAP types with an operator since the LHS includes a user entered key between the attribute name
     * and the operator. We fix this by sending the full string through our parsing logic to see if we can pull
     * the key value off the LHS and build a filter out of it. If so, its a valid match and we want to include
     * it as an autocomplete option.
     */
```

Let me know if that makes sense.

projects/components/src/filtering/filter/builder/filter-builder-lookup.service.test.ts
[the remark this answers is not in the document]
aaron-steinfeld: It's not wrong, I just think it's a lot weirder to have something like `My Pretty Display Name.somePath` vs `name.somePath` - it feels like we're mixing two things.

projects/components/src/filtering/filter/parser/types/contains-filter-parser.ts
```diff
@@ -33,9 +46,38 @@ export class ContainsFilterParser extends AbstractFilterParser<PossibleValuesTyp
...
   }
 
-  private parseStringMapValue(valueString: string): string[] {
-    return valueString.split(ContainsFilterParser.CONTAINS_DELIMITER);
+  private parseStringMapValue(
+    attribute: FilterAttribute,
+    splitFilter: SplitFilter<FilterOperator>
+  ): string[] | undefined {
+    if (splitFilter.lhs === attribute.displayName) {
+      return splitFilter.rhs.split(ContainsFilterParser.CONTAINS_RHS_DELIMITER);
```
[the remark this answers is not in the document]
aaron-steinfeld: I won't dig into this, but I think having two different syntaxes for this in different spots may become messy later.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
```diff
@@ -68,6 +68,10 @@ export abstract class AbstractFilterParser<TValue> {
       value: parsedValue
     };
   }
+
+  protected requiresNameParsing(_operator: FilterOperator): boolean {
```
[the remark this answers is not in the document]
aaron-steinfeld: It's defined here, and overridden there. But where is it actually called? If it's not used in the base class then it should be private to where it is used. That said, I don't see it being used in ContainsFilterParser either - maybe my browser text search is acting up though?

projects/components/src/filtering/filter/builder/filter-builder-lookup.service.test.ts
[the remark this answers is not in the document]
jake-bassett: Yeah, I agree it feels like we are mixing two different worlds but not sure what the alternative is since we expect users to work with display names. Once upon a time we were only using the attribute names here, but were asked to switch to display names. I can see both sides.

projects/components/src/filtering/filter/parser/types/contains-filter-parser.ts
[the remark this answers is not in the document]
jake-bassett: Not disagreeing at all. I have the same worry, but it seems to work well for current use cases.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
aaron-steinfeld: > The returned filters should always have a userString that includes the user provided text.

In other words, `isStringMatch` as defined on line 28 will always be true, right? 
```typescript
const isStringMatch = incompleteFilter.userString.toLowerCase().includes(text.trim().toLowerCase());
```
And if that's always true, then `filterMatchesUserText` will always return true and the filter is not needed.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
aaron-steinfeld: It does - logic organization wise though, I'd still recommend returning immediately if isStringMatch is true - it just convey the intent much better.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
jake-bassett: Since we are testing each attribute for a match, it won't always be true.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
jake-bassett: Fixed.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.service.ts
aaron-steinfeld: OK I dug a couple calls down and I think I see it now, but I do still find it very convoluted - if the text does match, we use it to build an incomplete filter, else it always returns an incomplete filter representing that attribute, right?

My expectation as a reader would be that autocompleteFilter would only return something that matches.

jake-bassett: Correct. We turn each attribute into an IncompleteFilter, and then filter the ones that don't match back out.

Agreed its a little convoluted, but the reason we don't filter first is because we need to build the userString and filter against that.

aaron-steinfeld: > the reason we don't filter first is because we need to build the userString and filter against that.

But doesn't that method build the userString as part of what it's doing? That's how it's able to build the incomplete filter in the first place

jake-bassett: Yep, that's what I mean. Ideally, we would filter first and then build the `IncompleteFilter`s, but the reason we don't swap the order of `autocompleteFilter()` and `filterMatchesUserText()` is because we need the `userString` from the `IncompleteFilter`s that `autocompleteFilter()` produces for the `filterMatchesUserText()` step.

aaron-steinfeld: That makes sense why it can't be done before going into autocomplete. I was more focusing on the implementation of autocomplete - it returns an array of IncompleteFilters, so it could filter them before returning so that it only returns matches.

That would be my expectation as a reader, but if you disagree we can leave it.

jake-bassett: Ah, I see. That's a fair concern. I do like the way the current method chain reads, but maybe because I've been staring at it for so long.

I just renamed those methods to maybe clear things up a bit.

projects/components/src/filtering/filter/parser/types/abstract-filter-parser.ts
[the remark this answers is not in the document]
jake-bassett: Oh, I remember now. I found a better way and this is dead code now. Thanks! Removing.

aaron-steinfeld (approved):

projects/components/src/filtering/filter/parser/types/in-filter-parser.ts
jake-bassett: Just saw this comment hanging out down here at the bottom.

Yeah, our current coercer doesn't do this. Was tempted to to change it but wasn't bold enough to deal with any fallout at the moment.

aaron-steinfeld: I don't think there should be any, but maybe we can do that in a separate PR

PR #241, opened 2020-10-07
feat: explorer service

arjunlalb:
Explorer service with a navparams builder.

Dependent on https://github.com/hypertrace/hypertrace-ui/pull/236


projects/observability/src/pages/explorer/explorer-service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
+import { NavigationParams, NavigationParamsType } from '@hypertrace/common';
+import { Filter, toUrlFilterOperator } from '@hypertrace/components';
+import { ScopeQueryParam } from './explorer.component';
+
+@Injectable({ providedIn: 'root' })
+export class ExplorerService {
+  public buildNavParamsWithFilters(scopeQueryParam: ScopeQueryParam, filters: UrlFilter[]): NavigationParams {
```
arjunlalb, line 8: This function is a straight forward one with no dependencies. Can make it a constant function. Or do we want to leave this as a service for possible future usecases? @aaron-steinfeld

aaron-steinfeld: In general, my rule of thumb is - would it ever make sense to mock out in a test? If so, leave it in a service. Or is it a utility function that does something like capitalization, transformation etc. that is silly to mock? If yes, make it constant.

Here, I think we should be adding dependencies (see below comment) so it would stay in a service.

aaron-steinfeld, line 8: We're still building the filter by hand here, and overloading what `filter.urlString` means.

If we kept it with the api I mentioned in the other PR, we'd' first go to metadata service to get the attribute, then grab the filter builder that matches that type and used it to build the filter, which should include a prebuilt url string. Something like this (no ide here, might be some issues)

```typescript
    this.metadataServce.getAttribute(scope, key)
       .pipe(
           map(attribute => this.filterBuilderLookupService
            .lookup(attribute.type)
            .buildFilter(attribute, operator, value)),
           map(filter => ({
              path: '/explorer',
              queryParams: {
                filter: filter.urlString,
                scope: scope
              }
          }))
        );
```

projects/common/src/external/external-url-navigator.ts
```diff
@@ -1,39 +1,46 @@
...
 @Injectable({ providedIn: 'root' })
 export class ExternalUrlNavigator implements CanActivate {
   public constructor(private readonly navService: NavigationService) {}
 
   public canActivate(route: ActivatedRouteSnapshot): Observable<boolean> {
-    const encodedUrl = route.paramMap.get(ExternalNavigationParams.Url);
-    const navType = route.paramMap.has(ExternalNavigationParams.NavigationType)
-      ? (route.paramMap.get(ExternalNavigationParams.NavigationType) as NavigationType)
+    const encodedUrl = route.paramMap.get(ExternalNavigationPathParams.Url);
+    const windowHandling = route.paramMap.has(ExternalNavigationPathParams.WindowHandling)
```
anandtiwary, line 17: Some of these changes I have merged already. Might wanna update branch to get a better diff.

anandtiwary: Sorry had to do it as some tests were failing. Hope I didn't duplicate your work

arjunlalb: I just had my branch based off of yours. No overlapping work @anandtiwary :)

projects/observability/src/pages/explorer/explorer-service.ts
arjunlalb: Done.

projects/components/src/public-api.ts
```diff
@@ -261,3 +261,7 @@ export { GreetingLabelComponent } from './greeting-label/greeting-label.componen
 export { MenuDropdownComponent } from './menu-dropdown/menu-dropdown.component';
 export { MenuItemComponent } from './menu-dropdown/menu-item/menu-item.component';
 export { MenuDropdownModule } from './menu-dropdown/menu-dropdown.module';
+
+// Filter builders
+export { NumberFilterBuilder } from './filtering/filter/builder/types/number-filter-builder';
```
aaron-steinfeld, line 266: These shouldn't need to be exported, thats why they're abstracted by the lookup service.

projects/observability/src/pages/explorer/explorer-service.test.ts
```diff
@@ -0,0 +1,74 @@
...
+            scope: scope,
+            requiresAggregation: false,
+            allowedAggregations: [],
+            groupable: false
+          }
+    )
+  );
+
+  const lookupMock = jest.fn(filterAttributeType =>
+    filterAttributeType === FilterAttributeType.Number ? new NumberFilterBuilder() : new StringFilterBuilder()
```
aaron-steinfeld, line 47: Follow on from my previous comment - we should either use the real filter builder service if we want to use the real builders and let it handle the details, or mock out the filter builders returned too.

projects/observability/src/pages/explorer/explorer-service.test.ts
```diff
@@ -0,0 +1,74 @@
...
+        mockProvider(FilterBuilderLookupService, { lookup: lookupMock }),
+        mockProvider(MetadataService, { getAttribute: getAttributeMock })
+      ]
+    });
+    spectator.service
+      .buildNavParamsWithFilters(ScopeQueryParam.EndpointTraces, [
+        { field: 'duration', operator: FilterOperator.GreaterThan, value: 200 },
+        { field: 'userIdentifier', operator: FilterOperator.Equals, value: 'test' }
+      ])
+      .subscribe(data =>
```
aaron-steinfeld, line 62: If putting an expectation inside a subscription, we need to use something like a `done()` callback to validate that it was actually verified. Alternatively, you can use the rxjs testing tools with marble syntax. Example:
```typescript
    runFakeRxjs(({ expectObservable }) => {
      const spectator = buildService();
      expectObservable(spectator.service.canLoad({}, [])).toBe('(x|)', { x: true });
```

Otherwise, this test would pass if the returned observable was empty or never completed.

projects/observability/src/pages/explorer/explorer-service.ts
```diff
@@ -0,0 +1,50 @@
...
+          map(
+            filterAttribute =>
+              this.filterBuilderLookupService
+                .lookup(filterAttribute.type)
+                .buildFilter(filterAttribute, filter.operator, filter.value).urlString
+          )
+        )
+    );
+
+    return combineLatest(filterStrings$).pipe(
```
aaron-steinfeld, line 37: Although functionally the same here, I think forkJoin is a better semantic fit here which conveys how this is meant to work. Basically:

forkJoin - When all observables complete, emit the last emitted value from each.

combineLatest - When any observable emits a value, emit the latest value from each.

projects/observability/src/pages/explorer/explorer-service.test.ts
arjunlalb: The real filter builder service gets injected objects for filter builders and registers them in the service. Mocking the numbers/string filter builders here would make the test irrelevant since we'd be mocking the filter construction logic itself. And the real filter builder service cannot be mocked without injecting all the necessary dependencies. What are the alternatives?

aaron-steinfeld: > Mocking the numbers/string filter builders here would make the test irrelevant since we'd be mocking the filter construction logic itself.

The idea of abstracting the construction of the filter in this service means we don't care what the construction logic is - just that it gets there, is invoked with the right args, and comes back with the corresponding result. So wouldn't mock a specific filter builder type and its logic, but instead just have lookup return a filter builder mock that has no specific implementation - such as returning "RESULT", which you can validate, as well as validating it was called with the appropriate arguments.

> And the real filter builder service cannot be mocked without injecting all the necessary dependencies. What are the alternatives?

Importing `FilterModule` would handle registering any builders for you. It would be tying the test to other implementations though - not explicitly like using string/number directly, but implicitly such that if they were to change the format of the filter, the test would break.

arjunlalb: @aaron-steinfeld Done

projects/observability/src/pages/explorer/explorer-service.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      metadata: attribute,
+      field: attribute.name,
+      operator: operator,
+      value: value,
+      userString: '',
+      urlString: `${attribute.name}${toUrlFilterOperator(operator)}${value}`
+    })
+  }));
+
+  test('creates nav params correctly', fakeAsync(() => {
```
aaron-steinfeld, line 52: nit: since we're using the rxjs test scheduler now, it shouldn't need to be fakeAsync any more.

aaron-steinfeld (dismissed):

projects/observability/src/pages/explorer/explorer-service.test.ts
arjunlalb: @aaron-steinfeld fixed this. reapproval please.

aaron-steinfeld (approved):

PR #243, opened 2020-10-08
fix: add new colors. refactor some existing colors

arjunlalb:
Adding some new colors to TS color constants.
While doing that, updated references to some existing colors as well.


projects/common/src/color/color.ts
```diff
@@ -1,7 +1,13 @@
 export const enum Color {
-  Purple2 = '#f2d0f5',
-  Purple5 = '#94209f',
+  Blue3 = '#70a7ff',
+  BlueGray4 = '#4b5f77',
```
skjindal93, line 3: Do we want to call it Blue4 or does the mocks say it as BlueGray4?

arjunlalb: I picked this from the color palette. Color palette has blues and blue grays separate.

skjindal93 (approved):

PR #245, opened 2020-10-08
feat: allow multiple filters of same attribute

jake-bassett:


projects/components/src/filtering/filter/parser/filter-parser-lookup.service.ts
```diff
@@ -1,32 +1,35 @@
...
+        return new InFilterParser();
+      case FilterOperator.ContainsKey:
+      case FilterOperator.ContainsKeyValue:
+        return new ContainsFilterParser();
+      default:
+        assertUnreachable(operator);
     }
 
-    return filterParser;
+    throw new Error(`Filter parser not found for operator '${operator}'.`);
```
aaron-steinfeld, line 32: if you return on line 29 this isn't needed (assertUnreachable returns never because it would throw)

projects/components/src/filtering/filter/filter-url.service.test.ts
```diff
@@ -69,13 +65,8 @@ describe('Filter URL service', () => {
...
         getAllValuesForQueryParameter: (param: string) => testQueryParamObject[param] ?? []
       }),
-      mockProvider(FilterBuilderLookupService, {
-        isBuildableType: (type: FilterAttributeType) => mockFilterBuilderLookup(type) !== undefined,
-        lookup: mockFilterBuilderLookup
-      }),
-      mockProvider(FilterParserLookupService, {
-        lookup: mockFilterParserLookup
-      })
+      FilterBuilderLookupService,
```
aaron-steinfeld, line 68: root injectables should be available by default

projects/components/src/filtering/filter/builder/filter-builder-lookup.service.ts
```diff
@@ -1,30 +1,33 @@
...
+        return new NumberFilterBuilder();
+      case FilterAttributeType.String:
+        return new StringFilterBuilder();
+      case FilterAttributeType.StringMap:
+        return new StringMapFilterBuilder();
+      case FilterAttributeType.StringArray: // Unsupported
+      case FilterAttributeType.Timestamp: // Unsupported
+        throw new Error(`Filter builder not found for attribute of type '${type}'.`);
+      default:
+        assertUnreachable(type);
```
aaron-steinfeld, line 28: return this instead of error

aaron-steinfeld (approved):

PR #262, opened 2020-10-13
fix: make breadcrumbs extensible

arjunlalb:


skjindal93 (dismissed):

skjindal93 (dismissed):

projects/components/src/breadcrumbs/breadcrumbs.service.ts
```diff
@@ -6,8 +6,8 @@ import { concat, defer, EMPTY, Observable, of } from 'rxjs';
 import { catchError, map, shareReplay, switchMap, toArray } from 'rxjs/operators';
 
 @Injectable({ providedIn: 'root' })
-export class BreadcrumbsService {
-  public readonly breadcrumbs$: ReplayObservable<Breadcrumb[]>;
+export class BreadcrumbsService<T extends Breadcrumb> {
```
aaron-steinfeld, line 9: may be easier to default T so references to breadcrumbs service don't change:
```suggestion
export class BreadcrumbsService<T extends Breadcrumb = Breadcrumb> {
```

aaron-steinfeld: Alternatively, could give breadcrumb an unknown payload and add the defaulted type param to breadcrumb so we can leave the whole usage exactly as is - in that scenario, the only places that would know about any custom data would be the producer and consumer of the breadcrumbs.

projects/observability/src/pages/apis/service-detail/service-detail-breadcrumb.resolver.ts
```diff
@@ -54,6 +54,10 @@ export class ServiceDetailBreadcrumbResolver implements Resolve<Observable<Bread
   }
 
   private getAttributeKeys(): string[] {
-    return ['name'];
+    return ['name', ...this.getExtraAttributeKeys()];
+  }
+
+  protected getExtraAttributeKeys(): string[] {
```
aaron-steinfeld, line 60: nit: I think it would make more sense to make `getAttributeKeys` protected rather than expose this unused function

arjunlalb: Thats what I had done in the first iteration. Then, @skjindal93 made a point that we can avoid repeating the base list/keys (just 'name' in this case) if we take this route.

Any strong reasons to prefer either way? I couldn't think of any except for probably the flexibility to override the whole list.

projects/components/src/breadcrumbs/breadcrumbs.service.ts
arjunlalb: I feel defaulting to Breadcrumb is cleaner.

aaron-steinfeld (dismissed):

PR #264, opened 2020-10-13
Make Explorer builder more extensible

jake-bassett:


projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -259,7 +259,7 @@ export class ExplorerDashboardBuilder {
           {
             type: 'table-widget-column',
             title: 'Service',
-            width: '20%',
+            width: '2',
```
anandtiwary, line 262: what does 2 represent here? like 2 fractions?

anandtiwary (approved):

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
jake-bassett: I saw it used elsewhere and played with it a bit. It appears to be the same as 20% here, but I'm guessing its like the flex format.

aaron-steinfeld: yes, flex of the non-explicit remaining width.

PR #267, opened 2020-10-15
fix: make api detail breadcrumb resolver extensible

arjunlalb:


projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
```diff
@@ -89,7 +89,11 @@ export class ApiDetailBreadcrumbResolver implements Resolve<Observable<Breadcrum
       ? [this.getParentNameAttribute(parentTypeMetadata), this.getParentIdAttribute(parentTypeMetadata)]
       : [];
 
-    return ['name', ...parentAttributes];
+    return ['name', ...this.getExtraAttributeKeys(), ...parentAttributes];
+  }
+
+  protected getExtraAttributeKeys(): string[] {
```
skjindal93, line 95: Didn't we just expose the `getAttributeKeys` for ServiceBreadcrumb?

arjunlalb: yea. that class had a static list. but for this resolver, there's more logic in the getAttributeKeys method. Didn't make sense to take those onto the child class.

arjunlalb: @skjindal93 Breadcrumb resolvers are independent resolvers and not based on a single base class.

aaron-steinfeld (approved):

PR #270, opened 2020-10-15
fix: container widget renderer title spacing and color

jake-bassett:


projects/dashboards/src/widgets/container/container-widget-renderer.component.scss
```diff
@@ -8,7 +8,7 @@
   flex-direction: column;
 
   .title {
-    @include widget-title($gray-7);
-    margin-top: 15px;
```
anandtiwary, line 12: Removing top margin may break some screens, wouldn't it?

jake-bassett: I went through most of the screens and didn't see any issues. There might be an odd one here or there that I missed, but if we see a case where this spacing doesn't work I can fix those places individually.

aaron-steinfeld (approved):

PR #271, opened 2020-10-15
refactor: move feature service to common

jake-bassett:


aaron-steinfeld (changes requested):
Discussed offline - the abstract resolver should be in common, the implementation should remain in the app.

PR #272, opened 2020-10-16
fix: container widget height with title

jake-bassett:


projects/dashboards/src/widgets/container/container-widget-renderer.component.scss
```diff
@@ -8,7 +8,16 @@
   flex-direction: column;
 
   .title {
+    height: 28px;
     @include widget-title($gray-9);
     padding-bottom: 16px;
   }
+
+  .content {
```
anandtiwary, line 16: `flex: 1 1 auto` for .content would also do the same right?

anandtiwary:
```
.title {
height: 28px
}

.content {
flex: 1 1 auto;
}
```

jake-bassett: That was my first attempt as well. I can't get `flex` to work here for some reason. When I try that, the `.content` just expands to fit its... well, content.

I could probably circle back to this, but this was the quickest/easiest solution without worrying about far reaching side effects.

anandtiwary: okay.. we can merge it if you are blocked. I can also play around with it later today.

anandtiwary (approved):

PR #280, opened 2020-10-20
feat: adding a toggle switch component

anandtiwary:


projects/components/src/toggle-switch/toggle-switch.component.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { createHostFactory, Spectator } from '@ngneat/spectator/jest';
+import { ToggleSwitchComponent } from './toggle-switch.component';
+
+describe('Toggle Switch Component', () => {
+  let spectator: Spectator<ToggleSwitchComponent>;
+
+  const createHost = createHostFactory({
+    component: ToggleSwitchComponent,
+    shallow: true,
+    imports: [MatSlideToggleModule, FormsModule]
```
anandtiwary, line 13: If I mock MatSlideToggle I won't be able to test the inputs as this component is using get/set. Hence I am importing the entire module.

jake-bassett (approved):

aaron-steinfeld:
How does this differ from toggle item? From checkbox? Why do we need both toggle item and toggle switch?

anandtiwary:
> How does this differ from toggle item? Or more accurately, from checkbox? Why do we need all these alternatives for the same input type?

Just visually. You will only have one toggle switch. But you can have multiple checkboxes. (btw Mat toggle switch uses checkbox inside)

PR #281, opened 2020-10-21
Time picker refactor

anandtiwary:


projects/assets-library/assets/styles/_font.scss
```diff
@@ -124,18 +129,21 @@ $font-family: 'Work Sans', sans-serif;
 }
 
 %body-1 {
+  @extend %fontfamily;
```
anandtiwary, line 132: Actually this was supposed to go with toggle switch. But i missed it over there. Need to set font family explicitly for material components.

aaron-steinfeld: We used to solve this through material typography, but I see that it got removed - @jake-bassett do you recall why?

The syntax:
```scss
$custom-typography: mat-typography-config(
  $font-family: $font-family
);

@include angular-material-typography($custom-typography);
```

projects/components/src/time-picker/time-picker.component.ts
```diff
@@ -0,0 +1,69 @@
...
+import { IconSize } from '../icon/icon-size';
+import { PredefinedTimeService } from '../time-range/predefined-time.service';
+
+@Component({
+  selector: 'ht-time-picker',
+  styleUrls: ['./time-picker.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="time-picker">
+      <ht-popover class="time-selector" [closeOnClick]="true">
```
aaron-steinfeld, line 13: Besides the values, how does this really differ from our select component? Could it just wrap that?

arjunlalb: +1

I think we can use the select component.

projects/assets-library/assets/styles/_font.scss
jake-bassett: That might have been removed with the sweep to remove the dead theming code. If that snippet is needed, we should add it back.

projects/components/src/time-picker/time-picker.component.ts
anandtiwary: It guess it could. But I don't want to do it now. have few other things to finish. This one is just extracted from date-time picker

projects/assets-library/assets/styles/_font.scss
anandtiwary: okay. So should I revert this and add above instead?

aaron-steinfeld: I think so

projects/components/src/time-picker/time-picker.component.ts
aaron-steinfeld: Please file a tech debt issue for this

anandtiwary: https://traceableai.atlassian.net/browse/ENG-6567

anandtiwary: Here you go

projects/assets-library/assets/styles/_font.scss
anandtiwary: This didn't work. I guess we will also have to call individual mixin to make it work.

```
@mixin mat-slide-toggle-typography($config-or-theme) {
  $config: mat-get-typography-config($config-or-theme);
  .mat-slide-toggle-content {
    font-family: mat-font-family($config);
  }
}
```
we will have to maintain a global mat scss where we can include these mixins. But don't you think the current approach of adding it to our style mixins is better? It works everywhere and won't be affected by any mat theme/scss changes

aaron-steinfeld: The proposed approach relies on our classes being used inside material's.

The approach I listed above relies on material's own maintenance, so we shouldn't have to worry about it. It would potentially get overridden if it was declared before the out of the box `include mat_core();` which wold reset the theme to the out of the box behavior. But you can also inline it so the whole change just becomes:
```diff
diff --git a/src/styles/main.scss b/src/styles/main.scss
index 203670c..bc0a890 100644
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -64,4 +64,4 @@ mark {
 // Include the common styles for Angular Material. We include this here so that you only
 // have to load a single css file for Angular Material in your app.
 // Be sure that you only ever include this mixin once!
-@include mat-core();
+@include mat-core(mat-typography-config($font-family: $font-family));
```

anandtiwary: done

src/styles/main.scss
```diff
@@ -1,6 +1,8 @@
 @import 'shared-globals';
 @import '@angular/material/theming';
 
+@include mat-core(mat-typography-config($font-family: $font-family));
```
aaron-steinfeld, line 4: mat-core is already included in this file on line 69, please update the parameters of that invocation.

anandtiwary: oops

anandtiwary: done

aaron-steinfeld (approved):

PR #282, opened 2020-10-21
fix: nav location as null

jake-bassett:


projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
```diff
@@ -65,7 +65,9 @@ export class NavigableDashboardComponent implements OnChanges {
   public ngOnChanges(changeObject: TypedSimpleChanges<this>): void {
     if (changeObject.navLocation) {
       const persistedDashboard$ =
-        this.navLocation === undefined ? EMPTY : this.dashboardPersistenceService.getForLocation(this.navLocation);
+        this.navLocation === undefined || this.navLocation === null
```
aaron-steinfeld, line 68: might be better to just check truthiness rather than doing a comparison, it implicitly handles any bad case here.

aaron-steinfeld (approved):

aaron-steinfeld (approved):

PR #285, opened 2020-10-23
feat: api flattening

jake-bassett:


projects/components/src/table/data/table-cdk-data-source-api.ts
```diff
@@ -5,6 +5,7 @@ import { TableDataSource } from './table-data-source';
 
 export interface TableDataSourceProvider {
   data?: TableDataSource<TableRow>;
+  flattenedTree?: boolean;
```
aaron-steinfeld, line 8: I'm confused (and I think this applies broadly across the PR), how is a flattened tree different than a regular table? why would table have to change at all, rather than changing the data and passing in regular table data?

PR #286, opened 2020-10-23
feat: flattened table mode for entity tree lists

jake-bassett:


aaron-steinfeld (changes requested):
Why modify table itself, which is already overly complex, rather than sending different data in?

jake-bassett:
> Why modify table itself, which is already overly complex, rather than sending different data in?

Because the toggle control belongs in the table. The table itself isn't changing much other than holding the toggle and emitting the mode. The table data source is doing the heavy lifting.

aaron-steinfeld:
> > Why modify table itself, which is already overly complex, rather than sending different data in?
> 
> Because the toggle control belongs in the table. The table itself isn't changing much other than holding the toggle and emitting the mode. The table data source is doing the heavy lifting.

Why does the control belong in table when we've been keeping other controls like filters and searches outside? The problem we've seen with putting any of the controls inside the table (for example, search initially) is that beyond the complexity it adds to the table, we're stuck moving it outside as soon as we want to put anything else external next to it.

PR #293, opened 2020-10-26
feat: add stand alone modal service

aaron-steinfeld:


projects/components/src/modal/default-modal-ref.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  public initialize(popoverRef: PopoverRef): void {
+    this.popoverRef = popoverRef;
+    this.popoverRef.closed$.subscribe({
+      complete: () => this.closedObserver.complete(),
+      error: err => this.closedObserver.error(err)
+    });
+  }
+
+  public close(result?: unknown): void {
```
anandtiwary, line 26: Why is result unknown? I would think it would be either a boolean for confirmation modal or undefined for generic modals.

projects/components/src/modal/default-modal-ref.ts
```diff
@@ -0,0 +1,32 @@
...
+export class DefaultModalRef extends ModalRef<unknown> {
+  public readonly closed$: Observable<unknown>;
+
+  private readonly closedObserver: Observer<unknown>;
+  private popoverRef?: PopoverRef;
+
+  public constructor() {
+    super();
+    const closedSubject = new ReplaySubject<unknown>(1);
+    this.closedObserver = closedSubject;
```
jake-bassett, line 14: Is there a reason to not just assign `closedObserver` directly rather than declaring `closedSubject` first?

Also, why name it `closedObserver` when its a subject?

projects/components/src/modal/default-modal-ref.ts
aaron-steinfeld: because we don't care what it is in this file. the ModalRef api uses type parameters here, but since this implementation is private and used for any modal with any type, by keeping the unknown internal to the implementation, it saves us from having to stamp around `DefaultModalRef<unknown>` when working with it in the service.

aaron-steinfeld: Also confirmation modal is just another modal, nothing special about it.

projects/components/src/modal/modal-container.component.ts
```diff
@@ -0,0 +1,83 @@
...
+        <ht-button
+          class="close-button"
+          icon="${IconType.Close}"
+          display="${ButtonStyle.Outlined}"
+          size="${ButtonSize.ExtraSmall}"
+          htTooltip="Close modal"
+          (click)="this.close()"
+        >
+        </ht-button>
+        <h3 class="header-title">{{ modalTitle }}</h3>
```
jake-bassett, line 25: Is there a reason to use h3 if we just style it ourselves anyway? Do we have global styles for h# elements?

projects/components/src/modal/default-modal-ref.ts
aaron-steinfeld: A subject is both an observable and an observer, so what I'm doing here is splitting it up into its component interfaces. If I directly assign it to closedObserver, I no longer can get the observable view of it. If I change closedObserver to be typed as a subject, I'm holding on to an implementation class that has more details than I want this class to know about.

projects/components/src/modal/modal-container.component.ts
aaron-steinfeld: Semantics mainly. We should always be using the most appropriate tags given the content they contain to better mark up the meaning of the document, the styling can still be done by classes. That said, this is pre-existing from the previous implementation: https://github.com/hypertrace/hypertrace-ui/blob/60d954c5b5e7991ce106870f163b49dd85fbea94/projects/components/src/overlay/modal/modal-overlay.component.ts#L24

projects/components/src/modal/default-modal-ref.ts
anandtiwary: Okay. Do you also plan to add a create confirmation modal method as a lot of the code over there is common and currently duplicated across many different implementations.

aaron-steinfeld: yep - it's in the initial PR of this that this code is based on ( https://github.com/hypertrace/hypertrace-ui/pull/216 ), but separating the confirmation modal into its own PR.

anandtiwary: Cool

projects/components/src/modal/modal.service.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  public createModal<TData, TResponse = never>(
+    config: ModalConfig<TData>,
+    injector: Injector = this.defaultInjector
+  ): ModalRef<TResponse> {
+    this.activeModal?.close();
+    const newModal = new DefaultModalRef();
+    this.activeModal = newModal;
+
+    const popover = this.popoverService.drawPopover({
```
anandtiwary, line 24: nit: this can be put into a separate function once we add the confirmation modal code to further improve readability.

anandtiwary (approved):
looks good to me.

PR #295, opened 2020-10-26
Confirmation

aaron-steinfeld:


projects/components/src/confirmation/confirmation-modal.component.ts
```diff
@@ -0,0 +1,55 @@
...
+import { ModalRef, MODAL_DATA } from '../modal/modal';
+
+@Component({
+  selector: 'ht-confirmation-modal',
+  styleUrls: ['./confirmation-modal.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="confirmation-modal">
+      <div class="description">
+        {{ this.descriptionText }}
```
anandtiwary, line 12: This one needs more thought. The content can be a description or it can be a form with select component (E.g Actor State -> Snooze).

The closing of popover also depends on the success/failure of any mutation that the caller is doing. So we need a way to notify the caller component that confirm is clicked, then wait for caller to return success/failure and then accordingly close the modal. 

If there is a form in the modal, then we need to keep the user on the modal so that user can modify the form data and resubmit. E.g Invalid email address in Settings -> add user.

aaron-steinfeld: So I can definitely be convinced that a confirmation modal should support a template beyond a simple string - but what you're describing is a full on modal with its own logic. I don't think it makes sense to try to make confirmation support complex use cases. The only thing they share is having success/cancel buttons.

anandtiwary: but we are duplicating code at many different places and the use case it supports is the primary use case right now. We should be able to use this new modal for the current use cases.

arjunlalb: I don't think confirmation modal should take anything more than a string and a boolean input (for the possible explicit "I agree" use case we might have for really critical usecases). Everything else counts as logic specific to one modal IMO.

projects/components/src/confirmation/confirmation-modal.component.ts
```diff
@@ -0,0 +1,55 @@
...
+          [label]="this.confirmButtonLabel"
+          [role]="this.confirmButtonRole"
+          (click)="this.onConfirmation()"
+        ></ht-button>
+      </div>
+    </div>
+  `
+})
+export class ConfirmationModalComponent {
+  private static readonly DEFAULT_CONFIRM_LABEL: string = 'OK';
```
arjunlalb, line 26: Shouldn't default be 'Confirm'?

arjunlalb:
Thanks for taking this and the underlying modal service to completion :) @aaron-steinfeld

projects/components/src/confirmation/confirmation-modal.component.ts
aaron-steinfeld: Probably - that feels so formal though! We can change this once we start actually using it and get a sense of which is a better fit.

projects/components/src/confirmation/confirmation-modal.component.ts
aaron-steinfeld: The code that's being duplicated is basically just
```html
      <div class="controls">
        <ht-button
          label="Cancel"
          role="${ButtonRole.Tertiary}"
          (click)="this.onCancel()"
        ></ht-button>
        <ht-button
          label="Confirm"
          role="${ButtonRole.Additive}"
          (click)="this.onConfirmation()"
          [disabled]="this.isButtonDisabled()"
        ></ht-button>
      </div>
```

We could pull that into its own little component, but by the time we passed in configuration I think you'd be saving a an insignificant number of lines of code - there's really not that much shared between these. Things like communicating back responses are already supported for any modal. Everything else is custom to each specific implementation - when to disable the confirm button, how to react to confirm, how to respond to network calls etc.

projects/components/src/confirmation/confirmation-modal.component.ts
arjunlalb: All our existing usecases are for 'Confirm'. Hence the suggestion.

arjunlalb (dismissed):

projects/components/src/confirmation/confirmation-modal.component.ts
aaron-steinfeld: Fair enough

arjunlalb (approved):

PR #296, opened 2020-10-27
feat: make highlight pipe customizable with highlight type

arjunlalb:


projects/common/src/utilities/formatters/string/highlight.pipe.test.ts
```diff
@@ -0,0 +1,27 @@
+import { HighlightPipe } from '@hypertrace/common';
```
aaron-steinfeld, line 1: local import

aaron-steinfeld (approved):

projects/common/src/utilities/formatters/string/highlight.pipe.ts
```diff
@@ -1,8 +1,30 @@
 import { Pipe, PipeTransform } from '@angular/core';
+import { assertUnreachable } from '../../lang/lang-utils';
 
 @Pipe({ name: 'htHighlight' })
 export class HighlightPipe implements PipeTransform {
-  public transform(fullText: string, highlightedText: string | undefined): string {
-    return highlightedText === undefined ? fullText : fullText.replace(highlightedText, '<mark>$&</mark>');
+  public transform(
```
anandtiwary, line 6: Quick question: Could we not have used these html(b/italiic etc) tags directly since we are not putting any custom highlighting?

PR #299, opened 2020-10-27
feat: adding support for multiple row selections for each depth

anandtiwary:


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
```diff
@@ -80,14 +81,30 @@ export class TableWidgetModel {
...
+   */
   @ModelProperty({
     key: 'selection-handler',
     displayName: 'Row selection Handler',
     type: ModelPropertyType.TYPE
   })
   public selectionHandler?: InteractionHandler;
 
+  @ModelProperty({
+    key: 'row-selection-handlers',
```
aaron-steinfeld, line 95: Rather than trying to encode this depth/matching handler logic, could we just make a custom handler that accepts either row type?

anandtiwary: This is more extendable. Creating a new custom handler will be cumbersome and we can't use it multiple places. We need to use this on both service and edge api tables.

aaron-steinfeld: It is more generic, but we need to balance that against the types of patterns we want to encourage and the readability/clarity of the json.

For this, the idea of having different handlers for different depths in the table feels like a clear antipattern to me. For a detail template, it can always have its own handler in the template. For a table, like the service/api view, why would clicking on an api have one behavior and clicking on a service another? That seems like questionable UX at best.

anandtiwary: Totally agree on questionable UX and that is why we have a JIRA to figure out and make necessary changes. We can either wait for that discussion or get this in during this time. Considering we want to make more incremental changes, my preference is to get this in.

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):

PR #300, opened 2020-10-27
feat: allow json config for adding classes to column renderers

jake-bassett:


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
```diff
@@ -51,6 +58,12 @@ export class TableWidgetColumnModel {
   })
   public alignment?: TableCellAlignmentType;
 
+  @ModelProperty({
+    key: 'classes',
+    type: ARRAY_PROPERTY.type
+  })
+  public classes: string[] = [];
```
anandtiwary, line 65: How would I know what classes I can add? Every cell renderer implementation may have its own set.

jake-bassett: They would need to be global or available in the renderer.

PR #302, opened 2020-10-28
feat: explorer filter on hover directive

arjunlalb:
A directive that would show the filter icon on hover on the host element, that can be used to navigate to explorer filtered by that data.

### DIFF
--- a/projects/observability/src/shared/components/explorer-filter-on-hover/explorer-filter-on-hover.component.ts
+++ b/projects/observability/src/shared/components/explorer-filter-on-hover/explorer-filter-on-hover.component.ts
@@ -1,0 +1,41 @@
+import { ChangeDetectionStrategy, Component, Inject } from '@angular/core';
+import { IconType } from '@hypertrace/assets-library';
+import { NavigationService } from '@hypertrace/common';
+import { IconSize, POPOVER_DATA } from '@hypertrace/components';
+import { ExplorerService } from '../../../pages/explorer/explorer-service';
+import { ExplorerFilterOnHoverData } from './explorer-filter-on-hover.directive';
+
+@Component({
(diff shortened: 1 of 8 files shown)


projects/observability/src/shared/components/explorer-filter-on-hover/explorer-filter-on-hover.component.ts
```diff
@@ -0,0 +1,41 @@
+import { ChangeDetectionStrategy, Component, Inject } from '@angular/core';
+import { IconType } from '@hypertrace/assets-library';
+import { NavigationService } from '@hypertrace/common';
+import { IconSize, POPOVER_DATA } from '@hypertrace/components';
+import { ExplorerService } from '../../../pages/explorer/explorer-service';
+import { ExplorerFilterOnHoverData } from './explorer-filter-on-hover.directive';
+
+@Component({
+  selector: 'ht-explorer-filter-on-hover',
```
aaron-steinfeld, line 9: I think separating the trigger and target can make this work much more precisely, and will allow decoupling the hover behavior with a specific component like an explorer drill down icon.

projects/observability/src/shared/components/explorer-filter-on-hover/explorer-filter-on-hover.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  styleUrls: ['./explorer-filter-on-hover.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<div class="hover-filter-container">
+    <ht-icon
```
anandtiwary, line 13: I feel this should be wrapped inside an ht-link so that users can just open it in a new tab.

PR #304, opened 2020-10-29
Separate table controls

jake-bassett:


projects/common/src/utilities/formatters/string/string-formatter.ts
```diff
@@ -1,5 +1,7 @@
 import { isEmpty } from 'lodash-es';
 
+export const titleCase = (str: string): string => (str.length === 0 ? str : str[0].toUpperCase() + str.slice(1));
```
anandtiwary, line 3: lodash has a method for this. Also there is a pipe `titleCase` from angular. Do we need our own?

projects/dashboards/src/widgets/widget-renderer.ts
```diff
@@ -45,7 +45,7 @@ export abstract class WidgetRenderer<TModel extends object, TData = unknown> imp
     this.fetchAndRunChangeDetection();
   }
 
-  private onDashboardRefresh(): void {
+  protected onDashboardRefresh(): void {
```
aaron-steinfeld, line 48: This should remain private - will suggest an alternative when I see how it's used

projects/distributed-tracing/src/shared/dashboard/data/graphql/table/spans/spans-table-data-source.model.ts
```diff
@@ -43,8 +43,4 @@ export class SpansTableDataSourceModel extends TableDataSourceModel {
       totalCount: response.total
     };
   }
-  protected getSearchFilterAttribute(): string {
-    // Todo: Figure this out
```
aaron-steinfeld, line 47: Did we not support paging on spans before?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
```diff
@@ -71,6 +93,20 @@ export class TableWidgetModel {
...
+    // tslint:disable-next-line: no-object-literal-type-assertion
+    type: {
+      key: ARRAY_PROPERTY.type,
+      subtype: {
+        key: ENUM_TYPE.type,
+        values: [TableMode.Flat, TableMode.Tree, TableMode.Detail]
+      }
+    } as ArrayPropertyTypeInstance
+  })
+  public modeToggles: TableMode[] = [];
```
aaron-steinfeld, line 108: this feels like it conflicts with mode - do we think it's feasible to unify the two properties? something like `modes` which accepts an array and defaults to a single mode?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
```diff
@@ -60,6 +61,27 @@ export class TableWidgetModel {
...
+    // tslint:disable-next-line: no-object-literal-type-assertion
+    type: {
+      key: ARRAY_PROPERTY.type,
+      subtype: {
+        key: ModelPropertyType.TYPE,
+        defaultModelClass: TableWidgetFilterModel
+      }
+    } as ArrayPropertyTypeInstance
+  })
+  public filterToggles: TableWidgetFilterModel[] = [];
```
aaron-steinfeld, line 83: I fear we're going to quickly be overrun by different controls, but let's see what happens

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -69,33 +97,36 @@ export class TableWidgetRendererComponent
...
-      /**
-       * Execute selection handler for single selection mode only
-       */
-      let selectedRow;
-      if (selections.length > 0) {
-        selectedRow = selections[0];
-        this.selectedRowInteractionHandler = this.getInteractionHandler(selectedRow);
-      }
+  protected onDashboardRefresh(): void {
+    this.initControls();
```
aaron-steinfeld, line 118: So this is where it's used - why does initControls need to be invoked on a dashboard refresh? The filter items and mode items are the same. Is the idea to reset their value? If so, I'd actually assume the opposite behavior, that they'd persist their state through a local refresh.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -118,10 +149,54 @@ export class TableWidgetRendererComponent
     );
   }
 
+  public onFilterChange(item: ToggleItem<TableWidgetFilterModel>): void {
+    if (item.value && item.value.attribute && item.value.operator && item.value.value) {
```
aaron-steinfeld, line 153: Is there a chance that these would be unset? What if one of the toggles was meant to clear the filters? Also, the value could be a falsy value like `0` or `false`, right?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -118,10 +149,54 @@ export class TableWidgetRendererComponent
...
+  public onSearchChange(text: string): void {
+    const searchFilter: TableFilter = {
+      field: this.api.model.searchAttribute!,
+      operator: FilterOperator.Like,
+      value: text
+    };
+    this.searchFilterSubject.next([searchFilter]);
+  }
+
+  public onModeChange(mode: TableMode): void {
```
aaron-steinfeld, line 177: We should never be writing back to the model - please store it locally for the current state, the model should be the persisted default.

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -0,0 +1,114 @@
...
+          class="control filter-toggle-group"
+          [items]="this.filterItems"
+          [activeItem]="this.activeFilterItem"
+          (activeItemChange)="this.onFilterChange($event)"
+        ></ht-toggle-group>
+      </div>
+
+      <!-- Mode Toggle -->
+      <ht-toggle-group
+        *ngIf="this.modeItems?.length"
```
aaron-steinfeld, line 36: should be greater than 1

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        [items]="this.modeItems"
+        [activeItem]="this.activeModeItem"
+        (activeItemChange)="this.onModeChange($event)"
+      ></ht-toggle-group>
+    </div>
+  `
+})
+export class TableControlsComponent implements OnChanges {
+  @Input()
+  public searchAttribute?: string;
```
aaron-steinfeld, line 47: This should either be used to emit a search filter, or be removed in favor of a boolean, since that's the only way it's being used currently

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  public readonly filterChange: EventEmitter<ToggleItem> = new EventEmitter<ToggleItem>();
+
+  @Output()
+  public readonly modeChange: EventEmitter<TableMode> = new EventEmitter<TableMode>();
+
+  private readonly searchDebounceSubject: Subject<string> = new Subject<string>();
+
+  public constructor(private readonly subscriptionLifecycle: SubscriptionLifecycle) {
+    this.subscriptionLifecycle.add(
+      this.searchDebounceSubject.pipe(debounceTime(200)).subscribe(text => this.searchChange.emit(text))
```
aaron-steinfeld, line 77: note to self - add this support natively to searchbox

projects/components/src/table/table.component.ts
```diff
@@ -212,6 +203,9 @@ export class TableComponent
   @Input()
   public data?: TableDataSource<TableRow>;
 
+  @Input()
+  public filters?: TableFilter[];
```
aaron-steinfeld, line 207: So I was just sitting there reflecting on past code reviews as I'm wont to do and started thinking - do we even need filters in table? Does table or the cdk data source do anything with filters besides spit them right back out with every request? It's a big ole `combineLatest`

jake-bassett: Yeah, I realized that as well once I got this working with the existing code path. The filters just get spit back out to the data source. But I didn't want to carve a new path for them since this PR is already a bit crazy.

I figured I could tackle that when we get back into breaking up the table some more and reworking the TableCdkDatatSource.

aaron-steinfeld: That's fair. No need in this PR, but I think it's a pretty small lift to remove them - just deletions inside table, remove the filters from the data request emitted by table, and then create a new request local to the widget that brings them together so the code beneath it doesn't change.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
jake-bassett: Yeah... I hear you. But there is only so much we can stuff in the header for a table. Not sure what else could be added (though I'm sure we'll find something).

I get wanting to compose bespoke headers so we don't have these uber components with a bunch of Input options that may not always make sense for each use case, but I also want to force us into a consistent experience. The single component has some pros and cons. For consistency, it makes sense to me to club all of these controls into a single component with our current use cases, but agree its a slippery slope.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Fixed the 0/false case. Thanks.

We do have a use case for "All" asset tags that clears the filter. This works because `value` is `undefined` for that filter, therefore the `if` check fails and an empty array is provided thus clearing the filter.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
jake-bassett: I don't think it should conflict. The `mode` property is adequate for 99% of our use cases so I would prefer not to make users build out an array and object when unnecessary.

projects/components/src/table/controls/table-controls.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import { NgModule } from '@angular/core';
+import { SubscriptionLifecycle } from '@hypertrace/common';
+import { TraceSearchBoxModule } from '../../search-box/search-box.module';
+import { ToggleGroupModule } from '../../toggle-group/toggle-group.module';
+import { TooltipModule } from '../../tooltip/tooltip.module';
+import { TableControlsComponent } from './table-controls.component';
+
+@NgModule({
+  imports: [CommonModule, TooltipModule, TraceSearchBoxModule, ToggleGroupModule],
+  providers: [SubscriptionLifecycle],
```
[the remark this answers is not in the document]
jake-bassett: Fixed

projects/dashboards/src/widgets/widget-renderer.ts
jake-bassett: Fixed

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Fixed

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Fixed

projects/components/src/table/controls/table-controls.component.ts
jake-bassett: Fixed

projects/components/src/table/controls/table-controls.component.ts
jake-bassett: Fixed

projects/distributed-tracing/src/shared/dashboard/data/graphql/table/spans/spans-table-data-source.model.ts
aaron-steinfeld: sorry, yep. I think we only use it via explorer, so not needed.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
aaron-steinfeld: Two that come to mind are the filter bar and then various action menus. So may be worth starting to think how those might fit in.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
aaron-steinfeld: OK fair. I guess with an array like this, you'd want a way to specify the default anyway. Thinking about it that way, maybe `modeOptions` is a better name (feel free to ignore if you're not back in here).

projects/components/src/table/controls/table-controls.component.ts
aaron-steinfeld: Yeah, the way I'd expect this to work would be an input like `debounceTime` that defaults to 0, so any debounce can be specified and it's backwards compat.

aaron-steinfeld (approved):

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
jake-bassett: I like that name better too. I'll change when I'm back in there. Same for filterOptions.

PR #307, opened 2020-10-29
feat: icon cell renderer accepts mapped args

jake-bassett:


projects/components/src/table/cells/data-renderers/icon/icon-table-cell-renderer.component.ts
```diff
@@ -21,4 +23,17 @@ import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
...
-export class IconTableCellRendererComponent extends TableCellRendererBase<string> {}
+export class IconTableCellRendererComponent extends TableCellRendererBase<CellData, Value> implements OnInit {
+  public iconSize: IconSize = IconSize.Small;
+
+  public ngOnInit(): void {
+    super.ngOnInit();
+
+    if (this.value && this.value.size) {
+      this.iconSize = this.value?.size;
+    }
```
aaron-steinfeld, line 38: nit:
```suggestion
    this.iconSize = this.value?.size ?? IconSize.Small;
```

projects/components/src/table/cells/data-renderers/icon/icon-table-cell-renderer.component.ts
```diff
@@ -11,8 +12,9 @@ import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
   styleUrls: ['./icon-table-cell-renderer.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="icon-cell" [ngClass]="{ clickable: this.clickable }">
-      <ht-icon [icon]="this.value" size="${IconSize.Small}" [showTooltip]="true"></ht-icon>
+    <div class="icon-cell" [ngClass]="{ clickable: this.clickable }" *ngIf="this.value">
+      <ht-icon [icon]="this.value.icon" [size]="this.iconSize" [showTooltip]="true" [ngClass]="this.value.color"
```
aaron-steinfeld: Why use color as a class? This relies on the json knowing the classes which is what we were trying to avoid. In the poc I shared, using `[style.color]` should allow us to not assume that knowledge.

jake-bassett: The reason I chose to have classes is because all of our colors are scss variables. We could allow the json user to give us a hex string, but that's not very maintainable. If the color palette changes, the json hexes remain the old color and will all have to be manually updated. If we have classes based on colors, those update automatically.

projects/components/src/table/cells/data-renderers/icon/icon-table-cell-renderer.component.ts
jake-bassett: Hah, this is exactly what I had before and changed it last moment.

projects/components/src/table/cells/data-renderers/icon/icon-table-cell-renderer.component.ts
aaron-steinfeld: I think there's definitely work to do in terms of color palette, but if the main concern is maintenance, I think that's accounted for by the enum we use elsewhere - and it'd be nice for the widgets to stay consistent. For example, in cartesian widgets:
```
                type: 'series',
                name: 'p99',
                'visualization-type': 'line',
                color: Color.BlueGray4,
```
Longer term, having known values that can match to any palette would be cool, but we're not there yet.

jake-bassett: Ah... that enum looks like a maintenance nightmare too. I disagree that is the right thing to do. We shouldn't be duplicating our hex codes this way.

But in the interest of consistency I can make the change.

aaron-steinfeld: OK, I'll file a tech debt to revisit how we specify color in JSON more broadly.

aaron-steinfeld: (another potential concern is that the css classes are dependent on how they're used. For example using a color in cartesian might mean as a `fill` class for svg, direct code access for canvas, or background-color/stroke in other scenarios)

aaron-steinfeld (approved):

aaron-steinfeld (approved):

PR #309, opened 2020-10-30
fix: fix table scrolling and loading issues

anandtiwary:
- Load Async is built to show loaders for content rendered inside it.
  We are using it in table as a placeholder where the loader wrapper
  stays after the cdk table has loaded all data. This creates lots of
  layout issues especially when using with flex container

- Introducing a new hide property in table loading state to remove the
  loader wrapper once loading is completed. In case of error or no data,
  the loader will automatically take the entire space.

- removing host properties as it interferes with flex properties for parent container

- adding flex grow to explorer table

### DIFF
--- a/projects/components/src/table/table.component.ts
+++ b/projects/components/src/table/table.component.ts
@@ -45,126 +45,133 @@ import {
 } from './table-api';
 import { TableColumnConfigExtended, TableService } from './table.service';
 
-// tslint:disable max-file-line-count
+// tslint:disable: template-cyclomatic-complexity component-max-inline-declarations max-file-line-count
 @Component({
   selector: 'ht-table',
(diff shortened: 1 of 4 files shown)


projects/components/src/table/table.component.ts
```diff
@@ -46,116 +46,129 @@ import {
 } from './table-api';
 import { TableColumnConfigExtended, TableService } from './table.service';
 
-// tslint:disable max-file-line-count
+// tslint:disable: template-cyclomatic-complexity component-max-inline-declarations max-file-line-count
```
aaron-steinfeld, line 49: 😨

aaron-steinfeld (approved):

jake-bassett (approved):

PR #311, opened 2020-10-30
fix: define vertical divider style in shared globals

arjunlalb:


projects/assets-library/assets/styles/_shared-globals.scss
```diff
@@ -4,3 +4,9 @@
   background: $gray-7;
   opacity: 0.62 !important;
 }
+
+.vertical-divider {
+  height: 16px;
```
aaron-steinfeld, line 9: Nit: I'd think it would be on the class user to add on a height, since each use case is likely different

arjunlalb: Right now, couldn't find another use case. But even in that case, I think we can make just the height parameterized.

aaron-steinfeld: We can remove it later, but I'd imagine we'd either use it like `class="vertical-divider" style:"height: 20px'` or as a parameterized mixin if we're writing our own css anyway - we'd need to include the height somewhere on the caller side though .

arjunlalb: Yea. I can make this a parameterized mixin.

aaron-steinfeld (approved):

skjindal93 (approved):

PR #312, opened 2020-10-30
feat: open in new tab component

arjunlalb:


projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  @Input()
+  public size?: ButtonSize = ButtonSize.Medium;
+
+  public constructor(
+    private readonly navigationService: NavigationService,
+    private readonly timeRangeService: TimeRangeService
+  ) {}
+
+  public openInNewTab(): void {
+    merge(this.navigationService.navigation$, this.timeRangeService.getTimeRangeAndChanges())
```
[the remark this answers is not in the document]
arjunlalb: Yes. I'm planning on adding that as well. Will be handy to create URLs for overlays and supply that in if individual pages exist for those.

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  public constructor(
+    private readonly navigationService: NavigationService,
+    private readonly timeRangeService: TimeRangeService
+  ) {}
+
+  public openInNewTab(): void {
+    this.navigationService.navigate({
+      navType: NavigationParamsType.External,
+      windowHandling: ExternalNavigationWindowHandling.NewWindow,
+      // Use input url if available. Else construct a shareable URL for the page
```
anandtiwary, line 42: Is the else part valid or should we throw an error if the url is missing?

anandtiwary: like, would we ever have a case when we would want to open the current url in a new tab through a button?

arjunlalb: I think its good to have. Example, I'm on the security detail page I want to open this in new tab and keep the history on the current tab. I think this button would be handy compared to copy-pasting the link(which isn't a permalink)  from address bar.

aaron-steinfeld (approved):

PR #313, opened 2020-11-01
feat: notification service

arjunlalb:
Success snackbar autodismisses in 5s.
Screenshot 2020-11-01 at 8 18 32 PM

Failure snackbar does not auto dismiss.
Screenshot 2020-11-01 at 8 19 35 PM

Info snackbar autodismisses in 5s. (This is the latest style. Border on the dismiss icon has been removed and the snackbar has rounded corners)
Screenshot 2020-11-01 at 10 04 36 PM


src/app/root.module.ts
```diff
@@ -18,7 +19,8 @@ import { NavigationModule } from './shared/navigation/navigation.module';
     NavigationModule,
     HttpClientModule,
     ApplicationFrameModule,
-    ObservabilityDashboardModule
+    ObservabilityDashboardModule,
+    MatSnackBarModule
```
aaron-steinfeld, line 23: This shouldn't be provided directly, components which use it should import our snackbar module - the usage of material should be isolated inside that mmodule.

projects/components/src/snackbar/snackbar.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    horizontalPosition: 'left',
+    verticalPosition: 'bottom',
+    duration: 5000 // 5 seconds in ms
+  };
+
+  private snackbarRef?: MatSnackBarRef<unknown>;
+  private readonly closedObserver$: Subject<unknown> = new Subject();
+
+  public constructor(private readonly snackbar: MatSnackBar) {
+    this.closedObserver$.subscribe(() => {
```
aaron-steinfeld, line 18: So this means we never support multiple concurrent snackbars. Generally the case, but is that the explicit intention? If not, we could keep the ref and subject local to each function call with the same effect (not sure if the underlying lib supports multiple or not anyway, I'm assuming it does).

projects/components/src/snackbar/snackbar.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  `
+})
+export class SnackbarComponent {
+  public iconType?: IconType;
+  public constructor(@Inject(MAT_SNACK_BAR_DATA) public readonly matData: SnackbarData) {
+    this.setIconType(this.matData.mode);
+  }
+
+  public setIconType(mode: SnackbarMode): void {
+    switch (mode) {
```
aaron-steinfeld, line 37: nit: functions that mutate instance variables are generally more brittle. This is a tiny nit in this case, but a good practice more generally. If this one returned an icon type, it'd be pure function, and the less functions that do mutation, the easier it is to understand the logic flows. Specifically in this cases, with a switch statement, allows avoiding `break` which is always good and it would allow iconType to be readonly because the assignment is all in the constructor.

Also, doesn't need to be public

projects/components/src/snackbar/snackbar.component.ts
```diff
@@ -0,0 +1,67 @@
...
+        assertUnreachable(mode);
+    }
+  }
+
+  public close(): void {
+    this.matData.closedObserver.next(undefined);
+  }
+}
+
+export enum SnackbarMode {
```
aaron-steinfeld, line 57:
```suggestion
export const enum SnackbarMode {
```

projects/components/src/snackbar/snackbar.component.ts
```diff
@@ -0,0 +1,67 @@
...
+      case SnackbarMode.Info:
+        this.iconType = IconType.Info;
+        break;
+      default:
+        assertUnreachable(mode);
+    }
+  }
+
+  public close(): void {
+    this.matData.closedObserver.next(undefined);
```
aaron-steinfeld, line 53: If we used an `Observer<void>` instead of an `Observer<unknown>` we should be able to avoid the unnecessary undefined arg.

aaron-steinfeld (commented):
Looks good, small nits

projects/components/src/snackbar/snackbar.service.ts
arjunlalb: the openFromComponent method in MatSnackbar library supports having one snackbar at a time. It dismisses the currently open one when a new one is opened. I was trying to have that single snackbar usecase in this iteration.

src/app/root.module.ts
arjunlalb: I was actually trying to do this. But interestingly enough, without doing this import at app level, the snackbar didn't work. I'm still trying to work out why.

projects/components/src/snackbar/snackbar.service.ts
arjunlalb: By the way, do we want to rename this to NotificationService? @aaron-steinfeld

src/app/root.module.ts
aaron-steinfeld: Hmm - this is likely to do with the a root service getting a root injector, which is not being able to find a lazy loaded component. Instead of providing the service in root, provide it in the module - see how snackbar itself is written: https://github.com/angular/components/blob/master/src/material/snack-bar/snack-bar.ts#L48

projects/components/src/snackbar/snackbar.service.ts
aaron-steinfeld: I do like that name better. We definitely don't need to build out support for multiple concurrent right now, I was just wondering if we were explicitly hiding it (but sounds like it never existed).

src/app/root.module.ts
arjunlalb: @aaron-steinfeld Tried that as well. Still doesn't work without importing MatSnackbarModule in the root module. Possible that I'm missing something.

projects/components/src/notification/notification.component.ts
```diff
@@ -0,0 +1,61 @@
...
+        assertUnreachable(this.matData.mode);
+    }
+  }
+
+  public close(): void {
+    this.matData.closedObserver.next();
+  }
+}
+
+export const enum SnackbarMode {
```
aaron-steinfeld, line 51: Let's update these names too

projects/components/src/notification/notification.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  public getStatusIconType(): IconType | undefined {
+    switch (this.matData.mode) {
+      case SnackbarMode.Success:
+        return IconType.Checkmark;
+      case SnackbarMode.Failure:
+        return IconType.Close;
+      case SnackbarMode.Info:
+        return IconType.Info;
+      default:
+        assertUnreachable(this.matData.mode);
```
aaron-steinfeld, line 42: return here so this always returns an IconType

projects/components/src/public-api.ts
```diff
@@ -285,3 +285,7 @@ export * from './modal/modal.service';
 
 export * from './confirmation/confirmation.module';
 export * from './confirmation/confirmation.service';
+
+// Snackbar
```
aaron-steinfeld, line 289: notification

src/app/root.module.ts
aaron-steinfeld: Will take a look

arjunlalb: Thank you!

projects/components/src/snackbar/snackbar.service.ts
anandtiwary: I do think we should only show one snack bar at a time. We can show one message for the overall status instead of the individual errors which we can just on the actual page/modal itself.

projects/components/src/notification/notification.service.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { NotificationModule, NotificationService } from '@hypertrace/components';
+import { createServiceFactory, mockProvider, SpectatorService } from '@ngneat/spectator';
+import { NotificationComponent, NotificationMode } from './notification.component';
+
+describe('NotificationService', () => {
+  let spectator: SpectatorService<NotificationService>;
+
+  const createService = createServiceFactory({
+    service: NotificationService,
+    imports: [NotificationModule],
```
aaron-steinfeld, line 11: Is the import needed? Shouldn't be to test a service

projects/components/src/notification/notification.service.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    providers: [mockProvider(MatSnackBar, { openFromComponent: jest.fn().mockReturnValue(undefined) })]
```
aaron-steinfeld, line 12: `mockReturnValue(undefined)` that's the default return value anyway for a mock

projects/components/src/notification/notification.service.test.ts
arjunlalb: not needed. removed.

projects/components/src/notification/notification.component.test.ts
```diff
@@ -0,0 +1,91 @@
...
+import { Subject } from 'rxjs';
+import { NotificationComponent, NotificationMode } from './notification.component';
+
+describe('NotificationComponent', () => {
+  let spectator: SpectatorHost<NotificationComponent>;
+
+  const createHost = createHostFactory({
+    component: NotificationComponent,
+    declareComponent: false,
+    imports: [IconLibraryTestingModule, NotificationModule],
```
aaron-steinfeld, line 16: Any reason to make this test deep?

projects/components/src/notification/notification.component.test.ts
```diff
@@ -12,9 +11,9 @@ describe('NotificationComponent', () => {
 
   const createHost = createHostFactory({
     component: NotificationComponent,
-    declareComponent: false,
-    imports: [IconLibraryTestingModule, NotificationModule],
-    providers: [...mockDashboardProviders]
+    shallow: true,
+    declarations: [IconComponent],
```
aaron-steinfeld, line 15: Can we mock the icon component?

aaron-steinfeld: (and remove the associated import)

arjunlalb: yes. done.

aaron-steinfeld (approved):

PR #321, opened 2020-11-03
feat: toggleable flat apis for domain and service lists

jake-bassett:


projects/components/src/table/data/table-cdk-data-source-api.ts
```diff
@@ -5,6 +5,7 @@ import { TableDataSource } from './table-data-source';
 
 export interface TableDataSourceProvider {
   data?: TableDataSource<TableRow>;
+  mode?: TableMode;
```
aaron-steinfeld, line 8: I thought the table controls change was all meant to avoid changing table to support the flattened use case.This appears to read like we're giving the mode to the cdk data source just so it can spit it back out with the request - but why is that necessary? Whatever is receiving the request presumably knows whether it's flattened or not (to provide this in the first place)

projects/distributed-tracing/src/shared/dashboard/data/graphql/table/table-data-source.model.ts
```diff
@@ -19,9 +26,9 @@ export abstract class TableDataSourceModel extends GraphQlDataSourceModel<TableD
 
   public getData(): Observable<TableDataSource<TableRow, SpecificationBackedTableColumnDef>> {
     return observableOf({
-      getData: request =>
-        this.query(filters => this.buildGraphQlRequest(filters, request)).pipe(
-          map(response => this.buildTableResponse(response, request))
+      getData: (request, mode) =>
```
aaron-steinfeld, line 29: So this is where it flushes out, but we have active mode available in the renderer that could pass this down - which we're already doing for getting columns. Or we could do something else like switch to a different data source for flattened data, set a variable, etc.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
```diff
@@ -65,6 +65,13 @@ export class TableWidgetColumnModel {
   })
   public filterable?: boolean = false;
 
+  @ModelProperty({
```
aaron-steinfeld, line 68: I'm really not a fan of introducing this concept at the generic table column level. Can we switch to whole different set of columns for a flattened list? Perhaps create a new flattenable tree table data source or something?

projects/observability/src/pages/apis/services/services-list.dashboard.ts
```diff
@@ -84,7 +100,15 @@ export const servicesListDashboard: DashboardDefaultConfiguration = {
     data: {
       type: 'entity-table-data-source',
       entity: 'SERVICE',
-      childEntity: 'API'
+      childEntity: 'API',
+      flattenedFilters: [
```
aaron-steinfeld, line 104: This does not appear related

projects/observability/src/pages/apis/services/services-list.dashboard.ts
```diff
@@ -19,6 +24,17 @@ export const servicesListDashboard: DashboardDefaultConfiguration = {
           type: 'entity-specification'
         }
       },
+      {
+        type: 'table-widget-column',
+        title: 'Service Name',
+        display: CoreTableCellRendererType.Text,
```
aaron-steinfeld, line 30: Shouldn't this be an entity renderer?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
```diff
@@ -172,8 +172,14 @@ export class TableWidgetModel {
...
-  public getColumns(): SpecificationBackedTableColumnDef[] {
-    return this.columns.map(column => column.asTableColumnDef());
+  public getColumns(mode?: TableMode): SpecificationBackedTableColumnDef[] {
+    return this.columns
+      .filter(column => this.filterFlattenedTreeColumns(column, mode ?? this.mode))
+      .map(column => column.asTableColumnDef());
+  }
+
+  private filterFlattenedTreeColumns(column: TableWidgetColumnModel, mode: TableMode): boolean {
+    return !column.flattenedOnly || (column.flattenedOnly && this.modeOptions.length > 0 && mode === TableMode.Flat);
```
aaron-steinfeld, line 182: Why the length check? Shouldn't a flattened only col always show up if mode is flat?

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
```diff
@@ -47,11 +46,33 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
   })
   public additionalChildSpecifications: Specification[] = [];
 
+  @ModelProperty({
+    key: 'flattenedFilters',
+    type: ARRAY_PROPERTY.type
+  })
+  public flattenedFilters: GraphQlFilter[] = [];
```
aaron-steinfeld, line 53: What do flattened filters do? Again, same concern about putting specific use case props on general purpose widgets.

projects/observability/src/pages/apis/services/services-list.dashboard.ts
aaron-steinfeld: If we make this an entity spec that fetches serviceName and serviceId, we'd be able to render this as a full entity supporting drilldown

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
aaron-steinfeld: may be moot given the other comments, but If I had a flat table, I'd expect flattenedOnly columns to show up regardless  of having a mode toggle or not.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
aaron-steinfeld: My suggestion would be something like (in pseudo-model):

```
{
   type: 'flattenable-tree-table',
   flat: {
       ...
   }
   tree: {
       ...
   }
}

```

PR #322, opened 2020-11-03
fix: adding event blocker and introducing disable property in menu component

anandtiwary:


projects/components/src/menu-dropdown/menu-dropdown.component.ts
```diff
@@ -32,6 +32,9 @@ export class MenuDropdownComponent {
   @Input()
   public label?: string;
 
+  @Input()
+  public disabled: boolean | undefined;
```
aaron-steinfeld, line 36: Any styling changes to indicate that it's disabled?

projects/components/src/menu-dropdown/menu-dropdown.component.ts
```diff
@@ -8,7 +8,7 @@ import { MenuItemComponent } from './menu-item/menu-item.component';
   styleUrls: ['./menu-dropdown.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="menu-dropdown">
+    <ht-event-blocker event="click" class="menu-container" [enabled]="this.disabled">
```
arjunlalb, line 11: same class for parent and child?

also, why are there multiple PRs modiying the same thing?

projects/components/src/menu-dropdown/menu-dropdown.component.ts
anandtiwary: yep. got sidetracked by insights issues

projects/components/src/menu-dropdown/menu-dropdown.component.ts
anandtiwary: yeah fixing the css. As to multiple PRs, just needed to change its base. Create separate PRs since they were touching different components

aaron-steinfeld (approved):

PR #325, opened 2020-11-04
refactor: adding a debouce to table datasource

anandtiwary:


projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -54,6 +54,7 @@ export class TableCdkDataSource implements DataSource<TableRow> {
     this.buildChangeObservable()
       .pipe(
         tap(() => this.loadingStateSubject.next({ loading$: NEVER })),
+        debounceTime(100),
```
arjunlalb, line 57: can we add a comment about the reasons for the debounce so that we don't forget later? :)

anandtiwary: Please refer to #326

PR #336, opened 2020-11-06
fix: toggle button group updates

arjunlalb:
## Description
Take justifyContent property as an input to toggle button group.
Minor CSS updates.

### Testing
Manual verification.

### DIFF
--- a/projects/components/src/toggle-button-group/toggle-button-group.component.ts
+++ b/projects/components/src/toggle-button-group/toggle-button-group.component.ts
@@ -21,7 +21,7 @@ import { ToggleButtonState, ToggleViewMode } from './toggle-button';
   providers: [SubscriptionLifecycle],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="toggle-button-group">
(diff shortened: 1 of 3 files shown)


projects/components/src/toggle-button-group/toggle-button-group.component.ts
```diff
@@ -39,6 +39,9 @@ export class ToggleButtonGroupComponent implements OnChanges, AfterViewInit {
   @Input()
   public disableInitialSelection?: boolean;
 
+  @Input()
+  public justifyContent?: string;
```
skjindal93, line 43: Should we follow an enum based mode approach, instead of passing styles directly as an input?

arjunlalb: I've already made those changes. will update PR soon :)

arjunlalb: @skjindal93 done

aaron-steinfeld (commented):
The implementation is fine, but I'd like to discuss where we might see using spaced out toggle buttons before adding this.

aaron-steinfeld (approved):
Discussed offline - in full page views, we generally want default justification, while modal views are centered/space between.

PR #346, opened 2020-11-10
fix: fix css for medium modal dimensions

arjunlalb:
## Description
Fixing CSS for medium size modal

### DIFF
--- a/projects/components/src/modal/modal-container.component.scss
+++ b/projects/components/src/modal/modal-container.component.scss
@@ -17,8 +17,8 @@
   }
 
   &.modal-size-medium {
-    height: 456px;
-    width: 530px;
+    height: 530px;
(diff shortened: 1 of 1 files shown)


projects/components/src/modal/modal-container.component.scss
```diff
@@ -17,8 +17,8 @@
   }
 
   &.modal-size-medium {
-    height: 456px;
-    width: 530px;
+    height: 530px;
```
aaron-steinfeld, line 20: @jake-bassett you added this size for your use case right? did you need it in the other orientation?

jake-bassett: This size works for what I'm doing too.

jake-bassett (approved):

PR #347, opened 2020-11-10
feat: adding gauge widget

anandtiwary:
## Description
Adding Gauge Dashboard widget

### Testing

Added unit tests for both widget and its model. Also did some manual test. We need to ensure the widget renders the data correctly.

### Checklist:

### Documentation


projects/observability/src/public-api.ts
```diff
@@ -210,3 +214,9 @@ export * from './shared/dashboard/data/graphql/specifiers/entity-specification.m
 
 // Explorer service
 export * from './pages/explorer/explorer-service';
+
+
+// Gauge
+export * from './shared/components/gauge/gauge.component';
```
aaron-steinfeld, line 220: Doubled up gauge component/module

projects/observability/src/shared/dashboard/widgets/gauge/gauge-widget-renderer.component.test.ts
```diff
@@ -0,0 +1,79 @@
...
+import { RENDERER_API } from '@hypertrace/hyperdash-angular';
+import { createComponentFactory } from '@ngneat/spectator/jest';
+import { MockComponent } from 'ng-mocks';
+import { EMPTY, of } from 'rxjs';
+import { DonutComponent } from '../../../components/donut/donut.component';
+import { LegendPosition } from '../../../components/legend/legend.component';
+import { DonutWidgetRendererComponent } from './gauge-widget-renderer.component';
+import { DonutWidgetModel } from './gauge-widget.model';
+
+describe('Donut widget renderer component', () => {
```
aaron-steinfeld, line 12: Guessing this is not ready for review

anandtiwary: nope

projects/observability/src/public-api.ts
anandtiwary: removed

anandtiwary: Ready now

projects/observability/src/shared/dashboard/widgets/gauge/gauge-widget-renderer.component.test.ts
anandtiwary: Ready now

projects/observability/src/shared/dashboard/widgets/gauge/gauge-widget.model.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { of } from 'rxjs';
+import { GaugeWidgetData } from './gauge-widget';
+import { GaugeWidgetModel } from './gauge-widget.model';
+
+describe('Gauge widget model', () => {
+  test('uses colors from color map', () => {
+    const modelFactory = createModelFactory();
+
+    const data: GaugeWidgetData = {
+      value: 5,
```
aaron-steinfeld, line 14: Hmm - I think I'd expect the max and thresholds to come from the widget model rather than the data model - those seem like they're independent of where we actually get the value from, and would allow dropping in our existing aggregation data sources.

anandtiwary: That's a valid point. The reason I did it this way is because the api - risk graphql exposes a risk value and a risk category. So the data source I built, would convert the risk category into one of these thresholds. But I agree, these two should come from widget model. Let me talk to Ines and establish different threshold values for the risk.

aaron-steinfeld (approved):
will follow up on the data/widget separation.

PR #348, opened 2020-11-11
feat: conditional container widget

jake-bassett:
## Description
Container widget that allows a datasource to provide one of two different children[] based on a conditional query or check.

### Testing
This is not the entire implementation. Just base code to enable an extended data source.

### Checklist:


projects/dashboards/src/widgets/container/conditional-container-widget.model.ts
```diff
@@ -0,0 +1,61 @@
...
+    } as ModelModelPropertyTypeInstance
+  })
+  public layout!: ContainerLayout;
+
+  @ModelProperty({
+    key: 'trueChildren',
+    type: ARRAY_PROPERTY.type,
+    required: true
+  })
+  public trueChildren: object[] = [];
```
anandtiwary, line 42: 🤣

projects/dashboards/src/widgets/container/conditional-container-widget.model.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  @ModelProperty({
+    key: 'falseChildren',
+    type: ARRAY_PROPERTY.type,
+    required: true
+  })
+  public falseChildren: object[] = [];
```
anandtiwary, line 49: 🤣

projects/dashboards/src/widgets/container/conditional-container-widget.model.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  @ModelInject(MODEL_API)
+  private readonly api!: ModelApi;
+
+  public getData(): Observable<boolean> {
+    return this.api.getData<boolean>();
```
anandtiwary, line 55: Container may not have any data source. So why would this ever emit?

projects/dashboards/src/widgets/container/conditional-container-widget.model.ts
jake-bassett: Shhhhh. Don't tell them! 😂

projects/dashboards/src/widgets/container/conditional-container-widget.model.ts
jake-bassett: A _conditional_ container widget will require a datasource to determine which children to return.

projects/dashboards/src/widgets/container/container-widget-renderer.component.ts
```diff
@@ -18,15 +20,17 @@ import { ContainerWidgetModel } from './container-widget.model';
     </div>
   `
 })
-export class ContainerWidgetRendererComponent extends WidgetRenderer<ContainerWidgetModel> implements AfterViewInit {
+export class ContainerWidgetRendererComponent
+  extends WidgetRenderer<ContainerWidgetModel | ConditionalContainerWidgetModel>
```
aaron-steinfeld, line 24: So two issues with this approach that I see -
1. The renderer now has to know about different types of models (not a huge deal)
2. The conditional works at a container level - this seems more problematic. If I want to condition one widget (which is more likely the case), does that mean it gets padded? What if the condition should change the layout? In my mind, I was thinking the conditional would always work with a single widget - that widget however could be a container itself, and handle its own layout.

I think one nice to have thing with the single widget approach is that if we do eventually add support for conditional models, it would use the exact same model.

jake-bassett: Yeah, I had the same thoughts but actually stopped and stashed the new `conditional-container-widget.model.ts` and `conditional-container-widget-renderer.component.ts` I was going to with to get feedback on this first.

Is that the suggestion then? We want this to have its own model widget model and renderer?

aaron-steinfeld: That'd be my suggestion, but of course alway welcome to push back. I would expect it to be no more code than this change (or close enough) and touch fewer existing files. It's trading a new pass through renderer for the changes to the existing renderer/container widget models. I also like the idea of going with simply `conditional.model.ts`- We can implement that by associating a renderer right now, but once we build out model support any JSON using this for widgets remains identical (because a widget is jut a model with an associated renderer)

projects/dashboards/src/widgets/conditional/conditional.model.ts
```diff
@@ -0,0 +1,34 @@
...
+  @ModelProperty({
+    key: 'true',
+    type: PLAIN_OBJECT_PROPERTY.type,
+    required: true
+  })
+  public true!: ModelJson;
+
+  @ModelProperty({
+    key: 'false',
+    type: PLAIN_OBJECT_PROPERTY.type,
```
aaron-steinfeld, line 19: ModelTemplatePropertyType.TYPE

projects/dashboards/src/widgets/conditional/conditional.model.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  @ModelInject(MODEL_API)
+  private readonly api!: ModelApi;
+
+  public getData(): Observable<boolean> {
+    return this.api.getData<boolean>();
+  }
+
+  public getModel(): Observable<ModelJson> {
+    return this.getData().pipe(map(result => (result ? this.true : this.false)));
```
aaron-steinfeld, line 32: Based on how we're using this in the renderer, we should hydrate this modelJson, and return a real model - something like `this.api.createChild(result ? this.true : this.false)`

projects/dashboards/src/widgets/conditional/conditional-widget-renderer.component.ts
```diff
@@ -0,0 +1,30 @@
...
+import { WidgetRenderer } from '../widget-renderer';
+import { ConditionalModel } from './conditional.model';
+
+@Renderer({ modelClass: ConditionalModel })
+@Component({
+  selector: 'ht-conditional-widget',
+  styleUrls: ['./conditional-widget-renderer.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="conditional-widget">
```
aaron-steinfeld, line 14: do we need the div? the instantiated widget should have its own container

projects/dashboards/src/widgets/conditional/conditional-widget-renderer.component.ts
```diff
@@ -0,0 +1,30 @@
...
+      <ng-container [hdaDashboardModel]="this.getModel() | async"> </ng-container>
+    </div>
+  `
+})
+export class ConditionalWidgetRendererComponent extends WidgetRenderer<ConditionalModel> {
+  @ViewChild('containerContent', { read: ViewContainerRef, static: true })
```
aaron-steinfeld, line 20: Doesn't appear used

projects/dashboards/src/widgets/conditional/conditional-widget-renderer.component.ts
```diff
@@ -8,23 +8,15 @@ import { ConditionalModel } from './conditional.model';
...
 @Component({
   selector: 'ht-conditional-widget',
-  styleUrls: ['./conditional-widget-renderer.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
-  template: `
-    <div class="conditional-widget">
-      <ng-container [hdaDashboardModel]="this.getModel() | async"> </ng-container>
-    </div>
-  `
+  template: ` <ng-container [hdaDashboardModel]="this.getChildModel() | async"> </ng-container> `
```
aaron-steinfeld, line 12: by virtue of making this our fetch data implementation, this is already available as `this.data$` - we could inline line 16 into fetchData.

projects/dashboards/src/widgets/conditional/conditional-widget-renderer.component.ts
```diff
@@ -8,23 +8,15 @@ import { ConditionalModel } from './conditional.model';
...
 })
 export class ConditionalWidgetRendererComponent extends WidgetRenderer<ConditionalModel> {
-  @ViewChild('containerContent', { read: ViewContainerRef, static: true })
-  public container!: ViewContainerRef;
-
-  public getModel(): Observable<ModelJson> {
-    return this.model.getModel().pipe(takeUntil(this.destroyed$));
+  public getChildModel(): Observable<object> {
+    return this.model.getChildModel().pipe(takeUntil(this.destroyed$));
```
aaron-steinfeld, line 16: nit: this takeUntil isn't a problem but not necessary ( we used to use these a lot more if you were looking at old code - now we try instead where possible to push the subscriptions into the DOM which will auto unsubscribe.

aaron-steinfeld (approved):

projects/dashboards/src/widgets/conditional/conditional-widget-renderer.component.ts
jake-bassett: I'll fix this locally. Thanks.

projects/dashboards/src/widgets/conditional/conditional-widget-renderer.component.ts
jake-bassett: Will fix locally. Thanks.

PR #352, opened 2020-11-12
fix: adding layout trigger to titled container

anandtiwary:
- Triggering layout initialization at the time of creation rather
 than after view init.

## Description

### Testing

### Checklist:

### Documentation


projects/components/src/layout/layout-change.directive.ts
```diff
@@ -1,19 +1,16 @@
...
+export class LayoutChangeDirective {
   @Output('htLayoutChange')
   public readonly changeEmitter: EventEmitter<void> = new EventEmitter();
 
   public constructor(private readonly layoutChange: LayoutChangeService, subscriptionLifecycle: SubscriptionLifecycle) {
     subscriptionLifecycle.add(layoutChange.layout$.subscribe(this.changeEmitter));
-  }
-
-  public ngAfterViewInit(): void {
     this.layoutChange.initialize();
```
anandtiwary, line 14: @aaron-steinfeld  This is the fix for the layout change issue. Since we initialize/register the bounding element after view init, during this time the bounding element already has the new width/height. When layout trigger emits a new dom change, such events get filtered out as it doesn't see any change in dimension.

projects/components/src/titled-content/titled-content.component.ts
```diff
@@ -10,7 +10,7 @@ import { TitledHeaderControlDirective } from './header-controls/titled-header-co
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div class="titled-content-container">
-      <div class="header">
+      <div class="header" [htLayoutChangeTrigger]="this.shouldShowHeader">
```
anandtiwary, line 13: Using this trigger here fails a lot of test. We now have to provide layout change service almost everywhere. Mocking the directive didn't help. Was wondering if we could avoid it.

projects/components/src/layout/layout-change.directive.ts
aaron-steinfeld: Hmm I think I understand, but if everything is sized correctly after view init, do we even need the layout change? We probably shouldn't do our first measurement in any rendering code until after view init.

anandtiwary: It's not sized correctly. Here is what happens:
1. Gauge component gets initialized.
2. At this time it gets 200px height
3. htlayoutChange directive gets initialized but the host element is not registered. 
3. On first input change, Gauge component renders the widget with 200px height
4. Titled content components view changes and it reduces the height of gauge from 200 to 170px
5. htlayoutChange now registers host element with layout service
6. Gauge is now registered in layout service with 170px height
7. htLayoutChangeTrigger triggers a layout change. This publish gets filtered out when it compares the current height of gauge with its registered height.
8. Nothings gets emitted.

projects/components/src/titled-content/titled-content.component.ts
anandtiwary: It didn't. It somehow still asking for LayoutService. Need to look more

projects/components/src/layout/layout-change.directive.ts
aaron-steinfeld: So if we move the first render in gauge afterViewInit would that fix the issue? And would it make sense? I'm thinking yes since it's sizing sensitive, it also prevents the double render.

anandtiwary: I have seen a delay when i tried this last week. But shouldn't we fix this from the layout directive itself as it is likely to affect other places as well?

aaron-steinfeld: Can we try it in onInit instead of the constructor? That would have the same affect, right?

anandtiwary: lemme try

projects/components/src/titled-content/titled-content.component.ts
anandtiwary: Added a new mockDashboardWidgetProviders.

projects/components/src/layout/layout-change.directive.ts
anandtiwary: Didn't work.

projects/common/src/layout/layout-change.service.ts
```diff
@@ -20,13 +20,15 @@ export class LayoutChangeService implements OnDestroy {
...
-      : this.getWindowResizeEvents();
 
     // Filter out layout changes from our parent if we didn't change size
-    parentChange$.pipe(filter(event => this.hasLayoutChanged(event))).subscribe(this.layoutChangeSubject);
+    (this.parentLayoutChange?.getLayoutChangeEventObservable() ?? this.getWindowResizeEvents())
+      .pipe(filter(event => this.hasLayoutChanged(event)))
+      .subscribe(this.layoutChangeSubject);
+  }
+
+  public getLayoutChangeEventObservable(): Observable<LayoutChangeEvent> {
```
anandtiwary, line 30: Changed this for testing purposes. Was unable to create mock service without it

aaron-steinfeld (approved):

PR #356, opened 2020-11-13
fix: get column should rebuild on model changes

anandtiwary:
## Description
The table widget should rebuild columns when model property changes. This PR adds that missing functionality.

### Testing

### Checklist:

### Documentation


jake-bassett (dismissed):

aaron-steinfeld (commented):
I'm not a huge fan of this - other changes (refresh, time range change etc) are handled at the renderer level. Change used to be too, and it was removed. Would love to figure out why that was before we merge this.

anandtiwary:
Sure we can look into that. But this change is not bad. The model is currently building the columns, so if it reacts and rebuilds on property changes then does not deviate much from the current design. If model was not building the columns, then I would have preferred to do this in the renderer itself. 

We do need this change to hide risk column. I can create a tech debt to deal with this in an iterative fashion

aaron-steinfeld (approved):

PR #362, opened 2020-11-16
fix: edit columns uses displayName when title not provided

jake-bassett:
## Description
The only functional change here is reverting the change that broke add/remove column and instead of filtering out invisible columns keeping them available to add/remove and grabbing the displayName from metadataService for the columns if a title is not provided in the json config.

Other changes were renaming "-dashboard" to ".dashboard" and cleaning up ConditionalWidgetRenderer comments from a previous PR.


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
```diff
@@ -185,13 +187,17 @@ export class TableWidgetModel {
...
+    );
 
     if (scope === undefined || !this.fetchEditableColumns) {
-      return of(modelColumns);
+      return modelColumns;
     }
 
-    return this.tableWidgetColumnsService.fetchColumn(scope, modelColumns);
+    return modelColumns.pipe(
+      switchMap(modelColumns => this.tableWidgetColumnsService.fetchColumn(scope, modelColumns))
```
anandtiwary, line 199: Isn't this shadowing the observable variable? Can we rename it to `modelColumns$` and `modelColumns` respectively?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
```diff
@@ -91,12 +95,21 @@ export class TableWidgetColumnModel {
   })
   public sort?: TableSortDirection;
 
-  public asTableColumnDef(): SpecificationBackedTableColumnDef {
+  @ModelInject(MetadataService)
+  private readonly metadataService!: MetadataService;
+
+  public asTableColumnDef(scope?: string): Observable<SpecificationBackedTableColumnDef>{
+    return scope ? this.metadataService.getAttributeKeyDisplayName(scope, this.value.name).pipe(
```
anandtiwary, line 102: Should we use `isEmpty` instead? scope can be '' .

anandtiwary (approved):

PR #364, opened 2020-11-19
refactor: adding a new label detail component

anandtiwary:
Screen Shot 2020-11-19 at 2 06 51 PM

On Hover:
Screen Shot 2020-11-19 at 2 06 57 PM

On Click: This is over the trigger element
Screen Shot 2020-11-19 at 2 07 03 PM

## Description
Adding a Label detail component. This would allow us to show some description and additional details of any label while ensuring it doesn't take a lot of space. The description and details are showing in a popover, so that makes it easier to view.

### Testing

### Checklist:

### Documentation


projects/components/src/popover/popover-position-builder.ts
```diff
@@ -74,18 +78,25 @@ export class PopoverPositionBuilder {
...
         return new ConnectionPositionPair({ originX: 'end', originY: 'bottom' }, { overlayX: 'end', overlayY: 'top' });
+
       case PopoverRelativePositionLocation.LeftCentered:
         return new ConnectionPositionPair(
           { originX: 'start', originY: 'center' },
           { overlayX: 'end', overlayY: 'center' }
         );
+
+      case PopoverRelativePositionLocation.OverLeftAligned:
+        return new ConnectionPositionPair({ originX: 'start', originY: 'top' }, { overlayX: 'start', overlayY: 'top' });
```
aaron-steinfeld, line 89: Is this different than aboveLeftAligned? The name seems to indicate the same thing to me.

projects/observability/src/shared/components/label-detail/label-detail.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  public icon?: IconType;
+
+  @Input()
+  public label?: string;
+
+  @Input()
+  public description?: string;
+
+  @Input()
+  public additionalDetails?: string[] = [];
```
aaron-steinfeld, line 87: Do we have a screenshot of what this is supposed to be? This seems a strange API. If anything, I'd expect it to take content or a template to show details, rather than a string array.

projects/components/src/popover/popover-position-builder.ts
anandtiwary: Yes. Above left aligned puts the bottom side of the popover container above the trigger element. So it makes the entire container above the trigger element

projects/observability/src/shared/components/label-detail/label-detail.component.ts
anandtiwary: Yes. Screenshot is in description

anandtiwary: Right now the use case is just a bunch of strings. But I agree, eventually it could be more than this.

aaron-steinfeld: Got it - the bulleted list still seems a strange API to show details though. Could we change that to content or a template input?

anandtiwary: What would be the dashboard api look like then? I am still figuring that out. There are two known use case for this. I can explain on slack further

aaron-steinfeld: Sure - depends on the use case, but any renderer can convert a list of data into content or a template anyway.

anandtiwary: LGTM. I would prefer to keep this component simple enough unless more use case come up. We try to make components very generic and then, from ux side, we switch to a completely new view. For simple components like these, it is better to iterate when supporting use case come up.

If you want I can move the list div into widget renderer in a **follow up**. Although that doesn't change much from the current state. May be the generic widget should support taking child template render the model inside. But dont want to do that now.

aaron-steinfeld: I think the problem arises when we make a generic component but don't give it a generic api - if this were a component built for a use case, then it makes sense to name it as such. If it's in our component lib as a label with details, then supporting a generic content for its detail content makes sense - we'll just break the existing usages when we have to update it. We can merge it as is, but this is a principle we need to get better at for exactly the reason you state, ux changes, so we want the api to reflect the concept (an api and a detail view) more than the implementation (a list of details).

aaron-steinfeld (approved):

projects/observability/src/shared/components/label-detail/label-detail.component.ts
anandtiwary: Although I am not sure how much we want to show in detail for a label. Is a list not enough? Do we think we may have to show a widget inside? It is just supposed to be an info box, which generally only has a text inside.

anandtiwary: May be i chose a wrong name for the component.

anandtiwary: how about `label-info` component? It is supposed to show a label with icon and show a text info box popover on click.

PR #367, opened 2020-11-20
feat: cartesian metric health range

jake-bassett:
## Description
Added support for Ranges to cartesian widget. The ranges will be displayed as a shaded band bounded between two upper and lower dashed lines. There will also be an "average" or similar solid line with pointed vertices associated with the range. That work is TBD based on API implementation. It should just translate into another series we add in.

Screen Shot 2020-11-19 at 5 46 20 PM


projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -16,19 +16,22 @@ import { LegendPosition } from '../legend/legend.component';
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<div #chartContainer class="fill-container" (htLayoutChange)="this.redraw()"></div> `
 })
 export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
   @Input()
-  public series: Series<TData>[] | undefined = [];
+  public series?: Series<TData>[] | undefined = [];
+
+  @Input()
+  public range?: Range<TData>;
```
aaron-steinfeld, line 34: we don't support multiple ranges? Also, I'd recommend changing this name, since range already has a meaning in the context of a cartesian, and shows up in a number of places in code.

projects/observability/src/shared/components/cartesian/chart.ts
```diff
@@ -50,6 +60,7 @@ export const enum RenderingStrategy {
 export const enum CartesianSeriesVisualizationType {
   Column,
   Line,
+  Dashed,
```
aaron-steinfeld, line 63: nit: DashedLine, as it's just a different stroke

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -319,19 +335,19 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
 
-    new CartesianNoDataMessage(this.chartBackgroundSvgElement, this.allSeries).updateMessage();
+    new CartesianNoDataMessage(this.chartBackgroundSvgElement, this.series).updateMessage();
   }
 
   private drawLegend(): void {
     if (this.chartContainerElement) {
       if (this.legendPosition !== undefined && this.legendPosition !== LegendPosition.None) {
-        this.legend = new CartesianLegend(this.allSeries, this.injector, this.intervalData, this.summaries).draw(
+        this.legend = new CartesianLegend(this.series, this.injector, this.intervalData, this.seriesSummaries).draw(
```
aaron-steinfeld, line 344: Should the range be part of the legend?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -51,8 +55,10 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
   protected timeRange?: TimeRange;
   protected readonly requestedAxes: Axis[] = [];
   protected intervalData?: CartesianIntervalData;
-  protected summaries: Summary[] = [];
-  protected readonly allSeries: Series<TData>[] = [];
+  protected readonly series: Series<TData>[] = [];
+  protected readonly seriesSummaries: Summary[] = [];
+  protected readonly ranges: Range<TData>[] = [];
+  protected readonly rangesSummaries: Summary[] = [];
```
aaron-steinfeld, line 61: where are the range summaries used?

projects/observability/src/shared/components/cartesian/d3/data/cartesian-data.ts
```diff
@@ -4,35 +4,40 @@ import {
...
+
   protected abstract buildMultiAxisDataLookupStrategy(
     tooltipStrategy: ChartTooltipTrackingStrategy
-  ): MouseDataLookupStrategy<TData, Series<TData>>;
+  ): MouseDataLookupStrategy<TData, TVisualization>;
+
+  protected abstract buildDataLookupStrategy(
+    visualization: TVisualization,
+    strategy: ChartTooltipTrackingStrategy
+  ): MouseDataLookupStrategy<TData, TVisualization>;
```
aaron-steinfeld, line 38: Why remove the default implementation? Does every class already implement this?

projects/observability/src/shared/components/cartesian/d3/data/range/cartesian-range.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    return this.buildMultiAxisDataLookupStrategy();
+  }
+
+  private getCombinedData(range: Range<TData>): TData[] {
+    return [range.upper.data, range.lower.data].flatMap(d => d);
+  }
+
+  private getPairedData(range: Range<TData>): [TData, TData][] {
+    return range.upper.data.map((upperData: TData, index: number) => [upperData, range.lower.data[index]]);
```
aaron-steinfeld, line 48: Do we enforce that these are matching anywhere?

projects/observability/src/shared/components/cartesian/d3/data/range/cartesian-range.ts
```diff
@@ -0,0 +1,172 @@
...
+  /*
+   * SVG
+   */
+
+  public drawSvg(element: BaseType): void {
+    if (this.range.hide) {
+      return;
+    }
+
+    const group = select(element).append('g').classed(CartesianRange.CSS_RANGE_CLASS, true);
```
aaron-steinfeld, line 60: Can we use our d3 selection wrapper here?

projects/observability/src/shared/components/cartesian/d3/scale/cartesian-scale-builder.ts
```diff
@@ -186,19 +190,25 @@ export class CartesianScaleBuilder<TData> {
...
 
-    return [];
+    return [...series, ...ranges];
   }
 
   private get data(): TData[] {
-    if (this.scaleState.series) {
-      return this.scaleState.series.flatMap(series => series.data);
-    }
+    const series = this.scaleState.series !== undefined ? this.scaleState.series.flatMap(s => s.data) : [];
```
aaron-steinfeld, line 203: seriesData, rangeData

projects/observability/src/shared/components/cartesian/d3/scale/cartesian-scale-builder.ts
```diff
@@ -186,19 +190,25 @@ export class CartesianScaleBuilder<TData> {
...
+    const ranges =
+      this.scaleState.ranges !== undefined
+        ? this.scaleState.ranges.flatMap(r =>
+            [r.upper.data.flatMap(d => d), r.lower.data.flatMap(d => d)].flatMap(d => d)
```
aaron-steinfeld, line 207: This doesn't really make sense (flatmap shouldn't take an identity fn arg, and we shouldn't need 3 levels of flatmap for two levels of data). Maybe something like `this.scaleState.ranges.map(r => [r.lower.data, r.upper.data]).flat(2)`

projects/observability/src/shared/components/cartesian/d3/scale/cartesian-scale-builder.ts
```diff
@@ -186,19 +190,25 @@ export class CartesianScaleBuilder<TData> {
...
-      return this.scaleState.series.filter(series => !series.hide);
-    }
+    const series = this.scaleState.series !== undefined ? this.scaleState.series.filter(s => !s.hide) : [];
+    const ranges =
+      this.scaleState.ranges !== undefined
+        ? this.scaleState.ranges.filter(r => !r.hide).flatMap(r => [r.upper, r.lower])
+        : [];
 
-    return [];
+    return [...series, ...ranges];
```
aaron-steinfeld, line 199: This isn't really allSeries any more based on how we use the term series elsewhere - that may be ok, depending on how it's used, but it should be renamed to reflect.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -46,22 +47,27 @@ export class CartesianWidgetRendererComponent<TData> extends InteractiveDataWidg
 
   public selectedInterval?: IntervalValue;
   public intervalOptions?: IntervalValue[];
-  private fetcher?: MetricSeriesFetcher<TData>;
+  private seriesFetcher?: MetricSeriesFetcher<TData>;
```
aaron-steinfeld, line 50: I'm going to defer reviewing the dashboard code now, since that's really dependent on the graphql api. Can we split that into a separate PR?

projects/observability/src/shared/components/cartesian/d3/data/range/cartesian-range.ts
jake-bassett: Perhaps. I can try. Do you have an example?

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
aaron-steinfeld: I'm quite bad at naming too - here's a bit of brainstorming:
`Band`
`SeriesPair`
`SeriesBand`
`SeriesBounds`

projects/observability/src/shared/components/cartesian/d3/data/range/cartesian-range.ts
aaron-steinfeld: Check out `EntityNodeBoxRendererService` - it's basically just using our service's `select` method rather than the global function.

projects/observability/src/shared/components/cartesian/d3/data/cartesian-data.ts
aaron-steinfeld: Hmm - I got confused because git shows it as a rename of series, which has many implementations. Looks like this is a separate class, and series is now showing up as an add. will re-review.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -93,8 +93,8 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
         )
       );
 
-    if (this.range) {
-      this.chart.withRanges(this.range);
+    if (this.ranges !== undefined) {
```
aaron-steinfeld, line 96: better to check this via truthiness explicitly (since this could be null if provided via an async pipe, even though we don't represent it in the type)

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
jake-bassett: Going to circle back to legend and tooltip separately.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
jake-bassett: I almost went with Band. It just didn't sound as cool. Will change to that.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -111,11 +114,11 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     return this;
   }
 
-  public withRanges(...ranges: Range<TData>[]): this {
+  public withBands(...ranges: Band<TData>[]): this {
     this.ranges.length = 0;
     this.ranges.push(...ranges);
 
-    this.scaleBuilder = this.scaleBuilder.withRanges(ranges);
+    this.scaleBuilder = this.scaleBuilder.withBands(ranges);
```
aaron-steinfeld, line 121: can come back to it, but we should rename ranges the property here too

projects/observability/src/shared/components/cartesian/d3/scale/cartesian-scale-builder.ts
```diff
@@ -192,21 +192,21 @@ export class CartesianScaleBuilder<TData> {
   private get allSeriesAndRangeSeries(): Series<TData>[] {
     const series = this.scaleState.series !== undefined ? this.scaleState.series.filter(s => !s.hide) : [];
     const ranges =
-      this.scaleState.ranges !== undefined
-        ? this.scaleState.ranges.filter(r => !r.hide).flatMap(r => [r.upper, r.lower])
+      this.scaleState.bands !== undefined
```
aaron-steinfeld, line 195: also rename ranges here (allSeriesAndBands or the like)

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
aaron-steinfeld: Band doesn't sound cool?! Bands are the coolest!

aaron-steinfeld (approved):
Couple missed spots on the rename, other than that, looks good. Your call if you want to fix those separately or here.

aaron-steinfeld (approved):

PR #381, opened 2020-11-24
feat: mode toggleable service list table

jake-bassett:
## Description
This change is to allow a Parent->Child table to be toggled to Child w/ Additional columns. The use case in this change is for the Services list with API children. The change here will allow that table to be toggled between tree and flat. The flat table will add one additional column to the second visible index that states the API's Service name

### Testing
This is only enabled in Services List, so tested everything there. Scanned other tables for sanity and no issues.


projects/components/src/table/data/table-cdk-data-source-api.ts
```diff
@@ -5,6 +5,7 @@ import { TableDataSource } from './table-data-source';
 
 export interface TableDataSourceProvider {
   data?: TableDataSource<TableRow>;
+  mode: TableMode;
```
aaron-steinfeld, line 8: The widget already knows the mode and when it changes right ? Can we remove the changes to the actual table component/data source here (except for the styling ones)?

projects/distributed-tracing/src/shared/dashboard/data/graphql/table/spans/spans-table-data-source.model.ts
```diff
@@ -19,7 +19,7 @@ export class SpansTableDataSourceModel extends TableDataSourceModel {
     return SPAN_SCOPE;
   }
 
-  protected buildGraphQlRequest(
+  public buildGraphQlRequest(
```
aaron-steinfeld, line 22: Why do these two methods need to be public now?

projects/distributed-tracing/src/shared/dashboard/data/graphql/table/table-data-source.model.ts
```diff
@@ -19,24 +26,26 @@ export abstract class TableDataSourceModel extends GraphQlDataSourceModel<TableD
...
+        this.query(filters => this.buildGraphQlRequest(filters, request, mode)).pipe(
+          map(response => this.buildTableResponse(response, request, mode))
         ),
-      getScope: () => this.getScope()
+      getScope: (mode?: TableMode) => this.getScope(mode)
     });
   }
 
-  public abstract getScope(): string | undefined;
+  public abstract getScope(mode?: TableMode): string | undefined;
```
aaron-steinfeld, line 37: mode and scope seem completely unrelated - It feels like we're just piggy backing off an existing method.

projects/observability/src/shared/dashboard/data/graphql/table/entity/mode-entity-table-data-source.model.ts
```diff
@@ -0,0 +1,65 @@
...
+import { ModelTemplatePropertyType } from '@hypertrace/dashboards';
+import {
+  GraphQlFilter,
+  SpecificationBackedTableColumnDef,
+  TableDataSourceModel
+} from '@hypertrace/distributed-tracing';
+import { Model, ModelJson, ModelProperty } from '@hypertrace/hyperdash';
+
+@Model({
+  type: 'mode-entity-table-data-source'
```
aaron-steinfeld, line 12: I was actually imagining this type of switching model a level up, switching between different table widget models based on mode that defined different sets of columns (and potentially different data sources, too). It'd still need a way of passing mode down from the renderer to the widget model, but it looks like that's already a requirement anyway.

projects/observability/src/shared/dashboard/data/graphql/table/entity/mode-entity-table-data-source.model.ts
```diff
@@ -0,0 +1,65 @@
...
+  @ModelProperty({
+    key: 'detail',
+    type: ModelTemplatePropertyType.TYPE
+  })
+  public detail!: ModelJson;
+
+  private getChildModel(mode: TableMode): TableDataSourceModel {
+    switch (mode) {
+      case TableMode.Flat:
+        return this.api.createChild<TableDataSourceModel>(this.flat);
```
aaron-steinfeld, line 36: we're creating new models here on every request and response

projects/observability/src/shared/dashboard/data/graphql/table/entity/mode-entity-table-data-source.model.ts
aaron-steinfeld: A more concrete pseudo-json:
```typescript
{
    type: 'mode-toggle-table-widget',
    flat: {
       type: 'table-widget',
       style: TableStyle.FullPage,
       searchAttribute: 'name',
       columns: [
          {
              type: 'table-widget-column',
              title: 'Name',
              width: '30%',
              display: ObservabilityTableCellType.Entity,
              value: {
                type: 'entity-specification',
                'entity-type': ObservabilityEntityType.Api
              }
         }
       ],
       data: {
        entity: 'API'
       }
    },
    tree: {
       ....
    }
  
}
```

The benefit I see in this approach is that the column models and table component can remain as is, with only a new model being added (and likely a couple changes in renderer to communicate the mode). It also leaves flexibility to use different search attributes based on mode, different data sources, change column orders, etc. - basically, it's two completely different tables, which is really how it is in the UI anyway.

aaron-steinfeld: Well I finally found the example I wrote up before and... I clearly led down the wrong path, sorry about that. https://github.com/hypertrace/hypertrace-ui/pull/321#discussion_r516949368
The goal is the same here, but because the columns aren't in the data source (my mistake there) the switch moves up a level.

projects/distributed-tracing/src/shared/dashboard/widgets/table/mode-toggle-table-widget.model.ts
```diff
@@ -0,0 +1,82 @@
...
+import { ARRAY_PROPERTY, Model, ModelJson, ModelProperty } from '@hypertrace/hyperdash';
+import { NEVER, Observable } from 'rxjs';
+import { SpecificationBackedTableColumnDef } from './table-widget-column.model';
+import { TableWidgetModel } from './table-widget.model';
+
+@Model({
+  type: 'mode-toggle-table-widget',
+  displayName: 'Mode Toggle Table Widget'
+})
+export class ModeToggleTableWidgetModel extends TableWidgetModel {
```
aaron-steinfeld, line 12: Given that this wouldn't use most of the values from the parent, polymorphism may not make sense here. That is, it wouldn't make sense to define columns on a mode-toggle-table-widget which always gets columns from its delegate. If you do want to use polymorphism over unions, I think it may make sense to make an abstract `TableWidgetModel` defining the apis needed by the renderer, then have two sibling implementations: `ModeToggleTableWidgetModel` and `BasicTableWidgetModel`

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -143,7 +146,8 @@ export class TableWidgetRendererComponent
       pairwise(),
       filter(([previous, current]) => !isEqualIgnoreFunctions(previous, current)),
       map(([_, current]) => current),
-      share()
+      share(),
+      tap(() => this.onDashboardRefresh())
```
aaron-steinfeld, line 150: Do we need this here or can we do it as part of onModeChange (which I'm guessing is why we now need it, to get new data when the mode changes too)

projects/distributed-tracing/src/shared/dashboard/widgets/table/mode-toggle-table-widget.model.ts
aaron-steinfeld: To that abstract class, we'd probably want to add a few more things (and to the delegate) - namely getters for things like searchAttribute and any other model properties that are accessed directly by the renderer.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Correct, putting it in onModeChange fixes the issue for this PR. I started with it here.

But there is an existing issue you can see in dev by starting from the Service list screen and refreshing. The list will be blank unless you hit the Refresh button at the top. It also works if you navigate in from another page. But adding the refresh here fixes the initial load issue.

aaron-steinfeld: Not sure I follow, but will try to follow up on that separately if it's existing.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
```diff
@@ -167,17 +90,17 @@ export class TableWidgetModel {
...
     key: 'fetchEditableColumns',
     displayName: 'Query for additional columns not provided',
     type: BOOLEAN_PROPERTY.type
   })
-  public fetchEditableColumns?: boolean = false;
+  public fetchEditableColumns: boolean = false;
 
   @ModelInject(MODEL_API)
-  private readonly api!: ModelApi;
+  protected readonly api!: ModelApi;
```
aaron-steinfeld, line 103: This is in the base class, so no longer needed here

aaron-steinfeld (commented):
Overall, looks good - can we revert the changes no longer needed in the data sources and component?

projects/components/src/table/table.component.ts
```diff
@@ -226,7 +227,7 @@ export class TableComponent
   public filters?: TableFilter[];
 
   @Input()
-  public mode?: TableMode = TableMode.Flat;
+  public mode: TableMode = TableMode.Flat;
```
aaron-steinfeld, line 230: nit: Inputs should always be optional to support things like an async pipe which will explicitly override a default with null.

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
```diff
@@ -47,6 +47,12 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
   })
   public additionalChildSpecifications: Specification[] = [];
 
+  @ModelProperty({
+    key: 'filters',
```
aaron-steinfeld, line 51: Filters already exists on the base gql data source model which this extends, and should be included in the filters passed in on line 61 (although it's possible there's a bug of course)

aaron-steinfeld (commented):
We should also do an inventory of all the tables in the product before merging IMO, since table's getting a reputation for being finicky

aaron-steinfeld (approved):

PR #382, opened 2020-11-24
fix: update css for entity renderer text to inherit font properties

arjunlalb:
## Description
update css for entity renderer text to inherit font properties

### Testing
Manually tested
### Checklist:
- [NA ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/projects/observability/src/shared/components/entity-renderer/entity-renderer.component.scss
+++ b/projects/observability/src/shared/components/entity-renderer/entity-renderer.component.scss
@@ -2,11 +2,13 @@
 @import 'font';
 
 .ht-entity-renderer {
+  @include body-1-regular();
   display: flex;
   flex-direction: row;
   align-items: center;
   font-weight: inherit;
-  @include body-1-regular();
+  font-size: inherit;
+  line-height: inherit;
 
   .icon {
     padding-right: 12px;


projects/observability/src/shared/components/entity-renderer/entity-renderer.component.scss
```diff
@@ -2,11 +2,13 @@
 @import 'font';
 
 .ht-entity-renderer {
+  @include body-1-regular();
```
aaron-steinfeld, line 5: Why declare the font if we're also trying to inherit it?

arjunlalb: for the defaults that are not overridden.

aaron-steinfeld: Is that anything beyond color? If it's not, can we expose a variable like default-body-color (and update the font mixins to something like)
```scss
$default-body-color:  $gray-9;
@mixin body-1-regular($color: $default-body-color) {
```
and this to just declare the color? I think it's confusing and prone to breaking to use a mixin for one property, and override the others intentionally.

arjunlalb: Anand is revamping the views. Dropping this PR.

PR #387, opened 2020-11-25
fix: extend get selection handler method

jake-bassett:


anandtiwary (approved):

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget.model.ts
```diff
@@ -102,6 +102,10 @@ export class TableWidgetModel extends TableWidgetBaseModel {
     return this.filterOptions;
   }
 
+  public getRowSelectionHandlers(): TableWidgetRowSelectionModel[] {
```
anandtiwary, line 105: Can we not access `this.rowSelectionHandlers` property directly? What do we gain by polymorphism here?

anandtiwary: I am not up to date with the new design. So feel free to ignore. I was thinking If model property is defined in the base model, then it should be made available without the function

PR #394, opened 2020-11-26
fix: mode toggle table selections

jake-bassett:


projects/distributed-tracing/src/shared/dashboard/widgets/table/mode-toggle-table-widget.model.ts
```diff
@@ -84,4 +85,9 @@ export class ModeToggleTableWidgetModel extends TableWidgetBaseModel {
   public getChildModel(row: TableRow): object | undefined {
     return this.delegateModel && this.delegateModel?.getChildModel(row);
   }
+
+  public getRowSelectionHandlers(row: TableRow): TableWidgetRowSelectionModel[] {
+    return (this.delegateModel && this.delegateModel?.getRowSelectionHandlers(row).length > 0) ?
```
anandtiwary, line 90: Is it same as `this.delegateModel?.getRowSelectionHandlers(row) ?? []` ?

anandtiwary (dismissed):

projects/distributed-tracing/src/shared/dashboard/widgets/table/mode-toggle-table-widget.model.ts
jake-bassett: Not quite since that could return undefined.

anandtiwary: if the first part returns undefined, then it would return [] to the caller

anandtiwary (approved):

PR #399, opened 2020-11-28
chore: update angular

aaron-steinfeld:
## Description
Update angular (this needs a dependency update pending review at https://github.com/hypertrace/hyperdash-angular/pull/109 ) first too.

Cleaned up some logging tests and self references, as well as a couple type issues from the stricter ts 4.x compiler.

### Testing
Ran tests and build. Verified UI locally.

### Checklist:


package.json
```diff
@@ -7,7 +7,7 @@
   },
   "license": "LicenseRef-LICENSE",
   "scripts": {
-    "postinstall": "ngcc --properties browser main --tsconfig './tsconfig.spec.json'",
+    "postinstall": "ngcc --tsconfig './tsconfig.app.json' --use-program-dependencies && ngcc --source 'node_modules/@angular'",
```
anandtiwary, line 10: What does these new options do?

projects/common/src/color/color.service.ts
```diff
@@ -38,11 +38,11 @@ export class ColorService {
   }
 
   public brighter(colorHex: string, basis: number): string {
-    return rgb(colorHex).brighter(basis).hex();
+    return rgb(colorHex).brighter(basis).formatHex();
   }
 
   public darker(colorHex: string, basis: number): string {
-    return rgb(colorHex).darker(basis).hex();
+    return rgb(colorHex).darker(basis).formatHex();
```
anandtiwary, line 45: Why this change? I dont see d3 being updated with this pr.

projects/components/src/notification/notification.component.test.ts
```diff
@@ -1,10 +1,10 @@
 import { fakeAsync } from '@angular/core/testing';
 import { MAT_SNACK_BAR_DATA } from '@angular/material/snack-bar';
 import { IconType } from '@hypertrace/assets-library';
-import { IconComponent } from '@hypertrace/components';
```
anandtiwary, line 4: Does the new compiler catches these circular dependencies in test?

anandtiwary (approved):

package.json
aaron-steinfeld: so previously, we were running ngcc across all test dependencies. The issue there is that spectator and spectator jest circularly reference each other (not sure why this wasn't an issue in the past) and break. We don't actually need spectator to be compiled for ivy though - so now we're using `--tsconfig './tsconfig.app.json' --use-program-dependencies` to only compile modules reference from the actual application, and then manually adding `ngcc --source 'node_modules/@angular'` to compile the angular testing modules used by the test bed (and we know all the first party angular ones are compatible with the ivy compiler anyway).

projects/common/src/color/color.service.ts
aaron-steinfeld: This was a deprecation called out by the linter. If I were to guess, we got a new locked version of d3-color that was still compatible with the version spec in package.json

projects/components/src/notification/notification.component.test.ts
aaron-steinfeld: Nope - I manually fixed all these because of the ngcc issue above. I thought the circular references breaking ngcc were our own, but it ended up being spectator. Since I had already fixed them, I left them in.

anandtiwary (approved):

PR #401, opened 2020-12-02
feat: update duration formatter to compute days

arjunlalb:
## Description
Update duration formatter to compute durations of >24 hours

### Testing
Unit tests added

### Checklist:

### Documentation
NA

### DIFF
--- a/projects/common/src/utilities/formatters/duration/duration-formatter.ts
+++ b/projects/common/src/utilities/formatters/duration/duration-formatter.ts
@@ -1,12 +1,19 @@
 export const durationFormatter = (duration?: number): string => {
+  const dayInMs: number = 1000 * 60 * 60 * 24;
   if (duration === undefined) {
     return '-';
   }
+  const days = Math.abs(Math.trunc(duration / dayInMs));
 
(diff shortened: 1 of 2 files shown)


skjindal93 (approved):

projects/common/src/utilities/formatters/duration/duration-formatter.ts
```diff
@@ -1,12 +1,19 @@
...
   if (duration === undefined) {
     return '-';
   }
+  const days = Math.abs(Math.trunc(duration / dayInMs));
 
-  const hours = new Date(duration).getUTCHours();
-  const minutes = new Date(duration).getUTCMinutes();
-  const seconds = new Date(duration).getUTCSeconds();
-  const milliseconds = new Date(duration).getUTCMilliseconds();
+  const date = new Date(duration);
```
aaron-steinfeld, line 8: Isn't this whole formatter already implemented in TimeDuration ? Why do we have a separate implementation here (looks like it was already here, but let's fix that). There's also inconsistent behavior here with the special case for date, but not other units.

PR #402, opened 2020-12-02
feat: define a generic no-op table cell parser

arjunlalb:
## Description
Generic object table cell parser

### Testing
None

### Checklist:

### Documentation
A generic table cell parser for no-op object parser usecases.


projects/components/src/table/cells/data-parsers/table-cell-object-parser.ts
```diff
@@ -0,0 +1,21 @@
+import { TableCellParser } from '../table-cell-parser';
+import { TableCellParserBase } from '../table-cell-parser-base';
+import { CoreTableCellParserType } from '../types/core-table-cell-parser-type';
+
+@TableCellParser({
+  type: CoreTableCellParserType.Object
+})
+export class TableCellObjectParser extends TableCellParserBase<ObjectCellData, ObjectCellData, string | undefined> {
```
aaron-steinfeld, line 8: Could we make this even more generic? For example, why does it have to even be an object?

The parseValue function could remain as is, and the parseFilterValue function could just have a bit of logic like
```typescript
if (typeof cellData === 'object' && cellData !== null && 'filterValue' in cellData) {
  return cellData.filterValue;
}
return String(cellData);
```

projects/components/src/table/cells/data-parsers/table-cell-object-parser.ts
```diff
@@ -0,0 +1,21 @@
...
+    return cellData;
+  }
+
+  public parseFilterValue(cellData: ObjectCellData): string | undefined {
+    return this.parseValue(cellData).filterValue;
+  }
+}
+
+// Since object is very generic, using an explicit key fr default filter value
+interface ObjectCellData extends Object {
```
aaron-steinfeld, line 19: `Object` should only be used as a constructor, instead use lowercased `object` when referencing a type (similar to `String` and `string`)

projects/components/src/table/cells/data-parsers/table-cell-object-parser.ts
arjunlalb: Made everything unknown.

projects/components/src/table/cells/types/core-table-cell-parser-type.ts
```diff
@@ -4,5 +4,6 @@ export const enum CoreTableCellParserType {
   Number = 'number',
   String = 'string',
   Timestamp = 'timestamp',
-  Icon = 'icon'
+  Icon = 'icon',
+  Object = 'object'
```
aaron-steinfeld, line 8: This value would update too. Do you prefer the name generic over NoOp ? I feel like Noop describes what it's doing better

projects/components/src/table/cells/types/core-table-cell-parser-type.ts
```diff
@@ -5,5 +5,5 @@ export const enum CoreTableCellParserType {
   String = 'string',
   Timestamp = 'timestamp',
   Icon = 'icon',
-  Object = 'object'
+  NoOp = 'noOp'
```
aaron-steinfeld, line 8: Nittiest nit:
```suggestion
  NoOp = 'noop'
```

aaron-steinfeld (approved):

aaron-steinfeld (approved):

PR #408, opened 2020-12-03
refactor: separating child entity as separate table data source

anandtiwary:
## Description
 refactor: separating child entity as separate table data source

- This would allow us to add custom filters and entity specific attributes
  to child data source

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/pages/apis/services/services-list.dashboard.ts
+++ b/projects/observability/src/pages/apis/services/services-list.dashboard.ts
@@ -82,7 +82,10 @@ const treeTableWidget: ModelJson = {
   data: {
     type: 'entity-table-data-source',
     entity: 'SERVICE',
-    childEntity: 'API'
+    'child-data-source': {
+      type: 'entity-table-data-source',
+      entity: 'API'
+    }
   }
 };
 
--- a/projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
@@ -14,7 +14,7 @@ import {
   STRING_PROPERTY
 } from '@hypertrace/hyperdash';
 import { EMPTY, Observable } from 'rxjs';
-import { map, mergeMap } from 'rxjs/operators';
+import { map } from 'rxjs/operators';
 import { Entity, EntityType } from '../../../../../graphql/model/schema/entity';
 import { GraphQlEntityFilter } from '../../../../../graphql/model/schema/filter/entity/graphql-entity-filter';
 import { EntitiesResponse } from '../../../../../graphql/request/handlers/entities/query/entities-graphql-query-builder.service';
@@ -57,14 +57,10 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
     return this.entityType;
   }
 
-  protected parentEntityFilter?: GraphQlEntityFilter;
-
   protected buildGraphQlRequest(
     filters: GraphQlFilter[],
     request: TableDataRequest<SpecificationBackedTableColumnDef>
   ): GraphQlEntitiesQueryRequest {
-    const inheritedFilters = [...filters, ...this.toGraphQlFilters(request.filters)];
-
     return {
       requestType: ENTITIES_GQL_REQUEST,
       entityType: this.entityType,
@@ -75,7 +71,7 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
         direction: request.sort.direction,
         key: request.sort.column.specification
       },
-      filters: this.parentEntityFilter ? [...inheritedFilters, this.parentEntityFilter] : inheritedFilters,
+      filters: [...filters, ...this.toGraphQlFilters(request.filters)],
       timeRange: this.getTimeRangeOrThrow(),
       includeTotal: true
     };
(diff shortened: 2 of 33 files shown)


projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
```diff
@@ -117,15 +105,26 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
 
   private queryEntityChildren(
     entity: Entity,
-    request: TableDataRequest<SpecificationBackedTableColumnDef>
+    parentRequest: TableDataRequest<SpecificationBackedTableColumnDef>
   ): Observable<TableRow[]> {
-    return this.query<EntitiesGraphQlQueryHandlerService>(this.buildChildEntityRequest(request, entity)).pipe(
-      map(response => this.resultsAsTreeRows(response.results, request, false))
+    this.childEntityDataSource!.parentEntityFilter = GraphQlEntityFilter.forEntity(entity);
```
aaron-steinfeld, line 110: Do we need to introduce state for this? We could just as easily introduce a private method like
`private getDataForParentEntity(entity: Entity): Observable<TableRow[]>`

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
```diff
@@ -117,15 +105,26 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
...
-  tableRequestType: 'root' | 'children';
-  tableRequest: TableDataRequest;
+  private buildChildTableRequest(
+    parentRequest: TableDataRequest<SpecificationBackedTableColumnDef>
+  ): TableDataRequest<SpecificationBackedTableColumnDef> {
+    return {
+      columns: parentRequest.columns,
+      position: {
+        startIndex: 0,
+        limit: 1000
```
aaron-steinfeld, line 125: can we bump this limit down to 40 to match prior behavior? Or at least 100. no reason to fetch 1000 children if there happen to be that many

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
anandtiwary: What would that do? I need a way for the parent to apply an entity filter on parent id.  This filter would change as user clicks on different rows.

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
anandtiwary: Yes. Will update this.

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
aaron-steinfeld: So I'm suggesting instead of
```typescript
this.childEntityDataSource!.parentEntityFilter = GraphQlEntityFilter.forEntity(entity);
return this.childEntityDataSource!.getData()...
```
we update it to
```typescript
return this.childEntityDataSource!.getDataForParentEntity(entity)...
```

It would work pretty much the same, it just prevents any bugs coming from missing assumptions between the caller and callee side since now those are represented in the function contract.

anandtiwary: If i can convert this entity filter into a table filter, then the code will be much cleaner. I can just add it to the request generated below.

aaron-steinfeld: That works too - It's the same code either way right? Just whether we convert the entity to a filter in the parent v child?
I was favoring passing an entity instead of passing a filter because it felt like it kept the responsibility for how to build a scoped request all in the child data source rather than split between the parent and child.

anandtiwary: Cool. Makes sense. 

Can we also modify the `getData` of the Entity-table-data-source or Table data source itself, to take an optional entity filter? Something like this

```
  public getData(entityFilter?: GraphqlFilter): Observable<TableDataSource<TableRow, SpecificationBackedTableColumnDef>> {
    return observableOf({
      getData: request =>
        this.query(filters => this.buildGraphQlRequest(filters.concat(entityFilter), request)).pipe(
          map(response => this.buildTableResponse(response, request))
        ),
      getScope: () => this.getScope()
    });
  }
```

anandtiwary: We would still have to build the child request in parent though. Example, this new request should have the correct page start index and limit and the column specifications.

aaron-steinfeld: I thought about overloading getData, but it felt potentially confusing since that api is public and called externally, while keeping a separate one allows us to keep it private and the argument required.

anandtiwary: i think the code will be much cleaner if I can convert the parent entity filter into a table filter and pass it in the child table request.

anandtiwary: But there is no easy way to do it (I think)

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
anandtiwary: done

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
anandtiwary: done

projects/observability/src/pages/apis/services/services-list.dashboard.ts
```diff
@@ -82,7 +82,10 @@ const treeTableWidget: ModelJson = {
   data: {
     type: 'entity-table-data-source',
     entity: 'SERVICE',
-    childEntity: 'API'
+    'child-data-source': {
```
jake-bassett, line 85: This should be much better. Nice job!

PR #410, opened 2020-12-04
Simplified onboarding

jake-bassett:
## Description
Mostly pre work to add icons and adjust styles. Only functional change is embedding the copy-to-clipboard within the snippet viewer. It will now show on hover in the top right and can be customized to show any message, tooltip, or icon only.


projects/components/src/button/button.component.scss
```diff
@@ -23,6 +23,7 @@
 
 .button-container {
   display: flex;
+  height: inherit;
```
anandtiwary, line 26: Should it inherit the height?

projects/components/src/panel/panel.component.scss
```diff
@@ -21,6 +21,7 @@
   .body {
     flex: 1;
     overflow: auto;
+    height: 100%;
```
anandtiwary, line 24: Any side effects due to this?

projects/components/src/viewer/snippet-viewer/snippet-viewer.component.ts
```diff
@@ -7,6 +7,13 @@ import { TypedSimpleChanges } from '@hypertrace/common';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div class="snippet-viewer">
+      <ht-copy-to-clipboard
+        *ngIf='this.enableCopy'
+        class='copy-to-clipboard'
+        [label]='this.copyLabel'
```
anandtiwary, line 13: nit: ""

projects/observability/assets/icons/nginx.svg
```diff
@@ -0,0 +1,3 @@
...
...38 14.4393 17.6905 14.3662 17.6868 14.2927V5.71691C17.6905 5.6434 17.6738 5.57029 17.6386 5.50567C17.6033 5.44104 17.5509 5.38742 17.4871 5.3507L10.0437 1.05464C9.91037 0.971385 9.76073 0.988262 9.62741 1.07152L2.21717 5.35127C2.14887 5.38401 2.09183 5.43629 2.05327 5.50148C2.01471 5.56668 1.99638 5.64185 2.00059 5.71748V14.3079ZM7.24566 13.4208V8.41425L7.24454 8.40806C7.26141 8.42494 7.27829 8.44181 7.27829 8.45869L11.4917 13.5046C11.5488 13.5679 11.601 13.6337 11.6529 13.6991C11.7376 13.8057 11.8216 13.9114 11.9248 14.0041C12.2282 14.279 12.6155 14.4433 13.024 14.4705C13.3064 14.4874 13.573 14.4542 13.8228 14.3208C14.1558 14.1375 14.3392 13.8483 14.3392 13.4545V6.36663C14.2728 5.88285 13.8397 5.41707 13.1905 5.50033C12.673 5.55039 12.3073 5.96667 12.3073 6.49938V11.5622C12.2884 11.5432 12.2747 11.5238 12.2633 11.5076C12.2551 11.496 12.248 11.486 12.241 11.4789L7.81157 6.16693C7.58374 5.90029 7.31766 5.68427 6.96777 5.58414C6.601 5.4677 6.23479 5.43395 5.86858 5.58414C5.50237 5.73434 5.21942 6.06623 5.21942 6.53314V13.4877C5.21942 13.6711 5.2526 13.8376 5.35274 13.9873C5.54737 14.3034 5.83089 14.453 6.19653 14.4699C6.79563 14.4862 7.24566 13.9867 7.24566 13.4208Z" fill="#009639"/>
```
anandtiwary, line 2: Should fill = currentColor?

projects/components/src/button/button.component.scss
jake-bassett: The buttons in the latest mocks have a height of 36px. We are using 32px in several places and that is what the medium button is hard coded to. 

I was too worried about the consequences of changing the button height everywhere to make the leap of setting 36px as the default, so I did this to allow overriding. Though, we should do a sweep and change them all at some point.

projects/components/src/panel/panel.component.scss
jake-bassett: No side effects that I see, but setting a parent height was required to make some scrolling flexbox stuff work correctly deeper down in the DOM.

projects/components/src/viewer/snippet-viewer/snippet-viewer.component.ts
jake-bassett: Hah. Don't worry, prettier check got me! Fix incoming.

projects/observability/assets/icons/nginx.svg
jake-bassett: I was matching Kubernetes icons. It is set to blue. It looks like a few of the Observability asset icons keep the branding colors, so I did the same here.

projects/components/src/viewer/snippet-viewer/snippet-viewer.component.test.ts
```diff
@@ -7,7 +9,7 @@ describe('Snippet Viewer Component', () => {
   const createHost = createHostFactory({
     declareComponent: false,
     component: SnippetViewerComponent,
-    imports: [SnippetViewerModule]
+    imports: [SnippetViewerModule, RouterTestingModule, HttpClientTestingModule]
```
aaron-steinfeld, line 12: Can we make this test shallow so it doesn't need a fake router?

projects/components/src/copy-to-clipboard/copy-to-clipboard.component.ts
```diff
@@ -51,6 +60,9 @@ export class CopyToClipboardComponent implements OnInit, OnDestroy {
   @Input()
   public text?: string;
 
+  @Input()
+  public tooltip?: string;
```
aaron-steinfeld, line 64: Should we default this to "Copy to Clipboard"?

projects/components/src/button/button.component.scss
aaron-steinfeld: Hmm so what are the desired sizes?
In the mocks, I see:
```
X-small: 24px
Small: 32px
Default: 36px
Large: 40px
```
In code, I see:
```
X-small: 18px
Small: 24px
Medium: 32px
Large: 40px
```

Rather than change medium, which I agree is a bit risky, can we introduce a Default size of 36px and deprecate medium?

We'd still need to resolve x small and small, but it would avoid having to override anything which we generally don't want to allow.

jake-bassett: I think I'd prefer to just go through and fix these correctly all at once in a separate PR. It shouldn't be too bad to do.

aaron-steinfeld: OK. Given that my lack-of-context opinion is that that we should just use the existing medium size until that's done, but leave it to your discretion.

jake-bassett: Yeah, thanks. Let's go with this override. I'll come back and fix the inconsistencies and remove the override separately.

aaron-steinfeld:
https://github.com/hypertrace/hypertrace-ui/pull/410#discussion_r536204570
https://github.com/hypertrace/hypertrace-ui/pull/410#discussion_r536204139

aaron-steinfeld (approved):

PR #414, opened 2020-12-04
style: update z-index of table header row

arjunlalb:
## Description
Update z-index of table header row

### DIFF
--- a/projects/components/src/table/table.component.scss
+++ b/projects/components/src/table/table.component.scss
@@ -37,7 +37,7 @@
   border-bottom: 1px solid $gray-2;
   border-top-left-radius: 3px;
   border-top-right-radius: 3px;
-  z-index: 2; // Allows content to use "isolation: isolate" to create a new stacking context with index below the header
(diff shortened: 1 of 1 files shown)


projects/components/src/table/table.component.scss
```diff
@@ -37,7 +37,7 @@
   border-bottom: 1px solid $gray-2;
   border-top-left-radius: 3px;
   border-top-right-radius: 3px;
-  z-index: 1; // Allows content to use "isolation: isolate" to create a new stacking context with index below the header
+  z-index: 2; // Allows content to use "isolation: isolate" to create a new stacking context with index below the header
```
arjunlalb, line 40: @jake-bassett @aaron-steinfeld whats the recommendation here? Should we use a higher value or does 2 suffice?

aaron-steinfeld: I _hate_ z index. The comment looks like it implies it just needs any z index to create a new context, so why do we need to change it?

aaron-steinfeld (approved):

PR #419, opened 2020-12-08
fix: inconsistent trace IDs

jake-bassett:
## Description
Fixing some naming strangeness around Trace IDs. Right now in our trace/span details we show the Symbol('traceId') for our Trace ID, but this is actually a Span ID for the entry into the Trace. So, fixed that mistake and also updated the ID icons so we have a different one for Spans vs Traces to hopefully further signal the difference in IDs to the user.

The underlying attributes for API_TRACE are still confusing so to summarize:

> `Symbol('traceId')` is the same as `apiTraceId` which is actually the entry span ID
> `traceId` is the actual Trace ID


projects/distributed-tracing/src/shared/dashboard/widgets/span-detail/data/span-detail-data-source.model.ts
```diff
@@ -49,6 +49,7 @@ export class SpanDetailDataSourceModel extends GraphQlDataSourceModel<SpanDetail
   protected constructSpanDetail(span: Span): SpanDetailData {
     return {
       id: span[spanIdKey],
+      traceId: span.apiTraceId as string,
```
anandtiwary, line 52: Should it be apiTraceId here? I remember working on a change where Buchi asked me to show Trace Id instead.

anandtiwary (approved):

projects/distributed-tracing/src/shared/dashboard/widgets/span-detail/data/span-detail-data-source.model.ts
aaron-steinfeld: apiTraceId is not trace ID - we should be using the real thing here.

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
```diff
@@ -57,6 +57,8 @@ export class TraceDetailDataSourceModel extends GraphQlDataSourceModel<TraceDeta
   protected constructTraceDetailData(trace: Trace): TraceDetailData {
     return {
       id: trace[traceIdKey],
+      traceId: trace.traceId as string, // The traceId is real Trace ID. NOT Symbol('traceId').
```
aaron-steinfeld, line 60: Can we replace or remove `id` then? I think it's confusing to have `id`, `traceId` and `entrySpanId` - there's only two distinct values, right?

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/trace-detail-widget-renderer.component.ts
```diff
@@ -15,11 +15,17 @@ import { TraceDetailWidgetModel } from './trace-detail-widget.model';
     <div class="trace-detail-widget-renderer fill-container">
       <div class="content" *htLoadAsync="this.data$ as data">
         <ht-span-detail [spanData]="data" [showTitleHeader]="false">
+          <ht-summary-value
+            class="summary-value"
+            icon="${IconType.SpanId}"
+            label="Entry Span ID"
```
aaron-steinfeld, line 21: What if there isn't an entry span? A regular trace doesn't really have this data - is it something we even want to show or just always show the (correct) trace id?

aaron-steinfeld (commented):
Have we tested both api-trace and trace visualizations to make sure they're consistent? The code they share and don't share is getting more and more confusing (at least to me)

aaron-steinfeld (changes requested):
blocking based on the span trace id issue highlighted by anand above

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
```diff
@@ -43,7 +43,7 @@ export class TraceDetailDataSourceModel extends GraphQlDataSourceModel<TraceDeta
   }
 
   protected getTraceAttributes(): string[] {
-    return ['tags', 'statusCode', ...this.attributes];
+    return ['tags', 'statusCode', 'traceId', 'apiTraceId', ...this.attributes];
```
aaron-steinfeld, line 46: So along the lines of my comment on waterfall, this is requesting attributes now that only exist on api traces - a regular trace would error. If this data source is only used for api traces (not sure, but the ambiguity and reuse of code between trace types has caused lots of headaches), then we should rename it as such - else, we either need to fork based on trace type or revert.

jake-bassett: So, we can get rid of apiTraceId here, but traceId we still need since it is the correct and actual Trace ID.

jake-bassett: Looking some more, I'm 99% sure that all of these are API_TRACE calls. The only place I see the full trace call is from the trace-waterfall-data-source.model. But you are right, this is a bit of a mess and I'm not 100%.

I'm going to rename these to api- to make them more specific.

jake-bassett: Tested this change in waterfall, full waterfall, and all of the trace tabs (and their links to waterfall). Seems to be fine.

aaron-steinfeld (approved):

PR #421, opened 2020-12-10
feat: divide trace data source into trace and api trace

jake-bassett:
## Description
Separate Trace and API Trace detail data sources so that Trace data source can be extended for other trace types.

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/api-trace-detail-data-source.model.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/api-trace-detail-data-source.model.ts
@@ -1,5 +1,5 @@
 import { Model } from '@hypertrace/hyperdash';
-import { Trace } from '../../../../graphql/model/schema/trace';
+import { Trace, traceIdKey } from '../../../../graphql/model/schema/trace';
 
 import { TraceDetailData, TraceDetailDataSourceModel } from './trace-detail-data-source.model';
 
@@ -8,13 +8,14 @@ import { TraceDetailData, TraceDetailDataSourceModel } from './trace-detail-data
 })
 export class ApiTraceDetailDataSourceModel extends TraceDetailDataSourceModel {
   protected getTraceAttributes(): string[] {
-    return [...super.getTraceAttributes(), 'apiTraceId'];
+    return [...super.getTraceAttributes(), 'traceId'];
   }
 
   protected constructTraceDetailData(trace: Trace): ApiTraceDetailData {
     return {
       ...super.constructTraceDetailData(trace),
-      entrySpanId: trace.apiTraceId as string // Symbol('traceId') same as apiTraceId which is actually Entry Span ID
+      traceId: trace.traceId as string,  // For API Trace traceId is real Trace ID. NOT Symbol('traceId').
+      entrySpanId: trace[traceIdKey] // API Trace Symbol('traceId') same as apiTraceId which is actually Entry Span ID
     };
   }
 }
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/api-trace-detail-data-source.model.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/api-trace-detail-data-source.model.ts
@@ -14,7 +14,7 @@ export class ApiTraceDetailDataSourceModel extends TraceDetailDataSourceModel {
   protected constructTraceDetailData(trace: Trace): ApiTraceDetailData {
     return {
       ...super.constructTraceDetailData(trace),
-      traceId: trace.traceId as string,  // For API Trace traceId is real Trace ID. NOT Symbol('traceId').
+      traceId: trace.traceId as string, // For API Trace traceId is real Trace ID. NOT Symbol('traceId').
       entrySpanId: trace[traceIdKey] // API Trace Symbol('traceId') same as apiTraceId which is actually Entry Span ID
     };
   }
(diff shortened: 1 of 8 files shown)


projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/api-trace-detail-data-source.model.ts
```diff
@@ -1,74 +1,24 @@
...
-      traceId: this.trace[traceIdKey],
-      spanLimit: 0,
-      timeRange: this.getTimeRangeOrThrow(),
-      traceProperties: this.getTraceAttributes().map(attribute =>
-        this.attributeSpecBuilder.attributeSpecificationForKey(attribute)
-      )
-    }).pipe(mergeMap(trace => this.mapResponseObject(trace)));
-  }
-
+export class ApiTraceDetailDataSourceModel extends TraceDetailDataSourceModel {
```
anandtiwary, line 9: I see a trace detail data source at this path in my local. Was it changed recently?

jake-bassett: I renamed trace- to api-trace. That was merged yesterday. Then, this change I'm adding back a trace- and making api-trace extend that instead.

anandtiwary: For an api trace, what does the `traceIdKey` in `Trace` object represent? Would it be traceId or apiTraceId/entrySpanId?

anandtiwary: For the query to work, It should represent `apiTraceId`. Right?
If we want consistency, then probably we always make traceIdKey represent traceId and for fetching api trace detail, we use entrySpanId in the getData query below.

jake-bassett: Its confusing. For an `API_TRACE`, the `traceIdKey` is the same value as the `apiTraceId` which is actually the Span ID.

anandtiwary: oh ohkay.. I thought we were doing the mapping in the handler somewhere. but nvm

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/api-trace-detail-data-source.model.ts
```diff
@@ -1,74 +1,24 @@
...
-    return ['tags', 'statusCode', 'traceId', 'apiTraceId', ...this.attributes];
-  }
-
-  private mapResponseObject(trace: Trace | undefined): Observable<TraceDetailData> {
-    if (trace === undefined) {
-      return EMPTY;
-    }
-
-    return of(this.constructTraceDetailData(trace));
+    return [...super.getTraceAttributes(), 'apiTraceId'];
```
aaron-steinfeld, line 11: shouldn't this be fetching the `traceId` (since `apiTraceId` is already returned as the default `id` for an api trace) - this also has implications for building the detail data below, where now we have to override the id set from super (which is actually the apitraceId, but should come from traceId for an api trace) and set the entrySpanId from the default id attribute.

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
```diff
@@ -0,0 +1,72 @@
...
+      spanLimit: 0,
+      timeRange: this.getTimeRangeOrThrow(),
+      traceProperties: this.getTraceAttributes().map(attribute =>
+        this.attributeSpecBuilder.attributeSpecificationForKey(attribute)
+      )
+    }).pipe(mergeMap(trace => this.mapResponseObject(trace)));
+  }
+
+  protected getTraceAttributes(): string[] {
+    return ['tags', 'statusCode', 'traceId', ...this.attributes];
```
aaron-steinfeld, line 46: I don't believe `traceId` exists on a regular trace - that would be returned by the default id attribute already fetched (see other comment for more context).

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (trace === undefined) {
+      return EMPTY;
+    }
+
+    return of(this.constructTraceDetailData(trace));
+  }
+
+  protected constructTraceDetailData(trace: Trace): TraceDetailData {
+    return {
+      traceId: trace.traceId as string, // The traceId is real Trace ID. NOT Symbol('traceId').
```
aaron-steinfeld, line 59: Sometimes, depends if called in trace or api_trace context (see above)

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/trace-detail-widget.model.ts
```diff
@@ -1,11 +1,11 @@
 import { Model, ModelApi, ModelProperty, STRING_PROPERTY } from '@hypertrace/hyperdash';
 import { ModelInject, MODEL_API } from '@hypertrace/hyperdash-angular';
 import { Observable } from 'rxjs';
-import { ApiTraceDetailDataSourceModel, TraceDetailData } from './data/api-trace-detail-data-source.model';
+import { TraceDetailData, TraceDetailDataSourceModel } from './data/trace-detail-data-source.model';
 
 @Model({
   type: 'trace-detail-widget',
-  supportedDataSourceTypes: [ApiTraceDetailDataSourceModel]
+  supportedDataSourceTypes: [TraceDetailDataSourceModel]
```
aaron-steinfeld, line 8: This should support either, right? Or do we have an api trace detail widget too?

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/api-trace-detail-data-source.model.ts
aaron-steinfeld: that may be a clearer way to handle a bunch of this - we could change the trace query handler to always match traceIdKey to the real trace id and introduce an optional entrySpanId (or just leave it to the requester)

aaron-steinfeld (approved):

PR #422, opened 2020-12-10
fix: a number of table layout bugs

aaron-steinfeld:
## Description
This fixes a few different table layout bugs:
- Empty table controls are still using up space due to padding
- Error state doesn't clear previous data
- In full screen mode (entity lists), the last row of data is hidden underneath the footer
- In embedded mode (explorer), the pagination controls are clipped

### Testing
Beyond the unit tests, I visually verified the full and embedded table modes in error, no data and normal states.

### Checklist:

### DIFF
--- a/projects/components/src/table/controls/table-controls.component.ts
+++ b/projects/components/src/table/controls/table-controls.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectionStrategy, Component, EventEmitter, Input, OnChanges, Output } from '@angular/core';
 import { SubscriptionLifecycle, TypedSimpleChanges } from '@hypertrace/common';
+import { isEmpty } from 'lodash-es';
 import { Subject } from 'rxjs';
 import { debounceTime } from 'rxjs/operators';
 import { ToggleItem } from '../../toggle-group/toggle-item';
@@ -11,7 +12,7 @@ import { TableMode } from '../table-api';
   providers: [SubscriptionLifecycle],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="table-controls">
+    <div class="table-controls" *ngIf="this.anyControlsEnabled">
       <!-- Filters -->
       <div class="filter-controls">
         <!-- Search -->
@@ -24,7 +25,7 @@ import { TableMode } from '../table-api';
 
         <!-- Filter Toggle -->
         <ht-toggle-group
-          *ngIf="this.filterItems?.length > 1"
+          *ngIf="this.filterItemsEnabled"
           class="control filter-toggle-group"
           [items]="this.filterItems"
           [activeItem]="this.activeFilterItem"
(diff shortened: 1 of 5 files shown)


projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -11,7 +12,7 @@ import { TableMode } from '../table-api';
   providers: [SubscriptionLifecycle],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="table-controls">
+    <div class="table-controls" *ngIf="this.anyControlsEnabled">
```
aaron-steinfeld, line 15: The changes in this file fix the issue where no controls are specified but extra padding is included

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -67,7 +67,12 @@ export class TableCdkDataSource implements DataSource<TableRow> {
     return this.rowsChange$.pipe(
       tap(rows => this.cacheRows(rows)),
       tap(rows => this.cacheFilterableValues(rows)),
-      tap(rows => this.loadingStateSubject.next({ loading$: of(rows), hide: rows.length > 0 }))
+      tap(rows => this.loadingStateSubject.next({ loading$: of(rows), hide: rows.length > 0 })),
```
aaron-steinfeld, line 70: The changes in this file allow us to clear the data on error (previously, we were returning an empty observable (`[]` as an observable input) which the cdk would ignore causing the error icon to overlap with the previous data. Returning actual empty data at the old catch location isn't feasible since the no data check came later.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
```diff
@@ -1,14 +1,14 @@
...
+
+.table-content-container {
+  @include flex-layout(column);
+}
 
-  .table {
-    flex: 1 1;
-    width: 100%;
-  }
+.table {
```
aaron-steinfeld, line 7: The previous flex controls were not in a flex container because the content of titled content is nested deeper in the dom. This caused clipping of the pagination controls for the embedded mode (because the table pushed it beyond the container bounds) and hidden data for the full screen mode (because it uses a fixed pagination controls, they were still visible. Instead, the table content went beneath it)

projects/observability/src/pages/explorer/explorer.component.scss
```diff
@@ -3,7 +3,7 @@
 .explorer-container {
   display: flex;
   flex-direction: column;
-  padding: 24px;
+  padding: 24px 24px 0;
```
aaron-steinfeld, line 6: with the appropriate table layout now, the bottom padding looks silly and eats up too much room, as the table itself is padded.

jake-bassett (approved):

PR #424, opened 2020-12-11
feat: support for multiple snippet highlight

arjunlalb:
## Description
Support multiple parts highlighting using highlightPipe.
Support in confirmation modal to highlight multiple parts of a description text.
Style update to description in confirmation modal to center align it with word break when applicable.

### Testing
UTs added

### Checklist:


projects/common/src/utilities/formatters/string/highlight.pipe.ts
```diff
@@ -1,20 +1,41 @@
 import { Pipe, PipeTransform } from '@angular/core';
+import { isArray } from 'lodash-es';
 import { assertUnreachable } from '../../lang/lang-utils';
 
 @Pipe({ name: 'htHighlight' })
 export class HighlightPipe implements PipeTransform {
   public transform(
     fullText: string,
-    highlightedText: string | undefined,
+    highlightSnippets: string | undefined | TextHighlightConfig[],
```
aaron-steinfeld, line 9: If highlight type is now provided here, does it make sense to keep the last arg? The highlightConfig is a superset of any other config, so leaving the string there makes sense as a simple default but any customization of mark can rely on the more complex api. Could also potentially accept a single `TextHighlightConfig` if you want to disambiguate.

projects/common/src/utilities/formatters/string/highlight.pipe.ts
```diff
@@ -1,20 +1,41 @@
...
     highlightType: HighlightType = 'mark'
   ): string {
+    if (isArray(highlightSnippets)) {
```
aaron-steinfeld, line 12: nit - rather than having a separate path, for array vs string, normalize the inputs first and then put everything through the same path. Also, this is a textbook use case for `reduce`  😉

projects/components/src/confirmation/confirmation-modal.component.ts
```diff
@@ -52,4 +53,5 @@ export interface ConfirmationModalData {
   confirmButtonLabel?: string;
   confirmButtonRole?: ButtonRole;
   descriptionText: string;
+  highlightParts?: TextHighlightConfig[];
```
aaron-steinfeld, line 56: I don't think this should be part of the confirmation modal. If we want to support completely custom content, we can do that - but I think making confirmation do the customization for us is unsustainable.

Alternative suggestion (we actually discussed doing this when confirmation was being written, but decided to hold off until needed) change `descriptionText` to `description`  or `message` and take `string | TemplateRef`, then use a template outlet to render the template if needed. That leaves the callers full control over the message, including using components, directives, pipes etc.

projects/common/src/utilities/formatters/string/highlight.pipe.ts
```diff
@@ -1,20 +1,41 @@
...
+        );
+      });
+
+      return textWithHighlight;
+    }
+
     const htmlTag = getHtmlTagForHighlightType(highlightType);
 
-    return highlightedText === undefined ? fullText : fullText.replace(highlightedText, `<${htmlTag}>$&</${htmlTag}>`);
+    return highlightSnippets === undefined
```
aaron-steinfeld, line 27: if i called the highlight pipe with no args, I'd expect the default case would be to highlight the whole thing.

projects/components/src/confirmation/confirmation-modal.component.ts
arjunlalb: Makes sense. Will drop these changes.

projects/common/src/utilities/formatters/string/highlight.pipe.ts
arjunlalb: combining this along with this comment (https://github.com/hypertrace/hypertrace-ui/pull/424#discussion_r540938447), I think it makes sense to keep the arg in case the whole string needs to be highlighted.

aaron-steinfeld: When designing an API, one of the most important things is making it have consistent, predictable behavior. That is, arguments should always make sense independently and never conflict or require each other (as in, if I optionally took one argument, I shouldn't require another optional argument to also be provided)

Here, it allows me to pass `text, undefined, <mark>` which has clear behavior expectations but is awkward to use (because you have to skip an optional arg to get to a different optional arg). If I pass `text, <config>, <mark>` it's unclear which mark will be used - the one from the config or the one from the mark. If we removed the mark arg entirely, every use case is still satisfied. I could highlight the whole string with `<text>, { <text>, <mark> }` - a bit awkward too with the repetition of text, but with clearly predictable behavior. Omitting the second arg entirely is just a syntax shortcut - but still reachable behavior when using the full api.

projects/common/src/utilities/formatters/string/highlight.pipe.ts
```diff
@@ -1,20 +1,37 @@
 import { Pipe, PipeTransform } from '@angular/core';
+import { isArray } from 'lodash-es';
 import { assertUnreachable } from '../../lang/lang-utils';
 
 @Pipe({ name: 'htHighlight' })
 export class HighlightPipe implements PipeTransform {
   public transform(
     fullText: string,
-    highlightedText: string | undefined,
+    highlightSnippets?: TextHighlightConfig | TextHighlightConfig[],
```
aaron-steinfeld, line 9: I think accepting a single text string here will make the most common use case easier (as seen from combo box change below). Not required though, just a shortcut.

projects/components/src/confirmation/confirmation.module.ts
```diff
@@ -1,10 +1,11 @@
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
+import { FormattingModule } from '@hypertrace/common';
 import { ButtonModule } from '../button/button.module';
 import { ConfirmationModalComponent } from './confirmation-modal.component';
 
 @NgModule({
   declarations: [ConfirmationModalComponent],
-  imports: [CommonModule, ButtonModule]
+  imports: [CommonModule, ButtonModule, FormattingModule]
```
aaron-steinfeld, line 9: The confirmation modal changes I'm guessing are happening separately (which makes sense), but allows us to roll back the changes to those files.

aaron-steinfeld (approved):

PR #425, opened 2020-12-12
feat: support for confirmation description template

arjunlalb:
## Description
Support for a template in place of description in confirmation modal.

### Testing
Manual test


projects/components/src/confirmation/confirmation-modal.component.ts
```diff
@@ -9,7 +9,10 @@ import { ModalRef, MODAL_DATA } from '../modal/modal';
   template: `
     <div class="confirmation-modal">
       <div class="description">
-        {{ this.descriptionText }}
+        <div *ngIf="this.descriptionText; else templateRenderer" class="text">{{ this.descriptionText }}</div>
+        <ng-template #templateRenderer>
```
aaron-steinfeld, line 13: I don't believe we need to rewrap this in a new template (although not positive) - can you just do `else.this.renderer` above?

projects/components/src/confirmation/confirmation-modal.component.ts
```diff
@@ -51,5 +56,6 @@ export interface ConfirmationModalData {
   cancelButtonLabel?: string;
   confirmButtonLabel?: string;
   confirmButtonRole?: ButtonRole;
-  descriptionText: string;
+  descriptionText?: string;
```
aaron-steinfeld, line 59: See comment [here](https://github.com/hypertrace/hypertrace-ui/pull/424#discussion_r541949306) about api design, this has a similar issue. If I provide both text and content, it's unclear which takes precedence. Instead, we should take them both in as a union arg so only one or the other can be provided. To do that, we likely need to make a breaking change because `descriptionText` doesn't make sense for a template, but since we're not actually publishing this we can just update the consumers (otherwise, I'd make descriptionText optional and deprecated).

projects/components/src/confirmation/confirmation-modal.component.ts
```diff
@@ -32,15 +32,18 @@ export class ConfirmationModalComponent {
...
+  public readonly isContentString: boolean;
 
   public constructor(private readonly modalRef: ModalRef<boolean>, @Inject(MODAL_DATA) config: ConfirmationModalData) {
     this.confirmButtonLabel = config.confirmButtonLabel ?? ConfirmationModalComponent.DEFAULT_CONFIRM_LABEL;
     this.confirmButtonRole = config.confirmButtonRole ?? ConfirmationModalComponent.DEFAULT_CONFIRM_ROLE;
     this.cancelButtonLabel = config.cancelButtonLabel ?? ConfirmationModalComponent.DEFAULT_CANCEL_LABEL;
-    this.descriptionText = config.descriptionText ?? '';
-    this.renderer = config.content;
+    this.isContentString = isString(config.content);
+    this.isContentString
```
aaron-steinfeld, line 44: nit: try to avoid assignments as expressions - can we move this to a traditional if/else?

aaron-steinfeld (approved):

projects/components/src/confirmation/confirmation-modal.component.ts
arjunlalb: done. @aaron-steinfeld

aaron-steinfeld (approved):

PR #427, opened 2020-12-14
feat: hide sort options in hover menu options if sortable is false

arjunlalb:
## Description
Show sort related options in hover options menu only if sortable is true for the column.

### Testing

### Checklist:

### DIFF
--- a/projects/components/src/table/header/table-header-cell-renderer.component.ts
+++ b/projects/components/src/table/header/table-header-cell-renderer.component.ts
@@ -42,26 +42,25 @@ import { TableColumnConfigExtended } from '../table.service';
           </ht-popover-trigger>
           <ht-popover-content>
             <div [style.min-width.px]="trigger.offsetWidth" class="popover-content">
-              <div class="popover-item" (click)="this.onFilterValues()" *ngIf="this.isFilterable">Filter Values</div>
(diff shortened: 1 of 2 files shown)


projects/components/src/table/header/table-header-cell-renderer.component.ts
```diff
@@ -44,11 +44,19 @@ import { TableColumnConfigExtended } from '../table.service';
             <div [style.min-width.px]="trigger.offsetWidth" class="popover-content">
               <div class="popover-item" (click)="this.onFilterValues()" *ngIf="this.isFilterable">Filter Values</div>
               <div class="popover-item-divider" *ngIf="this.isFilterable"></div>
-              <div class="popover-item" (click)="this.onSortChange(SORT_ASC)">
+              <div
```
jake-bassett, line 47: Are we also ensuring the divider lines match up with sorting turned both on and off. We would need to check with the other options - Filtering and Edit - on and off as well. A few combinations to test here.

arjunlalb: @jake-bassett done. used ng-containers to segregate a section and associated dividers.

jake-bassett (approved):

PR #428, opened 2020-12-14
feat: add util const to check for empty strings

jake-bassett:
## Description
This check comes up a lot when looking at strings passed in as Inputs() to components.

### DIFF
--- a/projects/common/src/utilities/lang/lang-utils.ts
+++ b/projects/common/src/utilities/lang/lang-utils.ts
@@ -32,8 +32,5 @@ const ignoreFunctions = (first: unknown, second: unknown) => {
   }
 };
 
-export const isFalsyString = (str: string | undefined | null): boolean =>
-  str === undefined || str === null || str === '';
-
 export const isNonEmptyString = (str: string | undefined | null): str is string =>
   str !== undefined && str !== null && str !== '';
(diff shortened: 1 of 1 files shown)


projects/common/src/utilities/lang/lang-utils.ts
```diff
@@ -31,3 +31,6 @@ const ignoreFunctions = (first: unknown, second: unknown) => {
     return first.toString() === second.toString();
   }
 };
+
+export const isEmptyString = (str: string | undefined | null): boolean =>
```
aaron-steinfeld, line 35: Do we not want to use `isEmpty` ? I'm generally for writing our own utils, but we've used lodash directly pretty extensively

jake-bassett: Ah, didn't realize that method picked up empty quotes. Just tested and it does. I'll remove that method. The other I added after you reviewed is more helpful anyway. I'll keep that one.

aaron-steinfeld (dismissed):

projects/common/src/utilities/lang/lang-utils.ts
```diff
@@ -31,3 +31,6 @@ const ignoreFunctions = (first: unknown, second: unknown) => {
     return first.toString() === second.toString();
   }
 };
+
+export const isNonEmptyString = (str: string | undefined | null): str is string =>
```
anandtiwary, line 35: Why not use lodash method for this? `isEmpty` ??

aaron-steinfeld (approved):

PR #430, opened 2020-12-15
Persistable table column configs

jake-bassett:
## Description
This adds a new table widget model property called `id` from a base class.  Setting that ID enables the table widget to persist column configs locally across browser refresh.

I've added IDs to all existing tables EXCEPT the toggleable tree/flat lists since those require a little more work and are going away _soon_ anyway.

This PR also fixes an existing bug I found where if a user removes all columns, they have no way to get them back without a refresh. This wasn't terrible before, but would have been a blocker with the table columns persisted since we would keep loading the bad config from browser storage. Put in a check to prevent last remaining column from being removed without first enabling another column.

NOTE: Renamed some pagination stuff in table.component, but no real functional changes there.


projects/components/src/table/columns/table-edit-columns-modal.component.ts
```diff
@@ -52,6 +53,16 @@ export class TableEditColumnsModalComponent {
       .sort((a, b) => (a.visible === b.visible ? 0 : a.visible ? -1 : 1));
   }
 
+  public isLastColumn(column: TableColumnConfigExtended): boolean {
+    let visibleCount = 0;
+
+    this.editColumns.forEach(editColumn => {
```
aaron-steinfeld, line 59: nit: `this.editColumns.filter(col => col.visible).length`

projects/components/src/table/table.component.ts
```diff
@@ -692,15 +696,18 @@ export class TableComponent
     this.pageChange.emit(pageEvent);
   }
 
-  private getPageData(params: ParamMap): Partial<PageEvent> | undefined {
+  private getPagination(params: ParamMap): Partial<PageEvent> | undefined {
```
aaron-steinfeld, line 699:
```suggestion
  private getPagination(params: ParamMap): Partial<PageEvent> {
```

can never emit undefined any more

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-base.model.ts
```diff
@@ -101,6 +101,13 @@ export abstract class TableWidgetBaseModel {
   })
   public pageable: boolean = true;
 
+  @ModelProperty({
+    key: 'persistId',
+    displayName: 'ID used to enable and identify persisted table preferences',
+    type: STRING_PROPERTY.type
+  })
+  public persistId?: string; // Format: <page>.<table-name> (e.g. "explorer.spans-table")
```
aaron-steinfeld, line 109: Hmmm don't love this, but I guess it's needed to identify the specific instance on the page. What are your thoughts to making this more generic and keeping this as a general model id (just in terms of naming)? We don't have to expose it on all models now, but would keep the concept consistent.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -109,7 +112,13 @@ export class TableWidgetRendererComponent
     this.onModeChange(this.model.mode);
 
     this.metadata$ = this.getScopeAttributes();
-    this.columnConfigs$ = this.getColumnConfigs();
+    this.columnConfigs$ = (isNonEmptyString(this.model.persistId)
+      ? this.preferenceService.get<string>(this.model.persistId, JSON.stringify([]))
```
aaron-steinfeld, line 116: can we make pref service handle objects (could be in a separate PR)

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -164,6 +176,22 @@ export class TableWidgetRendererComponent
...
+    columns: SpecificationBackedTableColumnDef[],
+    persistedColumns: TableColumnConfig[]
+  ): SpecificationBackedTableColumnDef[] {
+    return columns.map(column => {
+      const found = persistedColumns.find(persistedColumn => persistedColumn.id === column.id);
+
+      return {
+        ...column, // Apply default column config
+        ...(found ? found : {}), // Override with any saved properties
+        specification: column.specification, // Spec not user configurable and has nested methods, so apply those
```
aaron-steinfeld, line 189: Can we change it so we only persist the specific settings worth persisting - it feels like going default, persisted, some defaults is an easy way to introduce bugs if a new method for example is added to the col.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-base.model.ts
jake-bassett: So, we would want each model to have a separate `id` property, then we would change the `persistId` to just be a boolean for something like `enablePersist` or the like? We would then have a check to ensure there is a model ID set and use that as our key for the storage? 

Sounds fine.

aaron-steinfeld: Are there specific tables we explicitly wouldn't want to persist settings for? If not, I'd leave out the boolean entirely and assume we can safely persist if there's an id present.

jake-bassett: I think just tree tables for something like services list, but that is more of a technical detail. Since column edit is already disabled for those types, and that is what triggers the persist we should be fine.

I'll assume all tables can persist then and we can back that assumption out and add the boolean if we encounter one that shouldn't be.

aaron-steinfeld: Sure, I think we can build it out over time

projects/dashboards/src/widgets/widget-base.model.ts
```diff
@@ -0,0 +1,18 @@
+import { ModelProperty, STRING_PROPERTY } from '@hypertrace/hyperdash';
+
+export abstract class WidgetBaseModel {
```
aaron-steinfeld, line 3: Could we call this just `BaseModel` (or `Model`)? The idea being this isn't necessarily unique to widget models, but potentially any mode

aaron-steinfeld (approved):

PR #435, opened 2020-12-16
feat: adding start time to spans and trace queries

anandtiwary:
## Description
All spans and Traces queries may optionally use a startTime for building query time range.

### Testing

### Checklist:

### Documentation


projects/common/src/utilities/coercers/date-coercer.ts
```diff
@@ -23,7 +23,7 @@ export class DateCoercer extends Coercer<Date, DateCoercerOptions> {
       return undefined;
     }
 
-    const valueAsDate = new Date(value);
+    const valueAsDate = typeof value === 'string' && !isNaN(Number(value)) ? new Date(+value) : new Date(value);
```
anandtiwary, line 26: new Date("1608154354481") doesn't work.

projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
```diff
@@ -62,9 +71,10 @@ export class TraceWaterfallDataSourceModel extends GraphQlDataSourceModel<Waterf
       requestType: TRACE_GQL_REQUEST,
       traceId: this.traceId,
       spanLimit: 1000,
-      timeRange: this.getTimeRangeOrThrow(),
+      timeRange: this.getTraceTimeRangeOrThrow(),
       traceProperties: [],
-      spanProperties: this.spanSpecifications
+      spanProperties: this.spanSpecifications,
+      spansTimeRange: this.getTimeRangeOrThrow()
```
anandtiwary, line 77: Previously span time range was same as Selected time range. Keeping it the same now. Although, it is definitely incorrect. We should be able to fetch all spans here for the given trace.

anandtiwary:
We have trace detail, trace waterfall, api trace detail, api trace waterfall, extended api trace and extended api trace waterfall. Phew!

projects/common/src/utilities/coercers/date-coercer.ts
aaron-steinfeld: Could you add a test for that case?

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
```diff
@@ -109,15 +125,26 @@ export class TraceDetailService implements OnDestroy {
...
+    return startTimeAsDate !== undefined
+      ? new GraphQlTimeRange(startTimeAsDate.getTime() - 1, startTimeAsDate.getTime() + 1)
+      : new GraphQlTimeRange(timeRange.startTime, timeRange.endTime);
+  }
 }
 
 interface TraceDetailRouteIdParams {
   traceId: string;
   spanId?: string;
+  startTime?: unknown;
```
aaron-steinfeld, line 141: why unknown? params are always strings right?

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
```diff
@@ -90,6 +105,7 @@ export class TraceDetailService implements OnDestroy {
         this.specificationBuilder.attributeSpecificationForKey('serviceName'),
         this.specificationBuilder.attributeSpecificationForKey('displaySpanName')
       ],
+      spansTimeRange: this.buildGraphqlTimeRange(timeRange),
```
aaron-steinfeld, line 108: I think we could push this logic down in a couple ways:
- Use the same time range for both, and generate the span time range here, since the broader span time range should also be satisfactory for the trace (currently, it looks like we're not bounding the span request by the time stamp)
- Make the trace gql request only accept a timestamp, not a time range - move all time range generation logic inside the query handler and force all callers to provide a timestamp when querying a single trace.

projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
```diff
@@ -34,10 +35,18 @@ export class TraceWaterfallDataSourceModel extends GraphQlDataSourceModel<Waterf
   })
   public entrySpanId?: string;
 
+  @ModelProperty({
+    key: 'start-time',
+    required: true, // Todo
```
aaron-steinfeld, line 40: What's the todo here?

projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
aaron-steinfeld: Ah, commented on this above - we should be able to fix this.

projects/distributed-tracing/src/shared/services/navigation/tracing-navigation.service.ts
```diff
@@ -0,0 +1,32 @@
+import { Injectable } from '@angular/core';
+import { Dictionary, NavigationService } from '@hypertrace/common';
+import { Observable } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
+export class TracingNavigationService {
+  public constructor(private readonly navigationService: NavigationService) {}
+
+  public navigateToTraceDetail(traceId: string, spanId?: string, startTime?: unknown): Observable<boolean> {
```
aaron-steinfeld, line 9: same comment about unknowns

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
anandtiwary: It can be number, string or undefined. I kept it unknown since we are using dateCoercer anyway. But I can change it to just `string | number` if you like

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
anandtiwary: I don't think #1 would work. Trace Id can lie outside of the TR selected at the top. This TR is same as the span time range. So it is not really a broader time.

#2 seems a change with large diff. It scares me at this point.

projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
anandtiwary: oh sorry. I need to set required: false. Kept it to true so that I don't miss setting it in a dashboard json. Will update

anandtiwary:
updated

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
aaron-steinfeld: for 1, I think there was a miscommunication - I meant ignore the global TR, and given a timestamp, generate a window like `[-30m, +30m]` or so, and pass that in as the one time range for the request. Because we're filtering by IDs here, being extra generous on that window should be absolutely fine, even though not required. The behavior of the timestamps: 
- A trace timestamp should be equal to the earliest included span timestamp
- (assumption) A trace must complete in 30 minutes (I was curious what the platform assumption here is - it's configurable but defaults to 30s, so I think this is plenty of wiggle room)
- All spans belonging to a trace will have a timestamp within the window defined in the point above

2 is an extension of 1, but I think significantly less code since that window generation that is duplicated everywhere (in your PR as well as my proposal) is all left in one spot. It also has the added benefit of type safety - we're forcing the calls for a single trace to provide a timestamp and we can be assured b the compiler we didn't miss a spot.

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
aaron-steinfeld: Even if it's going to a date coercer, being more specific on the types earlier in the chain I think is still a gppd thing . These values are always originating from `paramMap.get` though right? https://angular.io/api/router/ParamMap#get looks like it can only return a string, if anything.

anandtiwary: done

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
anandtiwary: Yeah this makes sense. Should we do it separately or part of this PR only? This new change will likely affect other places as well. I have to focus on some other priority items as of now. So either we defer this api change for later or we defer this PR till that new commit is in.

anandtiwary: I have addressed your other comments.

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
```diff
@@ -33,7 +42,8 @@ export class TraceDetailService implements OnDestroy {
     this.routeIds$ = route.paramMap.pipe(
       map(paramMap => ({
         traceId: paramMap.get(this.getTraceIdParamName())!,
-        spanId: paramMap.get(this.getSpanIdParamName()) as string | undefined
+        spanId: paramMap.get(this.getSpanIdParamName()) as string | undefined,
+        startTime: paramMap.get(this.getStartTimeParamName()) as string | undefined
```
aaron-steinfeld, line 46: nit: this is returning the value or null and is typed, so rather than cast to get rid of the null, let's actually get rid of it
```suggestion
        startTime: paramMap.get(this.getStartTimeParamName()) ?? undefined
```

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
aaron-steinfeld: I think we should at a minimum do 1 as part of this PR (in other words, revert the separate span time range and update the calculations you've added so they're suitable for the request's "primary" time range) - that will be less code then the current diff and limits the tech debt by keeping the trace request API the same as it currently is. If you think changing the api to take a timestamp instead is a significantly larger diff/amount of work (it doesn't seem to me that it would be, but haven't looked at this code in a bit), that's fine to defer.

anandtiwary: Then I will continue this next week. Have few other things to take care of before it.

projects/distributed-tracing/src/shared/dashboard/widgets/span-detail/data/span-detail-data-source.model.ts
```diff
@@ -54,6 +63,14 @@ export class SpanDetailDataSourceModel extends GraphQlDataSourceModel<SpanDetail
       tags: span.spanTags as Dictionary<unknown>
     };
   }
+
+  protected getTimeRangeOrThrow(): GraphQlTimeRange {
```
anandtiwary, line 67: Should we do this SpanGraphQlQueryHandlerService as well? We are not using it at many places.

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
```diff
@@ -114,10 +120,13 @@ export class TraceDetailService implements OnDestroy {
 interface TraceDetailRouteIdParams {
   traceId: string;
   spanId?: string;
+  startTime?: string;
 }
+
 export interface TraceDetails {
   id: string;
   entrySpanId?: string;
+  startTime: unknown;
```
aaron-steinfeld, line 129: shouldn't this type match line 123?

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
```diff
@@ -21,21 +21,29 @@ export class TraceDetailDataSourceModel extends GraphQlDataSourceModel<TraceDeta
   })
   public trace!: Trace;
 
+  @ModelProperty({
+    key: 'start-time',
+    required: false,
+    type: UNKNOWN_PROPERTY.type
```
aaron-steinfeld, line 27: why unknown?

projects/distributed-tracing/src/shared/dashboard/widgets/span-detail/data/span-detail-data-source.model.ts
aaron-steinfeld: I think it's worth doing, but can happen in a separate PR

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
anandtiwary: Because i think we can't put union types in model properties. This field can be string, number or undefined

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
anandtiwary: done

projects/distributed-tracing/src/shared/dashboard/widgets/span-detail/data/span-detail-data-source.model.ts
anandtiwary: will put  up a separate PR

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
aaron-steinfeld: But why not just say number (or string)? Given that we can't be as broad as we want (the code would work with string | number), it seems better to be narrower and force the value provider to match (vs having an overly broad type where some values would result in an error)

aaron-steinfeld (approved):

projects/distributed-tracing/src/shared/dashboard/widgets/trace-detail/data/trace-detail-data-source.model.ts
anandtiwary: Surely we could do that. The unknown pattern was there (may be i used it initially for a valid reason). With current design, i agree we could just force it to be of a particular type.

PR #440, opened 2020-12-17
Bar gauge

jake-bassett:
## Description
Screen Shot 2020-12-21 at 1 08 31 PM


projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -0,0 +1,51 @@
...
+      </div>
+    </div>
+  `
+})
+export class BarGaugeComponent implements OnChanges {
+  @Input()
+  public title?: string;
+
+  @Input()
+  public value?: number;
```
aaron-steinfeld, line 31: Your use case will need multiple segments, right? Fine if you want to do it iteratively, but will break the current API.

aaron-steinfeld (approved):

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@Component({
+  selector: 'ht-bar-gauge',
+  styleUrls: ['./bar-gauge.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="bar-gauge">
+      <div *ngIf="this.title" class="title">{{ this.title | htDisplayTitle }}</div>
+      <div class="count">
+        {{ this.value?.toLocaleString() }} / {{ this.maxValue?.toLocaleString() }}
```
anandtiwary, line 12: This could lead to `undefined / <value>`

anandtiwary: Since value is optional

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: Yeah, true. I'll need to add those as well as a legend.

Actually, let me fix the multiple values now. I'll do the legend in a follow up.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
anandtiwary: We could also use `DecimalPipe` here instead of calling that function

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.scss
```diff
@@ -0,0 +1,53 @@
...
+    @include subtitle-1($gray-9);
+    padding-bottom: 8px;
+
+    .units {
+      @include body-small($gray-5);
+    }
+  }
+
+  .bar {
+    position: relative;
```
anandtiwary, line 19: Any reason for not using flexbox here?

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
anandtiwary: For multiple values, can the api be similar to the ht-gauge(For consistency)?

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.scss
jake-bassett: No real reason. This just seemed easier and more straight forward.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: I made it pretty close to ht-guage. I didn't wan the user to have to calculate min and max of each segment though, so that's a little different here.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  private checkNearMaxValue(): void {
+    setTimeout(() => {
+      /*
+       * We want to remove the white divider if we fill up the bar. Compare segments to
+       * max bar width minus 2px for the divider width.
+       */
+      this.nearMaxValue = this.segmentBars.nativeElement.offsetWidth >= this.maxValueBar.nativeElement.offsetWidth - 2;
```
aaron-steinfeld, line 74: This feels like a very confusing way to do this, but I'm not exactly sure what we're trying to do. Are we adding dividers that are adding a fixed width to the bar? Could we measure the segments to be like ` calc( percentage - dividerWidth )` ? Any time we start measuring, or setting timeouts (let alone both) my alarm bells start going off.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private calcTotalValueFromSegments(segments: Segment[]): number {
+    return segments.reduce((previousValue, currentValue) => previousValue + currentValue.value, 0);
+  }
+}
+
+export interface Segment {
+  value: number;
+  color?: string;
```
aaron-steinfeld, line 96: If color isn't provided, what are we defaulting to?

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -0,0 +1,102 @@
...
+
+@Component({
+  selector: 'ht-bar-gauge',
+  styleUrls: ['./bar-gauge.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="bar-gauge">
+      <div *ngIf="this.title" class="title">{{ this.title | htDisplayTitle }}</div>
+      <div class="count">
+        {{ this.totalValue | number }} / {{ this.maxValue | number }}
```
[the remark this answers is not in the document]
jake-bassett: This will work. The idea behind the component is that its more like a stacked bar, where each segment is a part of a whole to have meaning. The total and max here are for all of the segments together.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: $blue-7

Screen Shot 2020-12-18 at 1 20 50 PM

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: Yeah, it could be confusing to look at in code without playing with the component a bit. I'll add a clearer comment and see if that makes more sense.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -0,0 +1,123 @@
...
+      this.totalValue = this.calcTotalValueFromSegments(this.segments ?? []);
+      this.barSegments = this.toBarSegments(this.segments ?? [], this.maxValue ?? 0);
+
+      this.overMaxValue = this.maxValue !== undefined && this.totalValue > this.maxValue;
+
+      this.checkNearMaxValue();
+    }
+  }
+
+  private checkNearMaxValue(): void {
```
anandtiwary, line 80: Does this need to re-run if dom layout changes? We could use the layout change directive for it.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
aaron-steinfeld: Should we use color service instead to generate a palette so the segments are different colors?

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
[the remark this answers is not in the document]
aaron-steinfeld: It has meaning, but visually if there are three segments and the label is treating them as one, what is the actual meaning of the individual segments?

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
aaron-steinfeld: It just seems like, given the fact that we know the divider sizes, we don't need to do it dynamically like this - we could size the segments to leave room for dividers, or size the container to subtract the total divider width, right?

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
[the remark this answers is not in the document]
jake-bassett: Here is the mock. I can only assume that we are going to subdivide and categorize whatever we are counting here. If we are looking at calls, maybe we can subdivide and categorize on types of calls. For the bar gauge where we show byte count, maybe we could subdivide on request vs response.

Screen Shot 2020-12-21 at 9 31 43 AM

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: Maybe I'm not following... or maybe not explaining correctly. Probably both.

Each segment is capped with the white divider line to create a visual barrier between the different segments (or the rightmost segment and the gray background). But the point of this code is to remove that white line for the right most bar when it is close to maxing out and touching the end of the gray background bar. We want this so that the white divider doesn't break the visual edge of the bar's perimeter due to it being white on a white background. 

Screen Shot 2020-12-21 at 9 31 43 AM

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: I think that makes sense once we have a use case for it.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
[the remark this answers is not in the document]
aaron-steinfeld: ok gotcha - it looks like we have both the total and the individual segment data - I was focussing more on the bottom part/legend. Is the plan for that to come in later?

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: Eh, screw it. I'll just do this now. 👍

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
aaron-steinfeld: So I understood the general goal, but one thing that did confuse me is I thought all dividers were behind hid (since we're applying that to each divider) - but it looks like we then use css to ignore the non-final ones - keeping that class only on the last segment would help clarify that confusion. 

My general ask I think holds though - it seems like there are ways to do this without having to make the whole component async + layout sensitive. A few different ideas (may not all work):
- do the comparison based on the values and not the measured pixels (can do this part of the segment-building logic), removing the bar if the data is at 100%. There may be a couple of gauge size/data combinations that will slightly misbehave, but they'd be exceedingly rare (and would only result in a 1px visual issue).
- Make the divider transparent rather than white (may need to change the div layout, don't remember how the background colors were applied)
- Make the container clip any overflow

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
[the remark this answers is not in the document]
jake-bassett: I'm having just so much fun iterating on this "simple" component... you know what, I'm going to go ahead and do the legend now too. Weeeeeee!

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: Tried a few things. Managed to get rid of the setTimeout, but didn't like any other solutions.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -26,9 +34,9 @@ import { DomElementMeasurerService, getPercentage, TypedSimpleChanges } from '@h
           <div class="segment-bars">
             <div
               #segmentBars
-              *ngFor="let segment of this.barSegments"
+              *ngFor="let segment of this.barSegments; let index = index"
```
aaron-steinfeld, line 37: Doesn't look like index is used (but it reminds me, you could use `last` like `let last = last` and then change the condition to something like `last && this.isGaugeFull`- I took liberties with the name, could also leave it)

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -64,45 +84,52 @@ export class BarGaugeComponent implements OnChanges {
   public overMaxValue: boolean = false;
   public nearMaxValue: boolean = false;
 
-  public constructor(private readonly domElementMeasurerService: DomElementMeasurerService) {}
+  public constructor(
+    private readonly domElementMeasurerService: DomElementMeasurerService,
+    private readonly colorService: ColorService
+  ) {
+    this.colorService.registerColorPalette(BarGaugeComponent.BAR_GAUGE_COLORS, [Color.Blue7, Color.Blue4, Color.Blue2]);
```
aaron-steinfeld, line 91: This isn't specific to this PR but a metacomment since this pattern already exists: why is a color palette tied to a component vs a function (like visualization, health, etc.)? In other words, why would a pie vs bar vs line chart use different color palettes? I think this is a design system issue that's bothered me forever.

aaron-steinfeld (approved):

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
aaron-steinfeld: I'm satisfied enough without setTimeout, but I maintain there's a css-only way of accomplishing this 😉

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: Ah, thanks. That was left over from one of my other attempts. I'll remove.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: Yeah, I was wondering myself but saw this pattern elsewhere when I searched for how the colorService was being used. It does seem like something that should be global and tied to function, though I'm guessing that the divergence in color palettes is coming from mock colors changing over time.

Agree we should revisit this at some point.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
jake-bassett: There may be, let's come back to it if needed. Thanks! 👍

PR #472, opened 2020-12-31
feat: styles support for summary value

arjunlalb:
## Description
Support for styling summary value as link or text.

### Testing
UTs added. Manually tested.

### Checklist:


projects/components/src/summary-value/summary-value.component.ts
```diff
@@ -28,6 +28,9 @@ export class SummaryValueComponent implements OnChanges {
   @Input()
   public tooltip?: string;
 
+  @Input()
+  public summaryValueDisplayStyle?: SummaryValueDisplayStyle = SummaryValueDisplayStyle.Text;
```
skjindal93, line 32: Can get rid of the optional flag, since it's hardcoded

skjindal93 (approved):

aaron-steinfeld (approved):

projects/components/src/summary-value/summary-value.component.ts
```diff
@@ -28,6 +28,9 @@ export class SummaryValueComponent implements OnChanges {
   @Input()
   public tooltip?: string;
 
+  @Input()
+  public summaryValueDisplayStyle: SummaryValueDisplayStyle = SummaryValueDisplayStyle.Text;
```
anandtiwary, line 32: Probably this could have been avoided if we would just put summary-value inside an ht-link(as commented in the other PR)

```
<ht-link [params]="<>">
 <ht-summary-value></ht-summary-value>
 </ht-link>
```

anandtiwary: We may have to update the color to currentColor. I think I remember using this approach somewhere.

aaron-steinfeld: If that works, that's better and let's revert this

PR #484, opened 2021-01-07
fix table app frame and title content scss to work with safari

jake-bassett:
## Description
Topology, donut, radar, and table were not showing in Safari due to Safari not handling flexbox the same as other browsers.

### DIFF
--- a/projects/components/src/table/table.component.scss
+++ b/projects/components/src/table/table.component.scss
@@ -19,7 +19,7 @@
 
 .table {
   width: 100%;
-  flex: 1 1 0;
+  flex: 1 1;
   overflow: auto;
 }
 
@@ -65,7 +65,7 @@
 
 .header-cell {
   display: flex;
-  flex: 1 1 0;
+  flex: 1 1;
   overflow: hidden;
 
   .header-cell-renderer {
@@ -95,7 +95,7 @@
 }
 
 .data-cell {
-  flex: 1 1 0;
+  flex: 1 1;
   overflow: hidden;
   border-bottom: 1px solid $gray-1;
 
--- a/projects/components/src/titled-content/titled-content.component.scss
+++ b/projects/components/src/titled-content/titled-content.component.scss
@@ -32,6 +32,8 @@
   }
 
   .content {
+    height: 100%;
+    width: 100%;
     flex: 1 1 auto;
     overflow: hidden;
   }
--- a/src/app/application-frame/application-frame.component.scss
+++ b/src/app/application-frame/application-frame.component.scss
@@ -22,7 +22,7 @@
 }
 
 .app-content-with-header {
-  flex: 1 1 0;
+  flex: 1 1;
   height: 100%;
   width: 100%;
   display: flex;


aaron-steinfeld (approved):
Assume this was tested with all browsers? I feel like we sometimes go in circles between fixing for safari vs other browsers.

projects/components/src/table/table.component.scss
```diff
@@ -19,7 +19,7 @@
 
 .table {
   width: 100%;
-  flex: 1 1 0;
```
anandtiwary, line 22: AFAIR, I had to add this to take care of few layout issues. Must be worth testing with different layout scenarios.

jake-bassett:
These are pretty safe changes that we know work, but yeah tested in FF and Chrome.

projects/components/src/table/table.component.scss
jake-bassett: Which areas did you need this for? This didn't break the tables I looked at.

flex-basis: 0 doesn't work the way we expect in Safari, so we shouldn't use it like the above example anymore.

anandtiwary: I don't remember in detail, but few places to check would be:
1. Table with 2-3 rows should show pagination at the bottom
2. Loader should be shown correctly for all cases. Check User behavior table as well after applying a toggle filter
3. Explorer table

jake-bassett: Ah, thanks. So I checked 1 and 2 in a few places and that seems fine, but Explorer table is expanding beyond the bounds of the page and not showing the pagination now. I'll find an alternative solution with a follow up shortly.

anandtiwary: Just wondering why the browserlist and autoprefixer is not able to take care of flex-basis 0 for safari. Do we need to make any changes to `browserlistrc`? 
```
# This file is currently used by autoprefixer to adjust CSS to support the below specified browsers
# For additional information regarding the format and rule options, please see:
# https://github.com/browserslist/browserslist#queries
#
# For IE 9-11 support, please remove 'not' from the last line of the file and adjust as needed

> 0.5%
last 2 versions
Firefox ESR
not dead
not IE 9-11
```

anandtiwary: If i run `npx browserlist` i see following supported browsers:

```
npx browserslist
and_chr 86
and_ff 82
and_qq 10.4
and_uc 12.12
android 81
baidu 7.12
chrome 87
chrome 86
chrome 85
chrome 84
edge 87
edge 86
edge 85
firefox 83
firefox 82
firefox 81
firefox 78
ios_saf 14
ios_saf 13.4-13.7
ios_saf 13.3
ios_saf 12.2-12.4
kaios 2.5
op_mini all
op_mob 59
opera 72
opera 71
safari 14
safari 13.1
samsung 12.0
samsung 11.1-11.2
```

So it should take care of safari 14, safari 13.1 ? Either autoprefixer is buggy or the this bug showed up on a much older safari.

jake-bassett: I'm seeing the issues on version 14.0.2, so not sure its about adding webkit specific properties like browserlist (I think?) works. It seems more like the way webkit chooses to handle flex.

jake-bassett: https://github.com/hypertrace/hypertrace-ui/pull/485

PR #485, opened 2021-01-07
fix: explorer table expanding beyond page bounds

jake-bassett:
## Description
After fixing the Safari issues with flexbox, the Explorer table was expanding beyond the page limits. This PR fixes that and a few other small styling issues with table in explorer.

## Testing
Manual testing of this table and tables in other parts of the product, including entity lists, events, and api def.

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
@@ -7,6 +7,7 @@
 .table {
   flex: 1 1;
   width: 100%;
+  overflow: auto;
 }
 
 .header-margin {
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
@@ -5,9 +5,8 @@
 }
 
 .table {
-  flex: 1 1;
-  width: 100%;
-  overflow: hidden;
+  flex: 1;
+  min-height: 0;
 }
 
 .header-margin {
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
@@ -7,7 +7,7 @@
 .table {
   flex: 1 1;
   width: 100%;
-  overflow: auto;
+  overflow: hidden;
 }
 
 .header-margin {
--- a/projects/components/src/paginator/paginator.component.scss
+++ b/projects/components/src/paginator/paginator.component.scss
@@ -8,7 +8,7 @@
   align-items: center;
 
   height: $paginator-height;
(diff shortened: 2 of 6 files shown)


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.scss
```diff
@@ -7,6 +7,7 @@
 .table {
   flex: 1 1;
   width: 100%;
```
aaron-steinfeld, line 9: this scares me - shouldn't the table be sized not to overflow?

jake-bassett: Yeah, its odd. But this is the overflow for the table-widget-renderer. There is another overflow:auto in table itself that usually does the scrolling, but in explorer due to the way the page flexes, it is required here too.

jake-bassett: Changed it to hidden, because that's really more accurate since it isn't actually doing the scrolling.

aaron-steinfeld: I pulled it locally and chromium looks fine, but safari doesn't - 
image

aaron-steinfeld: Also, using `min-height: 0;` instead of  the overflow seems to have the same effect in chromium (but also, unfortunately the same effect in safari - perhaps that's a pre-existing issue?)

jake-bassett: How wonderful. 

Thanks. Checking now.

jake-bassett: After looking a bit, I'm guessing this must have been existing for a while. We just don't get Safari traffic so it hasn't been reported.

Can we get this fix in to at least get things fixed on Chrome and FF? I'll follow up with a fix for Safari.

aaron-steinfeld: Yep - do you think it makes min-height instead of overflow here? I was feeling it was better because overflow (auto or hidden) is describing behavior we don't actually want - it should never overflow by design. Min-height works because the issue that I was seeing is the default of min-height:auto in flex col layout was preventing shrinking.

jake-bassett: Sure, agreed. Let me test that change in the Safari fix and use min-height instead if it works.

jake-bassett: Think I found a solution that works for Safari. Just going to tack it on to this one.

aaron-steinfeld: nice!

aaron-steinfeld (approved):

PR #486, opened 2021-01-09
fix(gauge): y origin can be equal to radius

skjindal93:
## Description
- Added extra arc of `Math.PI / 12` at the bottom of the gauge component, instead of keeping it as semicircle. Adding the extra arc to the semicircle will require extra padding at the bottom, but it cannot be a fixed padding, because the larger the radius, the greater the padding required at the bottom. Approximating extra arc length as the extra padding required at the bottom
- Gauge y origin can be optimized to radius value always
- Reduced the arc width

### Testing
Tested with various width and height for the gauge component. It works beautifully auto computing the extra space required below

Smaller radius
Screen Shot 2021-01-09 at 5 58 41 PM

Larger radius
Screen Shot 2021-01-09 at 5 59 01 PM

### Checklist:


aaron-steinfeld (approved):

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -37,10 +37,10 @@ import { Arc, arc, DefaultArcObject } from 'd3-shape';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class GaugeComponent implements OnChanges {
-  private static readonly GAUGE_RING_WIDTH: number = 20;
+  private static readonly GAUGE_RING_WIDTH: number = 15;
   private static readonly GAUGE_ARC_CORNER_RADIUS: number = 10;
-  private static readonly GAUGE_AXIS_PADDING: number = 30;
   private static readonly GAUGE_MIN_RADIUS_TO_SHOW_PATH: number = 80;
+  private static readonly EXTRA_ARC_ANGLE: number = Math.PI / 12;
```
anandtiwary, line 43: Is the extra arc angle added to `start` and `end` so that the label appears more inside? Instead of this, could we have done this by updating the position of `text` on line 30?

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -107,8 +107,8 @@ export class GaugeComponent implements OnChanges {
     return this.buildArcGenerator()({
       innerRadius: radius - GaugeComponent.GAUGE_RING_WIDTH,
       outerRadius: radius,
-      startAngle: -Math.PI / 2,
-      endAngle: -Math.PI / 2 + (inputData.value / inputData.maxValue) * Math.PI
+      startAngle: -Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE,
+      endAngle: -Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE + (inputData.value / inputData.maxValue) * Math.PI
```
anandtiwary, line 111: This seems incorrect to me. 

case 1:  `inputData.value = 0`,  the end angle would be `-Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE` which is same as start angle. So the arc will be completely empty. Expected outcome.

case 2: `inputData.value = inputData.maxValue`, in this case the end angle of value arc should be equal to max end angle of background arc, that is `Math.PI / 2 + GaugeComponent.EXTRA_ARC_ANGLE`. Line 111, would make it `-Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE + 1 * Math.PI = Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE`

So the value arc would never be completely full.

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -117,13 +117,32 @@ export class GaugeComponent implements OnChanges {
...
+       * Since height > width > 2, radius can be treated as (width / 2), but only for semi-circles
+       *
+       * If there is an extra angle added to the semicircle,
+       * radius + extra arc length can over shoots the height
+       *
+       * If radius + extra arc length over shoots the height, then radius
+       * needs to be calculated using the provided height, instead of width / 2
+       */
+      const radius = width / 2;
+      const extraArcHeight = this.calculateExtraArcLength(radius);
```
anandtiwary, line 134: If we adjust the text position, could we avoid all this additional logic due to the extra arc length?

projects/observability/src/shared/components/gauge/gauge.component.ts
skjindal93: Good catch, I will fix that

PR #493, opened 2021-01-11
fix: add data to sheet overlay config

arjunlalb:
## Description
Add data to sheet overlay config.


projects/components/src/overlay/sheet/sheet.ts
```diff
@@ -1,7 +1,8 @@
 import { OverlayConfig } from './../overlay';
 
-export interface SheetOverlayConfig extends OverlayConfig {
+export interface SheetOverlayConfig<TData = unknown> extends OverlayConfig {
```
aaron-steinfeld, line 3: Hmm - so currently sheet contents would have access to this config by injecting @POPOVER_DATA, right?. I'd argue that's a bug in the injector design for sheet, which doesn't properly isolate the content from the container implementation. In other words, look at popover or modal, our other two constructs that work this way. They build a parallel injector for the content so the internal implementation of the container (that is, SheetOverlayConfig and POPOVER_DATA) doesn't get exposed to the content.

Ideally I'd like to fix the issue with injectors, as no sheet contents currently rely on POPOVER_DATA today. If not, can we avoid adding a dependency on it by using a template which already has the data passed in (see waterfall widget renderer for an example of this).

aaron-steinfeld (approved):

PR #494, opened 2021-01-11
fix(gauge): value arc angle computation

skjindal93:
Fixes https://github.com/hypertrace/hypertrace-ui/pull/486#discussion_r554518546


projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -104,11 +104,13 @@ export class GaugeComponent implements OnChanges {
...
 
   private buildValueArc(radius: number, inputData: GaugeInputData): string {
+    const completeAngle = Math.PI / 2 + GaugeComponent.EXTRA_ARC_ANGLE;
+
     return this.buildArcGenerator()({
       innerRadius: radius - GaugeComponent.GAUGE_RING_WIDTH,
       outerRadius: radius,
-      startAngle: -Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE,
-      endAngle: -Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE + (inputData.value / inputData.maxValue) * Math.PI
+      startAngle: -completeAngle,
```
aaron-steinfeld, line 112: Does this handle the first case described in the comment? If inputData.value is 0, we're doing `[-completeAngle, -completeAngle]` so there'd still be no arc

skjindal93: Yes. We want no value arc, when the inputData.value is 0. In fact @anandtiwary mentioned that's the expected outcome as well

>case 1: inputData.value = 0, the end angle would be -Math.PI / 2 - GaugeComponent.EXTRA_ARC_ANGLE which is same as start angle. So the arc will be completely empty. Expected outcome.

aaron-steinfeld: Ah I misread his comment - I thought our ux generally has a little arc for a 0-value gauge to show it's "emptyiness"? I'm pretty sure that's what we do in the linear case.

skjindal93: hmm. fair point. I can use a value of 0.5, if the value is 0 to display the 0-value arc

aaron-steinfeld: I think that makes sense - something like `percentFill = Math.min(0.01, inputData.value / inputData.maxValue)`

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -110,7 +110,7 @@ export class GaugeComponent implements OnChanges {
       innerRadius: radius - GaugeComponent.GAUGE_RING_WIDTH,
       outerRadius: radius,
       startAngle: -completeAngle,
-      endAngle: -completeAngle + 2 * (inputData.value / inputData.maxValue) * completeAngle
+      endAngle: -completeAngle + 2 * Math.max(0.05, inputData.value / inputData.maxValue) * completeAngle
```
aaron-steinfeld, line 113: `.05` feels high here, how does it look? It's showing a 5% fill in the empty case - I was thinking a 1% fill would be sufficient.

skjindal93: 0.01
Screen Shot 2021-01-11 at 9 56 20 PM

0.02
Screen Shot 2021-01-11 at 9 58 24 PM

0.05
Screen Shot 2021-01-11 at 9 56 42 PM

skjindal93: 0.05 looks high, as you said. Looks like a real value. 

0.02 seems to look good as compared to 0.01

projects/observability/src/shared/components/gauge/gauge.component.ts
anandtiwary: I am not sure it looks any better to show a little arc for 0. 

Also, could we rename completeAngle to just startAngle?

skjindal93: > Also, could we rename `completeAngle` to just startAngle and assign it `-completeAngle`

Didn't catch the comment

anandtiwary: `CompleteAngle` is confusing to me. Can we do `const startAngle = - (Math.PI / 2 + GaugeComponent.EXTRA_ARC_ANGLE);` ?

anandtiwary: So, `endAngle =   startAngle + 2 * (inputData.value / inputData.maxValue) * (- startAngle)

anandtiwary:
Anything more needed on this PR? For max risk score, the gauge would not appear full. This seems like an important bug

aaron-steinfeld:
> Anything more needed on this PR? For max risk score, the gauge would not appear full. This seems like an important bug

yes we should fix. I think the consensus was no nub, right @jake-bassett ?

skjindal93:
> > Anything more needed on this PR? For max risk score, the gauge would not appear full. This seems like an important bug
> 
> yes we should fix. I think the consensus was no nub, right @jake-bassett ?

got rid of the nib computation

aaron-steinfeld (approved):

anandtiwary (approved):

PR #499, opened 2021-01-13
Top level apis

jake-bassett:
## Description
We want to remove the nested Service -> APIs list and just use a flat list. This PR makes that change. 

It also:
* Adds the ability to configure dropdown filters to the table controls
* Introduces a secondary-entity-specification so we can have a second `display: Entity` column in the table that links to the service

Screen Shot 2021-01-13 at 2 31 54 PM

Screen Shot 2021-01-13 at 2 40 49 PM

Screen Shot 2021-01-13 at 2 32 10 PM


projects/common/src/utilities/types/types.ts
```diff
@@ -31,3 +31,4 @@ export interface Json {
 }
 
 export type JsonValue = Json | Json[] | string | number | boolean | null;
+export type PrimitiveValue = string | number | boolean;
```
jake-bassett, line 34: Oops. I ended up not using this. We can remove or keep.

projects/observability/src/shared/dashboard/data/graphql/specifiers/secondary-entity-specification.model.ts
```diff
@@ -0,0 +1,46 @@
...
+import { SpecificationModel } from '@hypertrace/distributed-tracing';
+import { Model, ModelProperty, STRING_PROPERTY } from '@hypertrace/hyperdash';
+import { Entity, EntityType } from '../../../../graphql/model/schema/entity';
+import { EntitySpecification } from '../../../../graphql/model/schema/specifications/entity-specification';
+import { SecondaryEntitySpecificationBuilder } from '../../../../graphql/request/builders/specification/entity/secondary-entity-specification-builder';
+
+@Model({
+  type: 'secondary-entity-specification'
+})
+export class SecondaryEntitySpecificationModel extends SpecificationModel<EntitySpecification> {
```
anandtiwary, line 11: Still trying to understand this. how will this be used?

projects/observability/src/shared/graphql/request/builders/specification/entity/secondary-entity-specification-builder.ts
```diff
@@ -0,0 +1,54 @@
+import { Dictionary } from '@hypertrace/common';
+import { GraphQlArgumentBuilder } from '@hypertrace/distributed-tracing';
+import { GraphQlSelection } from '@hypertrace/graphql-client';
+import { Entity, entityIdKey, EntityType, entityTypeKey } from '../../../../model/schema/entity';
+import { EntitySpecification } from '../../../../model/schema/specifications/entity-specification';
+
+export class SecondaryEntitySpecificationBuilder {
```
anandtiwary, line 7: Can we reuse existing ObservabilitySpec builder for this?

projects/observability/src/pages/apis/services/services-list.dashboard.ts
```diff
@@ -1,194 +1,129 @@
...
+      'enable-style': false
+    },
+    children: [
+      {
+        type: 'table-widget',
+        id: 'services-list.table',
+        mode: TableMode.Flat,
+        style: TableStyle.FullPage,
+        searchAttribute: 'name',
+        selectFilterOptions: [
```
aaron-steinfeld, line 25: After all that, we're only using a single filter?!

projects/observability/src/pages/apis/services/services-list.dashboard.ts
```diff
@@ -1,194 +1,129 @@
...
+          {
+            type: 'table-widget-select-filter',
+            attribute: 'serviceName',
+            placeholder: 'All Services',
+            data: {
+              type: 'entities-attribute-data-source',
+              entity: 'API',
```
aaron-steinfeld, line 32: Can we fetch `SERVICE.name` instead of `API.serviceName`? Looks like it should work the same way and prevents dupes/easier to query.

projects/observability/src/shared/dashboard/data/graphql/specifiers/secondary-entity-specification.model.ts
aaron-steinfeld: This seems identical in purpose to the existing `entity-specification-model` and has the same api (except casing - model's use `kebab-case`). Specs never really have a concept of "primary" or "secondary" since they don't provide the context - they just provide one or more selections under a parent context.

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -60,6 +76,9 @@ export class TableControlsComponent implements OnChanges {
   @Input()
   public searchPlaceholder?: string = 'Search...';
 
+  @Input()
+  public selectMap?: Map<string, SelectFilter>;
```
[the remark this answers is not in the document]
jake-bassett: I am cleaning up this code. Agreed its too confusing. Update incoming.

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -79,6 +98,9 @@ export class TableControlsComponent implements OnChanges {
   @Input()
   public checkboxChecked?: boolean;
 
+  @Output()
+  public readonly selectChange: EventEmitter<KeyValue<string, unknown>> = new EventEmitter<KeyValue<string, unknown>>();
```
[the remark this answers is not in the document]
jake-bassett: They are different though. The input is how we configure all of the select dropdown options. The output is only the one that changed.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -152,6 +166,41 @@ export class TableWidgetRendererComponent
     return this.data$!.pipe(map(data => data?.getScope()));
   }
 
+  private fetchSelectFilterValues(): void {
+    this.model.getSelectFilterOptions().forEach(option => {
```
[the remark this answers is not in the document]
jake-bassett: Yep! I just added some comments and cleaned up the code a bit more. Agreed it was a bit confusing before. Update incoming.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -152,6 +166,41 @@ export class TableWidgetRendererComponent
...
+      forkJoinSafeEmpty([
+        option.getData().pipe(
+          first(),
+          map((resultOptions: unknown[]) => {
+            const options: SelectOption<unknown>[] = resultOptions.map(value => ({
```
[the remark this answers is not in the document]
jake-bassett: Fixed the typings.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -152,6 +166,41 @@ export class TableWidgetRendererComponent
...
+            return {
+              attribute: option.attribute,
+              selectFilter: {
+                placeholder: option.placeholder,
+                options: options
+              }
+            };
+          })
+        )
+      ]).subscribe((selects: AttributeSelectFilter[]) => {
```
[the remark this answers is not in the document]
jake-bassett: Fixed. Thanks!

projects/observability/src/shared/graphql/request/builders/specification/entity/secondary-entity-specification-builder.ts
jake-bassett: Yes!

projects/observability/src/shared/dashboard/data/graphql/specifiers/secondary-entity-specification.model.ts
jake-bassett: Yeah, I can't remember what lead me down the road to this since I've picked up and shelved this PR so many times. Unneeded now though. Removing.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-select-filter.model.ts
```diff
@@ -0,0 +1,42 @@
...
+    type: STRING_PROPERTY.type
+  })
+  public placeholder?: string;
+
+  @ModelInject(MODEL_API)
+  protected readonly api!: ModelApi;
+
+  public getData(): Observable<PrimitiveValue[]> {
+    return this.api
+      .getData<Dictionary<PrimitiveValue>[]>()
```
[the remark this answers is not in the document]
aaron-steinfeld: I proposed one tweak in another comment, but alternately, do we even need a fetch? Could we use the fetched data to populate this and just pass the data to the filter model to transform into a filters? That's basically what we're doing with the existing table filters right? The downside would be the same, that our filter options are only based on the queried data.

projects/observability/src/pages/apis/services/services-list.dashboard.ts
aaron-steinfeld: moved this conversation to https://github.com/hypertrace/hypertrace-ui/pull/499#discussion_r558562306

projects/observability/src/shared/dashboard/data/graphql/entity/attribute/entities-attribute-data-source-model.ts
```diff
@@ -0,0 +1,38 @@
+import { AttributeSpecificationModel, Specification } from '@hypertrace/distributed-tracing';
```
aaron-steinfeld, line 1: file name `*.model.ts`

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -135,6 +155,51 @@ export class TableWidgetRendererComponent
       value: modeOption
     }));
 
+    this.selectFilterItems$ = forkJoinSafeEmpty(
```
aaron-steinfeld, line 158: Since this is fetching, we should make sure it gets refreshed when the widget refreshes (it also shouldn't fetch until fetchData is called, I think moving this into that function might solve both).

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-select-filter.model.ts
```diff
@@ -18,10 +18,10 @@ export class TableWidgetSelectFilterModel {
   public attribute!: string;
 
   @ModelProperty({
-    key: 'placeholder',
+    key: 'unsetOption',
```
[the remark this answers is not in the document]
jake-bassett: Ew, but fine. The mix of quoted and unquoted keys in the typescript drives me crazy. I know kabab is more correct, but I don't have to like it!

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -77,7 +77,7 @@ export class TableControlsComponent implements OnChanges {
   public searchPlaceholder?: string = 'Search...';
 
   @Input()
-  public selectMap?: Map<string, SelectFilter>;
+  public selectFilterItems?: SelectFilter[] = [];
```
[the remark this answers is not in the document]
jake-bassett: It shouldn't be a problem. The Output from the dropdown is the value, which is now an object keyValue pair of attribute:value so the listener knows which filter was changed.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-select-filter.model.ts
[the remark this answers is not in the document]
jake-bassett: Aaaaah, right. Agree they should be consistent. Fixed mine to just provide an array of attributes instead of the full Entity. Much better.

As for the single fetch, I think this the independent fetch is still a better route. I don't really like the way I did the header filters where we tie it to the table data. It couple everything too closely. Granted, if you are filtering on a table, the data between the filer and the table should match so technically they _are_ coupled. 

But by having them independently fetched we have more freedom to fetch results that might not make be fetched by the table due to pagination or other filters (maybe debatable if that second part is good or bad). It also allows us to have filters for columns we don't even show.

I think I'd like to go with what we have now (along with your other suggestions).

projects/observability/src/pages/apis/services/services-list.dashboard.ts
```diff
@@ -1,194 +1,129 @@
...
+        style: TableStyle.FullPage,
+        searchAttribute: 'name',
+        selectFilterOptions: [
+          {
+            type: 'table-widget-select-filter',
+            attribute: 'serviceName',
+            'unset-option': 'All Services',
+            data: {
+              type: 'entities-attribute-data-source',
+              entity: ObservabilityEntityType.Api,
```
aaron-steinfeld, line 32: Given the change in the attribute data source, this can change to SERVICE.name right? (the main reason I'm pushing down this path is actually the scale of api data - there could be thousands, so fetching only the services seems much more efficient and less likely to hit the hardcoded limit I think I saw - 100, right?)

projects/observability/src/pages/apis/services/services-list.dashboard.ts
```diff
@@ -1,194 +1,129 @@
...
+            width: '20%',
+            value: {
+              type: 'entity-specification',
+              'id-attribute': 'serviceId',
+              'name-attribute': 'serviceName',
+              'entity-type': ObservabilityEntityType.Service
+            }
+          },
+          {
+            type: 'table-widget-column',
```
aaron-steinfeld, line 63: These two are no longer needed (service id + name)

projects/observability/src/shared/dashboard/data/graphql/entity/entities-values-data-source.model.ts
```diff
@@ -0,0 +1,35 @@
...
+      map(response => response.results),
+      map(results => results.map(result => result[this.specification.resultAlias()]))
+    );
+  }
+
+  private buildRequest(specification: Specification, inheritedFilters: GraphQlFilter[]): GraphQlEntitiesQueryRequest {
+    return {
+      requestType: ENTITIES_GQL_REQUEST,
+      entityType: this.entityType,
+      limit: 100,
```
aaron-steinfeld, line 29: I suspect we'll want to expose the limit as a model prop at some point, but if we don't need it now that's fine.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -226,6 +297,22 @@ export class TableWidgetRendererComponent
...
+  public onSelectChange(select: KeyValue<string, unknown>): void {
+    const selectFilterModel = this.model.getSelectFilterOptions().find(option => option.attribute === select.key);
+
+    if (selectFilterModel === undefined) {
+      return; // This shouldn't happen
+    }
+
+    this.selectFilterSubject.pipe(first()).subscribe(filters => {
+      this.selectFilterSubject.next(
+        [...filters.filter(f => f.field !== select.key), selectFilterModel.getTableFilter(select.value)].filter(
```
aaron-steinfeld, line 309: I'm still really unclear what this is doing - removing the selected filter from the list and filtering out unset filters? Why is either necessary? the subscribe looks like it's just to get the current value too, so changing the declared type of `selectFilterSubject` to `BehaviorSubject` (which is already its implementation) will give you direct access to the current value.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-select-filter.model.ts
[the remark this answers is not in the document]
aaron-steinfeld: Yeah I hear you - I don't recall why we went with kebab originally, but now it's just consistency.

projects/components/src/table/controls/table-controls.component.ts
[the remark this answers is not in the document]
aaron-steinfeld: But that's in the caller - it's now on the caller to provide a value that it knows how to disambiguate from other values in other filters, right? I'm fine leaving it as is since the renderer is likely the only place we're using these controls in a complex way, but from a generic component perspective, I think that's a design issue.

(an alternative would be something like `{source: SelectFilter; value: SelectOption<..>}`)

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -23,6 +25,20 @@ import { TableMode } from '../table-api';
           (valueChange)="this.onSearchChange($event)"
         ></ht-search-box>
 
+        <!-- Selects -->
+        <ht-select
```
skjindal93, line 29: Shouldn't the select dropdown be a multi select dropdown? For example: Show me all the APIs which lies in Service A or Service B or Service C

jake-bassett: This is actually good feedback. I think multi select makes sense. That said, let me get the single select version out and demoed. I'll make the switch to IN filtering once everyone is onboard with the new changes.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Yeah, it was just to get the current filters so we can check. I'll use the BehaviorSubject instead and clean this up a bit more.

projects/components/src/table/controls/table-controls.component.ts
skjindal93: yes. that makes sense

aaron-steinfeld: > This is actually good feedback

What are you saying?!

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -226,6 +298,27 @@ export class TableWidgetRendererComponent
...
+      return; // This shouldn't happen
+    }
+
+    const existingSelectFiltersWithChangedRemoved = this.selectFilterSubject
+      .getValue()
+      .filter(f => f.field !== changed.value.key);
+    const changedSelectFilter = selectFilterModel.getTableFilter(changed.value.value);
+
+    const selectFilters = [...existingSelectFiltersWithChangedRemoved, changedSelectFilter].filter(
+      f => f.value !== undefined
```
aaron-steinfeld, line 316: does the undefined filtering mean we can't switch back to unfiltered (isn't undefined the "All Services" option) ?

aaron-steinfeld (approved):

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Correct. The select option returning a value of `undefined` signals to the widget to drop that filter, so we remove it from the request all together.

aaron-steinfeld: I'm confused - I would expect to not use that filter in the request, but it should still remain in the dropdown list if I want to return to see all apis across services. The behavior I read/you describe sounds like it is removed from the selection list once an option is chosen, never to return.

PR #500, opened 2021-01-14
feat: add support for long format in multi unit string

arjunlalb:
## Description
Add support for long string format in multi unit string

### Testing

### Checklist:


projects/common/src/time/time-duration.ts
```diff
@@ -16,20 +16,35 @@ export class TimeDuration {
...
+    displayZero: boolean = true,
+    unitStringType: UnitStringType = UnitStringType.Short
+  ): string {
     const mostSignificantPortion = this.getMostSignificantUnitOnly();
     const remainingMillis = this.millis - mostSignificantPortion.toMillis();
     if (mostSignificantPortion.getAmountForUnit(smallestUnit) < 1) {
-      return displayZero ? new TimeDuration(0, smallestUnit).toString() : '';
+      return displayZero
+        ? unitStringType === UnitStringType.Long
+          ? new TimeDuration(0, smallestUnit).toLongString()
```
aaron-steinfeld, line 29: nit: a private method that accepts the unitStringType (i.e `private toFormattedString(type: UnitStringType): string`) would allow removing all of the ternaries and particularly the nested one.

projects/common/src/time/time-duration.ts
```diff
@@ -141,3 +156,8 @@ type ConvertibleTimeUnit =
   | TimeUnit.Minute
   | TimeUnit.Second
   | TimeUnit.Millisecond;
+
+export enum UnitStringType {
```
aaron-steinfeld, line 160:
```suggestion
export const enum UnitStringType {
```

projects/common/src/time/time-duration.ts
arjunlalb: did a bit of clean up here.

projects/common/src/time/time-duration.ts
```diff
@@ -24,27 +24,22 @@ export class TimeDuration {
...
-        ).toMultiUnitString(smallestUnit, false, unitStringType)}`
-      : `${mostSignificantPortion.toString()}${new TimeDuration(
-          remainingMillis,
-          TimeUnit.Millisecond
-        ).toMultiUnitString(smallestUnit, false)}`;
+    const joiningStr = unitStringType === UnitStringType.Long ? ' ' : '';
+
+    return [
+      this.toFormattedString(mostSignificantPortion, unitStringType),
+      new TimeDuration(remainingMillis, TimeUnit.Millisecond).toMultiUnitString(smallestUnit, false, unitStringType)
```
aaron-steinfeld, line 37: nit - this is an instance method that takes in a separate instance. 
```typescript
 private toFormattedString(unitStringType: UnitStringType = UnitStringType.Short): string {
    return unitStringType === UnitStringType.Short ? this.toString() : this.toLongString();
  }	 
```

and allow the other code using it to be a tad cleaner like:
```diff
- this.toFormattedString(new TimeDuration(0, smallestUnit), unitStringType)
+ new TimeDuration(0, smallestUnit).toFormattedString(unitStringType)
```

aaron-steinfeld (dismissed):

projects/common/src/time/time-duration.ts
```diff
@@ -141,3 +151,8 @@ type ConvertibleTimeUnit =
   | TimeUnit.Minute
   | TimeUnit.Second
   | TimeUnit.Millisecond;
+
+export enum UnitStringType {
```
aaron-steinfeld, line 155: I noticed this got switched back from const again? was const causing issues?

aaron-steinfeld (approved):

projects/common/src/time/time-duration.ts
arjunlalb: yep. using const fails when the enum is used as a param value in a method.

aaron-steinfeld: hmm - shouldn't be. I suspect that's a local caching artifact from the switch between const and non-const. Could you trying clearing your jest cache `npm run test -- --clear-cache` and then running with the const version?

PR #501, opened 2021-01-14
feat: support for link in summary value component

arjunlalb:
## Description
Support for link in summary values.

### Testing
UTs added.

### Checklist:

### DIFF
--- a/projects/components/src/summary-value/summary-value.component.ts
+++ b/projects/components/src/summary-value/summary-value.component.ts
@@ -1,4 +1,5 @@
 import { ChangeDetectionStrategy, Component, Input, OnChanges } from '@angular/core';
+import { NavigationParams } from '@hypertrace/common';
 import { isEmpty } from 'lodash-es';
 import { IconSize } from '../icon/icon-size';
 
@@ -11,7 +12,15 @@ import { IconSize } from '../icon/icon-size';
       <ht-icon *ngIf="this.icon" [icon]="this.icon" size="${IconSize.Small}" class="icon"></ht-icon>
       <div class="dot" *ngIf="!this.icon"></div>
       <div class="label" *ngIf="this.label">{{ this.label }}:</div>
-      <div class="value" [ngClass]="this.summaryValueDisplayStyle">{{ this.value }}</div>
+      <ht-link
+        *ngIf="this.summaryValueDisplayStyle === '${SummaryValueDisplayStyle.Link}'; else textValue"
+        class="link"
+        [paramsOrUrl]="this.paramsOrUrl"
+        >{{ this.value }}</ht-link
+      >
+      <ng-template #textValue>
+        <div class="value">{{ this.value }}</div>
+      </ng-template>
     </div>
   `
 })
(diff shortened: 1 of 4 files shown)


arjunlalb:
@aaron-steinfeld @anandtiwary Addressing comments from https://github.com/hypertrace/hypertrace-ui/pull/472

projects/components/src/summary-value/summary-value.component.ts
```diff
@@ -11,7 +12,15 @@ import { IconSize } from '../icon/icon-size';
       <ht-icon *ngIf="this.icon" [icon]="this.icon" size="${IconSize.Small}" class="icon"></ht-icon>
       <div class="dot" *ngIf="!this.icon"></div>
       <div class="label" *ngIf="this.label">{{ this.label }}:</div>
-      <div class="value" [ngClass]="this.summaryValueDisplayStyle">{{ this.value }}</div>
+      <ht-link
```
anandtiwary, line 15: We would use this in template directly. The only change we need in this component is to set currentColor = inherit in the summary value root class on hover. 

<ht-link [params]="<>">
 <ht-summary-value></ht-summary-value>
 </ht-link>

arjunlalb: Why do we want to avoid taking in `paramsOrUrl` here? Any particular reason?

aaron-steinfeld: Just to reduce complexity. It's easier to support composing a link and other components together vs giving each component a link mode and adding more inputs to then support that link mode. It looks like this one already has the mode but is not linking.

My preference, is to:
- Remove the mode of this component entirely (this may not be feasible/part of this PR depending on how it's used)
- Come up with some convention class, like `link-target` that can be applied to what should be styled as a link if wrapped in a link container
- Add encapsulation-piercing style in `ht-link` for `link-target` so that it can correctly style that in any content (I know, who would have guessed I'd be promoting ng-deep - talk me down!)

anandtiwary: @arjunlalb Mainly because `ht-link` is designed to make its content a navigable link. It also brings high cohesion to the two components `ht-link` and `ht-summary-value`, where each component is doing a specific thing. 

@aaron-steinfeld I think the ht-link is already doing #2 and #3. It sets the color to blue on hover. The component inside its ng-content has to just inherit this color.  I think if we do add something like this then it would work. I have definitely tried it before.

```
.ht-summary-value {
// assign normal colors here for icon and value
}
.ht-summary-value:hover {
  color: inherit;
  .icon {
     color: currentColor;
   }

  .value {
     @include body-small(currentColor)
  }
}
```

aaron-steinfeld: @anandtiwary  the approach you're describing works too, but it's slightly flipped from my suggestion. In that approach, we're not piercing encapsulation, but instead the child is tactically saying "I'll take this value from the parent" which means the child knows what styles the parent is applying. In the approach I was suggesting, the child is instead saying "apply your styles here" and the parent is then pushing them down in lieu of inheriting.

My preference is the parent doing the styling so that it can change it's styles without potentially breaking children, but I think both are better than the existing approach, so I'll defer to the author :)

PR #503, opened 2021-01-14
fix: spelling

jake-bassett:
## Description
DataAccesor -> DataAccessor


aaron-steinfeld (approved):

anandtiwary:
Should we add cSpell linter? It would work similar to prettier and would ensure that spell mistakes are captures before hand. :)

jake-bassett:
> Should we add cSpell linter? It would work similar to prettier and would ensure that spell mistakes are captures before hand. :)

I'm all for any automation that makes our lives easier.

PR #512, opened 2021-01-19
Fix pagination UI

DmitiryPotychkin:
## Description
- Change navigation buttons size, border color, arrows icon
- Change select popup size and position

Previous:
Screenshot from 2021-01-19 13-11-43

Now:
Screenshot from 2021-01-19 13-11-06

### DIFF
--- a/projects/assets-library/assets/styles/_layout.scss
+++ b/projects/assets-library/assets/styles/_layout.scss
@@ -1,6 +1,7 @@
 @import 'color-palette';
 
 $paginator-height: 50px;
+$paginator-select-width: 74px;
 
 @mixin flex-layout($direction: column) {
   height: 100%;
--- a/projects/components/src/button/button.component.scss
+++ b/projects/components/src/button/button.component.scss
@@ -188,6 +188,22 @@
   }
 }
 
+.small-square {
+  height: 32px;
+  width: 32px;
+  min-width: auto;
+  border-radius: 6px;
+
+  &:not(.icon-only) {
+    padding: 0 16px;
+  }
+
+  .label {
+    @include body-1-medium();
+    color: inherit;
+  }
+}
+
 .extra-small {
   height: 24px;
   border-radius: 4px;
--- a/projects/components/src/paginator/paginator.component.scss
+++ b/projects/components/src/paginator/paginator.component.scss
@@ -28,7 +28,8 @@
   .previous-button {
     ::ng-deep {
       .button {
-        border-radius: 4px 0 0 4px;
+        border-radius: 6px 0 0 6px;
+        border-color: $gray-2;
       }
     }
   }
@@ -36,7 +37,8 @@
   .next-button {
     ::ng-deep {
       .button {
-        border-radius: 0 4px 4px 0;
+        border-radius: 0 6px 6px 0;
+        border-color: $gray-2;
         margin-left: -1px;
       }
     }
--- a/projects/components/src/paginator/paginator.component.scss
+++ b/projects/components/src/paginator/paginator.component.scss
@@ -16,7 +16,7 @@
   }
 
   .page-size-select {
-    min-width: 86px;
+    min-width: $paginator-select-width;
     margin-left: 12px;
   }
 
--- a/projects/components/src/paginator/paginator.component.ts
+++ b/projects/components/src/paginator/paginator.component.ts
@@ -10,7 +10,8 @@ import {
 import { IconType } from '@hypertrace/assets-library';
 import { TypedSimpleChanges } from '@hypertrace/common';
 import { Observable } from 'rxjs';
-import { ButtonStyle } from '../button/button';
+import { ButtonStyle, ButtonSize } from '../button/button';
+import { SelectType } from '../select/select-type';
 import { PageEvent } from './page.event';
 import { PaginationProvider } from './paginator-api';
 
@@ -27,12 +28,14 @@ import { PaginationProvider } from './paginator-api';
       </ht-label>
 
       <div class="pagination-buttons">
-        <!-- TODO: Change ToggleButtonComponent to be more generic ButtonGroupComponent to accommodate this use case -->
+        <!-- TODO: Change 
+         to be more generic ButtonGroupComponent to accommodate this use case -->
         <ht-button
           class="button previous-button"
           htTooltip="Go to previous page"
           display="${ButtonStyle.Bordered}"
-          icon="${IconType.Previous}"
+          size="${ButtonSize.SmallSquare}"
+          icon="${IconType.ArrowLeft}"
           [disabled]="!this.hasPrevPage()"
           (click)="this.gotoPreviousPage()"
         >
@@ -41,7 +44,8 @@ import { PaginationProvider } from './paginator-api';
           class="button next-button"
           htTooltip="Go to next page"
           display="${ButtonStyle.Bordered}"
-          icon="${IconType.Next}"
+          size="${ButtonSize.SmallSquare}"
+          icon="${IconType.ArrowRight}"
           [disabled]="!this.hasNextPage()"
           (click)="this.gotoNextPage()"
         >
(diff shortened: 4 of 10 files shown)


projects/components/src/button/button.component.scss
```diff
@@ -188,6 +188,22 @@
   }
 }
 
+.small-square {
+  height: 32px;
+  width: 32px;
+  min-width: auto;
```
anandtiwary, line 194: Is it only different from `small` class by width and min-width? Can we achieve the same result with the `small` class?

projects/components/src/paginator/paginator.component.ts
```diff
@@ -27,12 +28,14 @@ import { PaginationProvider } from './paginator-api';
       </ht-label>
 
       <div class="pagination-buttons">
-        <!-- TODO: Change ToggleButtonComponent to be more generic ButtonGroupComponent to accommodate this use case -->
+        <!-- TODO: Change 
+         to be more generic ButtonGroupComponent to accommodate this use case -->
         <ht-button
```
anandtiwary, line 33: We also have a toggle button component, can we use that here instead?

projects/components/src/select/select.component.ts
```diff
@@ -90,6 +91,9 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
   @Input()
   public highlightSelected: boolean = true;
 
+  @Input()
+  public type?: SelectType;
```
anandtiwary, line 95: Instead of introducing a new input, can the component set the width of the popover content dynamically to match the width of its trigger element? This would be useful at few other location.

aaron-steinfeld: I agree about not introducing a new input, but I would actually think set the width based on the width of the options provided - why make it wider than the widest option (with some max-width) ?

DmitiryPotychkin: Now popover content of select has min-width equal to width of trigger element. Example how it look not in pagination component.
Previous:
not-improved-select

Current:
improved-select

If option in popover wider than trigger element:
long-option

projects/components/src/paginator/paginator.component.ts
DmitiryPotychkin: Change ht-button to ht-toggle-button-group

projects/components/src/button/button.component.scss
DmitiryPotychkin: After implementing the-toggle-button-group instead of ht-button I removed all changes in button component and defined styles for pagination buttons in parent component

projects/components/src/paginator/paginator.component.scss
```diff
@@ -16,27 +16,82 @@
...
-    min-width: 86px;
+    min-width: $paginator-select-width;
     margin-left: 12px;
   }
 
   .pagination-buttons {
     display: flex;
     margin: 0 24px 0 12px;
+    ::ng-deep {
+      .toggle-button-group {
```
anandtiwary, line 27: I guess all of this is required because the paginator toggle has a slightly different style that toggle button component. @jake-bassett What should we do for this inconsistency?

projects/components/src/select/select.component.ts
```diff
@@ -115,6 +122,10 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
 
   public ngAfterContentInit(): void {
     this.selected$ = this.buildObservableOfSelected();
+
+    setTimeout(() => {
+      this.selectContainerWidth = `${this.selectContainer.nativeElement.offsetWidth}px`;
```
anandtiwary, line 127: Can we use `this.selectContainer.nativeElement.offsetWidth` directly in the template? Advantage would be that if `selectContainer`'s width changes after initial render(let's say it's width is 20% and a window resize happens) then the popover width would automatically match it, everytime

anandtiwary: You can do this directly in the dom itself. I played around with it an added a similar change to multi-select component which is a separate but similar component. Please take a look at this change here https://github.com/hypertrace/hypertrace-ui/pull/524/files

DmitiryPotychkin: Thanks, it's look far simply and code become more readable, I refactor this as it used to be in your commit

projects/components/src/paginator/paginator.component.scss
aaron-steinfeld: this is way too much reskinning IMO - there's a ton of class dependencies here that could easily change. I get that we sometimes need ng-deep, but at this point, I think using toggle is more likely to introduce breakages in the future and would rather either make the paginator style consistent with toggle or not use toggle in this component and implement the buttons as part of this component.

projects/components/src/select/select.component.scss
```diff
@@ -84,8 +84,7 @@
 }
 
 .select-content {
-  @include dropdown();
-  min-width: 120px;
+  @include dropdown(4px, 0, 1px 0 4px 0);
```
aaron-steinfeld, line 87: Why does the select have different styling than other dropdowns? It doesn't seem like it should be different/parameterizable.

projects/assets-library/assets/styles/_layout.scss
```diff
@@ -1,6 +1,7 @@
 @import 'color-palette';
 
 $paginator-height: 50px;
+$paginator-select-width: 74px;
```
aaron-steinfeld, line 4: Is this used outside of paginator?

projects/components/src/select/select.component.ts
```diff
@@ -100,6 +100,8 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
 
   public groupPosition: SelectGroupPosition = SelectGroupPosition.Ungrouped;
 
+  public selectContainerWidth!: string;
```
aaron-steinfeld, line 103: no longer used

projects/components/src/paginator/paginator.component.scss
jake-bassett: +1 on not reusing a component if we need to modify it so heavily.

anandtiwary: aah. I kind of suggested to use toggle since the behavior was similar. I didn't know the styling of pagination buttons is so different. Should we instead look into combining the two styles form ux side?

jake-bassett: Yeah, I can see the draw of using the toggle, but the use cases are just a little different. Pagination is really two buttons visually tied together, but functionality independent. Toggles have one or more active states.

projects/components/src/select/select.component.ts
```diff
@@ -40,21 +40,16 @@ import { SelectSize } from './select-size';
...
             <ht-icon class="trigger-icon" icon="${IconType.ChevronDown}" size="${IconSize.ExtraSmall}"> </ht-icon>
           </div>
         </ht-popover-trigger>
         <ht-popover-content>
-          <div class="select-content">
-            <div
-              *ngFor="let item of items"
-              (click)="this.onSelectionChange(item)"
-              class="select-option"
-              [ngClass]="this.size"
```
aaron-steinfeld, line 56: this got dropped - intentional?

DmitiryPotychkin: Yes, it was intentional, fixed

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):

anandtiwary (approved):

projects/components/src/select/select.component.ts
```diff
@@ -45,32 +45,36 @@ import { SelectSize } from './select-size';
...
+              <ht-icon *ngIf="this.icon" class="trigger-prefix-icon" [icon]="this.icon" [size]="this.iconSize">
+              </ht-icon>
+              <ht-label class="trigger-label" [label]="selected?.label || this.placeholder"> </ht-label>
+              <ht-icon class="trigger-icon" icon="${IconType.ChevronDown}" size="${IconSize.ExtraSmall}"> </ht-icon>
+            </div>
+            <div
+              *ngSwitchCase="'${SelectTriggerDisplayMode.Icon}'"
+              class="trigger-content icon-only"
+              [ngClass]="this.selected !== undefined ? 'selected' : ''"
+              #triggerContainer
```
aaron-steinfeld, line 63: this is insane the triggerContainer on 48. Should it be removed?

aaron-steinfeld (approved):

projects/components/src/select/select.component.ts
DmitiryPotychkin: yes, forgot to remove. Fixed now

anandtiwary (approved):

PR #514, opened 2021-01-20
feat: adding all option and fixing some styling

anandtiwary:
## Description

feat: adding "all" option and fixing some styling

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/multi-select/multi-select.component.scss
+++ b/projects/components/src/multi-select/multi-select.component.scss
@@ -54,7 +54,6 @@
 .multi-select-content {
   @include dropdown();
   min-width: 120px;
-  max-height: 204px;
 
   .multi-select-option {
     display: flex;
@@ -65,15 +64,15 @@
     align-items: center;
 
     .checkbox {
-      margin-right: 8px;
+      margin: 0px;
     }
 
     .icon {
-      margin-right: 8px;
+      margin-left: 8px;
     }
 
     .label {
-      margin-right: 8px;
+      margin-left: 8px;
     }
 
     &:hover {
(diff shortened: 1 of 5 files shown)


projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -54,7 +54,6 @@
 .multi-select-content {
   @include dropdown();
   min-width: 120px;
-  max-height: 204px;
```
aaron-steinfeld, line 57: Shouldn't it have some max height? I'd rather it scroll then run off the screen.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -40,6 +40,13 @@ import { SelectSize } from '../select/select-size';
         </ht-popover-trigger>
         <ht-popover-content>
           <div class="multi-select-content">
+            <div class="multi-select-option all-options" (click)="this.onAllSelectionChange()">
```
aaron-steinfeld, line 43: Can we make the presence of the all option controlled by an input?

anandtiwary: sure

projects/components/src/multi-select/multi-select.component.scss
anandtiwary: How do we choose this value? Different use cases may show different set of options. Do you want to expose an input for this?

anandtiwary: Or may be we can set a max height which is equal to available height, that is, from trigger element's position to the bottom on the window.

anandtiwary: Let me create a separate story for this change.

aaron-steinfeld (approved):

PR #515, opened 2021-01-21
Ng splitter

anandtiwary:
## Description

Adding a splitter component. Instead of using ng-content, this component looks for sibling elements(prev and next) and adjust the width/height according to the movement of the mouse. It supports both horizontal and vertical split.

https://user-images.githubusercontent.com/52081890/105258403-8e64c200-5b3e-11eb-914f-ea8475fd340d.mov

Screen Shot 2021-01-20 at 10 17 59 PM

### Testing

### Checklist:

### Documentation


anandtiwary:
Worked on it during holidays. Will add tests post first round of feedback.

projects/components/src/splitter/splitter.service.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  private calculateUpdatedDimensionForNextElement(
+    previousElementUpdatedDimension: number,
+    parentBounds: DOMRect,
+    direction: SplitterDirection
+  ): number {
+    return (
+      (direction === SplitterDirection.Horizontal ? parentBounds.width : parentBounds.height) -
+      previousElementUpdatedDimension -
+      4
```
anandtiwary, line 88: This is the splitter dom's width/height based on direction. This should be an input param .

projects/components/src/public-api.ts
```diff
@@ -186,6 +186,11 @@ export * from './select/select-option.component';
 export * from './select/select.component';
 export * from './select/select.module';
 
+// Splitter
+export * from './splitter/splitter';
```
aaron-steinfeld, line 190: Do we have a use case for this?

projects/components/src/splitter/splitter.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  private setupMouseActionSubscription(): void {
+    this.subscriptionLifecycle.unsubscribe();
+
+    this.subscriptionLifecycle.add(
+      this.buildSplitterLayoutChangeObservable()
+        .pipe(
+          debounceTime(this.debounceTime),
+          tap(layoutChange => this.layoutChange.emit(layoutChange)),
+          tap(layoutChange => layoutChange && this.layoutChangeService.publishLayoutChange())
```
aaron-steinfeld, line 48: rather than using a tap operator, it looks like these actions on layout change belong in the subscriber (which is currently empty)

projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.component.scss
```diff
@@ -17,16 +17,16 @@
   width: 100%;
 
   .table {
-    min-width: 500px;
-    flex: 1 1;
+    width: 500px;
+    //flex: 1 1;
```
aaron-steinfeld, line 21: remove commented out

projects/components/src/splitter/splitter.service.ts
```diff
@@ -0,0 +1,104 @@
...
+      }),
+      map(data => this.maybeUpdateLayoutOfSiblings(data, direction))
+    );
+  }
+
+  private maybeUpdateLayoutOfSiblings(data: UpdatedDimensionData[], direction: SplitterDirection): boolean {
+    let layoutChanged = false;
+    data.forEach(datum => {
+      if (direction === SplitterDirection.Horizontal) {
+        this.updateWidth(datum.element, datum.dimension);
```
aaron-steinfeld, line 56: this seems inverted to me. I'd expect a horizontal splitter updates height, and a vertical width. In other words, the orientation describes the splitter itself, not how it affects the dimensions of its siblings.

projects/components/src/splitter/splitter.service.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        return updatedDimensionData;
+      }),
+      map(data => this.maybeUpdateLayoutOfSiblings(data, direction))
+    );
+  }
+
+  private maybeUpdateLayoutOfSiblings(data: UpdatedDimensionData[], direction: SplitterDirection): boolean {
+    let layoutChanged = false;
```
aaron-steinfeld, line 53: this doesn't really seem necessary, it's just being set if there's any data, so we could just return `data.length > 0`

projects/components/src/splitter/splitter.service.ts
```diff
@@ -0,0 +1,104 @@
...
+  ): number {
+    return (
+      (direction === SplitterDirection.Horizontal ? parentBounds.width : parentBounds.height) -
+      previousElementUpdatedDimension -
+      4
+    );
+  }
+
+  private updateWidth(element: HTMLElement, width: number): void {
+    element.style.width = `${width}px`;
```
aaron-steinfeld, line 93: in general (for all the dom manipulation/reading) can we use angular's renderer class?

projects/components/src/public-api.ts
aaron-steinfeld: sorry meant to delete this when I saw it later :)

projects/components/src/splitter/splitter.component.ts
anandtiwary: sure.

projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.component.scss
anandtiwary: yep. Will remove it with tests

projects/components/src/splitter/splitter.service.ts
anandtiwary: Oh yeah. I thought such a return would look weird as it didn't depend on any work that happened in the method. I can change it.

projects/components/src/splitter/splitter.service.ts
anandtiwary: Okay, so i added a `direction` property which represents the direction in which the splitter could move. It does not represent the orientation of splitter. If this is causing confusion then I could rename the property.

projects/components/src/splitter/splitter.service.ts
anandtiwary: I can check that. would it bring this change inside the scope of Change detection?

projects/components/src/public-api.ts
anandtiwary: No prob. Waterfall is a strong use case for this. Another use case is the upcoming geo map.  

It also think we can use it at few places where we use the two panel approach like explorer. Splitter would smooth out the interaction there.

projects/components/src/splitter/splitter.service.ts
aaron-steinfeld: I don't think it would affect change detection (since it's not modifying inputs), and it may not give us any direct benefits right here, but it's a good practice for some reasons that are relevant and some that aren't. In descending order of importance:
1. Angular will do any sanitization (e.g. xss) that we'd normally get in a template but not when we directly modify the DOM.
2. With all changes going through a common renderer, we could in the future choose to link that up to a service like layout change (because we can insert custom renderers).
3. If we ever choose to use something like web workers or server side rendering, direct DOM access would break (while this abstraction would work).

projects/components/src/splitter/splitter.service.ts
aaron-steinfeld: It's the same purpose though right? When describing a splitter as vertical, I'd expect it to render... vertically. Kind of example: https://material.angular.io/components/divider/overview#vertical-divider

jake-bassett:
This would have been a good opportunity to start bringing in [PrimeNG](https://www.primefaces.org/primeng/showcase/#/splitter).

projects/components/src/splitter/splitter.service.ts
anandtiwary: Yeah, my take was that the direction it can move in is more important. But i can see it is confusion. Will update.

anandtiwary:
> This would have been a good opportunity to start bringing in [PrimeNG](https://www.primefaces.org/primeng/showcase/#/splitter).

Do we really want to bring in a new third party lib for small components? We already have a vast collection of in house components, switching to a third party seems a little late now :P.

projects/components/src/splitter/splitter.service.ts
anandtiwary: okay, let me check

jake-bassett:
> > This would have been a good opportunity to start bringing in [PrimeNG](https://www.primefaces.org/primeng/showcase/#/splitter).
> 
> Do we really want to bring in a new third party lib for small components? We already have a vast collection of in house components, switching to a third party seems a little late now :P.

I think it still makes sense since we are still building and improving components. That component lib brings in a lot more that we still need to build out beyond just the splitter.

projects/components/src/splitter/splitter.service.ts
anandtiwary: I don't see any method to access previous sibling using angular renderer2. I am not sure this would work with the current approach.

aaron-steinfeld: Yeah, perhaps not. For what it's worth - the approach seems very un-angulary and has some other pitfalls due to that too. It's reaching into the dom outside it and manipulating it, which is broken encapsulation and also makes assumptions on how those things are laid out (for example, if one sibling was using flex, this would break, right?)

The API I'd expect for something like this would be:
```
<ht-splitter [direction]="direction">
   <div *htSplitterContent>Left</div> 
   <div *htSplitterContent>Right</div>
</ht-splitter
```
This would keep everything internal to the splitter.

anandtiwary: Nope. it woudn't break if it is inside flex container. We could always nest it in various components and directives, but that just add additional structures to DOM. It doesn't do anything else apart from that. Take a deeply nested splitter as an example.

aaron-steinfeld: It would nest just fine in this example. It does add a dom layer to ensure we're modifying something we own rather than breaking encapsulation, but that seems like a tradeoff I'd take any day.

anandtiwary:
Addressed previous comments

projects/components/src/public-api.ts
```diff
@@ -296,6 +296,16 @@ export * from './select/select-control-option.component';
...
+export * from './splitter/splitter-content.directive';
+export * from './splitter/splitter.component';
+export * from './splitter/splitter.module';
+
+// Multi-select
+export * from './multi-select/multi-select.component';
+export * from './multi-select/multi-select.module';
```
itssharmasandeep, line 307: why this is needed, I guess it's not.

projects/components/src/splitter/splitter.component.scss
```diff
@@ -0,0 +1,81 @@
+@import 'mixins';
+
+:host {
+  width: 100%;
+  height: 100%;
```
itssharmasandeep, line 5: shouldn't require host properties

projects/components/src/splitter/splitter.component.scss
```diff
@@ -0,0 +1,81 @@
...
+    flex-direction: column;
+  }
+
+  .splitter-content {
+    display: flex;
+  }
+
+  .splitter-content > p {
+    width: 100%;
+    height: 100%;
```
itssharmasandeep, line 27: nit: can be replaced with `fill-container` and all the other places as well

projects/components/src/splitter/splitter.component.scss
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  .splitter {
+    display: flex;
+    align-items: center;
+    justify-content: center;
```
itssharmasandeep, line 33: nit: can be replaced with `center-contents` mixin

projects/components/src/splitter/splitter.component.scss
```diff
@@ -0,0 +1,81 @@
...
+        height: 50px;
+      }
+
+      &:hover {
+        cursor: col-resize;
+      }
+    }
+
+    &.vertical {
+      width: calc(100% - 72px);
```
itssharmasandeep, line 63: shouldn't require calc IMO, should find a different way instead

anandtiwary:
> > > This would have been a good opportunity to start bringing in [PrimeNG](https://www.primefaces.org/primeng/showcase/#/splitter).
> > 
> > 
> > Do we really want to bring in a new third party lib for small components? We already have a vast collection of in house components, switching to a third party seems a little late now :P.
> 
> I think it still makes sense since we are still building and improving components. That component lib brings in a lot more that we still need to build out beyond just the splitter.

mine is better ;)

(NgPrime splitter does not support dynamic content )

projects/components/src/splitter/splitter.service.ts
anandtiwary: Done

projects/components/src/splitter/splitter.service.ts
anandtiwary: removed

projects/components/src/splitter/splitter.component.ts
```diff
@@ -0,0 +1,294 @@
...
+          [ngStyle]="this.splitterSizeStyle"
+          (mousedown)="this.onGutterMouseDown($event, index)"
+        >
+          <div class="cursor"></div>
+        </div>
+      </ng-container>
+    </div>
+  `
+})
+export class SplitterComponent implements OnChanges, AfterContentInit {
```
itssharmasandeep, line 51: nit: Please use `protected` instead of `public`

projects/components/src/splitter/splitter.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import { ButtonModule } from '../button/button.module';
+import { IconModule } from '../icon/icon.module';
+import { PopoverModule } from '../popover/popover.module';
+import { TooltipModule } from '../tooltip/tooltip.module';
+import { SplitterComponent } from './splitter.component';
+import { SplitterContentDirective } from './splitter-content.directive';
+import { MemoizeModule } from '@hypertrace/common';
+
+@NgModule({
+  imports: [CommonModule, IconModule, PopoverModule, ButtonModule, TooltipModule, MemoizeModule],
```
itssharmasandeep, line 12: do we need `PopoverModule`?

projects/components/src/splitter/splitter.component.ts
```diff
@@ -0,0 +1,294 @@
...
+  template: `
+    <div class="splitter-container" [ngClass]="[this.direction | lowercase]">
+      <ng-container *ngFor="let content of this.contents$ | async as contents; let index = index">
+        <div class="splitter-content" [ngStyle]="{ flex: this.getFlex | htMemoize: content.dimension }">
+          <ng-container *ngTemplateOutlet="content.templateRef"></ng-container>
+        </div>
+
+        <div
+          class="splitter"
+          *ngIf="
```
itssharmasandeep, line 37: shouldn't the ngIf on `this.direction` should be on the parent div?

projects/components/src/public-api.ts
anandtiwary: it should be exported already. Let me remove it.

projects/components/src/splitter/splitter.component.scss
anandtiwary: Yeah, i forgot to remove it.

projects/components/src/splitter/splitter.component.scss
anandtiwary: I prefer to type it out for better readability.

projects/components/src/splitter/splitter.component.ts
anandtiwary: Let me set a default value and move the ngIf

projects/components/src/splitter/splitter.component.scss
jake-bassett: calc() offsets are often a sign that a flexbox is using 100% height incorrectly.

projects/components/src/splitter/splitter.module.ts
anandtiwary: no.

projects/components/src/splitter/splitter.component.ts
jake-bassett: Hah, this is one that @anandtiwary got me on recently! ;-)

projects/components/src/splitter/splitter.component.scss
anandtiwary: done

projects/components/src/splitter/splitter.component.ts
anandtiwary: done

anandtiwary: I did do it for a few properties :)

projects/components/src/splitter/splitter.service.ts
anandtiwary: Directive and Angular Renderer

jake-bassett (approved):
A few nits remaining, but @anandtiwary said he will return to those in a follow up. lgtm otherwise.

anandtiwary:
Merging this PR after more than 2 years. :)

anandtiwary:
> A few nits remaining, but @anandtiwary said he will return to those in a follow up. lgtm otherwise.

I think I addressed all the review comments. There are a few minor resize bugs, that we will work on in a follow up.

PR #516, opened 2021-01-21
Top level api nav

jake-bassett:
## Description
* Renames mode toggle in Table controls to a more generic view toggle.
* Changes the table mode toggle widget to a more generic view toggle widget **breaks dependencies**
* Updates some of the URL paths, eliminating the top level "services" lists and using "apis" instead ("endpoints" was already used and didn't want to overload)


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-view.model.ts
```diff
@@ -0,0 +1,18 @@
...
+})
+export class TableWidgetViewModel {
+  @ModelProperty({
+    key: 'view',
+    type: STRING_PROPERTY.type
+  })
+  public view!: string;
+
+  @ModelProperty({
+    key: 'json',
```
aaron-steinfeld, line 14: can we call this `template` (for consistency with other similar properties)

src/app/entity-metadata.ts
```diff
@@ -7,17 +7,17 @@ export const entityMetadata: EntityMetadataMap = new Map([
       entityType: ObservabilityEntityType.Api,
       icon: ObservabilityIconType.Api,
       detailPath: (id: string, sourceRoute?: string) => [sourceRoute ?? '', 'endpoint', id],
-      sourceRoutes: ['services']
+      sourceRoutes: ['apis']
```
aaron-steinfeld, line 10: I saw your comment, but unless I'm missing another spot, the only place `endpoints` is used is nested like `service/endpoints`. Rather than being confusing, I'd actually expect them to be consistent since they're showing the same list of entities in different scopes.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -124,7 +123,7 @@ export class TableWidgetRendererComponent
   public ngOnInit(): void {
     super.ngOnInit();
 
-    this.onModeChange(this.model.mode);
+    // This.onModeChange(this.model.mode);
```
aaron-steinfeld, line 126: no longer needed? how does the initial view get set? if no longer needed, remove comment

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-select-filter.model.ts
```diff
@@ -1,14 +1,15 @@
...
 import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { GraphQlDataSourceModel } from '../../data/graphql/graphql-data-source.model';
 
 @Model({
   type: 'table-widget-select-filter',
   displayName: 'Select Filter'
 })
-export class TableWidgetSelectFilterModel {
+export class TableWidgetSelectFilterModel extends GraphQlDataSourceModel<PrimitiveValue[]> {
```
aaron-steinfeld, line 12: this isn't a data source itself - it's calling it's api to get data.

src/app/entity-metadata.ts
jake-bassett: That's fine. I can go either way. I'll change `apis` to `endpoints`.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: For this iteration, I don't have the different views in the URL, so not deep linkable. We just show the first view option. 

We need to decide how/if we want the different view toggles to be deep linkable (I think we do) and we need to decide if we want them as a route or a url param. I'm leaning towards param, but am open.

projects/observability/src/public-api.ts
```diff
@@ -82,7 +82,7 @@ export * from './pages/apis/api-detail/traces/api-trace-list.component';
 export * from './pages/apis/api-detail/traces/api-trace-list.dashboard';
 export * from './pages/api-trace-detail/api-trace-detail.page.component';
 
-export * from './pages/apis/services/service-list.module';
+export * from './pages/apis/endpoints-list.module';
```
aaron-steinfeld, line 85: I'm shocked you didn't put this with the component and dashboard in the same dir 😉

src/app/entity-metadata.ts
```diff
@@ -7,17 +7,17 @@ export const entityMetadata: EntityMetadataMap = new Map([
...
   ],
   [
     ObservabilityEntityType.Service,
     {
       entityType: ObservabilityEntityType.Service,
       icon: ObservabilityIconType.Service,
-      detailPath: (id: string) => ['apis', 'service', id],
-      listPath: ['apis'],
-      apisListPath: (id: string) => ['apis', 'service', id, 'endpoints'],
+      listPath: ['endpoints'],
```
aaron-steinfeld, line 18: What are the desired breadcrumbs? Urls? I think things are a little off here (in terms of the entity metadata, the urls and the breadcrumb resolvers). 

I think I'd expect:
Endpoint List => '/endpoints'
Endpoint => '/endpoint/<id>' (should it be a different url if reached from service endpoint list vs endpoint list? different breadcrumb? If yes, then also '/service/endpoint/<id>' and matching source path)
Service => '/service/<id>' ; no list path, updated detail + api list path
Service Endpoint List => '/service/<id>/endpoints

projects/observability/src/public-api.ts
jake-bassett: Hah, fail. It was actually because it started out as the services file and the IDE updated it when I renamed. I just moved it to its proper home though. Thanks!

src/app/entity-metadata.ts
jake-bassett: I _think_ I agree. Here is what we will have after the incoming changes.

Screen Shot 2021-01-21 at 1 19 28 PM
Screen Shot 2021-01-21 at 1 19 38 PM

Screen Shot 2021-01-21 at 1 19 49 PM

aaron-steinfeld: I think the one that is still a bit of a question is endpoint.

My feelings are (assuming we want to show the service breadcrumb):
- When navigated from a service: `/service/endpoint/<endpoint-id>`. This would correspond to breadcrumbs of: `<Service Name> > Endpoints > <Endpoint name>`
- When navigated from elsewhere, such as EP list: `/endpoint/<endpoint-id>` This would correspond to breadcrumbs of: `Endpoints > <Endpoint name>`

skjindal93: - Are we getting rid of the service listing screen? 
- How are we navigating to a service, when there is no service listing page? Via the service column in API endpoints list?

JBAhire: Why did we get rid of service screen? This was major thing in a customer perspective and they are using service dashboards for insights. 

Is there an issue or discussion around this as I came to know about this because of latest UI release I cut yesterday. 

I am happy with moving APIs in first class navigation but that shouldn't result in removal of service nav completely.

aaron-steinfeld: This conversation has happened in a few spots now, but duplicating here:
So in the weeds of the discussions around the endpoint list, it was overlooked that it should be a sibling of the service list, which will now be flat. We'll get that cleaned up today.

aaron-steinfeld: I put up https://github.com/hypertrace/hypertrace-ui/pull/521 to fix the most pressing issues. It has significant, but not complete overlap with this PR (Sorry!)

jake-bassett:
@aaron-steinfeld did part of this work in another PR (thank you!). Going to close out this one.

PR #520, opened 2021-01-22
feat: cartesian metric health

jake-bassett:
## Description
Cartesian Metric Health


projects/observability/src/shared/graphql/model/schema/metric/graphql-metric-aggregation.ts
```diff
@@ -11,6 +11,8 @@ export const enum GraphQlMetricAggregationPath {
 
 export interface GraphQlMetricAggregation {
   value: number;
+  upperBound?: number;
```
aaron-steinfeld, line 14: just wondering - would it be better to model this as a separate interface? i.e. 
```typescript
export interface GraphQlMetricBaselineAggregation {
  value: number;
  upperBound: number;
  lowerBound: number;
}
```

Basically, instead of making the baseline props optional, model the whole baseline separately because it should either all be present, or none of it. Same idea shows up in a few other places.

PR #530, opened 2021-01-25
fix: side navigation padding

DmitiryPotychkin:
- change padding
- change location of toggle button


DmitiryPotychkin:
Also I suggest to add scroll into this component to handle cases when height of user's screen is smaller than height of navigation, rough example how it may look: 
side-nav-overflow

aaron-steinfeld:
> Also I suggest to add scroll into this component to handle cases when height of user's screen is smaller than height of navigation

We definitely should be handling that case, but I thought the plan was that the whole page should be scrolling, as we don't want separate scrollbars on the side nav and the content. Taking a quick look at the UI today it looks like the content already scrolls separately and the nav just gets clipped - so it should be addressed in some manner. If we want to throw a scroll on it for now, either in this PR or another - that's fine until we come up with something more elegant.

aaron-steinfeld (approved):

PR #532, opened 2021-01-25
refactor: rearrange styles for sharing in app header

aaron-steinfeld:
## Description
This rearranged some styles used by components in the app header so they can come from a single source of truth, rather than being redeclared in each component. The select component was also updated to actually respond to its size argument, which was previously ignored.

### Testing
Visual verification on top of unit tests.

### Checklist:

### DIFF
--- a/projects/components/src/header/application/application-header.component.scss
+++ b/projects/components/src/header/application/application-header.component.scss
@@ -45,9 +45,14 @@
     padding-left: 46px;
   }
 
-  .right-side {
+  .right-side-content {
     display: flex;
     margin-left: auto;
+    align-items: center;
+  }
+
+  .left-side-content {
+    display: flex;
     align-items: center;
   }
 
--- a/projects/components/src/header/application/application-header.component.ts
+++ b/projects/components/src/header/application/application-header.component.ts
@@ -12,12 +12,16 @@ import { GLOBAL_HEADER_HEIGHT, NavigationService } from '@hypertrace/common';
           <ng-content select="[logo]"></ng-content>
         </div>
         <div class="divider"></div>
+        <div class="left-side-content">
+          <ng-content select="[left]"></ng-content>
+        </div>
         <div class="time-range">
           <ht-time-range></ht-time-range>
         </div>
       </div>
-      <div class="right-side">
+      <div class="right-side-content">
         <ng-content></ng-content>
+        <ng-content select="[right]"></ng-content>
       </div>
     </div>
   `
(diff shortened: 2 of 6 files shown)


projects/components/src/header/application/application-header.component.scss
```diff
@@ -45,12 +45,17 @@
...
+  .right-side-content {
     display: flex;
     margin-left: auto;
     align-items: center;
   }
 
+  .left-side-content {
+    display: flex;
+    align-items: center;
+  }
```
anandtiwary, line 57: should we use justify content: flex-start here?

projects/components/src/header/application/application-header.component.ts
```diff
@@ -12,12 +12,16 @@ import { GLOBAL_HEADER_HEIGHT, NavigationService } from '@hypertrace/common';
...
+          <ng-content select="[left]"></ng-content>
+        </div>
         <div class="time-range">
           <ht-time-range></ht-time-range>
         </div>
       </div>
-      <div class="right-side">
+      <div class="right-side-content">
         <ng-content></ng-content>
+        <ng-content select="[right]"></ng-content>
```
anandtiwary, line 24: How is the `select="[right]"` property used here

aaron-steinfeld: This was a relabel of the existing one. Previously, header controls added via content always went to the right. Since I added a way to add them to the left too, I disambiguated via the `left` attribute, I relabeled the previously default selector to use `right` for clarity.

projects/components/src/header/application/application-header.component.scss
aaron-steinfeld: that's the default, it's already behaving that way.

projects/components/src/header/application/application-header.component.ts
anandtiwary: Still didn't get how two ng-content with one containing a select on `[right]` is working. What does this selector '[]'  means? I guess one of its child should have this selector defined somewhere.

anandtiwary: Fine to merge. I will play around with to understand what's happening here.

anandtiwary (approved):

projects/components/src/header/application/application-header.component.ts
aaron-steinfeld: Discussed offline (it's an attribute selector for header content, not yet used in this PR)

PR #534, opened 2021-01-26
Settings nav

DmitiryPotychkin:
Add parameter _skipLocationChange_ for not add certain location to location history stack. 

1. ht-navigation-list receive list of links, link may have parameter _skipLocationChange_
2. Value of this parameter with other arguments passed to navigate service method **navigateWithinApp** and after it to **navigate** method
3. if value of _skipLocationChange_ is true than service **Router** will receive _skipLocationChange: true_ in extras, and this location will not be added to location history stack after routing

### DIFF
--- a/projects/common/src/navigation/navigation.service.ts
+++ b/projects/common/src/navigation/navigation.service.ts
@@ -93,7 +93,8 @@ export class NavigationService {
         queryParams: params?.queryParams ?? this.buildQueryParam(),
         queryParamsHandling: params?.queryParamsHandling,
         replaceUrl: params?.replaceCurrentHistory,
-        relativeTo: params?.relativeTo
+        relativeTo: params?.relativeTo,
+        skipLocationChange: params?.shadow
       }
     };
   }
@@ -105,13 +106,15 @@ export class NavigationService {
   public navigateWithinApp(
     path: NavigationPath,
     relativeTo?: ActivatedRoute,
-    preserveParameters?: string[]
+    preserveParameters?: string[],
+    shadow?: boolean
   ): Observable<boolean> {
     return this.navigate({
       navType: NavigationParamsType.InApp,
       path: path,
       queryParams: this.buildQueryParam(preserveParameters ?? []),
-      relativeTo: relativeTo
+      relativeTo: relativeTo,
+      shadow: shadow
     });
   }
 
@@ -304,6 +307,7 @@ export interface InAppNavigationParams {
   queryParams?: QueryParamObject;
   queryParamsHandling?: 'merge' | 'preserve';
   relativeTo?: ActivatedRoute;
+  shadow?: boolean;
 }
 
 export interface ExternalNavigationParamsNew {
--- a/projects/common/src/navigation/navigation.service.ts
+++ b/projects/common/src/navigation/navigation.service.ts
@@ -93,8 +93,7 @@ export class NavigationService {
         queryParams: params?.queryParams ?? this.buildQueryParam(),
         queryParamsHandling: params?.queryParamsHandling,
         replaceUrl: params?.replaceCurrentHistory,
-        relativeTo: params?.relativeTo,
-        skipLocationChange: params?.skipLocationChange
+        relativeTo: params?.relativeTo
       }
     };
   }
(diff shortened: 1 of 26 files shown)


anandtiwary (dismissed):

projects/common/src/navigation/navigation.service.ts
```diff
@@ -105,13 +106,15 @@ export class NavigationService {
   public navigateWithinApp(
     path: NavigationPath,
     relativeTo?: ActivatedRoute,
-    preserveParameters?: string[]
+    preserveParameters?: string[],
+    skipLocationChange?: boolean
```
aaron-steinfeld, line 110: this is a shortcut method, instead of modifying this with more optional parameters, can you use `navigate` instead?

projects/common/src/navigation/navigation.service.ts
```diff
@@ -304,6 +307,7 @@ export interface InAppNavigationParams {
   queryParams?: QueryParamObject;
   queryParamsHandling?: 'merge' | 'preserve';
   relativeTo?: ActivatedRoute;
+  skipLocationChange?: boolean;
```
aaron-steinfeld, line 310: this is already captured on line 306 (`replaceCurrentHistory`)

aaron-steinfeld (changes requested):
Commented above

projects/common/src/navigation/navigation.service.ts
DmitiryPotychkin: What will be better - create new method similar to navigateWithinApp in navigation service (maybe it'll be helpful in future or it's a rare case?) or just directly use navigate method inside navigation-list component without any shortcut?

projects/common/src/navigation/navigation.service.ts
DmitiryPotychkin: Got it, thanks, will use it instead adding new

projects/common/src/navigation/navigation.service.ts
aaron-steinfeld: I would leave navigationService untouched, and call via:
```typescript
  this.navigationService.navigate({
      path: item.matchPaths[0],
      replaceCurrentHistory: item.skipLocationChange,
      relativeTo: this.activatedRoute
  });
```

Replacing the history, when changing something other than a query param (which is already supported) should be a rare case. I'm not convinced we need it here, but I definitely don't want to encourage it for future use cases.

aaron-steinfeld: (you beat me to this with the exact change 😄 )

aaron-steinfeld (dismissed):

projects/components/src/navigation/navigation-list.component.ts
```diff
@@ -83,7 +83,12 @@ export class NavigationListComponent {
   }
 
   public navigate(item: NavItemLinkConfig): void {
-    this.navigationService.navigateWithinApp(item.matchPaths[0], this.activatedRoute);
+    this.navigationService.navigate({
+      navType: NavigationParamsType.InApp,
+      path: item.matchPaths[0],
+      relativeTo: this.activatedRoute,
+      replaceCurrentHistory: item.skipLocationChange
```
aaron-steinfeld, line 90: So missed this the first time around, but we should clarify our naming (the test actually caught this). There's two separate booleans that the angular supports:
- `skipLocationChange`: When true, navigates without pushing a new state into history.
- `replaceUrl`: When true, navigates while replacing the current state in history.

To illustrate the distinction, imagine three nav items that are navigated in order (`first -> second -> third`)
```
first: { ...,  skipLocationChange: false},
second: { ...,  skipLocationChange: true},
third: { ...,  skipLocationChange: false}
```

with the behavior of `skipLocationChange`, when complete our history stack looks like:
```
<third> // Active
<first> // browser back
```

with the behavior of `replaceUrl`, when complete our history stack looks like:
```
<third> // Active
<second> // browser back
```



A subtle distinction, but our implementation uses `replaceUrl`, so we should name the item's property accordingly.

aaron-steinfeld: @DmitiryPotychkin can we update the naming here to reflect the behavior? In other words, `skipLocationChange` on line 124 should be named `replaceCurrentHistory`.

aaron-steinfeld (approved):

projects/components/src/navigation/navigation-list.component.ts
DmitiryPotychkin: @aaron-steinfeld thank you for the explanation!

PR #535, opened 2021-01-26
feat: cartesian metric health

jake-bassett:
## Description
Cartesian Metric Health


projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget.model.ts
```diff
@@ -24,7 +26,7 @@ import { MetricSeries, MetricSeriesDataFetcher, SeriesModel } from './series.mod
   type: 'cartesian-widget',
   displayName: 'Cartesian Widget'
 })
-export class CartesianWidgetModel<TData> {
+export class CartesianWidgetModel {
```
jake-bassett, line 29: Note the generic removed here. This is needed due to the datasource only providing an array of intervals. This means the baseline and series data gets combined in the data source model, then pulled apart again here in the CartesianWidgetModel so it can be separated into different series and charted correctly.

Not really in love with what I have here, so open to other suggestions.

projects/observability/src/shared/graphql/model/schema/metric/graphql-metric-aggregation.ts
```diff
@@ -11,6 +11,8 @@ export const enum GraphQlMetricAggregationPath {
 
 export interface GraphQlMetricAggregation {
   value: number;
+  upperBound?: number;
```
jake-bassett, line 14: Comment from @aaron-steinfeld on Draft PR on this about maybe using separate interfaces. Going to respond here.

In `observability-specificiation-builder.ts` in `extractFromServerData` we only return `MetricTimeseriesInterval[]`, so if we fetch `baselineSeries` and `series` together like we do we need to combine all of the values for each interval. Unless we want to change the signature of that method... which could be kind of big.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget.model.ts
aaron-steinfeld: It feels to me like some of these hoops are based on the new json model that we landed on here. Do you mind sharing an example of a full widget with the new model for comparison?

I was thinking it would end up something like below. I'm not positive this is any better since I haven't been working in the code enough to see any pitfalls/how much work this change would be/how much it would clean up, but since you were open to suggestions I figured I'd throw it out there. For future changes like this, I often find that starting with the desired json model is a good starting point - similar to proposing an api before implementing it.

```typescript
            {
                type: 'cartesian-widget',
                ...
                series: [
                  {  // Regular data
                    type: 'series',
                    ...
                    data: {
                      type: 'entity-metric-timeseries-data-source',
                      metric: {
                        type: 'metric-timeseries',
                        metric: 'duration',
                        aggregation: MetricAggregationType.P99
                      }
                    }
                  },
                  { // Baseline data
                    type: 'series',
                    'visualization-type': 'dashed',
                    ...
                    data: {
                      type: 'entity-metric-timeseries-data-source',
                      metric: {
                        type: 'metric-baseline-timeseries', // note the different spec
                        metric: 'duration',
                        aggregation: MetricAggregationType.P99
                      }
                    }
                  }
              ],
              bands: [
                { // Band data
                    type: 'band',
                    ...
                    data: {
                      type: 'entity-metric-band-timeseries-data-source',
                      metric: {
                        type: 'metric-baseline-timeseries',
                        metric: 'duration',
                        aggregation: MetricAggregationType.P99
                      }
                    }
                  }
```

jake-bassett:
```typescript
export const serviceMetricsExtendedDashboard: DashboardDefaultConfiguration = {
  ...serviceMetricsDashboard,
  json: {
    type: 'container-widget',
    layout: {
      type: 'auto-container-layout',
      rows: 4,
      'enable-style': false
    },
    children: [
      {
        type: 'cartesian-widget',
        'selectable-interval': true,
        'legend-position': LegendPosition.None,
        series: [
          {
            type: 'series',
            name: 'Average Duration',
            color: '#0076C7',
            'visualization-type': SeriesVisualizationType.Line,
            band: {
              type: 'band',
              name: ''
            },
            data: {
              type: 'entity-metric-timeseries-data-source',
              metric: {
                type: 'metric-timeseries',
                metric: 'duration',
                aggregation: MetricAggregationType.Average,
                showBands: true
              }
            }
          }
        ]
      }
    ]
  }
};
```

jake-bassett: Yeah, I started with what you have there and changed it to deal with that bug that I didn't know the timeline on. Now that the bug is fixed, I'm not too happy with how this workaround way turned out.

Let me see how much work it is to do this the right way.

projects/observability/src/shared/graphql/request/builders/selections/observability-specification-builder.ts
```diff
@@ -213,14 +216,66 @@ export class ObservabilitySpecificationBuilder extends SpecificationBuilder {
...
-    requireUnique: boolean = true
+    requireUnique: boolean = true,
+    fetchBounds: boolean = false
   ): GraphQlSelection {
     return {
       path: convertToGraphQlMetricAggregationPath(aggregation),
       alias: requireUnique ? aggregation : undefined,
-      arguments: this.argBuilder.forAggregationArgs(aggregation)
+      arguments: this.argBuilder.forAggregationArgs(aggregation),
+      children: fetchBounds ? [{ path: 'upperBound' }, { path: 'lowerBound' }] : []
```
aaron-steinfeld, line 278: I don't remember why this doesn't include the `value` path for children. I believe it's only used in this file like:
```typescript
              {
                ...aggregationSelection,
                children: [{ path: 'value' }]
              }
```

Can we either 
- Remove the fetchBounds arg, and handle the bounds similar to value (which, I believe is what we're already doing) 
- Clean up this bit of tech debt by: making this method private, add the `value` child selection in here in either case and updating the callers.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget.model.ts
```diff
@@ -127,58 +139,174 @@ export class CartesianWidgetModel<TData> {
   @ModelInject(ColorService)
   private readonly colorService!: ColorService;
 
-  public getSeriesFetcher(): Observable<MetricSeriesFetcher<TData>> {
+  public getSeriesFetcher(): Observable<MetricSeriesFetcher<TSeriesInterval>> {
```
aaron-steinfeld, line 142: Rather than having two separate fetchers, it may make more sense to have a single fetcher that returns both series and band data (so we don't have to parallel paths all the way up the stack). I would imagine something like

```typescript
export interface CartesianDataFetcher<TSeriesInterval> {
  getData(interval: TimeDuration): Observable<CombinedResults<TSeriesInterval>>;
  getRequestedInterval?(): TimeDuration | undefined;
}
```
This seems like it would greatly simplify the renderer (which basically has no changes if it's still working with a single fetcher), and the model would simplify too since it doesn't need to support the full fetcher building and interval resolution for both, but not sure if there's more issues here that's not coming through on read.

projects/observability/src/shared/graphql/model/metric/metric-timeseries.ts
```diff
@@ -1,4 +1,9 @@
 export interface MetricTimeseriesInterval {
-  value: number;
   timestamp: Date;
+  value: number;
+}
+
+export interface MetricTimeseriesBandInterval extends MetricTimeseriesInterval {
```
aaron-steinfeld, line 6: Should the band include the baseline? I was thinking the baseline is treated as a regular series (with a different spec) and the band is the one with its own unique config.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -46,22 +57,28 @@ export class CartesianWidgetRendererComponent<TData> extends InteractiveDataWidg
...
+      tap(fetchers => {
+        this.seriesFetcher = fetchers.seriesFetcher;
+        this.bandFetcher = fetchers.bandFetcher;
+
+        if (this.intervalSupported(this.seriesFetcher)) {
           this.intervalOptions = this.buildIntervalOptions();
           this.selectedInterval = this.getBestIntervalMatch(
             this.intervalOptions,
-            this.selectedInterval || fetcher.getRequestedInterval()
+            this.selectedInterval || this.seriesFetcher.getRequestedInterval()
```
aaron-steinfeld, line 81: if we're not asking the bands for intervals, they shouldn't be supported in their models.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -72,15 +74,29 @@ export class CartesianWidgetRendererComponent<TData> extends InteractiveDataWidg
     );
   }
 
-  protected buildDataObservable(): Observable<Series<TData>[]> {
-    return this.fetcher ? this.buildSeries(this.fetcher, this.selectedInterval) : NEVER;
+  protected buildDataObservable(): Observable<CartesianData<TSeriesInterval>> {
+    if (!this.fetcher) {
+      return of({ series: [], bands: [] });
```
aaron-steinfeld, line 79: if fetcher is not yet defined we should return never rather than empty (we want to show loading while waiting for the fetcher, not empty data state)

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/band.model.ts
```diff
@@ -0,0 +1,65 @@
...
+  public static readonly DEFAULT_OPACITY: number = 0.4;
+
+  @ModelProperty({
+    key: 'band-color',
+    type: STRING_PROPERTY.type
+  })
+  public bandColor: string = BandModel.BAND_COLOR;
+
+  @ModelProperty({
+    key: 'baseline-color',
```
aaron-steinfeld, line 31: So cycling back to my [comment](https://github.com/hypertrace/hypertrace-ui/pull/535#discussion_r565387690) about the expected json, I think the last strange thing is that the band is implicitly providing the baseline. I'm ok if you want to come back to that.

aaron-steinfeld (approved):

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/band.model.ts
jake-bassett: Yeah, while I agree that the current way is a little strange I don't really like explicitly calling out the baseline as another series either since its attached to the data call of the band. Maybe we should have just named `bands` something like `baselines` instead.

Let's revisit this one.

aaron-steinfeld: > another series either since its attached to the data call of the band.

This is the part I was referring to about separating in one of my other comments. They access different data, they should be separate specs (that get merged for perf).

jake-bassett: I must be missing something here.

The `baseline` and the `lower` and `upper` bounds come from the same single `baselineSeries` call. This is why I say they are associated. But I do still agree from a chart perspective there is more affinity between the baseline and the actual metric since they are charted as `Series`, while the lower and upper bounds are charted as `Band` so agree the json model is confusing in that respect.

aaron-steinfeld: So imagine a request for two different timeseries with different aggregations of the same metric. Say, p99 and p50 like this example from the service overview:
```typescript
 {
                    type: 'series',
                    name: 'p99',
                    'visualization-type': 'line',
                    color: Color.BlueGray4,
                    data: {
                      type: 'entity-metric-timeseries-data-source',
                      metric: {
                        type: 'metric-timeseries',
                        metric: 'duration',
                        aggregation: MetricAggregationType.P99
                      }
                    }
                  },
                  {
                    type: 'series',
                    name: 'p50',
                    'visualization-type': 'line',
                    color: Color.Blue3,
                    data: {
                      type: 'entity-metric-timeseries-data-source',
                      metric: {
                        type: 'metric-timeseries',
                        metric: 'duration',
                        aggregation: MetricAggregationType.P50
                      }
                    }
                  },
```

These are different series, specified differently, even though they're both asking for data thats contained within the series of a particular entity + metric. Merging them is the responsibility of the request merger (which could use some love). These `value` portion of the baselineSeries and the upper/lower bounds should work the same way in my mind. Completely separate requests as far as the model and specs are concerned.

PR #537, opened 2021-01-27
fix: add large size for Select component

arjunlalb:
## Description
Add large size support for select component

### Testing
N/A

### Checklist:


aaron-steinfeld (changes requested):
the height of the select component is specified by the size input, it should not be inherited

aaron-steinfeld (approved):

PR #541, opened 2021-01-28
fix: table scrolling issues

jake-bassett:
## Description
Fixes for some table scrolling issues

### DIFF
--- a/projects/components/src/table/table.component.scss
+++ b/projects/components/src/table/table.component.scss
@@ -3,6 +3,11 @@
 
 .table {
   height: 100%;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  flex: 1 1;
+  overflow: auto;
 }
 
 .title-row {
@@ -11,12 +16,6 @@
   height: 36px;
   min-height: 36px;
   @include overline($gray-4);
-}
-
-.table {
-  width: 100%;
-  flex: 1 1;
-  overflow: auto;
 }
 
 .bottom-margin {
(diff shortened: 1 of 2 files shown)


projects/components/src/table/table.component.scss
```diff
@@ -39,7 +38,7 @@
 .data-row {
   display: flex;
   flex-direction: row;
-  height: 44px;
+  min-height: 44px;
```
anandtiwary, line 41: rows can possibly expand with this change

jake-bassett: Fixed.

projects/components/src/table/table.component.scss
```diff
@@ -106,7 +107,7 @@
 }
 
 .state-watcher {
-  height: 100%;
+  height: calc(100% - #{$paginator-height});
```
anandtiwary, line 110: curious. what does #{} do?

projects/components/src/table/table.component.scss
anandtiwary: hmm.. why do we need both height and min-height?

projects/components/src/table/table.component.scss
jake-bassett: [Interpolation](https://sass-lang.com/documentation/interpolation). Its so we can use the SCSS variable in the CSS function. It works like html templates in typescript.

projects/components/src/table/table.component.scss
jake-bassett: The new flex layout was trying to shrink the row height, so we needed the min-height so flex doesn't override the height. The height was just to address your valid comment about the row expanding.

aaron-steinfeld (commented):
lgtm (except the scss, didn't review that)

anandtiwary (approved):

PR #543, opened 2021-01-29
Add ht-description component

DmitiryPotychkin:
Add description component, which truncate long text and display buttons "show more" or "show less"


projects/components/src/description/description.component.scss
```diff
@@ -0,0 +1,23 @@
+:host {
```
aaron-steinfeld, line 1: Is this doable without `:host`? we use it in some places, but it's a bad practice and we should try to get rid of it where possible (since the parent is generally responsible for styling the host, and will have greater precedence here)

DmitiryPotychkin: I'm try to do it without :host but for some reason style `min-width: 0 ` for `.description` not affected parent component. Originally this min-width: 0 added for fix such case as on screenshots.

Without min-width: 0 for host
Screenshot from 2021-01-29 03-04-54
Add min-width: 0
Screenshot from 2021-01-29 03-04-43

projects/components/src/description/description.component.ts
```diff
@@ -0,0 +1,39 @@
...
+        #eventDescription
+      >
+        {{ description }}
+        <span *ngIf="this.isDescriptionTextToggled" (click)="this.toggleDescriptionText()" class="description-button"
+          >show less</span
+        >
+      </div>
+      <div
+        class="description-button"
+        *ngIf="eventDescription.offsetWidth < eventDescription.scrollWidth && !this.isDescriptionTextToggled"
```
DmitiryPotychkin, line 22: Also I found this ngIf not recalculate on  window resize, I tried to use ```@ViewChild``` decorator and several other ways but ref always return undefined, so it's another one issue I working on

aaron-steinfeld: One option here is to use out layout change directive, which should give you a callback when the layout changes - either from a window resize, or a shift in other components. An example of this can be seen in `LabelComponent`

projects/components/src/description/description.component.scss
aaron-steinfeld: There's usually alternatives, but always the easiest to find. Don't spend too much time on it - if we need to use :host so be it.

projects/components/src/description/description.component.ts
```diff
@@ -1,39 +1,78 @@
...
   public description!: string;
 
+  public ngOnChanges(): void {
+    if (this.isInitialized) {
+      this.remeasure();
+    }
+  }
+
+  ngAfterViewInit() {
+    this.fullDescriptionTextWidth = this.eventDescriptionText.nativeElement.scrollWidth;
```
aaron-steinfeld, line 64: This should be reassigned in ngOnChanges (or easier, just move it into remeasure and don't persist it) as the description text is based on an input and can change.

projects/components/src/description/description.component.ts
```diff
@@ -1,39 +1,78 @@
...
   @Input()
   public description!: string;
 
+  public ngOnChanges(): void {
+    if (this.isInitialized) {
+      this.remeasure();
+    }
+  }
+
+  ngAfterViewInit() {
```
aaron-steinfeld, line 63: Missing visibility/type declarations

projects/components/src/description/description.component.ts
```diff
@@ -1,39 +1,78 @@
...
         show more
       </div>
     </div>
   `
 })
-export class DescriptionComponent {
+export class DescriptionComponent implements OnChanges, AfterViewInit {
+  public isInitialized: boolean = false;
   public isDescriptionTextToggled: boolean = false;
+  public readonly isDescriptionTruncated$: Subject<boolean> = new Subject();
```
aaron-steinfeld, line 45: Is there an advantage of using an async boolean here (vs a simple boolean that is reassigned)?

DmitiryPotychkin: The original problem was *ngIf don't re-rendered after the change condition, so I decide to use this method. But now I'm use `ChangeDetectorRef` for this purpose

projects/components/src/description/description.component.ts
DmitiryPotychkin: Got it, so now method look this:
```
// Called in ngOnChanges
  public remeasure(): void {  
    // This will add class for truncate text for get full width of text (without text wrapping)
    this.isDescriptionTextToggled = false;   
    this.cdRef.detectChanges();

   // Get width of parent element and text full width without text wrapping, compare it
    this.isDescriptionTruncated = 
      this.eventDescriptionContainer.nativeElement.offsetWidth < this.eventDescriptionText.nativeElement.scrollWidth;
    this.cdRef.detectChanges();
  }
```

projects/components/src/description/description.component.ts
```diff
@@ -71,8 +71,11 @@ export class DescriptionComponent implements OnChanges, AfterViewInit {
...
   public remeasure(): void {
-    this.isDescriptionTruncated$.next(
-      this.eventDescriptionContainer.nativeElement.offsetWidth < this.fullDescriptionTextWidth
-    );
+    this.isDescriptionTextToggled = false;
+    this.cdRef.detectChanges();
+
+    this.isDescriptionTruncated =
+      this.eventDescriptionContainer.nativeElement.offsetWidth < this.eventDescriptionText.nativeElement.scrollWidth;
+    this.cdRef.detectChanges();
```
aaron-steinfeld, line 79: Only run once, at the end of the remeasure method.

DmitiryPotychkin: It will cause bug, value of `isDescriptionTruncated` will be wrong as we try to get `this.eventDescriptionText.nativeElement.scrollWidth` before it truncated and DOM changed.

aaron-steinfeld: Hmm - that's one of the weird parts about using `detectChanges` explicitly instead of `markForCheck` (which just invalidates this node in the change detection tree to be picked up by angular's general change detection, which should ensure that the changes settle. Changing that here may be better - could you try that out, and if it's not working we'll go with this?

DmitiryPotychkin: I try to use `markForCheck` method but unfortunately it not work correctly

aaron-steinfeld (approved):

anandtiwary (approved):

anandtiwary:
@DmitiryPotychkin we should now use this component at the two locations mentioned in the other PR.

PR #545, opened 2021-01-29
feat: support to parse ISO 8601 duration string

arjunlalb:
## Description
Support for parsing ISO 8601 duration string

### Testing
UT added.

### Checklist:

### DIFF
--- a/projects/common/src/time/time-duration.ts
+++ b/projects/common/src/time/time-duration.ts
@@ -1,8 +1,10 @@
+import { parse, toSeconds } from 'iso8601-duration';
 import { assertUnreachable } from '../utilities/lang/lang-utils';
 import { TimeUnit } from './time-unit.type';
 
 export class TimeDuration {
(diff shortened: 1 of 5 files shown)


projects/common/src/time/time-duration.ts
```diff
@@ -12,6 +14,10 @@ export class TimeDuration {
     return this.millis;
   }
 
+  public static parse(durationString: string): TimeDuration {
+    return new TimeDuration(toSeconds(parse(durationString)), TimeUnit.Second);
+  }
+
```
anandtiwary, line 20: can we also update `toIso8601DurationString` to now use library methods?

arjunlalb: @anandtiwary the library does not have a method to convert to a duration string :)

aaron-steinfeld (dismissed):

anandtiwary (approved):

PR #546, opened 2021-01-29
feat: add global filtering that applies aross all queries

aaron-steinfeld:
## Description
This change adds support for registering global filter rules which will be applied to all graphql handlers. This allows rules to be managed independently of the various handlers, a more extensible design.

### Testing
Added and updated unit tests

### Checklist:

### DIFF
--- a/projects/distributed-tracing/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
+++ b/projects/distributed-tracing/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
@@ -1,0 +1,27 @@
+import { Injectable } from '@angular/core';
+import { GraphQlFilter } from './graphql-filter';
+
+@Injectable({ providedIn: 'root' })
+export class GlobalGraphQlFilterService {
+  private readonly filterRules: Map<symbol, FilterRule> = new Map();
+
+  public getGlobalFilters(scope: string): GraphQlFilter[] {
+    return Array.from(this.filterRules.values()).flatMap(rule => rule.filtersForScope(scope));
+  }
+
+  public mergeGlobalFilters(scope: string, localFilters: GraphQlFilter[] = []): GraphQlFilter[] {
+    return [...localFilters, ...this.getGlobalFilters(scope)];
+  }
+
+  public setGlobalFilterRule(ruleKey: symbol, rule: FilterRule): void {
+    this.filterRules.set(ruleKey, rule);
+  }
+
+  public clearGlobalFilterRule(ruleKey: symbol): void {
+    this.filterRules.delete(ruleKey);
+  }
+}
+
+interface FilterRule {
+  filtersForScope(scope: string): GraphQlFilter[];
+}
--- a/projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
+++ b/projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
@@ -4,6 +4,7 @@ import { GraphQlHandlerType, GraphQlQueryHandler, GraphQlSelection } from '@hype
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { MetadataService } from '../../../../services/metadata/metadata.service';
+import { GlobalGraphQlFilterService } from '../../../model/schema/filter/global-graphql-filter.service';
 import { GraphQlFilter } from '../../../model/schema/filter/graphql-filter';
 import { GraphQlSortBySpecification } from '../../../model/schema/sort/graphql-sort-by-specification';
 import { Specification } from '../../../model/schema/specifier/specification';
@@ -18,7 +19,10 @@ export class TracesGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
   private readonly selectionBuilder: GraphQlSelectionBuilder = new GraphQlSelectionBuilder();
   public readonly type: GraphQlHandlerType.Query = GraphQlHandlerType.Query;
 
-  public constructor(private readonly metadataService: MetadataService) {}
+  public constructor(
+    private readonly metadataService: MetadataService,
+    private readonly globalGraphQlFilterService: GlobalGraphQlFilterService
+  ) {}
 
   public matchesRequest(request: unknown): request is GraphQlTracesRequest {
     return (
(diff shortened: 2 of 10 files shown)


projects/distributed-tracing/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class GlobalGraphQlFilterService {
+  private readonly filterRules: Map<symbol, FilterRule> = new Map();
+
+  public getGlobalFilters(scope: string): GraphQlFilter[] {
+    return Array.from(this.filterRules.values()).flatMap(rule => rule.filtersForScope(scope));
+  }
+
+  public mergeGlobalFilters(scope: string, localFilters: GraphQlFilter[] = []): GraphQlFilter[] {
```
anandtiwary, line 12: Can this be used with table - filter bar interaction?

projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
```diff
@@ -37,7 +41,9 @@ export class TracesGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
         this.argBuilder.forTimeRange(request.timeRange),
         ...this.argBuilder.forOffset(request.offset),
         ...this.argBuilder.forOrderBy(request.sort),
-        ...this.argBuilder.forFilters(request.filters)
+        ...this.argBuilder.forFilters(
+          this.globalGraphQlFilterService.mergeGlobalFilters(resolveTraceType(request.traceType), request.filters)
```
anandtiwary, line 45: Putting this everywhere doesn't seem too scalable and may be error prone. Can we add it in idk `GraphQlQueryEventService` may be so that the inherited filters already have the desired filter?

projects/distributed-tracing/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
aaron-steinfeld: Not sure what you mean - will it work with with the filter bar (yes), or could we use the filter bar to do this instead of using the handlers (no, because this applies in places without a filter bar).

projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
aaron-steinfeld: I don't love it, but it seems like the best place we have right now. There's no place with central knowledge of all queries, by design - and filters aren't a concept that applies to every request in the same way. QueryEventService for example pushes filters down to the data sources to apply, which is even more decentralized (and only works in the dashboard code) - the service itself doesn't know the request shapes to provide them.

Open to other ideas, but I think the handlers are the narrowest (and most stable point) that still has the context on how to apply the filters without a larger redesign.

projects/distributed-tracing/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
anandtiwary: So we have `FilterUrlService` that keeps a filter bar and url in sync. I think table depends on it either directly or indirectly (via urls) to construct and apply filters in table requests. With the new service, I was thinking what if we combine the two so that only one service is responsible for applying filters in graphql.

projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
anandtiwary: Then can we do it using an abstract method which the concrete handler has to provide? I could see people easily missing to apply the right global filters in new handlers.

projects/distributed-tracing/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
aaron-steinfeld: So I think there's potential for a wider scale refactor, but there's nothing today that would solve the centralization issue. Filter bar and filter url service basically work, as you said, to keep that one component in sync with the url. The trickier part is to how to get those filters into queries. Right now, the filter bar is wired into certain dashboards by applying those filters (from the url or user interaction) directly to the dashboard's root data source. This approach really only works on homogenous screens (we really only use the filter bar on single-widget dashboards like service list), because it depends on every data source being able to apply the same filter. Even then, each data source has to opt in to merging the inherited filters - which is even more error prone, as data sources are much more numerous and change more frequently.

Given the time to do it, I think we could unify the systems, but it would be a major undertaking. I'm thinking of a hierarchical system of query contexts (not necessarily limited to filters), with a root that is applying global things like this, and with new contexts created underneath by things like filter bars - but even in that system, somewhere we need a way of applying that context in a way that is specific to each request, as they all have different shapes - and that, I think belongs on the handler side, not the requesting side.

anandtiwary: Sounds like a good project for you :)

aaron-steinfeld: hah - if only I had the time

projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
aaron-steinfeld: it wouldn't work very well as an abstract method. The transformation depends on the shape of the request, the logic of the handler and the shape of the constructed query. We could make a util, but at that point we're basically wrapping `this.globalGraphQlFilterService.mergeGlobalFilters` with another function, so no real purpose. Also, there's no base class for the handlers (and if there were, we wouldn't want to add an injection dependency to it for something that's only applicable to certain handlers).

anandtiwary: okay. I was thinking we could just define a function below. While it doesn't do much, it would just force the dev to consider the case of global filter application. Otherwise it is easy to miss. It is still a hack.
```
export interface GraphQlHandler<TRequest, TResponse> {
  readonly type: GraphQlHandlerType;

  matchesRequest(request: unknown): request is TRequest;
  /**
   * Converts the provided request into a single selection, or a map of selections with arbitrary keys.
   * If a map is provided, the response provided to the converter will use the same keys to disambiguate
   * the response for each selection.
   */
  convertRequest(request: TRequest): GraphQlSelection | Map<unknown, GraphQlSelection>;
  /**
   * Converts the provided response, or map of responses (if a map of selections were provided in the request
   * conversion), to a response object.
   */
  convertResponse(response: unknown | Map<unknown, unknown>, request: TRequest): TResponse | Observable<TResponse>;

  getRequestOptions?(request: TRequest): GraphQlRequestOptions;
}
```

We do need a better solution for this.

anandtiwary (approved):

PR #547, opened 2021-01-29
fix: initial checkbox filtering when using view toggle model

jake-bassett:
## Description
Fixes issue with initial checkbox filter not being applied to query due to table being wrapped in TableWidgetViewToggleModel. This is due to the the TableWidgetViewToggleModel requiring a view be set prior to it creating the delegate model.

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
@@ -142,14 +142,22 @@ export class TableWidgetRendererComponent
       ])
     );
 
-    this.filterItems = this.model.getFilterOptions().map(filterOption => ({
-      label: capitalize(filterOption.label),
-      value: filterOption
-    }));
-
     this.viewItems = this.model.getViewOptions().map(viewOption => ({
       label: capitalize(viewOption),
       value: viewOption
+    }));
+
+    this.populateStaticControls();
+  }
+
+  private populateStaticControls(): void {
+    /*
+     * CAUTION: If TableWidgetViewToggleModel is used, this.model is not hydrated until after setView() is called,
+     *  which is triggered by onViewChange().
+     */
+    this.filterItems = this.model.getFilterOptions().map(filterOption => ({
+      label: capitalize(filterOption.label),
+      value: filterOption
     }));
 
     this.maybeEmitInitialCheckboxFilterChange();
@@ -160,11 +168,11 @@ export class TableWidgetRendererComponent
   protected fetchData(): Observable<TableDataSource<TableRow> | undefined> {
     return this.model.getData().pipe(
       startWith(undefined),
-      tap(() => this.fetchFilterValues())
+      tap(() => this.fetchAndPopulateDynamicControls())
     );
   }
 
-  protected fetchFilterValues(): void {
+  protected fetchAndPopulateDynamicControls(): void {
     this.selectFilterItems$ = forkJoinSafeEmpty(
       this.model.getSelectFilterOptions().map(selectFilterModel =>
         // Fetch the values for the selectFilter dropdown
@@ -326,6 +334,7 @@ export class TableWidgetRendererComponent
 
   public onViewChange(view: string): void {
     this.model.setView(view);
+    this.populateStaticControls();
     this.columnConfigs$ = this.getColumnConfigs();
   }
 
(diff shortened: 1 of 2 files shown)


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -142,16 +142,24 @@ export class TableWidgetRendererComponent
...
-      label: capitalize(filterOption.label),
-      value: filterOption
-    }));
-
     this.viewItems = this.model.getViewOptions().map(viewOption => ({
       label: capitalize(viewOption),
       value: viewOption
     }));
 
+    this.populateStaticControls();
```
aaron-steinfeld, line 150: this delineation between static and dynamic controls seems like we're embedding too much knowledge of the controls into the renderer.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -142,16 +142,24 @@ export class TableWidgetRendererComponent
...
+  }
+
+  private populateStaticControls(): void {
+    /*
+     * CAUTION: If TableWidgetViewToggleModel is used, this.model is not hydrated until after setView() is called,
```
aaron-steinfeld, line 155: `this.model` should always be hydrated (first line of the constructor). It sounds like the issue is the view toggle isn't picking a delegate, by default, and forcing the renderer to do so?

jake-bassett: Correct. That's exactly what it is, but not sure how to default one in the model without a constructor. Just assigning one on the class property doesn't work.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Yeah, open to suggestions then. Preferably without going down a rabbit hole. :-)

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
aaron-steinfeld: sounds like a case for `ModelOnInit`!

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
aaron-steinfeld: of course, of course. When using on init to set the delegate, does this need go away?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Oh, lovely. Didn't know we had that. Looks like the fine gentlemen who built the dashboarding system thought of about everything.

aaron-steinfeld: you'd be amazed by all the cool features we have sitting on a github shelf somewhere 😉

aaron-steinfeld: I'm still waiting for the variable within a variable use case to come in handy.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
jake-bassett: Hmm. Not really since there are still some controls that fetch data and some that don't.

We could just always repopulate all controls on fetch so the renderer doesn't have to have knowledge of which filters are static vs dynamic, but that seems worse. Not sure how else to handle this since we kind of mix and match how we provide data to our control filters. I kind of think they should all just use the modelAPI to getData(), even if its just an observable that returns a static list.

aaron-steinfeld: > We could just always repopulate all controls on fetch so the renderer doesn't have to have knowledge of which filters are static vs dynamic, but that seems worse

Worse in that its an extra control render that's not necessary? If it's not noticeable, that seems preferable to me (not sure if that would cause things like losing control state on refresh, though).

>  I kind of think they should all just use the modelAPI to getData(), even if its just an observable that returns a static list.

Agreed, that way the caller doesn't know/care how the callee is implemented. Is this different than the "repopulate all on fetch" above?

jake-bassett: Yeah, okay. I think we agree. The populate on fetch does seem the least bad.

aaron-steinfeld (approved):

PR #550, opened 2021-02-01
feat: add 'right' as an option for popover fixed position

arjunlalb:
## Description
add **right** as an option for popover fixed position


projects/components/src/overlay/sheet/sheet.ts
```diff
@@ -1,10 +1,12 @@
 import { InjectionToken } from '@angular/core';
 import { Observable } from 'rxjs';
+import { PopoverFixedPositionLocation } from '../../popover/popover';
 import { OverlayConfig } from './../overlay';
 
 export interface SheetOverlayConfig<TData = unknown> extends OverlayConfig {
   size: SheetSize;
   data?: TData;
+  position?: PopoverFixedPositionLocation;
```
aaron-steinfeld, line 9: The overlay should only accept two position values - can you type this to `PopoverFixedPositionLocation.Right | PopoverFixedPositionLocation.RightUnderHeader`

arjunlalb: Done

aaron-steinfeld (approved):

PR #556, opened 2021-02-01
feat: introduce triggerLabelDisplayMode for select

arjunlalb:
## Description
Adding a display mode for trigger labels in select component.

### Testing
Screenshot 2021-02-01 at 4 41 06 PM
UT added. Screenshot above shows both modes side by side.

### Checklist:


projects/components/src/select/select.component.ts
```diff
@@ -164,3 +176,8 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
     return this.items.find(item => item.value === value);
   }
 }
+
+export const enum SelectTriggerDisplayMode {
+  MenuWithChevron = 'menu-with-chevron',
+  Button = 'button'
```
aaron-steinfeld, line 182: Based on the screenshot, this second mode seems more like an Icon display rather than button - I'd expect button to have borders

arjunlalb: It could be text without the icon as well or icon + text. Do you have an alternate name suggestion? Label?

aaron-steinfeld: Hmm - do we have a use case case for icon + text? That seems a bit strange to me. If we do have that, I'd make it a third mode - specifically because the icon mode that we need for that use case should still accept a label, but display it as the tooltip IMO.

arjunlalb: We currently don't have a use case for icon + text, but I was thinking it is a possibility we want to show just those two, without the chevron and borders.

aaron-steinfeld: I actually misread what the text is - it's not just a descriptor/placeholder, but the selection. In the icon case, we don't want the current selection to show, do we? I thought we actually want to style differently when theres a selection.
No selections:
image

Active selections (not sure if that border/background is meant to be there always, or if that's only when pressed)
image

arjunlalb: In case of icon, we do not want to show the current selection.

My understanding is that the highlighting is for when the menu is active. That'll need a bit more work. Since the icon display is not an independent standalone item there, showing the circular highlighting etc needs some more changes.

arjunlalb: Now that I think of this, I think it might be better to have an icon only mode, with the `trigger-content` section itself separate, so that the checks/stylings are cleaner. What do you think @aaron-steinfeld ?

aaron-steinfeld: I think that's what I was pushing for, depending on exactly what you mean. What I was thinking was two modes (for now) the existing one and an icon only one. The template for select could specify the trigger via an ngSwitch and switch between the correct template for the mode.

> My understanding is that the highlighting is for when the menu is active

I would think the blue should be when it's got a non-empty selection, and the blue with the border/background would be when it's pressed (i.e. the menu is active). We can confirm that.

arjunlalb: Yea. that makes more sense. discarding this.

PR #558, opened 2021-02-01
feat: icon only mode for select trigger

arjunlalb:
## Description
Icon only mode for select trigger

### Testing
UTs added. Manually tested.

### Checklist:

### DIFF
--- a/projects/components/src/select/select.component.ts
+++ b/projects/components/src/select/select.component.ts
@@ -34,8 +34,7 @@ import { SelectSize } from './select-size';
         this.groupPosition,
         selected ? selected.style.toString() : '',
         this.showBorder ? 'border' : '',
-        this.disabled ? 'disabled' : '',
-        this.triggerDisplayMode
+        this.disabled ? 'disabled' : ''
       ]"
       *htLetAsync="this.selected$ as selected"
     >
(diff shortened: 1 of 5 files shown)


projects/components/src/select/select.component.ts
```diff
@@ -34,18 +34,41 @@ import { SelectSize } from './select-size';
         this.groupPosition,
         selected ? selected.style.toString() : '',
         this.showBorder ? 'border' : '',
-        this.disabled ? 'disabled' : ''
+        this.disabled ? 'disabled' : '',
+        this.triggerDisplayMode
```
aaron-steinfeld, line 38: Is this needed? Can we contain the display mode styling fully within the trigger switch?

projects/components/src/select/select.component.ts
```diff
@@ -34,18 +34,41 @@ import { SelectSize } from './select-size';
...
+            *ngSwitchCase="'${SelectTriggerDisplayMode.Icon}'"
+            class="trigger-content icon-only"
+            [ngClass]="this.selected !== undefined ? 'selected' : ''"
+          >
+            <ht-icon
+              class="icon"
+              *ngIf="this.icon"
+              [icon]="this.icon"
+              [size]="this.iconSize"
+              [htTooltip]="this.placeholder"
```
aaron-steinfeld, line 68: This should be the selected label if selected, right? matching line 55 - could move that logic together if you want.

aaron-steinfeld (commented):
We'll need a similar change for multi select right? Realized the use case I was aware of at least is using this mode with multi-select, not the standard select.

aaron-steinfeld (approved):

PR #563, opened 2021-02-04
fix: update time duration <-> duration string conversion

arjunlalb:
## Description
Update time duration <-> iso 8601 duration string conversion logic.
Currently, all duration objects are converted to duration string after converting into seconds. And, all duration strings are converted to TimeDuration objects with seconds as unit.
This results in loss of unit selection and all values are forcefully converted to seconds.
This PR updates this logic.

TImeDuration to durationString conversion will now create a duration string using the unit of the duration object.
DurationString to TimeDuration conversion will now create a time duration object based on the biggest unit value specified in the duration string.

### Testing
UT added.

### Checklist:

### DIFF
--- a/projects/common/src/time/time-duration.ts
+++ b/projects/common/src/time/time-duration.ts
@@ -1,4 +1,4 @@
-import { parse, toSeconds } from 'iso8601-duration';
+import { Duration, parse } from 'iso8601-duration';
 import { assertUnreachable } from '../utilities/lang/lang-utils';
 import { TimeUnit } from './time-unit.type';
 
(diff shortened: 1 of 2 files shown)


projects/common/src/time/time-duration.ts
```diff
@@ -14,16 +14,55 @@ export class TimeDuration {
...
     return this.toMillis() / this.unitInMillis(unit);
   }
 
   public toIso8601DurationString(): string {
-    return `PT${this.toMillis() / 1000}S`;
+    if ([TimeUnit.Year, TimeUnit.Month, TimeUnit.Week, TimeUnit.Day].includes(this.unit)) {
+      return `P${this.value}${this.unit.toUpperCase()}`;
+    }
+
+    return `PT${this.value}${this.unit.toUpperCase()}`;
```
aaron-steinfeld, line 65: isn't this the same as the if case? and should those cases go down to minutes?

projects/common/src/time/time-duration.ts
```diff
@@ -14,16 +14,55 @@ export class TimeDuration {
...
+  /**
+   * Method to convert a duration string to TimeDuration object
+   * Currently we store the information as a single unit string.
+   * Will need to be updated to handle multiple units when needed.
+   */
+  public static parse(durationString: string): TimeDuration | undefined {
+    const duration: Duration = parse(durationString);
+    const isValid = (durationValue: number | undefined) => durationValue !== undefined && durationValue !== 0;
+    if (isValid(duration.years)) {
+      return new TimeDuration(duration.years!, TimeUnit.Year);
```
aaron-steinfeld, line 26: Since this is reading only the most significant portion of the duration, my concern is that this will make it really easy to introduce a bug - for example, imagine the server returns a duration of PT2H30M. So, rather than changing our duration to hold multiple units (which I suspect is out of scope right now), could we disregard trying to preserve the exact units of the duration string? If we're ok storing it in millis, we can change this logic to be something like
```typescript
   const totalMillis = new TimeDuration(duration.years ?? 0, TimeUnit.Year).toMillis()
   + new TimeDuration(duration.months ?? 0, TimeUnit.Month).toMillis()
   + ...

return new TimeDuration(totalMillis, TimeUnit.Millisecond);
```
The formatted strings can still handle converting that back to its most significant units.

projects/common/src/time/time-duration.ts
arjunlalb: @aaron-steinfeld the difference is the `T` in the middle. Section that follows `T` is time. Hence, all units can be uppercased.

PR #567, opened 2021-02-05
refactor: expose refresh method from time range service

aaron-steinfeld:
## Description
All data listened to refresh through the changes in time range currently. ideally, we'd have a separate refresh service that is decoupled from the specific causes of refresh, but given the scope of such a change, this will allow us to trigger refreshes that are not tied to the current time range.

### Testing
Visually verified

### DIFF
--- a/projects/common/src/time/time-range.service.ts
+++ b/projects/common/src/time/time-range.service.ts
@@ -59,6 +59,10 @@ export class TimeRangeService {
     return this.setTimeRange(TimeRangeService.toFixedTimeRange(startTime, endTime));
   }
 
+  public refresh(): void {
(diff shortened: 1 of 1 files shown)


anandtiwary:
finally i can close https://github.com/hypertrace/hypertrace-ui/pull/199

aaron-steinfeld:
> finally i can close #199 :P

Hah - got about that one! I noticed you did't approve 😉 I don't love this, but needed for the global filter service. When those change, we need some way of forcing everything to refetch (when nothing should be cached any more since filters have changed). Still don't want it for general components to use.

anandtiwary (approved):

PR #568, opened 2021-02-05
feat: add "network-only" cacheability option

arjunlalb:
## Description
Add "network-only" cacheability option

### Testing
NA

### Checklist:

### DIFF
--- a/projects/graphql-client/src/graphql-config.ts
+++ b/projects/graphql-client/src/graphql-config.ts
@@ -40,7 +40,7 @@ export interface GraphQlRequestOptions {
 export const enum GraphQlRequestCacheability {
   NotCacheable = 'no-cache', // Values used by apollo
   Cacheable = 'cache-first',
-  NetworkOnly = 'network-only'
+  RefreshCache = 'network-only'
 }
 
 export const enum GraphQlHandlerType {
--- a/projects/graphql-client/src/graphql-config.ts
+++ b/projects/graphql-client/src/graphql-config.ts
@@ -39,7 +39,8 @@ export interface GraphQlRequestOptions {
 
 export const enum GraphQlRequestCacheability {
   NotCacheable = 'no-cache', // Values used by apollo
-  Cacheable = 'cache-first'
+  Cacheable = 'cache-first',
+  NetworkOnly = 'network-only'
 }
 
 export const enum GraphQlHandlerType {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1091,9 +1091,9 @@
       }
     },
     "@apollo/client": {
-      "version": "3.3.6",
-      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.3.6.tgz",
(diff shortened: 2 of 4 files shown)


projects/graphql-client/src/graphql-config.ts
```diff
@@ -39,7 +39,8 @@ export interface GraphQlRequestOptions {
 
 export const enum GraphQlRequestCacheability {
   NotCacheable = 'no-cache', // Values used by apollo
-  Cacheable = 'cache-first'
+  Cacheable = 'cache-first',
+  NetworkOnly = 'network-only'
```
anandtiwary, line 43: What does it do?

aaron-steinfeld: NotCacheable means it doesn't read or write to the cache, this would be it always reads but will write back. I'd suggest changing the name though to something like `RefreshCache`

arjunlalb: Done.

aaron-steinfeld (approved):

projects/graphql-client/src/graphql-config.ts
anandtiwary: So it would return previous cached value immediately and later update that value once gql calls complete?

aaron-steinfeld: other way around. Instead of reading from the cache, it makes a regular network call. Unlike `no-cache`, the result of that call is written back to the cache when returned. To the caller, the two look the same - to the next caller, the cache has the latest value.

aaron-steinfeld: (the one you're describing sounds like https://www.apollographql.com/docs/react/data/queries/#cache-and-network - but since our code isn't set up to handle multiple responses it's not too useful to us)

anandtiwary: Okay.. but if we are making the call anyway, what is the advantage of updating the cache? When does it get read without waiting for call to complete?

aaron-steinfeld: It's basically the "force refresh" case. Sometimes, we might mutate something and want to force a refetch and ignore the cache, but later queries can read again from the cache. We should never use this at the handler level (because if we use it every time, it's the same as no-cache).

anandtiwary: got it. But the mutation query already does it ootb right? I think Apollo supports this. The requirement is that the mutation should return a valid object with id and all of the mutated properties in it. Apollo would find the relevant object in cache and add mutated properties there. This may require some caching rules to be configured but I think i tried this with a poc.

aaron-steinfeld: unfortunately a lot of our mutations are not returning the item they're mutating, or are weird things like single primitive flags etc.

anandtiwary: yeah. this is the reason I have that weird change in my other pr.  May be on the graphql service side, we can define a standard expectations from these `mutations`?

aaron-steinfeld: Really depends on the API though - some of the mutation apis are really just setting a single field of another object (or of multiple other objects) so I think it's going to be hard to do anything too standard here.

PR #574, opened 2021-02-08
feat: support for "None" in select component

arjunlalb:
## Description
Support for "None" in select component.

### Testing
UTs added.

### Checklist:

### DIFF
--- a/projects/components/src/select/select.component.ts
+++ b/projects/components/src/select/select.component.ts
@@ -74,6 +74,19 @@ import { SelectSize } from './select-size';
         </ht-popover-trigger>
         <ht-popover-content>
           <div class="select-content" [ngStyle]="{ 'minWidth.px': triggerContainer.offsetWidth }">
+            <ng-container *ngIf="this.showNoneSelection">
+              <div class="select-option none-option" (click)="this.onSelectionChange(this.noneOption)">
+                <span class="label">None</span>
+                <ht-icon
+                  class="status-icon"
+                  *ngIf="this.highlightSelected && this.isSelectedItem(this.noneOption)"
+                  icon="${IconType.Checkmark}"
+                  size="${IconSize.Small}"
+                ></ht-icon>
+              </div>
+
+              <ht-divider></ht-divider>
+            </ng-container>
             <div
               *ngFor="let item of items"
               (click)="this.onSelectionChange(item)"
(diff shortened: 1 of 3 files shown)


projects/components/src/select/select.component.ts
```diff
@@ -126,15 +139,22 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
   @Input()
   public highlightSelected: boolean = true;
 
+  @Input()
+  public showNoneSelection: boolean = false;
```
aaron-steinfeld, line 143: What benefit do we get from putting this in the component? How is it different from me, as a component consumer, supplier a value called `"None"` that I can then use however I want? This bakes in a lot of logic to the component for something that seems already doable.

arjunlalb: I feel this is cleaner done in the component because of the styling. We want to separate None from other options using the divider. Its like 'Select All` / 'Unselect All` in a multi select component.

aaron-steinfeld: So the extra styling is that separator - but supporting separators is a general requirement for selects anyway. My main concern, much more so than the extra complexity of explicitly implementing `None` is the breaking API change for existing usages by the `| undefined`.

So my preference would be to add divider support (we could do that by extending the `select-option` api, or adding a new type of select child), but if we can get away with not changing the existing component API, I'm fine with that too.

arjunlalb: Does adding logic in the component to identify an item with `undefined` value (if present) and show it at the top with the divider sounds like a better idea? That way we don't need to modify the component definition and the consumer can still control whats shown.

aaron-steinfeld: So if we want this base component to own "none" (specifically, where it's positioned and how it's styled), then to support that we could take in a little bit of config like `public noneSelectionConfig: { value: V };` - walking through my thinking here:
- This allows the API to still stick to `V` values (which may be defined to include `undefined` if the caller chooses)
- It supports but does not assume `undefined` for None. Taking in a value without wrapping it would be challenging to disambiguate between not set, and trying to search for undefined is making usage assumptions that aren't in the API.


But the reason I really am hesitant to build this support into the component at all is "None" is really tied to a specific use case. In other use cases it could be "Clear", "Unset", "Remove" etc. Sometimes we may want it at the top, others the bottom. Because of that, I think taking in the config object, at a minimum, allows us future remove to support these kind of customizations. But once we start doing that, the config object starts very much looking like another select-option with a special api, which is why I'd rather go down the path of supporting this through the content.

arjunlalb: https://github.com/hypertrace/hypertrace-ui/pull/578

PR #575, opened 2021-02-10
feat: adding notification operator

anandtiwary:
## Description
feat: adding notification operator

### Testing

### Checklist:

### Documentation


anandtiwary:
Two ways to use this:
1. `observable$.pipe(source => this.notificationService.withNotification('success', 'failure'), ...<other operators>)
2. `observable$.pipe(this.notificationService.withNotificationOperator('success', 'failure'), ...<other operators>)

With #2, In case `operator` type definition changes by Rxjs, we would only have to update one function.

projects/components/src/notification/notification.service.ts
```diff
@@ -48,4 +49,21 @@ export class NotificationService {
       data: { message: message, mode: NotificationMode.Info, closedObserver: this.closedObserver$ }
     });
   }
+
+  public withNotification<T>(source: Observable<T>, successMessage: string, failureMessage: string): Observable<T> {
+    return source.pipe(
+      tap(
```
aaron-steinfeld, line 55: How difficult is it to make a real operator (I've done it in rxjava where it's pretty confusing, haven't looked in rxjs)? My concern with this one is that it gets awkward to use in a pipe

aaron-steinfeld: Oh I take it back - it looks like rxjs operators are much simpler (should have read your other docs first). Given that, I may suggest switching the names around a bit (so the name looks more like an operator) but otherwise looks good.

something like `withNotification` -> `wrapWithNotification` and `withNotificationOperator` -> `withNotification`

aaron-steinfeld: For comparison, this is what an operator in java looked like 🙄 
https://github.com/hypertrace/query-service/blob/main/query-service-impl/src/main/java/org/hypertrace/core/query/service/RowChunkingOperator.java

aaron-steinfeld (approved):

projects/components/src/notification/notification.service.ts
anandtiwary: Yea I tried building our own operator. I got it to work too, but realized that it still has to go via notification service (since we need injected service). With that in consideration, the native implementation was not providing much value. I switched to the current approach which anyway goes via the service but is much more readable and maintainable.

anandtiwary: Will update the name

aaron-steinfeld: Ah right - so looking at an example, pipe takes a function of `UnaryFunction<Observable<T>, Observable<R>>` but the standard lib ones are generally implemented via lift and an operator. I'll have to page back in if there's any benefit to using lift + operator vs pipe, but nothing seems obvious (and if so we can do it without changing the pai.

aaron-steinfeld (approved):

projects/components/src/notification/notification.service.ts
aaron-steinfeld: Even their docs examples use pipe - I'd guess lift is the old way of doing it (before they used pipe when everything was on the prototype)
https://reactive.how/rxjs/pipeable-operators

anandtiwary: Yeah, the pipe approach is pretty popular. rxjs folks are also adding newer apis in 7.x, so this could change too. 

The lift approach would require us to build our own operator which comes with additional code. 

I think if we are doing something pretty involved then building our own operator makes sense. For something simple, Piggybacking on existing 
operators is much cleaner fnow.

PR #576, opened 2021-02-10
refactor: table controls more generic and common API

jake-bassett:
## Description
Refactor Table Controls to make them more generic and use common API.


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
```diff
@@ -119,6 +125,7 @@ export class TableWidgetColumnModel {
       editable: true,
       filterable: this.filterable,
       sort: this.sort,
+      sortable: this.sortable,
```
aaron-steinfeld, line 128: have we seriously never exposed this? I'm shocked

projects/distributed-tracing/src/shared/dashboard/data/graphql/graphql-options-data-source.model.ts
```diff
@@ -0,0 +1,10 @@
+import { TableControlOption } from '@hypertrace/components';
+import { Observable } from 'rxjs';
+import { GraphQlDataSourceModel } from './graphql-data-source.model';
+
+export abstract class GraphqlOptionsDataSourceModel<
```
aaron-steinfeld, line 5: nit: naming - this doesn't indicate anywhere that it has to do with table

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-control-checkbox-option-model.ts
```diff
@@ -0,0 +1,37 @@
+import { TableCheckboxOptions, TableControlOption } from '@hypertrace/components';
```
aaron-steinfeld, line 1: file name - `*.model.ts`

projects/components/src/table/table.component.ts
```diff
@@ -224,6 +225,9 @@ export class TableComponent
   @Input()
   public filters?: TableFilter[];
 
+  @Input()
+  public queryProperties?: Dictionary<unknown> = {};
```
aaron-steinfeld, line 229: what are query properties?

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,12 +1,56 @@
-import { KeyValue } from '@angular/common';
+import { Dictionary } from '@hypertrace/common';
 import { SelectOption } from '../../select/select-option';
+import { TableFilter } from '../table-api';
 
-export interface SelectFilter {
+export interface SelectControl {
   placeholder?: string;
-  options: SelectOption<KeyValue<string, unknown>>[];
+  default?: SelectOption<TableControlOption>;
```
aaron-steinfeld, line 7: should this option be specified in options too?

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,12 +1,56 @@
...
 export interface SelectChange {
-  select: SelectFilter;
-  value: KeyValue<string, unknown>;
+  select: SelectControl;
+  value: TableControlOption;
 }
+
+export interface CheckboxControl {
+  label: string;
+  value: boolean;
```
aaron-steinfeld, line 18: is this the default? If so, does it make sense to match shape with the select control?

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,12 +1,56 @@
...
+  options: TableCheckboxOptions;
+}
+
+export interface CheckboxChange {
+  checkbox: CheckboxControl;
+  value: boolean;
```
aaron-steinfeld, line 24: isn't the value contained in the option being sent back (similar to select)?

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,12 +1,56 @@
...
-  select: SelectFilter;
-  value: KeyValue<string, unknown>;
+  select: SelectControl;
+  value: TableControlOption;
 }
+
+export interface CheckboxControl {
+  label: string;
+  value: boolean;
+  options: TableCheckboxOptions;
```
aaron-steinfeld, line 19: should this be a trueOption and a falseOption both required - seems weird to ask for an array of two with known values in it. If you do leave it as the array, see comment below about type.

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,12 +1,56 @@
...
+export const enum TableControlOptionType {
+  Unset = 'unset',
+  Filter = 'filter',
+  Property = 'property'
+}
+
+export interface TableControlOption<TMetaValue = unknown, TValue = unknown> {
+  type: TableControlOptionType;
+  label: string;
+  metaValue: TMetaValue; // Used in a query - type based on TableWidgetControlOptionType
```
aaron-steinfeld, line 37: who reads this? Is the table just carrying it along I hope (in which case, it doesn't need to be part of the api)

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,12 +1,56 @@
...
+  type: TableControlOptionType.Filter;
+  metaValue: TableFilter;
+}
+
+export interface TablePropertyControlOption extends TableControlOption<Dictionary<unknown>> {
+  type: TableControlOptionType.Property;
+  metaValue: Dictionary<unknown>;
+}
+
+export type TableCheckboxOptions = [TableControlOption<unknown, boolean>, TableControlOption<unknown, boolean>];
```
aaron-steinfeld, line 56:
```suggestion
export type TableCheckboxOptions = [TableControlOption<unknown, true>, TableControlOption<unknown, false>];
```

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -213,6 +216,7 @@ export class TableCdkDataSource implements DataSource<TableRow> {
     columnConfigs: TableColumnConfigExtended[],
     pageEvent: PageEvent,
     filters: TableFilter[],
+    queryProperties: Dictionary<unknown>,
```
aaron-steinfeld, line 219: Similar to the comment about metadata way back - the query properties aren't being read here, we're just piping them through the table infrastructure for no clear reason. The table widget that holds them can append them to any requests for data, right?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-control-checkbox-option-model.ts
```diff
@@ -0,0 +1,37 @@
...
+  type: 'table-widget-checkbox-option'
+})
+export class TableWidgetControlCheckboxOptionModel extends TableWidgetControlModelBase {
+  @ModelProperty({
+    key: 'checked',
+    displayName: 'Checked',
+    type: BOOLEAN_PROPERTY.type,
+    required: false
+  })
+  public checked: boolean = false;
```
aaron-steinfeld, line 17: so a checkbox has two options, true and false. Is this model just to associate metadata with each one?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-control-model-base.ts
```diff
@@ -0,0 +1,48 @@
+import { TableControlOption, TableControlOptionType } from '@hypertrace/components';
+import { BOOLEAN_PROPERTY, ModelApi, ModelProperty } from '@hypertrace/hyperdash';
+import { ModelInject, MODEL_API } from '@hypertrace/hyperdash-angular';
+import { uniqWith } from 'lodash-es';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+export abstract class TableWidgetControlModelBase {
```
aaron-steinfeld, line 8: naming preference nit: `TableWidgetControlModel` (I just hate using `*Base`, `Abstract*`, `*Impl` and the like)

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-control-model-base.ts
```diff
@@ -0,0 +1,48 @@
...
+  @ModelProperty({
+    key: 'sort',
+    type: BOOLEAN_PROPERTY.type
+  })
+  public sort: boolean = true;
```
aaron-steinfeld, line 13: Do these really apply to all controls (such as checkboxes?)

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -158,60 +158,58 @@ export class TableWidgetRendererComponent
...
-                  key: selectFilterModel.attribute,
-                  value: undefined
-                }
-              });
-            }
-
             return {
-              placeholder: selectFilterModel.unsetOption,
-              options: options
+              placeholder: selectOptions[0]?.label,
```
aaron-steinfeld, line 178: Why is the first one the placeholder?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-view-toggle.model.ts
```diff
@@ -84,11 +83,16 @@ export class TableWidgetViewToggleModel extends TableWidgetBaseModel implements
...
-    return this.delegateModel?.getCheckboxFilterOption() ?? this.checkboxFilterOption;
+  public getCheckboxControlOptions(): TableWidgetControlCheckboxOptionModel[] {
+    return this.delegateModel?.getCheckboxControlOptions() ?? this.checkboxOptions;
   }
 
-  public getSelectFilterOptions(): TableWidgetSelectFilterModel[] {
-    return this.delegateModel?.getSelectFilterOptions() ?? this.selectFilterOptions;
+  public setCheckboxControlOptions(checkboxOptions: TableWidgetControlCheckboxOptionModel[]): void {
+    this.delegateModel?.setCheckboxControlOptions(checkboxOptions);
+    this.checkboxOptions = checkboxOptions;
```
aaron-steinfeld, line 92: why hold these in the toggle? Doesn't this always belong to the delegate?

projects/observability/src/pages/apis/endpoints/endpoint-list.dashboard.ts
```diff
@@ -22,17 +22,18 @@ export const endpointListDashboard: DashboardDefaultConfiguration = {
...
-            attribute: 'serviceName',
-            'unset-option': 'All Services',
+            type: 'table-widget-select-option',
+            'unique-values': true,
             data: {
-              type: 'entities-attribute-data-source',
-              entity: ObservabilityEntityType.Service,
+              type: 'entities-attribute-options-data-source',
+              'unset-label': 'All Services',
+              // Use API so we can inherit API filters
```
aaron-steinfeld, line 32: :( makes sense, just hate that we have to query so much more data for this

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: Yeah, the asymmetry bothers me too. `SelectControl` and `CheckboxControl` are similar, but just distinct enough that they need to be shaped a bit different. It's admittedly confusing. 

The `SelectControl` represents one component dropdown with **1 to n** `options`, each carrying its own `metaValue` for querying based on that selected option... and there can be **0 to n** dropdowns in the table control bar.

The `CheckboxControl` represents a single checkbox and the `options` represent the carried `metaValue` for querying `true` vs `false`. There can also be **1 to n** checkboxes, but they are all clubbed together under a single multiselect dropdown.

projects/components/src/multi-select/multi-select-justify.ts
```diff
@@ -0,0 +1,6 @@
+export const enum MultiSelectJustify {
```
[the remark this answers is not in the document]
jake-bassett: Different values to be used in css.

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: I knew this might be a sticking point because it just feels wrong in many ways, but the reason it is an array to so that all controls can have the same data fetching shape.

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: Oh, good eye. Returning `option` was a later change that supersedes returning `value`. Removing.

projects/components/src/table/data/table-cdk-data-source.ts
jake-bassett: Combined response from [this comment](https://github.com/hypertrace/hypertrace-ui/pull/576/files#r574155027).

So, the `metaValue`. What that is, is a json configurable way to provide the information that needs to be applied to the query in a generic way when a sinlge option is applied. The shape of `metaValue` is based on the the `type` of the `TableControlOption` There are three types right now.

```typescript
export const enum TableControlOptionType {
  Filter = 'filter',
  Property = 'property',
  UnsetFilter = 'unset-filter'
}
```

1. `Filter` expects `TableFIlter`
2. `Property` expects a `Dictionary<unknown>` of query properties and their associated values
3. `UnsetFilter` expects a `string`. It uses that string to find a currently applied filter and remove it

As for the second part about table carrying all of this through. This has come up before. Its not optimal and table is the only component that really takes this approach due to the history of building it with the `TableCdkDataSource` following Material guides. But, also, the way things are currently set up, its actually necessary for `filters' due to them being manipulated by table component internally in the column header filtering. 

That's not true for `queryProperties` right now (hopefully never is). Table component doesn't manipulate them in any way so we could pull those out and not pipe them through. I really don't want to do that though because then it splits `queryProperties` from `filters` and they are really very close in how we use them. I prefer to keep them being used the same way for consistency than split them up. Besides, the `filters` were so lonely before. Now they have a friend. Do you really want to do that to them? DO YOU!?!?

aaron-steinfeld: my god. I never thought about it like this (I guess we'll have to fix them both together 😉 )

aaron-steinfeld (approved):

PR #577, opened 2021-02-10
fix: support passing a templateRef for title tooltip via column model

arjunlalb:
## Description
Support passing templateRef for title tooltip via column model

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
@@ -1,3 +1,4 @@
+import { TemplateRef } from '@angular/core';
 import {
   CoreTableCellRendererType,
   TableCellAlignmentType,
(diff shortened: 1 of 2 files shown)


skjindal93 (approved):
unit tests? 🤔

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
```diff
@@ -37,9 +45,9 @@ export class TableWidgetColumnModel {
   @ModelProperty({
     key: 'titleTooltip',
     displayName: 'Title Tooltip',
-    type: STRING_PROPERTY.type
+    type: UNKNOWN_PROPERTY.type
   })
-  public titleTooltip?: string;
+  public titleTooltip?: string | TemplateRef<unknown>;
```
aaron-steinfeld, line 50: The api change makes sense for programmatic usages, but the model will never be a template (otherwise we're misusing it)

PR #578, opened 2021-02-10
feat: support control options in select

arjunlalb:
## Description
Support control options in select.

### Testing
UTs added.

### Checklist:

### DIFF
--- a/projects/components/src/select/select.component.ts
+++ b/projects/components/src/select/select.component.ts
@@ -15,6 +15,7 @@ import { LoggerService, queryListAndChanges$, SubscriptionLifecycle, TypedSimple
 import { EMPTY, merge, Observable, of } from 'rxjs';
 import { map, switchMap } from 'rxjs/operators';
 import { IconSize } from '../icon/icon-size';
+import { SelectControlOptionComponent, SelectControlOptionPosition } from './select-control-option.component';
 import { SelectGroupPosition } from './select-group-position';
 import { SelectJustify } from './select-justify';
 import { SelectOption } from './select-option';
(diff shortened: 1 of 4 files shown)


projects/components/src/select/select.component.ts
```diff
@@ -151,6 +177,7 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
 
   public ngAfterContentInit(): void {
     this.selected$ = this.buildObservableOfSelected();
+    this.topControlItems = this.controlItems?.filter(item => item.position === SelectControlOptionPosition.Top);
```
aaron-steinfeld, line 180: nit:  query list can change over time. Can we set this to something like:
```suggestion
    this.topControlItems$ = queryListAndChanges(this.controlItems).pipe(map(items => items.filter(item => item.position === SelectControlOptionPosition.Top)));
```

projects/components/src/select/select.component.ts
```diff
@@ -90,6 +99,18 @@ import { SelectSize } from './select-size';
               </ht-icon>
             </div>
           </div>
+
+          <ng-template #controlItemTemplate let-items="controlItems">
```
aaron-steinfeld, line 103: any reason to have the separate template rather than doing this inline? It's not reused or passed around, and it seems like it makes it a bit less readable. On the other hand, I could see using a template for the item and sharing it in both the control and standard branches.

arjunlalb: The reason for using a template is so that we can extend this to a `bottom` control group usecase later.
Reason to not combine the options and control items is because of the selected status icon. Control items do not need the selected status, but options do.

aaron-steinfeld: Ah right. I think we'd benefit from sharing the template though given the amount of duplication. For example, the control item here will render item.icon, but it doesn't look like the regular items render it. Could we use, in the same pattern you have, something like `let-show-selection-status`?

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):
neat

PR #579, opened 2021-02-10
Description component improve

DmitiryPotychkin:
- stop event bubbling after click on component's buttons
- add ability to pass content in component, display it in line with description text


projects/components/src/description/description.component.ts
```diff
@@ -21,18 +21,19 @@ import {
         data-sensitive-pii
         #eventDescriptionText
       >
+        <ng-content></ng-content>
```
aaron-steinfeld, line 24: I'm not against using content to pass in the description, but it's an alternative API to passing in description as an input - we shouldn't have both. If we have use cases that need more than a simple string passed in, let's make this the sole API to pass in content and remove the `description` input.

DmitiryPotychkin: Fixed

aaron-steinfeld (commented):
lgtm - will let @anandtiwary take a pass too

projects/components/src/description/description.component.ts
anandtiwary: Curious to know why we need ng-content? Is their an example usage I could see? Is it suppose to wrap any other component apart from a string?

aaron-steinfeld: I think the main use cases are links and formatted strings

anandtiwary: okay.. Links we should handle from outside, but i see the need for formatted strings.

anandtiwary (approved):

aaron-steinfeld (approved):

PR #580, opened 2021-02-10
Topology component

DmitiryPotychkin:
- Change size of  topology brush components
- Change min width of topology tool-tip


anandtiwary:
Is there anything more to this? Could you please share any screenshot, before and after?

PR #582, opened 2021-02-10
fix: parsing of string map values

jake-bassett:
## Description
Fix string map value parsing by only splitting on the first delimiter.


projects/components/src/filtering/filter/filter-delimiters.ts
```diff
@@ -1,3 +1,11 @@
+import { isNonEmptyString } from '@hypertrace/common';
+
 export const ARRAY_DELIMITER = ',';
 export const MAP_LHS_DELIMITER = '.';
 export const MAP_RHS_DELIMITER = ':';
+
+export const splitFirstOccurrence = (str: string, delimiter: string): string[] => {
+  const firstIndex = str.indexOf(delimiter);
+
+  return [str.substr(0, firstIndex), str.substr(firstIndex + 1)].filter(isNonEmptyString);
```
aaron-steinfeld, line 10: nit - I don't think this handles edge cases correctly (at least anything where a empty string is part of the expected split result such as a string that starts or ends with the delimiter, or an empty string)

projects/components/src/filtering/filter/parser/types/contains-filter-parser.ts
```diff
@@ -46,7 +46,7 @@ export class ContainsFilterParser extends AbstractFilterParser<PossibleValuesTyp
     splitFilter: SplitFilter<FilterOperator>
   ): string[] | undefined {
     if (splitFilter.lhs === attribute.displayName) {
-      return splitFilter.rhs.split(MAP_RHS_DELIMITER);
+      return splitFirstOccurrence(splitFilter.rhs, MAP_RHS_DELIMITER);
```
aaron-steinfeld, line 49: Could we add a test for this?

projects/components/src/filtering/filter/filter-delimiters.ts
jake-bassett: Added a test for that case.

aaron-steinfeld: the test cases look fine, but doesn't this still behave differently than split itself? If that's acceptable for how it's used then it doesn't need to change, but should be local to the file it's used vs exported like a generic util.

aaron-steinfeld: Just noticed the name changed in this rev - that's better, but I think it'd be cleaner to just keep it local

aaron-steinfeld (approved):

PR #587, opened 2021-02-11
feat: support html strings in title tooltip

arjunlalb:
## Description
Support html strings in title tooltip.

### Testing
UTs added.

### Checklist:


projects/components/src/table/header/table-header-cell-renderer.component.ts
```diff
@@ -64,13 +76,18 @@ import { TableColumnConfigExtended } from '../table.service';
...
+          <div [innerHTML]="this.sanitizedHtmlForTooltip"></div>
+        </ng-template>
       </ng-template>
     </div>
   `
 })
 export class TableHeaderCellRendererComponent implements OnInit, OnChanges {
   public readonly SORT_ASC: TableSortDirection = TableSortDirection.Ascending;
   public readonly SORT_DESC: TableSortDirection = TableSortDirection.Descending;
+  private readonly HTML_TOOLTIP_PREFIX: string = 'html:';
```
aaron-steinfeld, line 90: why the prefix? even a plain string is valid html so can we treat any tooltips that way?

projects/components/src/table/header/table-header-cell-renderer.component.ts
```diff
@@ -102,9 +119,14 @@ export class TableHeaderCellRendererComponent implements OnInit, OnChanges {
 
   public isFilterable: boolean = false;
 
+  @ViewChild('htmlTooltip')
+  public htmlTooltipTemplate?: TemplateRef<unknown>;
+  public sanitizedHtmlForTooltip?: string | null;
```
aaron-steinfeld, line 124: let's keep this as `?: string` (undefined or string) - we don't want to work with nulls, so wherever angular may return one, throw a `?? undefined` on.

projects/components/src/table/header/table-header-cell-renderer.component.ts
arjunlalb: I was being _clever_ with adding a differentiator to html strings. But you are right, its unnecessary. Removed.

projects/components/src/table/header/table-header-cell-renderer.component.ts
arjunlalb: Done

projects/components/src/table/header/table-header-cell-renderer.component.ts
```diff
@@ -64,6 +76,10 @@ import { TableColumnConfigExtended } from '../table.service';
             </div>
           </ht-popover-content>
         </ht-popover>
+
+        <ng-template #htmlTooltip>
+          <div [innerHTML]="this.sanitizedHtmlForTooltip"></div>
```
aaron-steinfeld, line 81: So I mentioned sanitization, but now that I think about it, angular should be doing that for us. Could you try removing the sanitizer and verifying that a tooltip is not executed (but still rendered?)

Ref: https://angular.io/guide/security#sanitization-example
Specifically:
> Angular recognizes the value as unsafe and automatically sanitizes it, which removes the onerror attribute but keeps safe content such as the <b> element.

projects/components/src/table/header/table-header-cell-renderer.component.ts
```diff
@@ -144,6 +166,25 @@ export class TableHeaderCellRendererComponent implements OnInit, OnChanges {
...
+
+    return this.htmlTooltipTemplate;
+  }
+
+  public getSanitizedHtml(tooltipString: string | undefined): string | null {
+    if (tooltipString === undefined) {
+      return '';
+    }
+
+    return this.sanitizer.sanitize(SecurityContext.HTML, tooltipString);
```
aaron-steinfeld, line 185: if we keep the sanitizer, we should put the `?? undefined` here - our methods should never return null (I thought we had a lint rule on that)

projects/components/src/table/header/table-header-cell-renderer.component.ts
arjunlalb: aah. cool. it does do the work for us.

arjunlalb: Updated.

aaron-steinfeld (approved):

PR #593, opened 2021-02-13
feat: feature flag cartesian metric health

jake-bassett:
## Description
Turn off all Cartesian Widget fetching of metric health unless specifically feature flagged to be on.


aaron-steinfeld (changes requested):
Doesn't belong in shared component

PR #605, opened 2021-02-17
feat: add variables input to navigable dashboard comp

jake-bassett:
## Description
Add variables input to navigable dashboard component.

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
@@ -1,5 +1,5 @@
 import { ChangeDetectionStrategy, Component, EventEmitter, Input, OnChanges, Output } from '@angular/core';
-import { TypedSimpleChanges } from '@hypertrace/common';
+import { Dictionary, TypedSimpleChanges } from '@hypertrace/common';
 import { Filter } from '@hypertrace/components';
 import { DashboardPersistenceService } from '@hypertrace/dashboards';
 import { Dashboard, ModelJson } from '@hypertrace/hyperdash';
@@ -37,10 +37,13 @@ import { GraphQlFilterDataSourceModel } from '../data/graphql/filter/graphql-fil
 })
 export class NavigableDashboardComponent implements OnChanges {
   @Input()
-  public readonly defaultJson?: ModelJson;
+  public navLocation?: string | null;
 
   @Input()
-  public navLocation?: string | null;
+  public defaultJson?: ModelJson;
+
+  @Input()
+  public variables?: Dictionary<unknown>;
 
   @Input()
   public filterConfig?: NavigableDashboardFilterConfig;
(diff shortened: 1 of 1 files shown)


projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
```diff
@@ -37,10 +37,13 @@ import { GraphQlFilterDataSourceModel } from '../data/graphql/filter/graphql-fil
 })
 export class NavigableDashboardComponent implements OnChanges {
   @Input()
-  public readonly defaultJson?: ModelJson;
```
aaron-steinfeld, line 40: man, this must be old :)

projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
```diff
@@ -37,10 +37,13 @@ import { GraphQlFilterDataSourceModel } from '../data/graphql/filter/graphql-fil
...
+  public navLocation?: string | null;
 
   @Input()
-  public navLocation?: string | null;
+  public defaultJson?: ModelJson;
+
+  @Input()
+  public variables?: Dictionary<unknown>;
```
aaron-steinfeld, line 46: I generally prefer Map over a plain object, but I don't have strong preferences - would rather be consistent. Do you know if we use either as inputs elsewhere?

projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
```diff
@@ -100,6 +104,13 @@ export class NavigableDashboardComponent implements OnChanges {
     }
   }
 
+  public applyVariablesToDashboard(dashboard: Dashboard, variables: Dictionary<unknown> = {}): void {
+    for (const key of Object.keys(variables)) {
+      dashboard.setVariable(key, variables[key]);
+    }
+    dashboard.refresh();
```
aaron-steinfeld, line 111: The refresh shouldn't be needed, but let's verify that this applies before the first fetch.

projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
aaron-steinfeld, line 46: Should we react to changes in this in `ngOnChanges`?

jake-bassett: I used Dictionary to pass in the request properties to table CDK (though we will eventually remove that and filters).

I kind of prefer Dictionary to Map because it doesn't need to be instantiated and is a little lighter weight. Can change though if we feel strongly about using Map instead.

jake-bassett: Hmm. I thought about it, but doesn't really seem necessary. I don't think they should change and I figured we don't for the filters so we could add later if needed.

aaron-steinfeld (approved):

PR #612, opened 2021-02-18
feat: additional dependencies for time agnostic apis

jake-bassett:
## Description
Some additional changes for Time Agnostic APIs. This fixes an issue with drilling into an API that is fetched in a list using `includeInactive`. We now include that property whenever querying for a single API. The second change here is to display null data correctly.


projects/distributed-tracing/src/shared/components/table/data-cell/metric/metric-table-cell-parser.ts
```diff
@@ -25,7 +25,7 @@ export class MetricTableCellParser extends TableCellParserBase<number, CellValue
       case 'number':
         return cellData;
       case 'object':
-        return cellData === null ? undefined : cellData.value;
+        return cellData === null ? undefined : cellData.value === null ? undefined : cellData.value;
```
anandtiwary, line 28: nit: `cellData?.value ?? undefined`

projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
```diff
@@ -15,7 +15,7 @@ import { EntityNavigationService } from '../../services/navigation/entity/entity
       class="ht-entity-renderer"
       [ngClass]="{ navigable: this.navigable }"
       [htTooltip]="this.name"
-      (click)="this.onClickNavigate()"
+      (click)="this.navigable && this.onClickNavigate()"
```
anandtiwary, line 18: Should this check be put inside?

projects/observability/src/shared/graphql/request/handlers/entities/query/entity/entity-graphql-query-handler.service.ts
```diff
@@ -56,7 +56,9 @@ export class EntityGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
       limit: 1,
       timeRange: request.timeRange,
       properties: request.properties,
-      filters: [new GraphQlEntityFilter(request.id, request.entityType)]
+      filters: [new GraphQlEntityFilter(request.id, request.entityType)],
+      // If querying for a single API, then always want to includeInactive
+      includeInactive: request.entityType === ObservabilityEntityType.Api
```
aaron-steinfeld, line 61: This would apply always right? Even if we disabled this functionality elsewhere

projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
jake-bassett: Eh, its a fair ask but not sure its really necessary here. Just more lines of code and I don't feel it is any cleaner. If you feel strongly I can change though.

projects/observability/src/shared/graphql/request/handlers/entities/query/entity/entity-graphql-query-handler.service.ts
jake-bassett: Hmm. Good point. It would. Not sure how to feature flag this cleanly. Open to suggestions.

projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
anandtiwary: cool

aaron-steinfeld (approved):

PR #614, opened 2021-02-19
feat: add time ago pipe for last called column

jake-bassett:
## Description
Add time ago pipe and  time ago cell renderer for Last Called column.


projects/common/src/utilities/formatters/time/display-time-ago.pipe.ts
```diff
@@ -0,0 +1,48 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Dictionary } from '../../types/types';
+
+@Pipe({
+  name: 'htDisplayTimeAgo'
+})
+export class DisplayTimeAgo implements PipeTransform {
+  private readonly intervals: Dictionary<number> = {
+    year: 31536000,
```
aaron-steinfeld, line 9: This stuff is already built into duration, can't we just reuse that (`TimeDuration.getMostSignificantUnitOnly`)

projects/common/src/utilities/formatters/time/display-time-ago.pipe.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    return '-';
+  }
+
+  private calcSecondsAgo(timestamp: DateTimeValue): number {
+    return Math.floor((+new Date() - +new Date(timestamp)) / 1000);
+  }
+}
+
+type DateTimeValue = Date | number;
```
aaron-steinfeld, line 48: Is this assumed to be millis?

projects/common/src/utilities/formatters/time/display-time-ago.pipe.ts
```diff
@@ -0,0 +1,48 @@
...
+      if (counter > 0) {
+        return `${counter} ${key}${counter === 1 ? '' : 's'} ago`;
+      }
+    }
+
+    return '-';
+  }
+
+  private calcSecondsAgo(timestamp: DateTimeValue): number {
+    return Math.floor((+new Date() - +new Date(timestamp)) / 1000);
```
aaron-steinfeld, line 44: `Date.now()` is a lot clearer than `+new Date()`. Likewise, `+new Date(timestamp))` vs `new Date(timestamp).getTime()`. That said, forget javascript coercing! Now our DateCoercer here would make more sense. It handles more value types and is already what we use to perform this logic in other places

projects/common/src/utilities/formatters/time/display-time-ago.pipe.ts
```diff
@@ -0,0 +1,48 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Dictionary } from '../../types/types';
+
+@Pipe({
+  name: 'htDisplayTimeAgo'
```
aaron-steinfeld, line 5: This is not a pure pipe - the results will change over time. How do we want to handle that from a change detection point of view?

projects/components/src/table/cells/data-renderers/time-ago/time-ago-table-cell-renderer.component.ts
```diff
@@ -10,7 +20,7 @@ import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
   styleUrls: ['./time-ago-table-cell-renderer.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="cell-value" [htTooltip]="this.value | htDisplayTimeAgo">
+    <div class="cell-value" [htTooltip]="this.value | htDisplayDate: this.dateFormat">
```
jake-bassett, line 23: Changed the tooltip back to timestamp. I think showing the underlying data is more important.

projects/common/src/utilities/formatters/time/display-time-ago.pipe.ts
aaron-steinfeld: Yeah I go back and forth here. For the column, I agree with you. When data reloads, we blow away the whole table and redraw anyway. For a generic pipe, that's where I hesitate. For example, if we were to use this in the refresh button (which actually does show "time ago"), it wouldn't update.

tl;dr - I'm ok leaving it pure as long as you comment about this

projects/common/src/time/time-duration.ts
```diff
@@ -26,6 +26,25 @@ export class TimeDuration {
...
+
+    if (secondsAgo < 59) {
+      return 'Just now';
+    }
+
+    const mostSignificantPortion = this.getMostSignificantUnitOnly();
+
+    const value = mostSignificantPortion.value;
+    const units = mostSignificantPortion.toUnitString();
+
```
aaron-steinfeld, line 40: line 38-41 are basically a reimplementation of `toLongString()` I believe. Rather than adding any code in time duration, I'd expect the timeAgo pipe to use duration (illustrated below). Putting this logic into time duration doesn't feel right because a duration has no reference to a point in time - it's only a length of time.

projects/common/src/utilities/formatters/time/display-time-ago.pipe.ts
jake-bassett: Comment added.

projects/common/src/utilities/formatters/time/display-time-ago.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+export class DisplayTimeAgo implements PipeTransform {
+  private readonly dateCoercer: DateCoercer = new DateCoercer();
+
+  public transform(value?: DateOrMillis | null): string {
+    if (value === null || value === undefined) {
+      return '-';
+    }
+
+    return new TimeDuration(this.calcSecondsAgo(value), TimeUnit.Second).toTimeAgoString();
```
aaron-steinfeld, line 23:
```suggestion
    const durationAgo = new TimeDuration(this.calcSecondsAgo(value), TimeUnit.Second);
    if (durationAgo.getAmountForUnit(TimeUnit.Seconds) >= 60) {
       return 'Just now';
     }
     
    return `${durationAgo.getMostSignificantUnitOnly().toLongString()} ago`;
```

aaron-steinfeld (approved):

PR #615, opened 2021-02-19
feat: adding a new summary values component

anandtiwary:
## Description
feat: adding a new summary values component 

### Testing

### Checklist:

### Documentation


projects/components/src/summary-values/summary-values.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  @Input()
+  public tooltip?: string;
+
+  public displayValue?: string;
+  public additionalValues?: string[];
+
+  public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (changes.values) {
+      this.displayValue = this.values && this.values?.length > 0 ? this.values[0] : undefined;
```
aaron-steinfeld, line 52:
```suggestion
      this.displayValue = this.values?.[0];
```

projects/components/src/summary-values/summary-values.component.ts
```diff
@@ -0,0 +1,56 @@
...
+        summaryValueDisplayStyle="${SummaryValueDisplayStyle.Text}"
+        class="summary-value"
+      ></ht-summary-value>
+      <ng-container *ngIf="this.additionalValues && this.additionalValues!.length > 0">
+        <div class="additional-values" [htTooltip]="detailsTemplate">
+          <span class="count">+{{ this.additionalValues!.length }}</span>
+        </div>
+        <ng-template #detailsTemplate>
+          <div class="tooltip-contents">
+            <span *ngFor="let value of this.additionalValues" class="value">
```
aaron-steinfeld, line 25: So the tooltip is all the values in sibling spans? Should these be comma separated or anything? If this looks like how I'm imagining, I'd think one tooltip should apply across the whole component listing all values together, rather than separating into two different tooltips.

projects/components/src/summary-values/summary-values.component.ts
```diff
@@ -0,0 +1,56 @@
...
+        [icon]="this.icon"
+        [value]="this.displayValue"
+        [label]="this.label"
+        [tooltip]="this.tooltip"
+        summaryValueDisplayStyle="${SummaryValueDisplayStyle.Text}"
+        class="summary-value"
+      ></ht-summary-value>
+      <ng-container *ngIf="this.additionalValues && this.additionalValues!.length > 0">
+        <div class="additional-values" [htTooltip]="detailsTemplate">
+          <span class="count">+{{ this.additionalValues!.length }}</span>
```
aaron-steinfeld, line 21: So this would be `+3` for example, if given 4 values? Not really sure what this should look like, can you add a screenshot? here for example, I'd expect it to say `+3 more`

projects/components/src/summary-value/summary-value.component.ts
```diff
@@ -41,7 +43,7 @@ export class SummaryValueComponent implements OnChanges {
     } else if (!isEmpty(this.value)) {
       this.tooltipText = this.value;
     } else {
-      this.tooltipText = undefined;
+      this.tooltipText = '-';
```
aaron-steinfeld, line 46: I'm not sure this belongs here. This behavior can easily be had by putting `value` through the pipe `htDisplayString`, but doing it this way is a breaking change for any places we rely on this not appearing for undefined.

projects/components/src/summary-values/summary-values.component.ts
anandtiwary: The mocks says +3 only. Please check the dependent PR for screenshot

projects/components/src/summary-value/summary-value.component.ts
anandtiwary: I think we are apply ngIf separately. But I will verify this again to be sure

projects/components/src/summary-values/summary-values.component.ts
anandtiwary: Check the screenshot :)

projects/components/src/summary-values/summary-values.component.ts
anandtiwary: This would fail i think? If this.values = [], wouldn't it throw error?

aaron-steinfeld: accessing an index that's not present will just returned undefined; there's no array out of bounds in javascript

projects/components/src/summary-values/summary-values.component.ts
aaron-steinfeld: I'll give you this one - the shading of the background conveys more, even if I prefer it to say it - but I think the single tooltip and keeping the default undefined behavior are good changes that don't affect your new use.

anandtiwary: I think the child summary value component should show '-' if the value is undefined. If we want to hide the entire component then that should be the responsibility of its parent component. Reason being, we may have use case where we are displaying a label and tooltip. In those case, displaying '-' instead of disappearing would be a good indication to the user.

projects/components/src/summary-value/summary-value.component.ts
anandtiwary: fine I will revert this

anandtiwary: reverted

projects/components/src/summary-values/summary-values.component.ts
anandtiwary: done

projects/components/src/summary-values/summary-values.component.ts
aaron-steinfeld: I don't think it's wrong for the summary value to do that, just that it's a change that's just as easy to do in the parent as the child. If you want to check all the usages to see if it makes sense though, I'm fine with it. Please combine the tooltip though - this should produce a single tooltip across all 3 elements.

anandtiwary: I have reverted the change for modifying summary value component. However, I am not sure about combining the tooltips. Mainly because, the tooltip on the first value can just represent the label. Like "IP Address" over the first value if we are not displaying the label itself. Combining the two also works, but we would need to define a display format for input tooltip and extra values. I think this is something we can tackle/iterate later on.

anandtiwary: would this work? I would prefer to merge this now and iterate on the other change. for now, it is matching the mocks.

aaron-steinfeld: Combining them just makes the tooltip work like the inputs. Multiple values have been provided, so they should not be split into separate tooltips. If I'm going to take the effort to hover with my mouse, I want to see the whole list as one unit. In my mind, it's analogous to hiding part of a text string ellipses - we don't only show the hidden part when hovering, we show the whole string.

anandtiwary: done.

aaron-steinfeld (approved):

PR #625, opened 2021-02-22
feat: support searching in multi select component

arjunlalb:
## Description
Support searching in multi select component.

### Testing
UTs added. 

### Checklist:


projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -104,6 +112,9 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
   @Input()
   public showBorder: boolean = false;
 
+  @Input()
+  public enableSearch: boolean = false;
```
aaron-steinfeld, line 116: I'm really questioning the need here - it seems to me that if we've got so many items in a multi-select that we need to search, we're probably doing something wrong.

arjunlalb: I think its a natural progression for a component like this. In cases where we might have a long list, we want to support search.

aaron-steinfeld: Discussed offline, yes there is

aaron-steinfeld (approved):

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -56,16 +57,23 @@ import { MultiSelectJustify } from './multi-select-justify';
...
               <div class="multi-select-option all-options" (click)="this.onAllSelectionChange()">
                 <input class="checkbox" type="checkbox" [checked]="this.areAllOptionsSelected()" />
                 <span class="label">Select All</span>
               </div>
-
-              <ht-divider></ht-divider>
             </ng-container>
 
-            <div *ngFor="let item of items" (click)="this.onSelectionChange(item)" class="multi-select-option">
+            <ht-divider *ngIf="this.showAllOptionControl || this.enableSearch"></ht-divider>
```
anandtiwary, line 74: The new mocks supports "Select All" And "Clear Selected" as buttons (instead of it as an option). Is this something you are taking care of too?

arjunlalb: Not doing it at the moment. Currently doing things that would unblock me. Can circle back to it in the end.

PR #629, opened 2021-02-23
Expander toggle - icon styling

DmitiryPotychkin:
Change color of icon "Collapsed" to $gray-7

### DIFF
--- a/projects/components/src/expander/expander-toggle.component.scss
+++ b/projects/components/src/expander/expander-toggle.component.scss
@@ -13,3 +13,11 @@
 .expander-toggle:hover {
   color: $gray-5;
 }
+
+.collapse {
+  color: $gray-7;
+
+  &:hover {
+    color: $gray-7;
+  }
+}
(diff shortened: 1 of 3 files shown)


projects/components/src/expander/expander.component.test.ts
```diff
@@ -43,6 +43,7 @@ describe('Expander component', () => {
     );
 
     expect(spectator.query('ht-icon')).toExist();
+    expect(spectator.query('collapse')).toExist();
```
aaron-steinfeld, line 46: probably looking for `.collapse`

projects/components/src/expander/expander-toggle.component.scss
```diff
@@ -13,3 +13,11 @@
 .expander-toggle:hover {
   color: $gray-5;
 }
+
+.collapse {
+  color: $gray-7;
```
aaron-steinfeld, line 18: So a different color when collapsed and no hover color? That seems off to me

PR #630, opened 2021-02-23
Tooltip directive - onChange 

DmitiryPotychkin:
- Remove tooltip popover at onChanges


projects/components/src/tooltip/tooltip.directive.ts
```diff
@@ -43,6 +43,12 @@ export class TooltipDirective implements OnDestroy {
     this.mouseLeave$.next(event);
   }
 
+  public ngOnChanges(): void {
+    if (this.popover) {
```
anandtiwary, line 47: If tooltip is already displayed and the mouse is already over the element, then let's say the tooltip content changes. In this case, I think the tooltip should update rather that getting removed.

aaron-steinfeld: Is there a bug behavior we're seeing? The motivation of this change isn't obvious. If it's to handle the situation anand mentioned, when content changes while a tooltip is being displayed, I think there's more to do here.

PR #631, opened 2021-02-23
Add back table control data source base

jake-bassett:
## Description
Add GraphqlTableControlOptionsDataSourceModel back since we are now ready to use it.

### DIFF
--- a/projects/components/src/table/controls/table-controls-api.ts
+++ b/projects/components/src/table/controls/table-controls-api.ts
@@ -35,10 +35,6 @@ export type TableControlOption<T = unknown> =
   | TableFilterControlOption<T>
   | TablePropertyControlOption<T>;
 
-export type LabeledTableControlOption = TableControlOption & {
-  label: string;
-};
-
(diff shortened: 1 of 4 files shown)


anandtiwary (dismissed):

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -35,6 +35,10 @@ export type TableControlOption<T = unknown> =
   | TableFilterControlOption<T>
   | TablePropertyControlOption<T>;
 
+export type LabeledTableControlOption = TableControlOption & {
```
aaron-steinfeld, line 38: Does it make sense to put this in the api? Removing the label from the api (since it's only used by dashboarding) was kind of the goal of that refactor

jake-bassett: Ah, I see what you were doing there now. I like it. I'll move it back. One sec.

aaron-steinfeld (approved):

PR #633, opened 2021-02-24
feat: adding a generic explore cartesian group by ds

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
@@ -1,25 +1,98 @@
-import { ColorService, forkJoinSafeEmpty, RequireBy } from '@hypertrace/common';
-import { GraphQlDataSourceModel, MetadataService } from '@hypertrace/distributed-tracing';
-import { Model } from '@hypertrace/hyperdash';
+import { GraphQlGroupBy } from './../../../../graphql/model/schema/groupby/graphql-group-by';
+import { GraphQlFilter } from '../../../../../../../distributed-tracing/src/shared/graphql/model/schema/filter/graphql-filter';
+import { ColorService, forkJoinSafeEmpty, RequireBy, TimeDuration } from '@hypertrace/common';
+import { GraphQlDataSourceModel, GraphQlSortBySpecification, MetadataService, Specification } from '@hypertrace/distributed-tracing';
+import {
+  ARRAY_PROPERTY,
+  Model,
+  ModelModelPropertyTypeInstance,
+  ModelProperty,
+  ModelPropertyType,
+  NUMBER_PROPERTY,
+  PLAIN_OBJECT_PROPERTY,
+  STRING_PROPERTY
+} from '@hypertrace/hyperdash';
 import { ModelInject } from '@hypertrace/hyperdash-angular';
 import { isEmpty } from 'lodash-es';
-import { EMPTY, Observable } from 'rxjs';
-import { map, switchMap } from 'rxjs/operators';
+import { Observable, of, NEVER } from 'rxjs';
+import { map, mergeMap } from 'rxjs/operators';
 import { Series } from '../../../../components/cartesian/chart';
-import { ExploreVisualizationRequest } from '../../../../components/explore-query-editor/explore-visualization-builder';
+import {
+  ExploreRequestState,
+  ExploreSeries
+} from '../../../../components/explore-query-editor/explore-visualization-builder';
 import { MetricTimeseriesInterval } from '../../../../graphql/model/metric/metric-timeseries';
 import { ExploreSpecification } from '../../../../graphql/model/schema/specifications/explore-specification';
 import {
   ExploreGraphQlQueryHandlerService,
+  EXPLORE_GQL_REQUEST,
+  GraphQlExploreRequest,
   GraphQlExploreResponse
 } from '../../../../graphql/request/handlers/explore/explore-graphql-query-handler.service';
 import { CartesianDataFetcher } from '../../../widgets/charts/cartesian-widget/cartesian-widget.model';
 import { ExploreResult } from './explore-result';
+import { ExploreSelectionSpecificationModel } from '../specifiers/explore-selection-specification.model';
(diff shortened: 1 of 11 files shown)


projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -1,25 +1,98 @@
...
+  // })
+  // public groupBySpecifications: Specification[] = [];
+
+  @ModelProperty({
+    key: 'series',
+    displayName: 'Explore Series',
+    required: false,
+    type: ARRAY_PROPERTY.type
+  })
+  public series: ExploreSeries[] = [];
```
anandtiwary, line 65: Need thoughts on this API. Goal here is to take group by specs and a metric spec as inputs and have a way to build cartesian series configuration that we can use with cartesian widget.

An example usage is Security events cartesian which needs to show three series for each `securityEventScore` (High, medium and low). The current change works for this use case. Problem comes up when we want to assign a specific color to each series. For this to happen, we need a way to specify the series config from the model json.

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -19,7 +32,51 @@ import { ExploreResult } from './explore-result';
   type: 'explore-cartesian-data-source'
 })
 export class ExploreCartesianDataSourceModel extends GraphQlDataSourceModel<CartesianDataFetcher<ExplorerData>> {
-  public request?: ExploreVisualizationRequest;
+  @ModelProperty({
```
anandtiwary, line 35: I think I am settling on this api now. I plan to create child datasources that would extend from this class and would contain the logic of converting the `CartesianDataFetcher<ExplorerData>` and `CartesianResult<TInterval>` to its specific series configuration info.

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -19,7 +32,51 @@ import { ExploreResult } from './explore-result';
...
+    key: 'group-by-include-rest',
+    displayName: 'Group By',
+    required: false,
+    type: BOOLEAN_PROPERTY.type
+  })
+  public groupByIncludeRest: boolean = true;
+
+  @ModelProperty({
+    key: 'order-by',
```
aaron-steinfeld, line 74: probably don't need order by

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -19,7 +32,51 @@ import { ExploreResult } from './explore-result';
...
+  @ModelProperty({
+    key: 'metric-explore-series',
+    displayName: 'Metric explore series',
+    type: ARRAY_PROPERTY.type,
+    required: true
+  })
+  public metricExploreSeries: ExploreSeries[] = [];
+
+  @ModelProperty({
+    key: 'group-by',
```
aaron-steinfeld, line 51: I'm confused - don't these each apply to individual series? But when grouping we also don't know the series until the results are retrieved, which is why we've never made a data source for them other than the one dynamically populated with a result object

aaron-steinfeld (commented):
As a starting point, can you give an example of what the json might look like for a cartesian widget using explorer data?

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
anandtiwary: Yeah. So if we have group by, one series will be generated for each group. I am trying to make a DS for such a case.

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
anandtiwary: won't it be useful to order the various groups?

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
anandtiwary: Going to convert this into an abstract class

projects/observability/src/shared/dashboard/data/graphql/explorer-visualization/explorer-visualization-cartesian-data-source.model.ts
```diff
@@ -0,0 +1,28 @@
...
+export class ExplorerVisualizationCartesianDataSourceModel extends ExploreCartesianDataSourceModel {
+  public request?: ExploreVisualizationRequest;
+
+  protected buildRequestState(interval?: TimeDuration | 'AUTO'): ExploreRequestState | undefined {
+    if (this.request?.series.length === 0 || this.request?.context === undefined) {
+      return undefined;
+    }
+
+    return {
+      series: this.request.series,
```
aaron-steinfeld, line 21: So previously, from what I recall, the model read off the request stream producing new data whenever the request changed - saving us from having to rebuild the dashboard on every explorer query change.

This is why the previous implementation piped off `this.request.exploreQuery$` rather than reading the current state. Does the change, as implemented, still respond to changes in the explorer query as the user interacts with explorer? And if so, is the dashboard being rebuilt each time (maybe it always was?)?

anandtiwary: Yeah. this was confusing. I have updated this in the latest commit.

aaron-steinfeld: ok, so table should be fine with getData emitting new data over time.

aaron-steinfeld (approved):
I'm sure it's great

anandtiwary:
thank you sir!

PR #642, opened 2021-02-26
feat: Add explore table cell parser

anandtiwary:
## Description
feat: Add explore table cell parser

### Testing

### Checklist:

### Documentation


projects/observability/src/shared/components/table/data-cell/explore/explore-table-cell-parser.ts
```diff
@@ -0,0 +1,45 @@
+import { TableCellParser, TableCellParserBase } from '@hypertrace/components';
+import { MetricAggregation } from '@hypertrace/distributed-tracing';
+import { ObservabilityTableCellType } from '../../observability-table-cell-type';
+
+@TableCellParser({
+  type: ObservabilityTableCellType.ExploreValue
+})
+export class ExploreValueTableCellParser extends TableCellParserBase<number, CellValue, CellValue> {
```
aaron-steinfeld, line 8: Not sure why these need their own parser. Explorer can return different data types potentially depending on the query, so there's no real one size fits all parser. I would expect you'd pick a text or metric renderer depending on you query, and use the parser associated with that. If those need modifications to support an explorer result (though I doubt it, looking at what this is doing), that may be a better approach.

anandtiwary: It returns ExploreValue which is an object. So all existing cell renderers fail to work.

aaron-steinfeld: They should support that:
Metric parser, for example - https://github.com/hypertrace/hypertrace-ui/blob/a15f0f42e904cfd8f1b5fc2b9ab23c383a47349a/projects/distributed-tracing/src/shared/components/table/data-cell/metric/metric-table-cell-parser.ts#L28

If there's a specific parser you're trying to use that doesn't, we should update there

anandtiwary: Metric parser yes. But it also converts value into number. Since we have group by, we would want to access string values as well

aaron-steinfeld: That was the point I was attempting to make - if we want to display as text, we should use the text parser (since the parser is picked by the cell renderer, anyway). So we should update the text parser to handle objects in the same way some of the other parsers already do, by falling back to `input.value` if it's given an object instead of a string, and then for string specifically, finally falling back to `String(input)`

anandtiwary:
I think I messed up something. https://github.com/hypertrace/hypertrace-ui/pull/644 has this change and also adds a row parser. Closing this PR

PR #644, opened 2021-02-26
feat: Add row parser

anandtiwary:
## Description
feat: Add row parser

### Testing

### Checklist:

### Documentation


projects/components/src/table/cells/data-parsers/table-cell-row-parser.ts
```diff
@@ -0,0 +1,21 @@
+import { TableRow } from '../../table-api';
+import { TableCellParser } from '../table-cell-parser';
+import { TableCellParserBase } from '../table-cell-parser-base';
+import { CoreTableCellParserType } from '../types/core-table-cell-parser-type';
+
+@TableCellParser({
+  type: CoreTableCellParserType.Row
+})
+export class TableCellRowParser extends TableCellParserBase<unknown, TableRow, unknown> {
+  public parseValue(_: unknown, rowData: TableRow): TableRow {
```
aaron-steinfeld, line 10: Given that the renderers have access to both the row and cell, do we need a row parser? Would `TableCellNoOpParser` serve this use case?

projects/components/src/table/cells/data-parsers/table-cell-string-parser.ts
```diff
@@ -1,16 +1,35 @@
...
 
-  public parseFilterValue(cellData: string): string {
+  private extractValue(cellData: CellData): CellValue {
+    switch (typeof cellData) {
+      case 'string':
+        return cellData;
+      case 'object':
+        return cellData?.value ?? undefined;
+      default:
+        return undefined;
```
aaron-steinfeld, line 21: I think it would be safer to make this return cellData in cases other than `cellData?.value` - that way, it should be more or less backwards compatible. Here, if it previously returned a number for example, it would have been correctly rendered as text. Now it would be turned to undefined. Alternatively, we could use `String()` to stringify values (although I'd still pass undefined and null through for the renderer to deal with)

projects/components/src/table/cells/data-parsers/table-cell-row-parser.ts
anandtiwary: But it wouldn't assign rowData to value. I don't need to use value property, but this is how it is designed

projects/components/src/table/cells/data-parsers/table-cell-string-parser.ts
anandtiwary: makes sense

projects/components/src/table/cells/data-parsers/table-cell-row-parser.ts
anandtiwary: Lemme know. I wouldn't it be weird for the design? I can use fields from constructor, but i hope it doesn't break any other use case

aaron-steinfeld: It shouldn't. All the base class does is set some values, but the renderer ultimately can use whatever it wants. If we have a renderer that's doing something at the row level, it's likely custom anyway

aaron-steinfeld (approved):

PR #652, opened 2021-03-01
fix: some styling issues

jake-bassett:
## Description
Fix some styling issues.

### DIFF
--- a/projects/components/src/select/select.component.scss
+++ b/projects/components/src/select/select.component.scss
@@ -17,7 +17,7 @@
   display: flex;
   align-items: center;
   width: 100%;
-  height: 34px;
+  height: 36px;
 
   &.small {
     height: 32px;
--- a/projects/components/src/table/controls/table-controls.component.scss
+++ b/projects/components/src/table/controls/table-controls.component.scss
@@ -15,7 +15,7 @@
     display: flex;
 
     .search-box {
-      width: 256px;
+      min-width: 256px;
     }
   }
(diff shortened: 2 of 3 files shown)


projects/components/src/table/controls/table-controls.component.scss
```diff
@@ -15,7 +15,7 @@
     display: flex;
 
     .search-box {
-      width: 256px;
+      min-width: 256px;
```
aaron-steinfeld, line 18: Why min width here? Do we want it to grow on a table?

projects/components/src/select/select.component.scss
```diff
@@ -17,7 +17,7 @@
   display: flex;
   align-items: center;
   width: 100%;
-  height: 34px;
+  height: 36px;
```
aaron-steinfeld, line 20: This seems like it could have more far reaching affects. have we checked all of the current selects that they're not being messed up by resizing?

jake-bassett: I spot checked several. 36px is correct and showing in the checked locations. I see no issues with overflow or misalignment.

projects/components/src/table/controls/table-controls.component.scss
jake-bassett: I noticed the search box shrank in width, I believe from some parent flex-boxing that went in. Changing to min-width fixed the issue.

PR #654, opened 2021-03-02
Modal and textarea styling

jake-bassett:
## Description
* Update styling for textarea (first usage)
* Add taller large modal size

### DIFF
--- a/projects/components/src/modal/modal-container.component.scss
+++ b/projects/components/src/modal/modal-container.component.scss
@@ -26,6 +26,11 @@
     width: 640px;
   }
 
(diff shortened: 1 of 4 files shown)


anandtiwary (approved):

projects/components/src/modal/modal-container.component.scss
```diff
@@ -26,6 +26,11 @@
     width: 640px;
   }
 
+  &.modal-size-large-tall {
```
anandtiwary, line 29: We probably need to include Response size. We have done it for sheet and it would be useful to add it here. It is straightforward.

PR #655, opened 2021-03-03
feat: string array cell renderer

arjunlalb:
## Description
Cell renderer for string arrays.

Expected behaviour:
Should show the first item in the cell, along with a `+x` representing the number of remaining values. Tooltip should show a list of these values on hovering on `+x`.

Screenshot 2021-03-03 at 2 02 22 PM

Overflow
Screenshot 2021-03-03 at 8 26 23 PM

### Testing
Manually verified. Added UTs.

### Checklist:

### DIFF
--- a/projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
+++ b/projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
@@ -19,12 +19,12 @@ import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
   styleUrls: ['./string-array-table-cell-renderer.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="string-array-cell">
+    <div class="string-array-cell" [htTooltip]="this.summaryTooltip">
       <span class="first-item">{{ this.firstItem }}</span>
-      <span class="summary-text" [htTooltip]="this.summaryTooltip">{{ this.summaryText }}</span>
+      <span class="summary-text">{{ this.summaryText }}</span>
 
       <ng-template #summaryTooltipElement>
-        <div *ngFor="let value of this.value.slice(1, this.value.length)">{{ value }}</div>
+        <div *ngFor="let value of this.value">{{ value }}</div>
       </ng-template>
     </div>
   `
--- a/projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
+++ b/projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
@@ -1,0 +1,60 @@
+import { ChangeDetectionStrategy, Component, Inject, OnInit, TemplateRef, ViewChild } from '@angular/core';
+import { TableColumnConfig, TableRow } from '../../../table-api';
+import {
+  TABLE_CELL_DATA,
+  TABLE_COLUMN_CONFIG,
+  TABLE_COLUMN_INDEX,
+  TABLE_DATA_PARSER,
+  TABLE_ROW_DATA
+} from '../../table-cell-injection';
+import { TableCellParserBase } from '../../table-cell-parser-base';
+import { TableCellRenderer } from '../../table-cell-renderer';
+import { TableCellRendererBase } from '../../table-cell-renderer-base';
+import { CoreTableCellParserType } from '../../types/core-table-cell-parser-type';
+import { CoreTableCellRendererType } from '../../types/core-table-cell-renderer-type';
+import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
+
+@Component({
+  selector: 'ht-string-array-table-cell-renderer',
+  styleUrls: ['./string-array-table-cell-renderer.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="string-array-cell">
+      <span class="first-item">{{ this.firstItem }}</span>
+      <span class="summary-text" [htTooltip]="this.summaryTooltip">{{ this.summaryText }}</span>
+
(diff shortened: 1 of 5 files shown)


projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  selector: 'ht-string-array-table-cell-renderer',
+  styleUrls: ['./string-array-table-cell-renderer.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="string-array-cell">
+      <span class="first-item">{{ this.firstItem }}</span>
+      <span class="summary-text" [htTooltip]="this.summaryTooltip">{{ this.summaryText }}</span>
+
+      <ng-template #summaryTooltipElement>
+        <div *ngFor="let value of this.value.slice(1, this.value.length)">{{ value }}</div>
```
aaron-steinfeld, line 27: will using a function to get the array here cause this to continually re-render? Seems easy enough to avoid

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,60 @@
...
+import { CoreTableCellRendererType } from '../../types/core-table-cell-renderer-type';
+import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
+
+@Component({
+  selector: 'ht-string-array-table-cell-renderer',
+  styleUrls: ['./string-array-table-cell-renderer.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="string-array-cell">
+      <span class="first-item">{{ this.firstItem }}</span>
```
aaron-steinfeld, line 23: How's this handle overflow?

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,60 @@
...
+      <span class="summary-text" [htTooltip]="this.summaryTooltip">{{ this.summaryText }}</span>
```
aaron-steinfeld, line 24: So I think the tooltip should be across the whole cell, and show all values. This actually just came up when @anandtiwary implemented basically exactly this component... can we share an underlying component? let me go find it

aaron-steinfeld (commented):
Please investigate how we can share this pattern with https://github.com/hypertrace/hypertrace-ui/pull/615 (either by using that component inside this renderer, or pulling a base component out of the two of them

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
aaron-steinfeld: Linked in the summary comment

arjunlalb: @aaron-steinfeld the link you used is for this PR. :)

aaron-steinfeld: I don't think so? Linked to #615, this is #655 no matter how similar they might look :)

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
arjunlalb: Handled it. Updated description with screenshot.

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
arjunlalb: done.

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
arjunlalb: Sorry. My browser was not redirecting when I clicked on the link. So I assumed its this page. I'll check that PR.

arjunlalb: @aaron-steinfeld Checked that PR. I'm not sure if we should use that here. #655 adds them as summary values which has a certain format. If there are no icons present, it shows a dot. To circumvent that usecase we'll have to modify the summary value component's behaviour. Also override fonts etc. I don't think we should do that just to reuse it. Our usecase here is plain string array display. 

We cannot take this out into a component and have it used there either because that component is built for summary values usecase. The API is more complex with optional icons, labels etc.

One thing we can do is to write a wrapper which covers both cases that'd use summary value / plain text based on requirement. But then we'd be writing a component with a pretty confusing/complicated API to share one template and couple lines of code.

aaron-steinfeld: OK - definitely not worth maintaining that level of difference in one component. I just wish we'd stop using similar but slightly unique components everywhere...

aaron-steinfeld (approved):

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
anandtiwary: We should really consolidate the two styles. I think per michael it is the same component. This would lead to inconsistency, right?

arjunlalb: I'll file a tech debt for this.

PR #656, opened 2021-03-04
feat: email validator and a string formatter

arjunlalb:
## Description
Validator for email.
String formatter to split a comma separated list into a string array.

### Testing
UTs added.

### Checklist:

### DIFF
--- a/projects/common/src/utilities/formatters/string/string-formatter.test.ts
+++ b/projects/common/src/utilities/formatters/string/string-formatter.test.ts
@@ -1,4 +1,9 @@
-import { displayString, titleCaseFromKebabCase, titleCaseFromSnakeCase } from './string-formatter';
+import {
+  displayString,
+  getStringsFromCommaSeparatedList,
+  titleCaseFromKebabCase,
+  titleCaseFromSnakeCase
+} from './string-formatter';
 
 describe('String formatter', () => {
   test('can convert from kebab case to title case', () => {
@@ -25,5 +30,13 @@ describe('String formatter', () => {
     expect(displayString(() => 'hi')).toBe('Function');
     expect(displayString(Symbol('test symbol'))).toBe('Symbol(test symbol)');
     expect(displayString(false)).toBe('false');
+  });
+
+  test('creates string array correctly from comma separated list', () => {
+    expect(getStringsFromCommaSeparatedList('first,second,   third   ')).toEqual(
+      expect.arrayContaining(['first', 'second', 'third'])
+    );
+    expect(getStringsFromCommaSeparatedList('first,second,   ')).toEqual(expect.arrayContaining(['first', 'second']));
+    expect(getStringsFromCommaSeparatedList('first')).toEqual(expect.arrayContaining(['first']));
   });
 });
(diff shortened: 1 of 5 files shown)


projects/common/src/utilities/formatters/string/string-formatter.ts
```diff
@@ -42,3 +42,10 @@ export const displayString = (provided?: unknown): string => {
 export const collapseWhitespace = (str: string): string =>
   // Replace all whitespace with a single space
   str.replace(/\s\s+/g, ' ');
+
+export const getStringsFromCommaSeparatedList = (text: string): string[] =>
+  text
+    .trim()
```
aaron-steinfeld, line 48: nit: this is redundant, would function the same without since we're trimming the components

projects/common/src/utilities/formatters/string/string-formatter.test.ts
```diff
@@ -26,4 +31,12 @@ describe('String formatter', () => {
     expect(displayString(Symbol('test symbol'))).toBe('Symbol(test symbol)');
     expect(displayString(false)).toBe('false');
   });
+
+  test('creates string array correctly from comma separated list', () => {
+    expect(getStringsFromCommaSeparatedList('first,second,   third   ')).toEqual(
+      expect.arrayContaining(['first', 'second', 'third'])
```
aaron-steinfeld, line 37: use a regular equality match in these tests, since we're looking for this to return a specific array (i.e. it would be wrong if it returned `['first', 'second', 'third', 'fourth']` but this matcher would accept it.
```suggestion
      ['first', 'second', 'third']
```

projects/common/src/utilities/validators/email-validator.ts
```diff
@@ -0,0 +1,10 @@
+export const isEmailAddressValid = (emailAddress: string) => {
+  // Regex as per W3C spec for email field
+  // Source: https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
```
aaron-steinfeld, line 3: That's a hefty regex. Email validation is always tough since it's been described across various RFCs and there's inconsistencies across some providers, but let's go with it for now. In prior lives, I remember we've stripped down email validation more and more with weird cases in the wild.

arjunlalb: Yea. Even this one does not cover all the cases. But the w3c spec seemed like a good enough sample.

projects/common/src/utilities/formatters/string/string-formatter.test.ts
arjunlalb: done

projects/common/src/utilities/formatters/string/string-formatter.ts
arjunlalb: done

projects/common/src/utilities/validators/email-validator.ts
aaron-steinfeld: Is it feasible to use the built in angular one? https://angular.io/api/forms/Validators#email looks like it's based on the same thing

aaron-steinfeld: It is not

aaron-steinfeld (approved):

anandtiwary:
We can also look into https://github.com/ngneat/forms-manager . @arjunlalb Could you please check if it can simplify your use case?

PR #659, opened 2021-03-06
feat: mutation tracker service

arjunlalb:
## Description
Mutation tracker service to facilitate better lower level cache invalidation.

<-- Will fill the sections below when the PR is out of draft mode -->

### Testing
Manually tested.

### Checklist:

### DIFF
--- a/projects/graphql-client/src/graphql-config.ts
+++ b/projects/graphql-client/src/graphql-config.ts
@@ -30,6 +30,7 @@ export interface GraphQlQueryHandler<TRequest, TResponse> extends GraphQlHandler
 
 export interface GraphQlMutationHandler<TRequest, TResponse> extends GraphQlHandler<TRequest, TResponse> {
   readonly type: GraphQlHandlerType.Mutation;
+  reportMutation?(request: TRequest): void;
 }
 
 export interface GraphQlRequestOptions {
--- a/projects/graphql-client/src/mutation-tracker/mutation-tracker.service.ts
+++ b/projects/graphql-client/src/mutation-tracker/mutation-tracker.service.ts
@@ -3,46 +3,73 @@ import { Injectable } from '@angular/core';
 @Injectable({ providedIn: 'root' })
 export class MutationTrackerService {
   /**
-   This tracker service assumes that a mutation affects two types of queries.
-   1. Query by ID for the mutated object.
-   2. List query that contains the mutated object.
-
-   A separate mutation type is introduced so that the mutation handler and query handler can stay independent.
-
-   Currently does not support refreshing composite nested queries.
+   * This tracker service assumes that a mutation affects two types of queries.
+   *  1. Query for mutated object by ID.
+   *  2. List query that contains the mutated object.
+   *
+   * - Correlation between the object/list is made via the MutationType construct.
+   * - Both query/mutation handlers should use the same key for MutationType.
+   * - Expects requestTypeOrId to be a Symbol for requests from list queries and string (ID) for individual object queries.
+   * - Can be extended to support complex nested query use cases later.
    */
 
-  private mutationTypesAffected: Set<MutationType> = new Set<MutationType>();
-  private objectsAffected: Set<string> = new Set<string>();
+  // Set of objects that has been mutated.
+  private mutatedObjects: Set<MutationType> = new Set<MutationType>();
 
-  // When an action is marked as executed, clear the set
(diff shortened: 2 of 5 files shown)


projects/graphql-client/src/graphql-config.ts
```diff
@@ -30,6 +30,7 @@ export interface GraphQlQueryHandler<TRequest, TResponse> extends GraphQlHandler
 
 export interface GraphQlMutationHandler<TRequest, TResponse> extends GraphQlHandler<TRequest, TResponse> {
   readonly type: GraphQlHandlerType.Mutation;
+  reportMutation?(request: TRequest): void;
```
aaron-steinfeld, line 33: Do we need to even modify the mutation handlers? The main request service knows every time it invokes a mutation, right?

projects/graphql-client/src/mutation-tracker/mutation-tracker.service.ts
```diff
@@ -0,0 +1,75 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({ providedIn: 'root' })
+export class MutationTrackerService {
+  /**
+   * This tracker service assumes that a mutation affects two types of queries.
```
aaron-steinfeld, line 6: So some (if not all) of this is certainly part of apollo, but I'm not sure how configurable/easy to use that is - did you explore that path (and if so, what were the issues)?

projects/graphql-client/src/mutation-tracker/mutation-tracker.service.ts
```diff
@@ -0,0 +1,75 @@
...
+   *  1. Query for mutated object by ID.
+   *  2. List query that contains the mutated object.
+   *
+   * - Correlation between the object/list is made via the MutationType construct.
```
aaron-steinfeld, line 10: nit: naming - it's really just the object type, right? Wouldn't this just use the typename from gql? That would probably work in some cases - likely update (which usually takes in the same type as a query returns). Delete often just takes an ID though.

projects/graphql-client/src/mutation-tracker/mutation-tracker.service.ts
```diff
@@ -0,0 +1,75 @@
...
+   * - Both query/mutation handlers should use the same key for MutationType.
+   * - Expects requestTypeOrId to be a Symbol for requests from list queries and string (ID) for individual object queries.
+   * - Can be extended to support complex nested query use cases later.
+   */
+
+  // Set of objects that has been mutated.
+  private mutatedObjects: Set<MutationType> = new Set<MutationType>();
```
aaron-steinfeld, line 17: Seems representable by the map keys. Is this holding some other info?

aaron-steinfeld (commented):
Overall, I like the goal + effort here - just wondering if we can bury it in even deeper

projects/graphql-client/src/mutation-tracker/mutation-tracker.service.ts
arjunlalb: I was trying to get this done last day and seems like this service approach is the better solution.

We have 3 types of queries in our system.
1. ID based queries
2. List queries
3. Root queries (have no ID at all. configs)

I built a solution where we contain the extend of changes to the mutation method. Apollo lets us have an update call back executed after a mutation, using this we can clear cache entries for the objects that were mutated. This works for ID based queries and a modified version of it can work for root queries. However, this approach leads to inconsistent data for list queries. After mutating an object, if we delete its entry, the list query will be unaware of the change and will continue to use the cached list. This means the entry we just updated will be missing from the list until an explicit over the network query is made. So, this solution does not really work well unless we deploy our own custom logic for identifying which entry is associated with which queries and so on.

Given that, I feel like this tracker service approach is safer and cleaner. It doesn't modify any data directly. 

We could explore how to push down the markAsConsumed part down to the graphql service.

Thoughts? 

@aaron-steinfeld

aaron-steinfeld: I think the main goal is to keep any real logic out of the handlers. If mutation handlers have an abstract method/field they need to implement to say what they affect, and query handlers on what they read, then any cache decisions or evictions can happen at the service, as you indicated. Later, if we find a way to push that logic into apollo, it's a local impl change to that lower level service. I agree there's likely no way to void handlers providing those two pieces of data given how some apis are structured.

PR #665, opened 2021-03-08
feat: adding size option to toggle switch components

palbizu:
## Description

Added an optional input to modify the default styles of the toggle switch that were defined as small. The styles were encapsulated in relation to the new class.

### Testing

Visual Verification.

image

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/toggle-switch/toggle-switch.component.ts
+++ b/projects/components/src/toggle-switch/toggle-switch.component.ts
@@ -10,7 +10,7 @@ import { ToggleSwitchSize } from './toggle-switch-size';
     <mat-slide-toggle
       color="primary"
       [(ngModel)]="this.checked"
-      [ngClass]="{ 'small-slide-toogle': this.size === '${ToggleSwitchSize.Small}' }"
+      [ngClass]="{ 'small-slide-toogle': this.size === "${ToggleSwitchSize.Small}" }"
       [disabled]="this.disabled"
       (change)="this.onToggle($event)"
     >
--- a/projects/components/src/toggle-switch/toggle-switch.component.ts
+++ b/projects/components/src/toggle-switch/toggle-switch.component.ts
@@ -10,7 +10,7 @@ import { ToggleSwitchSize } from './toggle-switch-size';
     <mat-slide-toggle
       color="primary"
       [(ngModel)]="this.checked"
-      [ngClass]="{ 'small-slide-toogle': this.size === ${ToggleSwitchSize.Small} }"
+      [ngClass]="{ 'small-slide-toogle': this.size === '${ToggleSwitchSize.Small}' }"
       [disabled]="this.disabled"
       (change)="this.onToggle($event)"
     >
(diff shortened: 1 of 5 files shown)


projects/components/src/toggle-switch/toggle-switch.component.ts
```diff
@@ -26,6 +27,9 @@ export class ToggleSwitchComponent {
   @Input()
   public disabled?: boolean;
 
+  @Input()
+  public size?: string = 'small';
```
anandtiwary, line 31: So default it "small". Do we know we still have to support medium(now or in future)?

Also, if we are taking size as input, better to use `const enum` as we can type it to the two specific options. Example in Icon component.

arjunlalb: Since the small use case is new, I'd say lets default to medium and use small as the override.

palbizu: Ok. I've just set medium as default and small as the new case. I also added the enum.

projects/components/src/toggle-switch/toggle-switch.component.ts
```diff
@@ -9,6 +10,7 @@ import { MatSlideToggleChange } from '@angular/material/slide-toggle';
     <mat-slide-toggle
       color="primary"
       [(ngModel)]="this.checked"
+      [ngClass]="{ 'small-slide-toogle': this.size === ${ToggleSwitchSize.Small} }"
```
aaron-steinfeld, line 13: Is it feasible to apply a size class either way (`small-slide-toggle` or `medium-slide-toggle`? That way, we can update the scss such that small styles don't just apply as overrides to the medium styles. The overriding behavior makes the css harder to understand and more error prone, in my opion.

palbizu: By default, the slide-toggle has a medium size, for that reason, I'm overriding the styles. Tell me a bit more about your approach, I think that I don't understand your point.

aaron-steinfeld: > By default, the slide-toggle has a medium size

So in my mind, that knowledge of the default should not extend beyond the line `public size?: string = 'small';`. If we were to change that to default to medium, that shouldn't require css changes (currently it wouldn't _require_ them, but the css wouldn't make much sense having the override behavior it does).

Regardless of the default though, the reason is just to have symmetry in the implementation for maintainability. A code-level equivalent of this (with the same issue) would be to say create a `Shape` class, which returns an area of `length * width`, a `Rectangle` subclass which doesn't override area, and a `Circle` subclass which does (to `pi*r^2`). Similarly, there it would make sense to make area abstract and have each shape implement it itself.

aaron-steinfeld (dismissed):
discussed offline - explicit medium style means inlining all the material default width/height values (since that's what our medium corresponds to). That's still my preference, but that makes it more debatable so approving.

aaron-steinfeld (approved):

PR #668, opened 2021-03-09
refactor: fixing table css so that it can work inside flex container

anandtiwary:
## Description
Please include a summary of the change, motivation and context.
refactor: fixing table css so that it can work inside flex container

### Testing
Please describe the tests that you ran to verify your changes. Please summarize what did you test and what needs to be tested e.g. deployed and tested helm chart locally. 

### Checklist:
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

### Documentation
Make sure that you have documented corresponding changes in this repository or [hypertrace docs repo](https://github.com/hypertrace/hypertrace-docs-website) if required.

PR #671, opened 2021-03-09
Revert "refactor: fixing table css so that it can work inside flex container"

anandtiwary:
Reverts hypertrace/hypertrace-ui#668


aaron-steinfeld (approved):
Take it that it didn't work?

anandtiwary:
It had create another PR to fix a side effect. I later realized it is not worth the time to make further changes. There is an alternative which is to set `min-height:0` in the parent container (at the usage side).

PR #679, opened 2021-03-10
fix: add padding to the topology

palbizu:
## Description

Adding padding between legend and the top nodes. 

### Testing
Visual Testing.

### Checklist:

### Documentation


projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -167,7 +167,8 @@ export class D3Topology implements Topology {
       .append('svg')
       .classed(D3Topology.SVG_CLASS, true)
       .attr('width', '100%')
-      .attr('height', '100%');
+      .attr('height', '100%')
+      .style('padding-top', '24px');
```
aaron-steinfeld, line 171: Is this padding we want everywhere? We use this both as a full screen and embedded widget in different places, so may make more sense to put it on the specific usage

palbizu: Yes, it applies everywhere.

aaron-steinfeld: I think this will end up just looking weird - especially on topology when nodes are draggable, adding padding here means nodes will magically cut off. 
image

So I guess I'm not sure what we're trying to accomplish with the padding

palbizu: The approach was changed.

aaron-steinfeld (approved):

PR #680, opened 2021-03-10
feat: add large short modal size and fix cloud colors

jake-bassett:
## Description
New modal size for LargeShort.

### DIFF
--- a/projects/components/src/modal/modal.ts
+++ b/projects/components/src/modal/modal.ts
@@ -13,6 +13,7 @@ export interface ModalConfig<TData = unknown> {
 export const enum ModalSize {
   Small = 'small',
   Medium = 'medium',
+  LargeShort = 'large-short',
   Large = 'large',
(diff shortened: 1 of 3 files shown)


projects/components/src/modal/modal.ts
```diff
@@ -13,6 +13,7 @@ export interface ModalConfig<TData = unknown> {
 export const enum ModalSize {
   Small = 'small',
   Medium = 'medium',
+  LargeShort = 'large-short',
```
aaron-steinfeld, line 16: hah... I don't get why we bother

aaron-steinfeld (approved):

projects/components/src/modal/modal.ts
jake-bassett: Yeah...

I almost think we should have size inputs for height and width. Then we can just mix and match. It will keep us semi-consistent at least.

PR #684, opened 2021-03-11
fix: remove host max width for select component

arjunlalb:
## Description
Remove host max width for select component.
Containers of the component should be able to control the width.


aaron-steinfeld (approved):
Agree with this change, please check select usages before merging though.

PR #686, opened 2021-03-12
fix: update display string conversion logic

arjunlalb:
## Description
Undefined / null should return '-'.
Blank string should be rendered as is.

### Testing
UT updated.

### Checklist:

### DIFF
--- a/projects/common/src/utilities/formatters/string/string-formatter.ts
+++ b/projects/common/src/utilities/formatters/string/string-formatter.ts
@@ -17,11 +17,7 @@ export const displayString = (provided?: unknown): string => {
 
   switch (typeof provided) {
     case 'object':
-      return Array.isArray(provided)
-        ? `[${provided.map(displayString).join(', ')}]`
-        : provided === null
-        ? 'Unknown'
-        : 'Object';
+      return Array.isArray(provided) ? `[${provided.map(displayString).join(', ')}]` : 'Object';
     case 'undefined':
       return '-';
     case 'function':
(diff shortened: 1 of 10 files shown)


projects/common/src/utilities/formatters/string/string-formatter.ts
```diff
@@ -25,11 +23,11 @@ export const displayString = (provided?: unknown): string => {
         ? 'Unknown'
```
aaron-steinfeld, line 23: a stray unknown (didn't show up in tests because this is unreachable - nullness is checked on line 14 so we can remove this nested ternary.

arjunlalb: done

projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
```diff
@@ -51,7 +51,7 @@ export class StringArrayTableCellRendererComponent extends TableCellRendererBase
   public ngOnInit(): void {
     super.ngOnInit();
 
-    this.firstItem = this.value[0] ?? '-';
+    this.firstItem = this.value[0] ?? '';
```
aaron-steinfeld, line 54: if this is null, it should still be `-` right? But better, could we just use display string pipe in the template here?

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):

PR #698, opened 2021-03-16
refactor: use new group limit in explorer

aaron-steinfeld:
## Description
Uses the new group limit parameter in explorer api instead of the deprecated row limit behavior.

### Testing
Visual verification, updated tests

### Checklist:

### DIFF
--- a/projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
@@ -63,7 +63,7 @@ export abstract class ExploreCartesianDataSourceModel extends GraphQlDataSourceM
       requestType: EXPLORE_GQL_REQUEST,
       selections: requestState.series.map(series => series.specification),
       context: requestState.context,
-      limit: requestState.groupByLimit ?? 100,
+      limit: requestState.resultLimit,
       timeRange: this.getTimeRangeOrThrow(),
       interval: requestState.interval as TimeDuration,
       filters: filters,
--- a/projects/observability/src/shared/dashboard/data/graphql/explorer-visualization/explorer-visualization-cartesian-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/explorer-visualization/explorer-visualization-cartesian-data-source.model.ts
@@ -1,4 +1,3 @@
-import { TimeDuration } from '@hypertrace/common';
 import { GraphQlFilter } from '@hypertrace/distributed-tracing';
 import { Model } from '@hypertrace/hyperdash';
 import { NEVER, Observable } from 'rxjs';
@@ -19,10 +18,8 @@ import { ExploreCartesianDataSourceModel, ExplorerData } from '../explore/explor
 export class ExplorerVisualizationCartesianDataSourceModel extends ExploreCartesianDataSourceModel {
   public request?: ExploreVisualizationRequest;
 
-  protected fetchResults(interval: TimeDuration | 'AUTO'): Observable<CartesianResult<ExplorerData>> {
-    const requestState = this.buildRequestState(interval);
-
-    if (this.request === undefined || requestState === undefined) {
+  protected fetchResults(): Observable<CartesianResult<ExplorerData>> {
+    if (this.request === undefined) {
       return NEVER;
     }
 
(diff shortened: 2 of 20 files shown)


projects/observability/src/shared/dashboard/data/graphql/explorer-visualization/explorer-visualization-cartesian-data-source.model.ts
```diff
@@ -46,17 +43,8 @@ export class ExplorerVisualizationCartesianDataSourceModel extends ExploreCartes
...
-
-    return {
-      series: this.request.series,
-      context: this.request.context,
-      interval: interval,
-      groupBy: this.request.groupBy,
-      groupByLimit: this.request.groupByLimit
-    };
+  protected buildRequestState(): ExploreRequestState | undefined {
+    // Unused since fetchResults is overriden, but abstract so requires a def
```
aaron-steinfeld, line 47: This one's a bit weird. This is an abstract method used in the parent's fetchResults which we override in this class. It's not really relevant to this implementation since fetchResults uses an observable of state rather than building its own.

This was actually breaking the explorer, (although no one noticed) since the previous implementation was mapping the response data based on the initial state, not the current one.

anandtiwary: Yeah. AFAIR, the `ExploreVisualizationRequest` did not have all the latest request data. I think we emit on `exploreQuery$: Observable<TimeUnaware<GraphQlExploreRequest>>;` with latest interval data (possibly more). When i tested it last, using `ExploreVisualizationRequest` directly did break few use cases for me. So for correctness, I kept it the same way for this data source (but fixed the dependency in the parent data source).

anandtiwary: If this is working, then you don't need to override `fetchResults`. Just `return this.request` inside `buildRequestState`

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -63,7 +63,7 @@ export abstract class ExploreCartesianDataSourceModel extends GraphQlDataSourceM
       requestType: EXPLORE_GQL_REQUEST,
       selections: requestState.series.map(series => series.specification),
       context: requestState.context,
-      limit: requestState.groupByLimit ?? 100,
+      limit: requestState.resultLimit,
```
anandtiwary, line 66: On the consumer side, do we need to change the existing limit in modelJsons (may be) since this will now limit the number of results?

projects/observability/src/shared/dashboard/data/graphql/explorer-visualization/explorer-visualization-cartesian-data-source.model.ts
aaron-steinfeld: As a user interacts with the explorer, all of the fields on `ExploreVisualizationRequest` may change, and the new queries are emitted on that observable - so we should never really be reading from `ExploreVisualizationRequest` directly (it's basically like the angular router snapshot vs observable state, but with more confusing naming), but we also can't build state from it either - we should always use the latest from the observable. One way to see the current issue is going to the explorer, selecting a group by key and removing the interval. Instead of getting a segmented aggregation, nothing renders (because we process the result thinking it still has an interval).

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
aaron-steinfeld: None exist in the UI as of today (besides what you've got in flight), but in general yes - new data models that extend this will have to use the two types of limit appropriately.

projects/observability/src/shared/dashboard/data/graphql/explorer-visualization/explorer-visualization-cartesian-data-source.model.ts
anandtiwary: ah okay. It will still be confusing to a new developer. Do you think we should change the design a little bit (as a tech debt)?

aaron-steinfeld: yeah, for sure.

anandtiwary (approved):

PR #699, opened 2021-03-16
fix: issues with checkbox diffing

jake-bassett:
## Description
Fixes issue with table control multi select checkboxes not diffing correctly.

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
@@ -29,12 +29,13 @@ import { Renderer } from '@hypertrace/hyperdash';
 import { RendererApi, RENDERER_API } from '@hypertrace/hyperdash-angular';
 import { capitalize, isEmpty, pick } from 'lodash-es';
 import { BehaviorSubject, combineLatest, Observable, of, Subject } from 'rxjs';
-import { filter, first, map, pairwise, share, startWith, switchMap, tap } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -310,10 +312,10 @@ export class TableWidgetRendererComponent
     this.checkboxControls$ = forkJoinSafeEmpty(
       this.model.getCheckboxControlOptions().map(checkboxControlModel =>
         checkboxControlModel.getOptions().pipe(
-          first(),
+          take(1),
```
aaron-steinfeld, line 315: Any behavior difference for take vs first here? Any reason for either of them, for that matter (do we getOptions to emit more than once, and if it did, why ignore it)?

jake-bassett: I just changed to take(1) after our previous conversation about them reacting to empty observables differently. No real reason other than that. I don't even think they are needed here if the data sources are correct. The Observables should complete on their own, but since this part of code doesn't control that, I thought this was a safety. Open to removing them.

aaron-steinfeld: got it - take(1) does seem better than first. As to safety, it depends on how we're using it in the consumer. Here, because we're using forkjoin, which only listens to the last emission take is fine (changing forkjoin to listen to the first rather than last emission). If we were using this in a more standard case (not in a forkjoin) I think we'd rather take the full observable and let the component react to the data - whether it's empty, a single emit or N emits.

aaron-steinfeld (approved):

PR #703, opened 2021-03-18
fix: remove redundant string formatter methods

arjunlalb:
## Description
Remove redundant string formatter methods. Usages to be replaced with startCase in lodash.

### Testing
N/A

### Checklist:


projects/common/src/utilities/formatters/string/string-formatter.ts
```diff
@@ -1,14 +1,15 @@
...
     .join(' ');
 
-export const titleCaseFromSnakeCase = (snakeCaseString: string): string =>
-  snakeCaseString
-    .split('_')
-    .map(str => (str.length === 0 ? str : str[0].toUpperCase() + str.slice(1)))
-    .join(' ');
+export const titleCaseFromKebabCase = (kebabCaseString: string): string => titleCaseFromString(kebabCaseString, '-');
+
+export const titleCaseFromWhiteSpacedString = (whiteSpacedString: string): string =>
```
anandtiwary, line 9: Is this method similar to Lodash's 'startCase' ?

arjunlalb: Yep

anandtiwary: Do we still need this new method? 🙂

arjunlalb: startCase works for the other cases for which we've defined util methods as well (kebab/snake case to title case). Going by that, my understanding was that if its light enough, we'll define our own?

aaron-steinfeld: > Going by that, my understanding was that if its light enough, we'll define our own?

I think that goes more for components (as the styling on these evolves) or for util function that we'd need to bring in a new dependency for (such as the whole iso duration thing). If we had no lodash dependency, we shouldn't bring it in for this - but since we do have it and it's already in widespread use, we should be ok to use it where it does exactly what we want.

They might have been written without knowing about start case - or maybe they behave a little differently? If the behavior is identical, then we need not implement it ourselves - although we likely leave the existing ones here for backwards compatibility (could deprecate them, though)

anandtiwary: I would vote for deprecating them if they are exactly the same.

arjunlalb: I'd also prefer deprecating them then. startCase works for whitespaced, kebab case and snake cased strings.

arjunlalb: Deprecated.

aaron-steinfeld (approved):

PR #718, opened 2021-03-24
feat: add spacing between title and table when title is set

jake-bassett:
## Description
Add spacing between title and table content if a title is provided.

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
@@ -47,11 +47,12 @@ import { TableWidgetModel } from './table-widget.model';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <ht-titled-content
+      class="table-title"
       [title]="this.model.header?.title | htDisplayTitle"
       [link]="this.model.header?.link?.url"
       [linkLabel]="this.model.header?.link?.displayText"
     >
-      <div class="table-content-container">
+      <div class="table-content-container" [class.titled]="this.model.header?.title !== undefined">
         <ht-table-controls
           class="table-controls"
           [searchEnabled]="!!this.api.model.getSearchAttribute()"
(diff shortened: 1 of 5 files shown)


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -47,11 +47,12 @@ import { TableWidgetModel } from './table-widget.model';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <ht-titled-content
+      class="table-title"
```
aaron-steinfeld, line 50: I see there's already a `[ngClass]="{ 'header-margin': this.model.header?.topMargin }"` - is that the same thing? If the two don't have different purposes, can we resolve to one way of doing it?

jake-bassett: The header margin is for the table controls header. The new one is for the title that sits above that.

jake-bassett: Screen Shot 2021-03-24 at 11 07 30 AM

aaron-steinfeld: I think that's a bug though - the header model where this top margin is defined is used for the title, not table controls. I'm guessing when table controls were added, it created an unintentional separation between the title and where that margin was applied. If you look where this is used today, it's only in one spot - backend overview, on a table that doesn't have controls (which is likely why the bug was never caught - it was still rendering the same).

aaron-steinfeld: So the places where the header is used without topMargin being set just look bad - they should have this property set. If we want to remove that property entirely and automatically add the appropriate padding if title is set, I think that works:

image

We just want to make sure we're not double padding it if that property is set, so we should do one or the other.

aaron-steinfeld (approved):

PR #719, opened 2021-03-24
feat: request includeInactive for all entity id queries

jake-bassett:
## Description
All single entity queries by ID will set `includeInactive` to true.

### DIFF
--- a/projects/observability/src/shared/graphql/request/handlers/entities/query/entity/entity-graphql-query-handler.service.ts
+++ b/projects/observability/src/shared/graphql/request/handlers/entities/query/entity/entity-graphql-query-handler.service.ts
@@ -57,8 +57,8 @@ export class EntityGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
       timeRange: request.timeRange,
       properties: request.properties,
       filters: [new GraphQlEntityFilter(request.id, request.entityType)],
-      // If querying for a single API by ID, then always want to includeInactive
-      includeInactive: true
+      // If querying for a single API by ID, then usually want to includeInactive
+      includeInactive: request.includeInactive !== false
     };
   }
 }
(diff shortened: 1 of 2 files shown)


projects/observability/src/shared/graphql/request/handlers/entities/query/entity/entity-graphql-query-handler.service.ts
```diff
@@ -57,8 +57,8 @@ export class EntityGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
       timeRange: request.timeRange,
       properties: request.properties,
       filters: [new GraphQlEntityFilter(request.id, request.entityType)],
-      // If querying for a single API, then always want to includeInactive
-      includeInactive: request.entityType === ObservabilityEntityType.Api
+      // If querying for a single API by ID, then always want to includeInactive
```
aaron-steinfeld, line 60: @skjindal93 does this work for non-api entities? Just wondering if there was a reason it was limited before.

jake-bassett: It does in my testing, buy maybe @skjindal93 can confirm. I was the one that put the original limitation in just because I was only working with APIs. It may have not been needed even then.

skjindal93: Yes. `includeInactive` works entity type agnostic. Can be used with any of the entity types

aaron-steinfeld (dismissed):

aaron-steinfeld:
So the failing test reminds me - we probably don't want this flag on for things that use the entity query handler like fetching topology or interactions data - can we make it configurable defaulting to true?

aaron-steinfeld (approved):

PR #720, opened 2021-03-24
fix: Adding input variables in the Gauge component

palbizu:
## Description
Two input variables (label and color) were added in the Gauge to be able to generate it from them and not only from the thresholds.

### Testing
Visual Testing

### Checklist:

### Documentation


projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.scss
```diff
@@ -0,0 +1,23 @@
+@import "font";
+@import "color-palette";
+
+.input-data {
```
arjunlalb, line 4: 2 space indentation.

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
```diff
@@ -0,0 +1,44 @@
+import { Component, Input } from '@angular/core';
+import { Color } from '@hypertrace/common';
+
+@Component({
+  selector: 'ht-basic-gauge',
+  template: `
```
arjunlalb, line 6: specify changeDetection strategy as well.

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  public defaultLabel: string = '';
+
+  @Input()
+  public defaultColor: Color | string = '';
+
+  @Input()
+  public hasRadius: boolean = false;
+
+  @Input()
+  public valueArc?: string | undefined = '';
```
arjunlalb, line 43: you can skip 'undefined' in the type def. Marking it as optional using `?` does the job already

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
```diff
@@ -0,0 +1,44 @@
...
+      ></path>
+      <text x="0" y="-4" class="value-display" [attr.fill]="this.defaultColor">
+        {{ this.value }}
+      </text>
+      <text x="0" y="20" class="label-display">{{ this.defaultLabel }}</text>
+    </g>
+  `,
+  styleUrls: ['./basic-gauge.component.scss']
+})
+export class BasicGaugeComponent {
```
arjunlalb, line 25: why is this called Basic Gauge?

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  @Input()
+  public value: number = 0;
+
+  @Input()
+  public maxValue: number = 10;
+
+  @Input()
+  public defaultLabel: string = '';
```
anandtiwary, line 34: may be just label and color. I would think most of the logic can live in this component right? The other component can take thresholds and assign it the value, maxValue, label and color to this component. hasRadius, valueArc should be internal to basic(may be a better name) component

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -13,22 +13,16 @@ import { Arc, arc, DefaultArcObject } from 'd3-shape';
...
-              *ngIf="rendererData.radius > ${GaugeComponent.GAUGE_MIN_RADIUS_TO_SHOW_PATH}"
-              [attr.d]="rendererData.data.valueArc"
-              [attr.fill]="rendererData.data.threshold.color"
-            />
-            <text x="0" y="-4" class="value-display" [attr.fill]="rendererData.data.threshold.color">
-              {{ rendererData.data.value }}
-            </text>
-            <text x="0" y="20" class="label-display">{{ rendererData.data.threshold.label }}</text>
-          </g>
+          <ng-template *ngIf="rendererData.data">
```
aaron-steinfeld, line 16: `ng-template` wouldn't be appropriate here (a structural directive like *ngIf already makes a template) - maybe this was meant to be `ng-container` ?

aaron-steinfeld (commented):
So I think there may be some confusion from talking to different folks about this one with different ideas. Maybe a quick sync could help clarify - I really would like to avoid having two components here - the existing gauge only needs some small API tweaks.

projects/observability/src/shared/components/gauge/gauge.component.ts
palbizu: Removed it

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
palbizu: I removed the file because I couldn't find a way to do the subdivision

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
palbizu: Deleted

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
palbizu: Deleted

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.ts
palbizu: Deleted

projects/observability/src/shared/components/gauge/basic-gauge/basic-gauge.component.scss
palbizu: Deleted

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -22,12 +18,12 @@ import { Arc, arc, DefaultArcObject } from 'd3-shape';
               class="value-ring"
               *ngIf="rendererData.radius > ${GaugeComponent.GAUGE_MIN_RADIUS_TO_SHOW_PATH}"
               [attr.d]="rendererData.data.valueArc"
-              [attr.fill]="rendererData.data.threshold.color"
+              [attr.fill]="rendererData.data.color"
             />
-            <text x="0" y="-4" class="value-display" [attr.fill]="rendererData.data.threshold.color">
+            <text x="0" y="0" class="value-display" [attr.fill]="rendererData.data.color">
```
aaron-steinfeld, line 23: Other than the changes in the render data access, were the changes to y values and the radius threshold above intended?

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -158,9 +160,22 @@ export class GaugeComponent implements OnChanges {
...
           maxValue: this.maxValue,
-          threshold: currentThreshold
+          label: currentThreshold.label,
+          color: currentThreshold.color
         };
       }
+    } else if (
+      this.value !== undefined &&
+      this.maxValue !== undefined &&
+      this.defaultLabel !== undefined &&
```
aaron-steinfeld, line 170: The condition on line 154 takes care of the value and maxValue checks already I believe - it also enforces that we have thresholds defined, which is no longer required. My thinking is that if we get in here (that is, have a value and maxValue), we should always calculate an input - the label and color are just nice to haves.

that might look something like
```
return {
  value: this.value,
  maxValue: this.maxValue,
  label: currentThreshold?.label ?? this.defaultLabel ?? '',
  color: currentThreshold?.color ?? this.defaultColor ?? <some_gauge_default_color>
};
```

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -201,11 +216,13 @@ interface GaugeData {
   valueArc: string;
   value: number;
   maxValue: number;
-  threshold: GaugeThreshold;
+  color: Color | string;
```
aaron-steinfeld, line 219: preexisting nit: gaugeData, even in it's definition, is just an extension of gaugeInputData - may as well define it that way to save the duplicate definition of each property.

```ts
interface GaugeData extends GaugeInputData {
  valueArc: string;
}
```

projects/observability/src/shared/components/gauge/gauge.component.ts
palbizu: Already rolled back

projects/observability/src/shared/components/gauge/gauge.component.ts
palbizu: Good approach, thanks!

projects/observability/src/shared/components/gauge/gauge.component.ts
palbizu: done

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -197,15 +199,13 @@ interface GaugeSvgRendererData {
...
-  maxValue: number;
-  threshold: GaugeThreshold;
 }
 
 interface GaugeInputData {
   value: number;
   maxValue: number;
-  threshold: GaugeThreshold;
+  label: string;
+  color: Color | string;
```
arjunlalb, line 210: can we force only Color here? what non color string are we expecting here?

palbizu: There is no problem to force Color here, but GaugeThreshold could be Color or string, so I think that the better idea would be to maintain the same type in the rest of the interfaces. If you think that we can lose the functionality previously developed I can remove it without major issues.

aaron-steinfeld: I suspect it would bubble up and be a pain - the gauge itself, less of a problem, but this is used in a widget where colors are coming from json. They're likely `Color`s (which is just our enum of predefined color hexes), but the type safety isn't there since they're coming from unknown strings. Since we use them in the exact same way anyway (this accepting a string just means it can take a css-compatible color string), I'm guessing that's where the current union comes from.

projects/observability/src/shared/components/gauge/gauge.component.test.ts
```diff
@@ -22,7 +22,7 @@ describe('Gauge component', () => {
     ]
   });
 
-  test('render all data', () => {
+  test('render data from threasholds', () => {
```
aaron-steinfeld, line 25:
```suggestion
  test('render data from thresholds', () => {
```

projects/observability/src/shared/components/gauge/gauge.component.ts
```diff
@@ -8,11 +8,7 @@ import { Arc, arc, DefaultArcObject } from 'd3-shape';
     <div class="gauge-container" (htLayoutChange)="this.onLayoutChange()">
       <svg class="gauge" *ngIf="this.rendererData">
         <g attr.transform="translate({{ rendererData.origin.x }}, {{ rendererData.origin.y }})">
-          <path
-            class="gauge-ring"
-            [attr.d]="rendererData.backgroundArc"
-            *ngIf="rendererData.radius > ${GaugeComponent.GAUGE_MIN_RADIUS_TO_SHOW_PATH}"
-          />
+          <path class="gauge-ring" [attr.d]="rendererData.backgroundArc" *ngIf="rendererData.radius > 80" />
```
aaron-steinfeld, line 11: Should this still be `GAUGE_MIN_RADIUS_TO_SHOW_PATH` ? Where did the hardcoded 80 come from?

aaron-steinfeld (commented):
overall, lgtm

projects/observability/src/shared/components/gauge/gauge.component.ts
palbizu: It was previously developed. It's used as an edge to allow render the corresponding arc, there is a relation between the arc and this radius. I can apply the change but probably @anandtiwary has a deeper understanding of this specific constant.

anandtiwary: let me cherry-pick and see.

projects/observability/src/shared/components/gauge/gauge.component.test.ts
palbizu: Fixed

projects/observability/src/shared/components/gauge/gauge.component.ts
palbizu: I will keep it as it currently is, thanks for the clarification

projects/observability/src/shared/components/gauge/gauge.component.ts
aaron-steinfeld: radius check was there previously, but in this PR we've changed it from `rendererData.radius > GaugeComponent.GAUGE_MIN_RADIUS_TO_SHOW_PATH` to `rendererData.radius > 80` - that change, more than the check itself is what I'm wondering about.

palbizu: I see. I will keep the constant. Thanks!

aaron-steinfeld (approved):

PR #723, opened 2021-03-25
feat: adding server side search support

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/multi-select/multi-select.component.ts
+++ b/projects/components/src/multi-select/multi-select.component.ts
@@ -56,9 +56,9 @@ import { MultiSelectJustify } from './multi-select-justify';
         </ht-popover-trigger>
         <ht-popover-content>
           <div class="multi-select-content" [ngStyle]="{ 'min-width.px': triggerContainer.offsetWidth }">
-            <ng-container *ngIf="this.enableSearch">
+            <ng-container *ngIf="this.searchMode !== '${MultiSelectSearchMode.Disabled}'">
               <ng-container *ngIf="this.allOptions$ | async as allOptions">
-                <ng-container *ngIf="!this.enableImplicitOptionsSearch || allOptions.length > 5">
+                <ng-container *ngIf="this.searchMode === '${MultiSelectSearchMode.EmitOnly}' || allOptions.length > 5">
                   <ht-search-box
                     class="search-bar"
                     (valueChange)="this.searchOptions($event)"
@@ -88,9 +88,9 @@ import { MultiSelectJustify } from './multi-select-justify';
               </ng-container>
             </ng-container>
 
-            <div class="multi-select-options">
+            <div class="multi-select-options" *htLoadAsync="this.filteredOptions$ as filteredOptions">
               <div
-                *ngFor="let item of this.filteredOptions$ | async"
+                *ngFor="let item of filteredOptions"
                 (click)="this.onSelectionChange(item)"
                 class="multi-select-option"
               >
@@ -134,10 +134,7 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
   public showBorder: boolean = false;
 
   @Input()
-  public enableSearch: boolean = false;
-
-  @Input()
-  public enableImplicitOptionsSearch: boolean = true;
+  public searchMode: MultiSelectSearchMode = MultiSelectSearchMode.Disabled;
 
   @Input()
   public justify: MultiSelectJustify = MultiSelectJustify.Left;
@@ -149,7 +146,7 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
   public readonly selectedChange: EventEmitter<V[]> = new EventEmitter<V[]>();
 
   @Output()
-  public readonly optionsSearchChange: EventEmitter<string> = new EventEmitter<string>();
+  public readonly searchValueChange: EventEmitter<string> = new EventEmitter<string>();
 
   @ContentChildren(SelectOptionComponent)
   private readonly allOptionsList?: QueryList<SelectOptionComponent<V>>;
@@ -176,10 +173,16 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
   }
 
   public searchOptions(searchText: string): void {
-    if (!this.enableImplicitOptionsSearch) {
-      this.optionsSearchChange.emit(searchText);
-    } else {
+    if (this.searchMode === MultiSelectSearchMode.CaseInsensitive) {
       this.searchSubject.next(searchText);
+
+      return;
+    }
+
+    if (this.searchMode === MultiSelectSearchMode.EmitOnly) {
+      this.searchValueChange.emit(searchText);
+
+      return;
     }
   }
 
@@ -242,4 +245,10 @@ export const enum TriggerLabelDisplayMode {
   Placeholder = 'placeholder-mode',
   Selection = 'selection-mode',
   Icon = 'icon-mode'
+}
+
+export const enum MultiSelectSearchMode {
+  Disabled = 'disabled', // Search is not available
+  CaseInsensitive = 'case-insensitive', // Current available values are filtered in a case insensitive way and an emit is triggered
+  EmitOnly = 'emit-only' // Current available values not filtered, but an emit still triggered
 }
--- a/projects/components/src/multi-select/multi-select.component.ts
+++ b/projects/components/src/multi-select/multi-select.component.ts
@@ -245,6 +245,6 @@ export const enum TriggerLabelDisplayMode {
 
 export const enum MultiSelectSearchMode {
   Disabled = 'disabled', // Search is not available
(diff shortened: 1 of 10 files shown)


anandtiwary:
Need feedback on the api. 

Goal:
Support server side search through the multi select.

Questions: 
- Naming. 
- How should we combine enable client Search and enable server search? We need both
- Should we show no data message in case options list is empty
Screen Shot 2021-03-25 at 11 52 58 AM

- Should we show search box when number of options are less (example < 5). IMO it seems unnecessary.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -136,6 +136,9 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
   @Input()
   public enableSearch: boolean = false;
 
+  @Input()
+  public enableImplicitOptionsSearch: boolean = true;
```
aaron-steinfeld, line 140: What if we took a page out of the table api (hah!) and took in some like a searchProvider object instead. We could default it to a client side version, but provide a server implementation in other places.

anandtiwary: oh god. let me check

anandtiwary: That sounds complicated. Can we do it with just may be a `searchMode` enum (Implicit and external)? I think eventually, we would have a separate basic filter multi select component, but we need this as a stepping stone.

aaron-steinfeld: OK looked at multiselect - the search provider wouldn't work as well because the data is actually coming in as content children vs an array like some of the other components, so an output is the only way that comes to mind to accomplish this - but let's clean up the ambiguity between the two enable search inputs as you mentioned.

anandtiwary: so does this look good? 
```
@Input()
public enableSearch: boolean = false;

@Input()
public serachMode: MultiSelectSearchMode = MultiSelectSearchMode.Implicit;

...


export const enum MultiSelectSearchMode {
  Implicit = 'implicit,
  External = 'external
}
```

aaron-steinfeld: I'd rather boil down another input and make the enum values more explicit, but same idea:
```
@Input()
public searchMode: MultiSelectSearchMode = MultiSelectSearchMode.Disabled;

...

export const enum MultiSelectSearchMode {
  Disabled = 'disabled', // Search is not available
  CaseInsensitive = 'case-insensitive', // Current available values are filtered in a case insensitive way and an emit is triggered
  EmitOnly = 'emit-only' // Current available values not filtered, but an emit still triggered
}
```

anandtiwary: Cool.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -170,7 +173,17 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
   }
 
   public searchOptions(searchText: string): void {
-    this.searchSubject.next(searchText);
+    if (this.searchMode === MultiSelectSearchMode.CaseInsensitive) {
+      this.searchSubject.next(searchText);
+
+      return;
```
aaron-steinfeld, line 179: this case should still emit. There's no harm in doing so - it could be useful for a component's consumer to know when the search has changed. I'd expect that from a lib component and here it's actually more complex to avoid. Also, it says so in the enum comment :)

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -56,9 +56,9 @@ import { MultiSelectJustify } from './multi-select-justify';
         </ht-popover-trigger>
         <ht-popover-content>
           <div class="multi-select-content" [ngStyle]="{ 'min-width.px': triggerContainer.offsetWidth }">
-            <ng-container *ngIf="this.enableSearch">
+            <ng-container *ngIf="this.searchMode !== '${MultiSelectSearchMode.Disabled}'">
               <ng-container *ngIf="this.allOptions$ | async as allOptions">
-                <ng-container *ngIf="allOptions.length > 5">
+                <ng-container *ngIf="this.searchMode === '${MultiSelectSearchMode.EmitOnly}' || allOptions.length > 5">
```
aaron-steinfeld, line 61: this condition is non obvious to me - the two search modes don't really dictate when the search box should or should be shown - I'd think it would be either the same logic for both, or no logic and let the caller dictate by explicitly disabling filter if they're not providing 5+ options (can't remember where exactly that conversation landed in the previous PR).

projects/components/src/multi-select/multi-select.component.ts
anandtiwary: I missed the comment. I think this can be confusing. Until there is a known use case, I think we should not emit for this mode since filtering is applied internally. That's why i was keeping `Implicit` as it's name.

projects/components/src/multi-select/multi-select.component.ts
anandtiwary: That problem comes up when you search and get less than 5 results. You are stuck at that point. We can't do it with an option, since this logic will not work with server side search.

projects/components/src/multi-select/multi-select.component.ts
anandtiwary: I would prefer calling it Implicit for now and keep the emit logic as above.

aaron-steinfeld: What part do you find confusing? A component is expected to emit when its state changes - the caller decides whether they care or not.

The name implicit really doesn't indicate anything to the caller - it just says that something is not directly expressed. But since we're directly expressing that, why not have it describe what it's actually doing?

projects/components/src/multi-select/multi-select.component.ts
aaron-steinfeld: That's actually why I'd rather take the logic out entirely - if I express a search mode, show it. If I want to hide the search box in certain conditions, that's trivially done on the caller side.

anandtiwary: Cool. I will take it out

anandtiwary: The only thing is, for consistent behavior we will have to put `searchMode= options.length > 5 ? MultiSelectSearchMode.Implicit : MultiSelectSearchMode.Disabled` in every usage.  It looks really weird to show search with just 1 option.

projects/components/src/multi-select/multi-select.component.ts
anandtiwary: If search is happening implicitly then why emit on the search value. As a consumer am I supposed to act on it? The name `case-insensitive` doesn't convey well that there is a auto filtering applied

aaron-steinfeld: As a consumer you may want to or you may not (the same as any other output). In my mind that's a much clearer name than implicit as it not only indicates a sort is occurring but how it works - if there's a clearer name though, I'm game.

projects/components/src/multi-select/multi-select.component.ts
aaron-steinfeld: There's very places where we have a variable number of options, anyway - and I think with any static small number (such as 6) search doesn't really make sense.

projects/components/src/multi-select/multi-select.component.ts
anandtiwary: I am fine with that name, but with that name I don't think we should emit. Can we keep the same name but emit it only when there is a valid use case?

projects/components/src/multi-select/multi-select.component.ts
anandtiwary: We need the options.length > 1 logic for displaying buttons 'select all' and 'clear selected' buttons. I can remove `allOptions.length > 5` conditions if you feel it is not useful and condition is non obvious.

anandtiwary: Removed

projects/components/src/multi-select/multi-select.component.ts
aaron-steinfeld: The reason I'm caught up on this is that it's not that we're omitting something until we need it - that, I'm pretty much always for. Here, we're adding extra logic to make the behavior of an output inconsistent until we need to use it. In that case, I'd much rather have consistency even if we're not always using it. Outputs are meant to tell you about a state change, in this case, it's telling us that the search string has changed - the host is the one deciding whether to do anything with that or not.

anandtiwary: oh come on :)

anandtiwary: updated

aaron-steinfeld (approved):

projects/components/src/multi-select/multi-select.component.ts
aaron-steinfeld: you're my hero!

PR #730, opened 2021-03-29
fix: updating checkbox component in multi-select

palbizu:
## Description
The checkbox used within the multi-select component was replaced by the component belonging to the library and also some minor styles were modified.

### Testing
- Visual test

image

image

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/multi-select/multi-select.component.scss
+++ b/projects/components/src/multi-select/multi-select.component.scss
@@ -80,11 +80,16 @@
     height: 34px;
     padding: 4px 16px;
     cursor: pointer;
-    font-size: 14px;
+    font-size: 15px;
     align-items: center;
 
     .checkbox {
       margin: 0px;
+      color: $gray-3;
+
+      ::ng-deep .mat-ripple {
+        display: none;
+      }
     }
 
     .icon {
--- a/projects/components/src/multi-select/multi-select.component.test.ts
+++ b/projects/components/src/multi-select/multi-select.component.test.ts
@@ -93,10 +93,8 @@ describe('Multi Select Component', () => {
     });
 
     spectator.tick();
-    const selectedElements = spectator.queryAll('ht-checkbox', { root: true });
-    expect(
-      selectedElements.filter(checkboxOption => checkboxOption.getAttribute('ng-reflect-checked') === 'true').length
-    ).toBe(2);
+    const checkboxElements = spectator.queryAll('ht-checkbox', { root: true });
+    expect(checkboxElements.length).toBe(6);
   }));
 
   test('should display provided options with icons when clicked', fakeAsync(() => {
@@ -125,10 +123,8 @@ describe('Multi Select Component', () => {
     expect(spectator.query('.multi-select-content', { root: true })).toExist();
     expect(optionElements.length).toBe(6);
 
-    const selectedElements = spectator.queryAll('ht-checkbox', { root: true });
-    expect(
-      selectedElements.filter(checkboxOption => checkboxOption.getAttribute('ng-reflect-checked') === 'true').length
-    ).toBe(2);
+    const checkboxElements = spectator.queryAll('ht-checkbox');
+    expect(checkboxElements.length).toBe(6);
 
     optionElements.forEach((element, index) => {
       expect(element).toHaveText(selectionOptions[index].label);
@@ -223,11 +219,7 @@ describe('Multi Select Component', () => {
 
     spectator.tick();
 
-    expect(
-      spectator
-        .queryAll('ht-checkbox', { root: true })
-        .filter(checkboxOption => checkboxOption.getAttribute('ng-reflect-checked') === 'true').length
-    ).toBe(0);
+    expect(spectator.queryAll('ht-checkbox', { root: true }).length).toBe(6);
     expect(onChange).toHaveBeenCalledTimes(1);
     expect(onChange).toHaveBeenLastCalledWith([]);
     expect(spectator.query(LabelComponent)?.label).toEqual('Select options');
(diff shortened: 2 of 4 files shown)


projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -226,7 +228,7 @@ describe('Multi Select Component', () => {
     spectator.click(allOptionElement!);
 
     spectator.tick();
-    const selectedElements = spectator.queryAll('input:checked', { root: true });
+    const selectedElements = spectator.queryAll(CheckboxComponent);
```
anandtiwary, line 231: won't we still need to query from root?

projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -80,11 +80,16 @@
...
-    font-size: 14px;
+    font-size: 15px;
     align-items: center;
 
     .checkbox {
       margin: 0px;
+      color: $gray-3;
+
+      ::ng-deep .mat-ripple {
+        display: none;
```
aaron-steinfeld, line 91: Do we not want the ripple here? I don't have strong feelings one way or another, but it seems like the ripple should be consistent across all our checkboxes - inside a multiselect or otherwise.

projects/components/src/multi-select/multi-select.component.test.ts
aaron-steinfeld: Yeah, I'd expect so - @palbizu the reason we need to query from root here is that when opened, the multiselect attaches the floating content overlay to the dom root (or more precisely, it uses popover to do so), rather than within the multiselect component (which queries are generally scoped to). Adding the extra `{ root: true }` to the queries with the new component selector would fix that, but further, the previous query on the line at least was looking not just for checkboxes, but checkboxes that were selected. To make this query equivalent, we'd end up with something like
`spectator.queryAll(lCheckboxComponent, { root: true }).filter(checkbox => checkbox.checked);`

palbizu: Resolved. thanks!

projects/components/src/multi-select/multi-select.component.scss
palbizu: I'm ignoring the ripple there because I don't want to have the hover effect because the complete line has its own hover.

projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -91,8 +93,10 @@ describe('Multi Select Component', () => {
     });
 
     spectator.tick();
-    const selectedElements = spectator.queryAll('input:checked', { root: true });
-    expect(selectedElements.length).toBe(2);
+    const selectedElements = spectator.queryAll('ht-checkbox', { root: true });
+    expect(
+      selectedElements.filter(checkboxOption => checkboxOption.getAttribute('ng-reflect-checked') === 'true').length
```
anandtiwary, line 98: @palbizu I think you are forced to use this function since queryAll does not support using the directive `CheckboxComponent` with the root option.  I know you want to test whether `checked` property is set to true or not. 

Mmm, may be we can just test whether `selectedElements` (better rename it to checkboxes) length ===2. No need to test checked property in this case, since using a public function for a child component is even worse.

aaron-steinfeld: Ah, I didn't realize that wasn't supported - what about `spectator.queryAll('ht-checkbox', { root: true, read: CheckboxComponent })` ?

palbizu: @aaron-steinfeld your approach is not supported

aaron-steinfeld: ah, sounds like a good improvement to spectator :) without that, if we don't need to verify the checked property for correctness, we can probably drop it - but somewhere, we should have a test verifying that the checkboxes are set correctly. As much as I dislike it, it's better than not testing it at all if we don't have a better way.

aaron-steinfeld: Another option is to use the real checkbox instead of mocking it, which would allow you to query its contents - a tradeoff of bad options though.

projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -91,8 +93,8 @@ describe('Multi Select Component', () => {
     });
 
     spectator.tick();
-    const selectedElements = spectator.queryAll('input:checked', { root: true });
-    expect(selectedElements.length).toBe(2);
+    const checkboxElements = spectator.queryAll('ht-checkbox', { root: true });
+    expect(checkboxElements.length).toBe(6);
```
anandtiwary, line 97: How did this change to 6? You are only migrating to checkbox component, so the count shouldn't change

palbizu: The content is changing because in the first case are testing if the input is checked and in the second one it has all checkboxElements. In this case, there are 6 checkbox elements and two of them are checked.

palbizu: I'll go back to the last PR, renaming the variables to have better clarity on the elements I'm manipulating.

aaron-steinfeld (commented):
lgtm; will let anand take a look and see if he has any other ideas

projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -91,8 +93,11 @@ describe('Multi Select Component', () => {
     });
 
     spectator.tick();
-    const selectedElements = spectator.queryAll('input:checked', { root: true });
-    expect(selectedElements.length).toBe(2);
+    const selectedCheckboxElements = spectator.queryAll('ht-checkbox', { root: true });
+    expect(
+      selectedCheckboxElements.filter(checkboxElement => checkboxElement.getAttribute('ng-reflect-checked') === 'true')
```
anandtiwary, line 98: sorry @palbizu for the flip flp

anandtiwary (approved):

projects/components/src/multi-select/multi-select.component.test.ts
anandtiwary: Sure, I will find time for it. Should be easy to fix

PR #731, opened 2021-03-30
fix: Unable to click topology

palbizu:
## Description
The popover is positioned near the node or edge where the user hovers or clicks, dynamically adjusting to the screen.

### Testing
 Visual Testing 

Hovering:

image

After click:

image

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/components/topology/renderers/tooltip/topology-tooltip-popover.ts
+++ b/projects/observability/src/shared/components/topology/renderers/tooltip/topology-tooltip-popover.ts
@@ -20,7 +20,7 @@ export class TopologyTooltipPopover implements TopologyTooltip {
     private readonly injector: Injector,
     private readonly popoverService: PopoverService
   ) {
-    this.popoverSubject = new BehaviorSubject(this.buildPopover(false));
+    this.popoverSubject = new BehaviorSubject(this.buildPopover(false, null));
     this.hidden$ = this.popoverSubject.pipe(switchMap(popover => merge(popover.hidden$, popover.closed$)));
   }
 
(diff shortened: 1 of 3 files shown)


projects/observability/src/shared/components/topology/renderers/tooltip/topology-tooltip-popover.ts
```diff
@@ -86,9 +85,10 @@ export class TopologyTooltipPopover implements TopologyTooltip {
     }
 
     popover.updatePositionStrategy({
-      type: PopoverPositionType.Relative,
-      locationPreferences: [PopoverRelativePositionLocation.InsideTopLeft],
-      origin: this.container
+      type: PopoverPositionType.FollowMouse,
```
aaron-steinfeld, line 88: So this would follow the mouse, but how do we then close the popover? You had mentioned that relative anchored on the node would look better and I tend to agree - should we go with that option?

aaron-steinfeld (dismissed):
I think this is an improvement at least, so if we want to go further with it later, that's fine

projects/observability/src/shared/components/topology/renderers/tooltip/topology-tooltip-popover.ts
palbizu: I think this can work as an improvement, I would like to know what @anandtiwary thinks, but I think it solves in almost all cases the problem we had.

anandtiwary: Looks fine. Could you also please share screenshot of topology when you click on a node and when you hover on a node?

palbizu: I adjusted some changes that @aaron-steinfeld suggested to me and leave the evidence in the comment (images of the cases)

aaron-steinfeld (approved):

PR #738, opened 2021-04-02
Table expander checkbox

anandtiwary:
## Description
Fix: Separating Table Expander and checkbox header renderers

### Testing

### Checklist:

### Documentation


projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -59,7 +59,7 @@ export class TableCdkDataSource implements DataSource<TableRow> {
       .pipe(
         tap(() => this.loadingStateSubject.next({ loading$: NEVER })),
         /**
-         * Below debouce is needed to handle multiple emission from buildChangeObservable.
+         * Below debounce is needed to handle multiple emission from buildChangeObservable.
```
jake-bassett, line 62: I assume this is the real reason for the PR and all this other stuff is just because you were working in the area.

jake-bassett (dismissed):
lgtm

aaron-steinfeld (commented):
lgtm will let @jake-bassett take a look since he's more familiar

aaron-steinfeld:
> lgtm will let @jake-bassett take a look since he's more familiar

damn, he beat me

projects/components/src/table/cells/state-parsers/table-cell-state-checkbox.ts
```diff
@@ -0,0 +1,17 @@
...
+import { TableCellParser } from '../table-cell-parser';
+import { TableCellParserBase } from '../table-cell-parser-base';
+import { CoreTableCellParserType } from '../types/core-table-cell-parser-type';
+
+@TableCellParser({
+  type: CoreTableCellParserType.Checkbox
+})
+export class TableCellStateParserCheckbox extends TableCellParserBase<TableRowState, TableRowState, undefined> {
+  public parseValue(_: TableRowState, rowData: StatefulTableRow): TableRowState {
+    return rowData.$$state;
```
jake-bassett, line 11: I'm not sure how this is working. Shouldn't you need to map this to the specific property you are looking for? In this case, the boolean for `selected`?

projects/components/src/table/data/table-cdk-data-source.ts
anandtiwary: Yep. Totally

projects/components/src/table/cells/state-parsers/table-cell-state-checkbox.ts
jake-bassett: This approach looks right, so if its working all good - approved. Just curious where the final mapping to the property is happening.

anandtiwary: So, the renderer actually depends on the entire state. So i didn't change that part. Just using the row to find state object instead of depending on the cell value and deal with id conflicts

jake-bassett: Cool. Run with it.

projects/components/src/table/cells/state-parsers/table-cell-state-expander.ts
```diff
@@ -0,0 +1,17 @@
+import { StatefulTableRow, TableRowState } from '../../table-api';
+import { TableCellParser } from '../table-cell-parser';
+import { TableCellParserBase } from '../table-cell-parser-base';
+import { CoreTableCellParserType } from '../types/core-table-cell-parser-type';
+
+@TableCellParser({
+  type: CoreTableCellParserType.RowExpander
+})
+export class TableCellStateParserExpander extends TableCellParserBase<TableRowState, TableRowState, undefined> {
```
aaron-steinfeld, line 9: Are these two parsers the same? Do we need to dupe it?

anandtiwary: They are same right now. I wanted to get feedback on whether I should make this return the whole object (more safer) or make it return the specific property.  I can update this PR depending on this.

aaron-steinfeld: I'm fine with either:
- one copy of this
- no copies (and just use the no-op parser and update the renderers to read state off the row)

I think splitting to two specific ones is overkill - it's not _really_ safer because the renderer has the row anyway.

anandtiwary: Will move to one copy. Don't want to remove the parser.

anandtiwary: updated

aaron-steinfeld (approved):

PR #745, opened 2021-04-05
fix: table column and content misalignment

palbizu:
## Description
Column bars padding was removed and content was centered.

### Testing
Visual Testing

Before: 

image

After: 

image

image

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/table/table.component.scss
+++ b/projects/components/src/table/table.component.scss
@@ -71,7 +71,7 @@
   }
 
   .header-column-resize-handle {
-    padding: 6px 2px;
+    padding: 6px 0px;
     height: 100%;
     cursor: col-resize;
 
--- a/projects/components/src/table/table.component.scss
+++ b/projects/components/src/table/table.component.scss
@@ -67,11 +67,15 @@
 
   .header-cell-renderer {
     width: 100%;
-    padding: 10px 12px;
+    padding: 10px 12px 10px 7px;
+  }
+
+  .header-cell-renderer:first-child {
+    padding-left: 12px;
   }
 
   .header-column-resize-handle {
-    padding: 6px 0px;
+    padding: 6px 2px;
     height: 100%;
     cursor: col-resize;
 
--- a/projects/components/src/table/cells/data-renderers/table-data-cell-renderer.component.scss
+++ b/projects/components/src/table/cells/data-renderers/table-data-cell-renderer.component.scss
@@ -5,7 +5,7 @@
   align-items: center;
   height: calc(100% - 8px); // subtract margin
   margin: 4px 4px;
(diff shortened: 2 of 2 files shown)


anandtiwary:
Can you also share a before screenshot? I guess you can get it from app-dev for the same url

palbizu:
> Can you also share a before screenshot? I guess you can get it from app-dev for the same url

Done

projects/components/src/table/table.component.scss
```diff
@@ -71,7 +71,7 @@
   }
 
   .header-column-resize-handle {
-    padding: 6px 2px;
+    padding: 6px 0px;
```
aaron-steinfeld, line 74: doesn't this put the width of the resize handle down to 1px? I think it would make it very hard to grab. The screenshots look good but can we accomplish the same thing by reducing the header padding rather than the resize handles?

aaron-steinfeld:
We also should check how this looks on a screen using the tree mode, like the trace waterfall - the indentation may change things a little bit there.

projects/components/src/table/table.component.scss
```diff
@@ -67,7 +67,11 @@
 
   .header-cell-renderer {
     width: 100%;
-    padding: 10px 12px;
+    padding: 10px 12px 10px 7px;
```
aaron-steinfeld, line 70: Where's the odd value (-5px) from? Aren't we moving 4px in the data cell?

projects/components/src/table/table.component.scss
palbizu: done

projects/components/src/table/table.component.scss
```diff
@@ -67,7 +67,11 @@
...
+  }
+
+  .header-cell-renderer:first-child {
```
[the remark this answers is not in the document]
palbizu: because it is not affected by the header-column-resize-handle padding

projects/components/src/table/table.component.scss
palbizu: I could do it, I understand that visually it looks a little better this way.

aaron-steinfeld: I misunderstood - the issue is that the data was moved over 8px (4 margin, 4 padding) and the header was moved over 17px (5px from the resize handle - 2px padding, 1px width - except in the first col) and 12px padding from this style.

With your change, it's now 12px consistently in the header, and 12px consistently in data - right?

aaron-steinfeld (approved):

projects/components/src/table/table.component.scss
palbizu: Yes. It's consistently now.

PR #747, opened 2021-04-06
feat: support displaying beta tag in left nav item

arjunlalb:
## Description
Support displaying beta tag on left nav items.

### Testing
Manually verified

### Checklist:

### DIFF
--- a/projects/components/src/navigation/navigation-list.component.scss
+++ b/projects/components/src/navigation/navigation-list.component.scss
@@ -85,4 +85,6 @@
 
 .expanded {
   width: 220px;
+  overflow-x: hidden;
+  resize: horizontal;
 }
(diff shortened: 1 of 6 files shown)


projects/components/src/navigation/navigation-list.component.scss
```diff
@@ -85,4 +85,6 @@
 
 .expanded {
   width: 220px;
+  overflow-x: hidden;
+  resize: horizontal;
```
aaron-steinfeld, line 89: do we want dynamic resize on the left nav? Since the widths are known ahead of time, we should have already sized appropriately (and I'm curious how the app responds to this changing size)

arjunlalb: updated

aaron-steinfeld (approved):

PR #750, opened 2021-04-06
fix: updating page data after modify an implicit filter

palbizu:
## Description
Updating data after clicking on a node in the topology or modifying an implicit filter.

### Testing
Visual Testing.

### Checklist:

### Documentation


projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
```diff
@@ -84,6 +84,10 @@ export class NavigableDashboardComponent implements OnChanges {
     }
 
     if (changeObject.filterConfig) {
+      if (changeObject.filterConfig.currentValue?.implicitFilters) {
```
aaron-steinfeld, line 87: I think we should likely drop the condition - we can always refresh filters if the filter config changed - it may refresh in certain situations where it's unnecessary, but it's likely rare enough to not matter. In this form, it would miss the case where filters are cleared (implicitFilters goes from defined to undefined)

palbizu: I agree.

aaron-steinfeld (approved):

aaron-steinfeld (approved):

PR #752, opened 2021-04-07
Display string enum pipe

jake-bassett:
## Description
We keep using the same code over and over. Turning it into a const method and pipe.

### DIFF
--- a/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
+++ b/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
@@ -2,7 +2,7 @@ import { Pipe, PipeTransform } from '@angular/core';
 import { displayStringEnum } from './display-string-enum';
 
 @Pipe({
-  name: 'htDisplayEnum'
+  name: 'htDisplayStringEnum'
 })
 export class DisplayStringEnumPipe implements PipeTransform {
   public transform(value: string): string {
--- a/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
+++ b/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
@@ -1,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { displayStringEnum } from './display-string-enum';
+
+@Pipe({
+  name: 'htDisplayEnum'
+})
+export class DisplayStringEnumPipe implements PipeTransform {
+  public transform(value: string): string {
+    return displayStringEnum(value);
+  }
+}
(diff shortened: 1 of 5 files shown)


projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
```diff
@@ -0,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { displayStringEnum } from './display-string-enum';
+
+@Pipe({
+  name: 'htDisplayStringEnum'
+})
+export class DisplayStringEnumPipe implements PipeTransform {
```
anandtiwary, line 7: Why not use displayString?

jake-bassett: Hmm. Not sure that will work, or at least it shouldn't. The `displayString` method shouldn't replace `-` or `_`. It should just basically String(<arg>) any input but not format it... with some extra logic for null checks, arrays, etc.

aaron-steinfeld: I don't think we want to overload display string with this behavior (for example "my-value" would become "My Value", which we don't want the main string pipe doing)

aaron-steinfeld (commented):
lgtm, tests please (at least on the const function, the pipe is trivial enough to not matter)

projects/common/src/utilities/formatters/enum/display-string-enum.ts
```diff
@@ -5,5 +5,9 @@ export const displayStringEnum = (provided?: string): string => {
     return '-';
   }
 
-  return startCase(provided.toLowerCase());
+  // This removes dashes and underscores and gives all words initial caps
+  const startCased = startCase(provided);
+
+  // We only want the first word capitalized.
```
aaron-steinfeld, line 11: If that's what we want, what about `capitalize(lowerCase(provided))` - believe that works the same way as startCase:
```
_.lowerCase('--Foo-Bar--');
// => 'foo bar'
```

aaron-steinfeld (approved):

projects/common/src/utilities/formatters/enum/display-string-enum.ts
jake-bassett: That does look cleaner, but I think this one fails for camelCase and PascalCase.

aaron-steinfeld: image

anandtiwary: lodash should have ng pipes for each method. `'text' | lowercase | capitalize`

PR #756, opened 2021-04-08
feat: show exit calls break up on hover on exit calls cell in trace list

itssharmasandeep:
## Description
feat: show exit calls break up on hover on exit calls cell in trace list

### Testing
Local testing is done 

### Checklist:

Working example for 0 exit calls
Screenshot 2021-04-07 at 12 44 50 PM

Working example for more than 0 exit calls
Screenshot 2021-04-07 at 12 46 32 PM


projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.scss
```diff
@@ -0,0 +1,31 @@
...
+.api-callee-name-count {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 2px;
+
+  .api-callee-name {
+    @include ellipsis-overflow();
+    max-width: 200px;
+    font-size: 13.5px;
```
aaron-steinfeld, line 17: Is this a font size from a mock (if so, could you slack it to me?) our fonts should all be coming from mixins which include all the various font styles in the design system, but I don't recall any at fractional pixels.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,84 @@
...
+})
+@TableCellRenderer({
+  type: EXIT_CALLS_CELL,
+  alignment: TableCellAlignmentType.Left,
+  parser: CoreTableCellParserType.NoOp
+})
+export class ExitCallsTableCellRendererComponent extends TableCellRendererBase<ExploreValue, any> implements OnInit {
+  public readonly apiCalleeNameCount: any;
+  public readonly apiExitCalls: number;
+  public readonly maxShowAPICalleeNameCount: number = 10;
```
aaron-steinfeld, line 54: nit - casing consistency, should be `maxShowApiCalleeNameCount` (acronyms are treated the same as words)

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,84 @@
...
+    </div>
+  `
+})
+@TableCellRenderer({
+  type: EXIT_CALLS_CELL,
+  alignment: TableCellAlignmentType.Left,
+  parser: CoreTableCellParserType.NoOp
+})
+export class ExitCallsTableCellRendererComponent extends TableCellRendererBase<ExploreValue, any> implements OnInit {
+  public readonly apiCalleeNameCount: any;
```
aaron-steinfeld, line 52: no any in this codebase! I bet the linter agrees :) Please use explicit types throughout this file

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (apiCalleeNameCount) {
+      const showAPICalleeNameCount: any = {};
+      let count = 0;
+      Object.keys(apiCalleeNameCount).forEach((key: string) => {
+        if (count < this.maxShowAPICalleeNameCount) {
+          showAPICalleeNameCount[key] = apiCalleeNameCount[key];
+          count++;
+        }
+      });
+      this.totalCountOfDifferentAPICallee = Object.keys(apiCalleeNameCount).length;
```
aaron-steinfeld, line 79: as a getFunction, I would not expect this function alter the object state - can we move this assignment into the constructor too (this will also help strengthen the types, since `public totalCountOfDifferentAPICallee!: number;` could then become `public readonly totalCountOfDifferentAPICallee: number;`

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,84 @@
...
+    super(columnConfig, index, parser, cellData, rowData);
+    this.apiCalleeNameCount = this.getMaxShowAPICalleeNameCount(rowData.apiCalleeNameCount);
+    this.apiExitCalls = Number(rowData.apiExitCalls);
+  }
+
+  public getMaxShowAPICalleeNameCount(apiCalleeNameCount: any): any {
+    if (apiCalleeNameCount) {
+      const showAPICalleeNameCount: any = {};
+      let count = 0;
+      Object.keys(apiCalleeNameCount).forEach((key: string) => {
```
aaron-steinfeld, line 73: suggestion: a more straightforward way of doing this could be to use `Object.entries` and then slice the array down to a max length. Then, the copy + count isn't needed and you could just ngFor over the entry items.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  public constructor(
+    @Inject(TABLE_COLUMN_CONFIG) columnConfig: TableColumnConfig,
+    @Inject(TABLE_COLUMN_INDEX) index: number,
+    @Inject(TABLE_DATA_PARSER) parser: TableCellParserBase<ExploreValue, Trace, unknown>,
+    @Inject(TABLE_CELL_DATA) cellData: ExploreValue,
+    @Inject(TABLE_ROW_DATA) rowData: Trace
+  ) {
+    super(columnConfig, index, parser, cellData, rowData);
+    this.apiCalleeNameCount = this.getMaxShowAPICalleeNameCount(rowData.apiCalleeNameCount);
```
aaron-steinfeld, line 65: The only assumption we want to make with data types is that the cell data is what we're trying to render - so instead of looking for a known key off the row like `rowData.apiCalleeNameCount`, we can assume that value has already been assigned to cellData. This allows this renderer to be reused for any column with this data, no matter the name (or the rest of the row contents - we should avoid the assumption about rowData too if we want to reuse this).

aaron-steinfeld (commented):
Could we see a screenshot of this in action?

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.scss
anandtiwary: Yeah, @itssharmasandeep usually the mocks have font mixin name written next to it. If not, we need to ask ux to use one of the existing ones.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  TABLE_CELL_DATA,
+  TABLE_COLUMN_CONFIG,
+  TABLE_COLUMN_INDEX,
+  TABLE_DATA_PARSER,
+  TABLE_ROW_DATA
+} from '@hypertrace/components';
+import { Trace } from '@hypertrace/distributed-tracing';
+import { ExploreValue } from '@hypertrace/observability';
+
+export const EXIT_CALLS_CELL = 'EXIT_CALLS_CELL';
```
arjunlalb, line 18: this type should be added to ObservabilityTableCellType object.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,84 @@
...
+
+@Component({
+  selector: 'exit-calls-table-cell-renderer',
```
arjunlalb, line 21: all selectors start with `ht-` prefix

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.test.ts
```diff
@@ -37,7 +41,11 @@ describe('Exit Calls table cell renderer component', () => {
       providers: [tableCellRowDataProvider({ apiExitCalls: 2, apiCalleeNameCount: value })]
     });
 
-    expect(spectator.component.getMaxShowAPICalleeNameCount(value)).toMatchObject(value);
-    expect(spectator.component.totalCountOfDifferentAPICallee).toBe(2);
+    expect(spectator.component.apiCalleeNameCount).toMatchObject([
```
aaron-steinfeld, line 44: This test has value, but what would be even better would be to verify that this data renders - for example by querying for the rendered spans and checking their content (e.g. `expect(spectator.queryAll('.api-callee-name')[0]).toContainText('key1');`)

That checks both how the data is processed but also that the template is correctly binding what we expect.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -62,23 +61,9 @@ export class ExitCallsTableCellRendererComponent extends TableCellRendererBase<E
     @Inject(TABLE_ROW_DATA) rowData: Trace
   ) {
     super(columnConfig, index, parser, cellData, rowData);
-    this.apiCalleeNameCount = this.getMaxShowAPICalleeNameCount(rowData.apiCalleeNameCount);
+    const apiCalleeNameCount: string[][] = Object.entries(Object(rowData.apiCalleeNameCount));
```
aaron-steinfeld, line 64: nit: Are these extra constructors (`Object()` here and `Number()` below) necessary? Or is the data already in those types?

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,69 @@
...
+    @Inject(TABLE_COLUMN_INDEX) index: number,
+    @Inject(TABLE_DATA_PARSER) parser: TableCellParserBase<ExploreValue, Trace, unknown>,
+    @Inject(TABLE_CELL_DATA) cellData: ExploreValue,
+    @Inject(TABLE_ROW_DATA) rowData: Trace
+  ) {
+    super(columnConfig, index, parser, cellData, rowData);
+    const apiCalleeNameCount: string[][] = Object.entries(Object(rowData.apiCalleeNameCount));
+    this.totalCountOfDifferentApiCallee = apiCalleeNameCount.length;
+    this.apiCalleeNameCount = apiCalleeNameCount.slice(0, this.maxShowApiCalleeNameCount);
+    this.apiExitCalls = Number(rowData.apiExitCalls);
```
aaron-steinfeld, line 67: Is this `rowData.apiExitCalls` derivable from  `rowData.apiCalleeNameCount` ? Where possible, we want one column to correspond to one attribute - this is a follow on to my previous comment about avoid using rowData and instead use cellData.

This PR doesn't use this new renderer, but this comes back to how we'll eventually use it - in a table where we want each column definition should be self contained. The renderer can be used in many tables, so we don't want to make assumptions about other columns and what values they make available. For example, this might be used such as:
```ts
{
        type: 'table-widget-column',
...
        display: ObservabilityTableCellType.ExitCalls,
        value: {
          type: 'attribute-specification',
          attribute: 'apiCalleeNameCount'
        },
``` 

If we need the second attribute, we should instead use a composite specification to bring that one in, too:
```ts
        {
          type: 'composite-specification',
          specifications: [
            {
              type: 'attribute-specification',
              attribute: 'apiCalleeNameCount'
            },
            {
              type: 'attribute-specification',
              attribute: 'apiExitCalls'
            }
          ],
          'order-by': 'apiCalleeNameCount' // This specifies which attribute should be used for server-side ordering if we sort by this col
        }
```

In both cases, this will provide a `cellData` that has the specific data this renderer needs.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.test.ts
itssharmasandeep: Okay, this change I can make.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
itssharmasandeep: In this data is coming as a type called **unknown**
So I had to type cast these in appropriate values. Or is there any better way to handle this type called **unknown**?

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
itssharmasandeep: Yeah
looking into this

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
aaron-steinfeld: This would be fixed by the other comment thread below - the reason we don't have types is that the rest of the row isn't known (another indication we shouldn't be using it). We can and should declare the type of the data we expect so - instead of `TableCellRendererBase<ExploreValue, Trace>` it would be, for example, `TableCellRendererBase<Dictionary<number>>` (if we expect to receive an object like `{ "some-service": 10 }`)

This would change the type on the incoming data to `@Inject(TABLE_CELL_DATA) cellData: Dictionary<number>`

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  public readonly apiCalleeNameCount: string[][];
+  public readonly apiExitCalls: number;
+  public readonly maxShowApiCalleeNameCount: number = 10;
+  public readonly totalCountOfDifferentApiCallee!: number;
+
+  public constructor(
+    @Inject(TABLE_COLUMN_CONFIG) columnConfig: TableColumnConfig,
+    @Inject(TABLE_COLUMN_INDEX) index: number,
+    @Inject(TABLE_DATA_PARSER) parser: TableCellParserBase<ExploreValue, Trace, unknown>,
+    @Inject(TABLE_CELL_DATA) cellData: ExploreValue,
```
anandtiwary, line 60: based on current specification, the cell data shouldn't be 'exploreValue'

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
itssharmasandeep: Done

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,74 @@
...
+@Component({
+  selector: 'ht-exit-calls-table-cell-renderer',
+  styleUrls: ['./exit-calls-table-cell-renderer.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="exit-calls-cell" [htTooltip]="exitCallsTooltip">
+      <span class="exit-calls-count">{{ this.apiExitCalls }}</span>
+
+      <ng-template #exitCallsTooltip>
+        <ng-container *ngIf="this.apiExitCalls > 0">
```
arjunlalb, line 32: This is an if-else use case. You can clean this up like. There are many examples in the codebase.

```
<ng-container *ngIf="this.apiExitCalls > 0; else noExitCalls">
  Content goes here
</ng-container>
<ng-container #noExitCalls></ng-container>
```

arjunlalb (commented):
LGTM

aaron-steinfeld:
lgtm other than existing comment on test

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.test.ts
aaron-steinfeld: Nevermind, discussed offline - because this is a template that's not rendered in the scope of this component, it's ugly to test and not worth it.

aaron-steinfeld (approved):

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  public constructor(
+    @Inject(TABLE_COLUMN_CONFIG) columnConfig: TableColumnConfig,
+    @Inject(TABLE_COLUMN_INDEX) index: number,
+    @Inject(TABLE_DATA_PARSER)
+    parser: TableCellParserBase<CellData, Trace, unknown>,
+    @Inject(TABLE_CELL_DATA) cellData: CellData,
+    @Inject(TABLE_ROW_DATA) rowData: Trace
+  ) {
+    super(columnConfig, index, parser, cellData, rowData);
+    const apiCalleeNameCount: string[][] = Object.entries(cellData.value[1]);
```
anandtiwary, line 69: - This name is confusing to me mainly because of 'Count' suffix. So  `apiCalleeNameCount` is a dictionary? 

can we call it `apiCalleeNameEntires` instead? 

- Is `[string, string][]` a better type? 

- totalCountOfDifferentApiCallee -> uniqueApiCallee

- Line 71. Why do we need to use `slice` here?

aaron-steinfeld: Agree naming is confusing/type suggestions. Line 71 is meant to limit the amount of pairs to show in the tooltip, but again naming would help clarify (`maxShowApiCalleeNameCount` as a const value could be `MAX_API_CALLEE_TO_SHOW`, for example).

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  TABLE_CELL_DATA,
+  TABLE_COLUMN_CONFIG,
+  TABLE_COLUMN_INDEX,
+  TABLE_DATA_PARSER,
+  TABLE_ROW_DATA
+} from '@hypertrace/components';
+import { Trace } from '@hypertrace/distributed-tracing';
+import { ObservabilityTableCellType } from '../../observability-table-cell-type';
+
+interface CellData {
```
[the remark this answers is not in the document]
anandtiwary: Did this change recently? 

I remember using composite spec where I have used value as array directly(let me find it). And regardless, imo the specification should reflect this object structure. 
```

export interface CompositeSpecification extends Specification {
  extractFromServerData(resultContainer: Dictionary<unknown>): unknown[];
}
```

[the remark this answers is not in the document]
anandtiwary:
```
      {
        type: 'table-widget-column',
        title: 'Risk',
        display: SecurityTableCellType.RiskScore,
        // tslint:disable-next-line: no-invalid-template-strings
        visible: '${showRiskColumn}',
        value: {
          type: 'composite-specification',
          specifications: [
            {
              type: 'attribute-specification',
              attribute: 'riskScore'
            },
            {
              type: 'attribute-specification',
              attribute: 'riskScoreCategory'
            },
            {
              type: 'attribute-specification',
              attribute: 'likelihoodElements'
            },
            {
              type: 'attribute-specification',
              attribute: 'impactElements'
            }
          ],
          'order-by': 'riskScore'
        },
        sort: TableSortDirection.Descending
      }
 ```
 And its parser. 
 ```
 import { TableCellParser, TableCellParserBase } from '@hypertrace/components';
import { SecurityTableCellType } from '../security-table-cell-type';

@TableCellParser({
  type: SecurityTableCellType.RiskScore
})
export class RiskScoreTableCellParser extends TableCellParserBase<
  CellData,
  RiskTableCellParsedData | undefined,
  number | undefined
> {
  public parseValue(cellData: CellData): RiskTableCellParsedData | undefined {
    if (Array.isArray(cellData) && cellData.length === 4) {
      return {
        score: cellData[0],
        category: cellData[1],
        likelihoodElements: cellData[2],
        impactElements: cellData[3]
      };
    }

    return undefined;
  }

  public parseFilterValue(cellData: CellData): number | undefined {
    return this.parseValue(cellData)?.score;
  }
}

type CellData = [number, string, string[], string[]];

export interface RiskTableCellParsedData {
  score: number;
  category: string;
  likelihoodElements: string[];
  impactElements: string[];
}

```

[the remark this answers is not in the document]
anandtiwary: Possibly, table is calling parser's `parseValue` is called after stripping units.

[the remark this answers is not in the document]
aaron-steinfeld: This does seem a bit suspicious - the wrapping object should only be there if units are defined, which they shouldn't be for a composite spec.

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
anandtiwary: Let's also rename `maxShowApiCalleeNameCount` to `MAX_API_CALLEE_TO_SHOW`

projects/observability/src/shared/components/table/data-cell/exit-calls/exit-calls-table-cell-renderer.component.ts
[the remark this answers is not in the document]
anandtiwary: discussed offline with Aaron. I think the implemented structure is fine for now.

PR #768, opened 2021-04-13
feat: adding default interval for cartesian

anandtiwary:
## Description
feat: adding default interval for cartesian

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/components/interval-select/interval-select.component.ts
+++ b/projects/observability/src/shared/components/interval-select/interval-select.component.ts
@@ -23,7 +23,7 @@ import { SelectOption, SelectSize } from '@hypertrace/components';
 })
 export class IntervalSelectComponent implements OnChanges {
   @Input()
-  public interval: IntervalValue = 'AUTO';
+  public interval?: IntervalValue;
 
   @Input()
   public intervalOptions?: IntervalValue[];
--- a/projects/observability/src/shared/components/interval-select/interval-select.component.ts
+++ b/projects/observability/src/shared/components/interval-select/interval-select.component.ts
@@ -23,7 +23,7 @@ import { SelectOption, SelectSize } from '@hypertrace/components';
 })
 export class IntervalSelectComponent implements OnChanges {
   @Input()
-  public interval?: IntervalValue;
+  public interval: IntervalValue = 'AUTO';
 
   @Input()
   public intervalOptions?: IntervalValue[];
@@ -54,8 +54,9 @@ export class IntervalSelectComponent implements OnChanges {
     }
   }
 
-  public onIntervalChange(value: IntervalValue): void {
-    this.intervalChange.next(value);
+  public onIntervalChange(interval: IntervalValue): void {
+    this.interval = interval;
+    this.intervalChange.next(interval);
   }
 
   public isDisabledOrUnselectable(): boolean {
--- a/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
+++ b/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
@@ -58,13 +58,16 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
     return this.model.getDataFetcher().pipe(
       tap(fetcher => {
         this.fetcher = fetcher;
+        const defaultInterval =
+          this.model.defaultInterval?.value !== undefined ? this.model.defaultInterval.getDuration() : undefined;
 
         if (this.intervalSupported()) {
           this.intervalOptions = this.buildIntervalOptions();
-          this.selectedInterval = this.getBestIntervalMatch(this.intervalOptions, this.selectedInterval);
+          this.selectedInterval =
+            defaultInterval ?? this.getBestIntervalMatch(this.intervalOptions, this.selectedInterval);
         } else {
           this.intervalOptions = undefined;
-          this.selectedInterval = undefined;
+          this.selectedInterval = defaultInterval;
         }
       }),
       switchMap(() => this.buildDataObservable())
@@ -87,7 +90,7 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
   }
 
   private intervalSupported(): boolean {
-    return this.model.selectableInterval;
+    return this.model.selectableInterval || this.model.defaultInterval !== undefined;
   }
 
   private resolveInterval(value?: IntervalValue): TimeDuration {
--- a/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
+++ b/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
@@ -58,13 +58,14 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
     return this.model.getDataFetcher().pipe(
       tap(fetcher => {
         this.fetcher = fetcher;
-        const defaultInterval =
-          this.model.defaultInterval?.value !== undefined ? this.model.defaultInterval.getDuration() : undefined;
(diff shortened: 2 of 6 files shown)


projects/observability/src/shared/components/interval-select/interval-select.component.ts
```diff
@@ -23,7 +23,7 @@ import { SelectOption, SelectSize } from '@hypertrace/components';
 })
 export class IntervalSelectComponent implements OnChanges {
   @Input()
-  public interval?: IntervalValue;
+  public interval: IntervalValue = 'AUTO';
```
aaron-steinfeld, line 26: Why change this default? It's a breaking change and I'm not sure what it adds since the widget is providing a default if it wants to.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -58,13 +58,16 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
     return this.model.getDataFetcher().pipe(
       tap(fetcher => {
         this.fetcher = fetcher;
+        const defaultInterval =
+          this.model.defaultInterval?.value !== undefined ? this.model.defaultInterval.getDuration() : undefined;
 
         if (this.intervalSupported()) {
           this.intervalOptions = this.buildIntervalOptions();
-          this.selectedInterval = this.getBestIntervalMatch(this.intervalOptions, this.selectedInterval);
+          this.selectedInterval =
```
aaron-steinfeld, line 66: The default behavior belongs inside `getBestIntervalMatch` for two reasons:
1. in this implementation, a default interval is being given precedence over a selected interval
2. That method enforces that the requested interval has an allowed resolution. Even if configured, we need to reject for example a 15s resolution requested at a a 1 week interval.

I think this just looks like
 ```ts
this.selectedInterval = this.getBestIntervalMatch(this.intervalOptions, this.selectedInterval ?? this.model.defaultInterval?.getDuration());
```

File: projects/observability/src/shared/components/interval-select/interval-select.component.ts (hunk 1 above)
Comment (anandtiwary) [reply projects/observability/src/shared/components/interval-select/interval-select.component.ts:26]: Yeah, I can revert it.

File: projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts [hunk 3]
```diff
@@ -58,13 +58,16 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
     return this.model.getDataFetcher().pipe(
       tap(fetcher => {
         this.fetcher = fetcher;
+        const defaultInterval =
+          this.model.defaultInterval?.value !== undefined ? this.model.defaultInterval.getDuration() : undefined;
```
anandtiwary, line 62: `this.model.defaultInterval?.value !== undefined ? this.model.defaultInterval.getDuration() : undefined;`

One weird thing i noticed is that our model library creates empty model object even when we don't define them in the model json. It means I can't used something like this ` this.model.defaultInterval?.getDuration()`

aaron-steinfeld: Not by default it shouldn't - the newly added property requests a default instance via the `defaultModelClass` key:

```
defaultModelClass: TimeDurationModel
```
from docs:
```ts
export interface ModelModelPropertyTypeInstance extends ModelPropertyTypeInstance {
    /**
     * If the property is undefined, by default, use this class
     */
    defaultModelClass?: Constructable<object>;
}
```

anandtiwary: So i can safely remove this property then?

aaron-steinfeld: yep, it's there explicitly to create a default instance if none is provided, which is the opposite of what you want.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -58,13 +58,17 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
     return this.model.getDataFetcher().pipe(
       tap(fetcher => {
         this.fetcher = fetcher;
+        const defaultInterval = this.model.defaultInterval?.getDuration();
 
         if (this.intervalSupported()) {
           this.intervalOptions = this.buildIntervalOptions();
-          this.selectedInterval = this.getBestIntervalMatch(this.intervalOptions, this.selectedInterval);
+          this.selectedInterval = this.getBestIntervalMatch(
```
anandtiwary, line 65: With this, if `defaultInterval` is not part of one of the supported options then it would choose AUTO. I think this is acceptable for now.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -58,13 +58,17 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
...
+            this.intervalOptions,
+            this.selectedInterval ?? defaultInterval
+          );
         } else {
           this.intervalOptions = undefined;
-          this.selectedInterval = undefined;
+          this.selectedInterval = defaultInterval;
```
aaron-steinfeld, line 71: If interval isn't selectable (which I'm guessing is what `intervalSupported()` represents), does this mean we're setting the interval without verifying that it's allowable for the time range?  It almost feels like this if elsecan change to something like
```ts
          const intervalOptions = this.buildIntervalOptions();
          this.selectedInterval = this.getBestIntervalMatch(
             intervalOptions,
             this.selectedInterval ?? defaultInterval
          );
          if (this.isIntervalSelectable()) {
            this.intervalOptions = intervalOptions; // The only thing this flag controls is whether options are available (and thus, the selector)
          }
```

projects/observability/src/shared/components/interval-select/interval-select.component.ts
anandtiwary: done

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
anandtiwary: done

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
anandtiwary: done

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
anandtiwary: Yeah. Isn't that acceptable behavior? IMO, we should support setting any arbitrary interval explicitly by the user.

anandtiwary: I think if it is a nit issue then LGTM. we can tackle this later.

aaron-steinfeld: Nope - the problem is that at high resolutions, it will put extra load on the system and/or be an invalid request. So since the system can't support it, we shouldn't here either.

aaron-steinfeld (approved):

PR #769, opened 2021-04-13
fix: clicking on checkbox does not select the option

palbizu:
## Description

Prevent the default action in the checkbox located in the multi-select component to prevent the checkbox element from being cleared.

### Testing

Visual Testing

### Checklist:

### Documentation


projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -92,7 +92,11 @@ import { MultiSelectJustify } from './multi-select-justify';
                 (click)="this.onSelectionChange(item)"
                 class="multi-select-option"
               >
-                <ht-checkbox class="checkbox" [checked]="this.isSelectedItem(item)"></ht-checkbox>
+                <ht-checkbox
```
aaron-steinfeld, line 95: We actually already have a component that handles this for us - look for `<ht-event-blocker event="click">`

aaron-steinfeld: But now that I look at this isn't the issue that we're listening on the parent to broadcast the change? So wouldn't blocking this event prevent the broadcast?

palbizu: In this case, I'm calling to prevent-default function to block the click on the checkbox allowing to call the parent function because the values are set from the parent component.

anandtiwary: we need to use checkedChange right?

anandtiwary: And possibly put `<ht-event-blocker event="click">` inside checkbox?

aaron-steinfeld: Clarified offline - we're preventing default but not stopping propagation. Given that, the event blocker is probably more of a pain to use (we'd have to wire up checkbox's change listener in addition to the outer click listener), so I'm ok sticking with this approach - but let's clarify the naming a bit: `(click)="this.preventClickDefault($event)"` or similar (since propagation isn't being blocked)

anandtiwary: @aaron-steinfeld would one of the following also work? 
```
- Put ht-event-blocker inside ht-checkbox component. then click won’t be propagated. Now use `checkedChange` output in multi select to call appropriate selection method
- Inside multi select, wrap  `<ht-event-blocker><ht-checkbox></ht-checkbox></ht-event-blocker> `. Now use `checkedChange`  output in multi select to call appropriate selection method
```

palbizu: I tested both approaches and they are blocking the propagation of the checkbox, so the  checkbox value is not updated

aaron-steinfeld: Either should work - but requires using both a click listener (to get label clicks) and the checkedChange to get checkbox clicks. This is what I was getting at above - it's doable, but just using prevent default ends up cleaner in this situation. The real ideal fix is a bigger overhaul of checkbox to support more complex labels so it handles the whole click internally, but out of scope IMO.

palbizu: done

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -92,7 +92,11 @@ import { MultiSelectJustify } from './multi-select-justify';
...
+                  class="checkbox"
+                  (click)="preventClickDefault($event)"
```
aaron-steinfeld, line 97:
```suggestion
                  (click)="this.preventClickDefault($event)"
```

projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -138,6 +138,27 @@ describe('Multi Select Component', () => {
...
+          options: selectionOptions,
+          selected: []
+        }
+      }
+    );
+
+    spectator.tick();
+    spectator.click('.trigger-content');
+    const selectedCheckboxElement = spectator.queryAll('ht-checkbox', { root: true })[0];
+    expect(spectator.dispatchFakeEvent(selectedCheckboxElement, 'click', true).defaultPrevented).toBe(true);
```
aaron-steinfeld, line 159: This part looks good, but can we add a second validation here that `selectedChange` emits similar to the "should notify and update selection when selection is changed" test - we want to both make sure we're blocking this, but still processing the selection.

palbizu: done. Thanks!

projects/components/src/multi-select/multi-select.component.ts
palbizu: fixed

aaron-steinfeld (approved):

anandtiwary (approved):

PR #770, opened 2021-04-13
chore: update to jfrog

aaron-steinfeld:
## Description
Bumping dependency fetching to use jfrog instead of bintray


anandtiwary (dismissed):

aaron-steinfeld:
nuking the package-lock is a bad idea.

skjindal93 (approved):

PR #774, opened 2021-04-15
fix: hover over a selected table changed color

palbizu:
## Description
The background color and borders were changed in hover over on a selected table.

### Testing
Visual Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/table/table.component.scss
+++ b/projects/components/src/table/table.component.scss
@@ -49,9 +49,9 @@
     border-top: 1px solid $blue-2;
 
     &.hovered-row {
-      background: $blue-2;
-      border-bottom: 1px solid $blue-5;
-      border-top: 1px solid $blue-5;
+      background: $gray-1;
+      border-bottom: 1px solid $gray-2;
+      border-top: 1px solid $gray-2;
     }
   }
 


projects/components/src/table/table.component.scss
```diff
@@ -49,9 +49,9 @@
     border-top: 1px solid $blue-2;
 
     &.hovered-row {
-      background: $blue-2;
```
anandtiwary, line 52: This is for hover on a selected row. Has the ux changed for this? At all other place, we make the color darker on hover. This would make it inconsistent with similar behavior on a card list

michael-traceable: Yeah. So hover on these rows right now is a little off. We go darker on buttons typically, but with rows and other navigation items the hover I think would be better suited to always be the same. 

Alternatively, I could accept that a hover on a selected row doesn't change anything. Instead it stays as a selected state.

anandtiwary: I am fine with whatever you decide on visual aspects. I just want to share that this change require similar changes to
- Sequence chart: d3 row hover. Without this waterfall would appear broken.
- Card lists

@michael-traceable Could you confirm we need to apply styling at these two locations too?

michael-traceable: Good call! Yes, waterfall should match because it's similar structure (table format)... The card list we can match as well, however, I believe I also have a ticket that the hover on that should just be the shadow and no gray, but it doesn't bother me too much.

jake-bassett: Should we bundle all of these changes at once? I kind of lean towards yes so there are no inconsistencies.

anandtiwary: I agree. @palbizu Can we bundle all the changes together in the same PR?

palbizu: @anandtiwary just to double-check. I should apply the same pattern to
1- Tables (I already did it in this PR)

michael-traceable: Yes... but with 3, please see the intended hover for a card item. https://www.figma.com/file/tVP4INp6sTDVPx6H7y5efk/%F0%9F%93%90-Components?node-id=2841%3A1

image

PR #785, opened 2021-04-19
feat: adding rows count to textarea

palbizu:
## Description
- Adding rows count feature to Textarea component. 
- Adding minHeight feature to Textarea component. 

### Testing
Visual Testing

Old:

image

New:

image

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/textarea/textarea.component.test.ts
+++ b/projects/components/src/textarea/textarea.component.test.ts
@@ -33,16 +33,16 @@ describe('Textarea Component', () => {
     expect(logSpy).not.toHaveBeenCalled();
   });
 
-  test('should apply disabled attribute when disabled', () => {
-    spectator = createHost(`<ht-textarea [placeholder]="placeholder" [disabled]="disabled"></ht-textarea>`, {
+  test('should match the rowsCount with the number of rows in the textarea', () => {
+    const rows = 10;
+
+    spectator = createHost(`<ht-textarea [placeholder]="placeholder" [rowsCount]="rowsCount"></ht-textarea>`, {
       hostProps: {
         placeholder: 'TEST',
-        disabled: true
+        rowsCount: rows
       }
     });
 
-    const disabled = spectator.query('textarea')!.getAttribute('ng-reflect-disabled');
-
-    expect(disabled).toBe('true');
+    expect((spectator.query('.textarea') as HTMLTextAreaElement).rows).toBe(rows);
   });
 });
--- a/projects/components/src/textarea/textarea.component.test.ts
+++ b/projects/components/src/textarea/textarea.component.test.ts
@@ -33,6 +33,19 @@ describe('Textarea Component', () => {
     expect(logSpy).not.toHaveBeenCalled();
   });
 
+  test('should apply disabled attribute when disabled', () => {
+    spectator = createHost(`<ht-textarea [placeholder]="placeholder" [disabled]="disabled"></ht-textarea>`, {
+      hostProps: {
+        placeholder: 'TEST',
+        disabled: true
+      }
+    });
+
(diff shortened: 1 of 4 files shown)


projects/components/src/textarea/textarea.component.ts
```diff
@@ -29,6 +30,9 @@ export class TextareaComponent implements OnInit {
   @Input()
   public disabled: boolean | undefined;
 
+  @Input()
+  public rowsCount: number = 0;
```
aaron-steinfeld, line 34: nit: `rowCount` (or `rows`)

palbizu: done. I need to replace the default count because 0 was failing.

projects/components/src/textarea/textarea.component.ts
```diff
@@ -29,6 +30,9 @@ export class TextareaComponent implements OnInit {
   @Input()
   public disabled: boolean | undefined;
 
+  @Input()
+  public rows: number = -1;
```
anandtiwary, line 34: is '-1' a legit input for `mat text area`? It seems odd.

anandtiwary: what if we keep the default value of rows to '0' ?

palbizu: I didn't have problems with this value. I did with 0. Probably I should use a positive value as a default.

palbizu: If I use '0' the test threw an " IndexSizeError " DOMException

aaron-steinfeld: or just `undefined` should let the browser default hold, right?

palbizu: Good advise!. Done

aaron-steinfeld (approved):

projects/components/src/textarea/textarea.component.ts
palbizu: Using rows as undefined generates  the next error

palbizu: I think the best option is to use 2 as default since it is the default that textarea has by definition.

projects/components/src/textarea/textarea.component.ts
```diff
@@ -29,6 +31,12 @@ export class TextareaComponent implements OnInit {
   @Input()
   public disabled: boolean | undefined;
 
+  @Input()
+  public rows: number = 2;
+
+  @Input()
+  public minHeight?: number;
```
aaron-steinfeld, line 38: Rather than having an explicit input, can we have the text area stick to the size of its parents so you can style the `<ht-textarea>` the same way you would style it directly? In other words, assigning a minHeight to the component is the desired way of accomplishing this rather than pushing styles through inputs.

palbizu: Done. Let me know your thoughts about my approach.

aaron-steinfeld:
So my main goal here is that we can style our component in the same way we'd style the text area if we were using it directly. Inheriting individual properties is a bit brittle, but if we can't come up with a better way I think it's better than the input.

projects/components/src/textarea/textarea.component.test.ts
```diff
@@ -45,4 +45,17 @@ describe('Textarea Component', () => {
 
     expect(disabled).toBe('true');
   });
+
+  test('should match the rowsCount with the number of rows in the textarea', () => {
```
aaron-steinfeld, line 49:
```suggestion
  test('should match the rows with the number of rows in the textarea', () => {
```

aaron-steinfeld (approved):

aaron-steinfeld (approved):

PR #787, opened 2021-04-20
feat: string enum table cell renderer

jake-bassett:
## Description
Added table cell renderer to format string enums into more human readable text.

### DIFF
--- a/projects/components/src/table/cells/data-renderers/enum/string-enum-table-cell-renderer.component.ts
+++ b/projects/components/src/table/cells/data-renderers/enum/string-enum-table-cell-renderer.component.ts
@@ -1,0 +1,27 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { TableCellRenderer } from '../../table-cell-renderer';
+import { TableCellRendererBase } from '../../table-cell-renderer-base';
+import { CoreTableCellParserType } from '../../types/core-table-cell-parser-type';
+import { CoreTableCellRendererType } from '../../types/core-table-cell-renderer-type';
+import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
+
+@Component({
+  selector: 'ht-string-enum-table-cell-renderer',
+  styleUrls: ['./string-enum-table-cell-renderer.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div
+      [ngClass]="{ clickable: this.clickable, 'first-column': this.isFirstColumn }"
+      class="string-enum-cell"
+      [htTooltip]="this.value"
+    >
+      {{ this.value | htDisplayStringEnum }}
+    </div>
+  `
+})
+@TableCellRenderer({
+  type: CoreTableCellRendererType.StringEnum,
+  alignment: TableCellAlignmentType.Left,
+  parser: CoreTableCellParserType.String
+})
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -52,6 +52,21 @@ h1 {
   box-sizing: border-box;
 }
 
+strong {
```
jake-bassett, line 55: Unrelated to this change, but needed for another.

projects/components/src/table/cells/data-renderers/enum/string-enum-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,27 @@
...
+import { CoreTableCellRendererType } from '../../types/core-table-cell-renderer-type';
+import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
+
+@Component({
+  selector: 'ht-string-enum-table-cell-renderer',
+  styleUrls: ['./string-enum-table-cell-renderer.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div
+      [ngClass]="{ clickable: this.clickable, 'first-column': this.isFirstColumn }"
```
arjunlalb, line 14: what advantage do we get compared to directly supplying the startcased enum value to a regular string cell renderer?

projects/components/src/table/cells/data-renderers/enum/string-enum-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,27 @@
...
+      {{ this.value | htDisplayStringEnum }}
+    </div>
+  `
+})
+@TableCellRenderer({
+  type: CoreTableCellRendererType.StringEnum,
+  alignment: TableCellAlignmentType.Left,
+  parser: CoreTableCellParserType.String
+})
+export class StringEnumTableCellRendererComponent extends TableCellRendererBase<string> {}
```
anandtiwary, line 27: nit: IMO, the name 'String Enum' is not very clear. Or we can at least add similar comments from `htDisplayStringEnum` here too

```
  // This removes dashes and underscores and gives all words initial caps

  // We only want the first word capitalized.
```

src/styles/main.scss
```diff
@@ -52,6 +52,21 @@ h1 {
...
+}
+
+em {
+  font-style: normal;
+  background: $gray-2;
+  border-radius: 4px;
+  padding: 2px 4px;
+}
+
+q {
```
arjunlalb, line 66: what are these for?

projects/components/src/table/cells/data-renderers/enum/string-enum-table-cell-renderer.component.ts
jake-bassett: Nothing other than this does the text presentation for you. It has been coming up more and more, so this is a convenience, but it also feels (at least to me) more correct to have the presentation logic in the cell renderer rather than pre-formatting it before passing it in.

projects/components/src/table/cells/data-renderers/enum/string-enum-table-cell-renderer.component.ts
jake-bassett: I always struggle with names, but also always open to suggestions.

I've added the comment in my local. That should help some. Not worth holding up this PR for, but I'll commit them in whatever next HT PR I have.

PR #794, opened 2021-04-24
fix: reset paging if table filters or sorts

aaron-steinfeld:
## Description
Tables should switch page to page one if the user searches, applies a sort or filters. Currently, any offset is preserved meaning a search when on page two will show you page two of search results etc. The cause of this is three parts - 
1. the paginator, which actually pushes changes to the data, only pushed if the change was caused by the component itself, not if the inputs to the paginator changed.
2. The table only pushed changes to the paginator when the url/route changed
3. Because changes emitting from the paginator never made it back to the paginator's inputs, change detection in the reset case wasn't firing (because the reset page was equivalent to the last page input)

### Testing
Visual/E2E

### DIFF
--- a/projects/components/src/paginator/paginator.component.ts
+++ b/projects/components/src/paginator/paginator.component.ts
@@ -9,7 +9,7 @@ import {
 } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
 import { TypedSimpleChanges } from '@hypertrace/common';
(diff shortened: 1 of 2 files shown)


jake-bassett (approved):
Could have sworn we fixed this once before.

aaron-steinfeld:
> Could have sworn we fixed this once before.

Probably!

PR #800, opened 2021-04-26
feat: adding badge to tab header

itssharmasandeep:
## Description
Adding badge to tab header

### Testing
Local testing done.

When tab is not selected
Screenshot 2021-04-26 at 12 50 09 PM

When tab is selected
Screenshot 2021-04-26 at 12 50 34 PM


projects/assets-library/assets/styles/_color-palette.scss
```diff
@@ -84,6 +84,8 @@ $cloud-3: #a1c5e0;
 
 $brown-1: #9e4c41;
 
+$grey-200: #e5e6e6;
```
arjunlalb, line 87: we no longer use the colors with suffix 100, 200 etc. That used to be the case earlier. We converted all such values into a correspoding gray-1, 2 ... n pattern when we redesigned. This might be a remnant from the earlier design system. All the old gray-200 values are now gray-1 in the system. You can either use that if its suitable or consult with @michael-traceable to get a suitable color for this.

projects/components/src/tabs/content/tab-group.component.scss
```diff
@@ -75,9 +75,25 @@
...
+  flex-direction: row;
+  align-items: center;
+
+  .sub-text {
+    @include font-label($gray-5);
+    display: flex;
+    justify-content: center;
+    text-align: center;
+    margin-left: 6px;
+    width: 20px;
```
arjunlalb, line 87: fixed width might not scale well if the numbers are big right? I'd recommend against using fix width. Use padding to fill the area around.

itssharmasandeep: fixed

projects/assets-library/assets/styles/_color-palette.scss
itssharmasandeep: Currently changed this to `$grey-1`

projects/assets-library/assets/styles/_color-palette.scss
```diff
@@ -84,6 +84,8 @@ $cloud-3: #a1c5e0;
 
 $brown-1: #9e4c41;
 
+$grey-1: #e5e6e6;
```
anandtiwary, line 87: We already have these colors below. The color you shared doesn't belong to this list. @michael-traceable Could you please chime in here?
```
$gray-1: #f4f5f5;
$gray-2: #e1e4e5;
$gray-3: #b7bfc2;
$gray-4: #889499;
$gray-5: #657277;
$gray-6: #40474a;
$gray-7: #272c2e;
$gray-8: #171a1c;
$gray-9: #080909;
```

projects/components/src/tabs/content/tab-group.component.ts
```diff
@@ -10,7 +10,16 @@ import { TabComponent } from './tab/tab.component';
       <mat-tab-group animationDuration="0ms" disableRipple (selectedTabChange)="this.activeTabIndex = $event.index">
         <mat-tab *ngFor="let tab of this.tabs; index as i">
           <ng-template mat-tab-label>
-            <div class="tab-label">{{ tab.label }}</div>
+            <div class="tab-label">
+              {{ tab.label }}
+              <div
+                *ngIf="tab.subText && tab.subText !== undefined"
```
anandtiwary, line 16: Just `tab.subText` this condition should be sufficient right?

projects/components/src/tabs/content/tab/tab.component.ts
```diff
@@ -10,4 +10,7 @@ import { ContentHolder, CONTENT_HOLDER_TEMPLATE } from '../../../content/content
 export class TabComponent extends ContentHolder {
   @Input()
   public label!: string;
+
+  @Input()
+  public subText?: string;
```
anandtiwary, line 15: should we call it `summaryText` ? Or what was the other word you mentioned?  `chipText` ?

projects/assets-library/assets/styles/_color-palette.scss
michael-traceable: These look correct, yes.

michael-traceable: Actually, $gray-6 should be #3F474A. I wonder where that discrepancy came from

projects/assets-library/assets/styles/_color-palette.scss
arjunlalb: I meant we should use one of the existing colors from the color palette. A `$gray-1` already exists.

michael-traceable: Blargh... If it's used... it should be `gray-2`... if it's not used, then just remove it.

projects/components/src/tabs/content/tab-group.component.ts
jake-bassett: Yep. That should work.

projects/components/src/tabs/content/tab/tab.component.ts
jake-bassett: Is this actually subtext? From the screenshot it looks more like a badge.

jake-bassett:
This looks more like a badge than sub-text. Can we rename the code appropriately?

projects/components/src/tabs/content/tab/tab.component.ts
itssharmasandeep: Yeah, The name was open for suggestion. Thanks!
I have fixed this

projects/components/src/tabs/content/tab-group.component.ts
itssharmasandeep: fixed!!

projects/assets-library/assets/styles/_color-palette.scss
itssharmasandeep: @anandtiwary @jake-bassett do we want to make this change as well?
So this `$gray-6` is being used at many places.

projects/assets-library/assets/styles/_color-palette.scss
itssharmasandeep: Yeah, made the changes

projects/assets-library/assets/styles/_color-palette.scss
arjunlalb: we have $gray-6 used at about 7-8 components today. we might want to review them before updating it. might be worth doing as a separate PR.

jake-bassett: Yeah, let's run with what we feel is correct as we know it today. If we break some colors in the mean time, that's fine. They weren't necessarily correct to begin with and its not worth blocking this PR for.

Let's create a ticket to do a follow up audit on those as @arjunlalb suggests. Do you mind doing that work next, @itssharmasandeep? Should be quick.

jake-bassett (approved):

arjunlalb (approved):

projects/assets-library/assets/styles/_color-palette.scss
itssharmasandeep: Yeah, I can pickup this.

PR #801, opened 2021-04-26
feat: adding note icon

itssharmasandeep:
## Description
Adding note icon

### Testing
Local testing done.

Screenshot 2021-04-26 at 11 45 47 AM


projects/assets-library/src/icons/icon-type.ts
```diff
@@ -78,6 +79,7 @@ export const enum IconType {
   NoData = 'svg:noData',
   Node = 'node',
   None = 'none',
+  Note = 'svg:note',
```
arjunlalb, line 82: both svgs `page` and `note` looks to be the same. If it already exists, could we reuse that and avoiding a new `note` icon?

itssharmasandeep: They are same, I liked `note` better name for that icon and more generic as it is used as `description` icon in the security tab inside settings.
I will be changing that on the traceable side once this is merged.

arjunlalb (approved):

PR #803, opened 2021-04-27
fix: waterfall undefined strings

jake-bassett:
## Description
This fixes undefined strings in several places for waterfall.

### DIFF
--- a/projects/components/src/sequence/sequence-chart.component.ts
+++ b/projects/components/src/sequence/sequence-chart.component.ts
@@ -115,7 +115,6 @@ export class SequenceChartComponent implements OnChanges {
       id: segment.id,
       start: segment.start - minStart,
       end: segment.end - minStart,
-      label: segment.label,
       color: segment.color
     }));
   }
--- a/projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
+++ b/projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
@@ -103,14 +103,14 @@ export class TraceDetailService implements OnDestroy {
   private buildTimeString(trace: Trace, units: string): string {
     return `${new DateFormatter({ mode: DateFormatMode.DateAndTimeWithSeconds }).format(
       trace.startTime as number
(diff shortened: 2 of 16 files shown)


projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
```diff
@@ -103,14 +103,14 @@ export class TraceDetailService implements OnDestroy {
...
-    )} for ${trace.duration as string} ${units}`;
+    )} for ${trace.duration as string} ${units ?? ''}`.trim();
   }
 
   private buildTitleString(trace: Trace): string {
     if (trace.spans?.length === 1) {
       const entrySpan = trace.spans[0];
 
-      return `${entrySpan.serviceName as string} ${entrySpan.displaySpanName as string}`;
+      return `${entrySpan.serviceName as string} ${(entrySpan.displaySpanName as string) ?? ''}`.trim();
```
palbizu, line 113: Wouldn't it be correct to do the trim() before generating the string?

projects/components/src/sequence/sequence-chart.component.ts
```diff
@@ -115,7 +115,6 @@ export class SequenceChartComponent implements OnChanges {
       id: segment.id,
       start: segment.start - minStart,
       end: segment.end - minStart,
-      label: segment.label,
```
jake-bassett, line 118: I couldn't see anywhere this was used, so removed it. cc: @anandtiwary to confirm this is okay.

palbizu (approved):
Looks good!. I left a comment to check beforehand to merge it.

projects/distributed-tracing/src/pages/trace-detail/trace-detail.service.ts
jake-bassett: In this case since we also have a space in the middle of the template, I `trim()` once the complete string is built to remove any space between (since there might not be a `displaySpanName`).

arjunlalb (approved):
LGTM

PR #806, opened 2021-04-30
feat: header functionality for the list view

itssharmasandeep:
## Description
Header functionality for the list view

### Testing
Local testing done

### Checklist:

Screenshot 2021-04-30 at 12 34 48 PM

### DIFF
--- a/projects/components/src/list-view/list-view.component.scss
+++ b/projects/components/src/list-view/list-view.component.scss
@@ -7,6 +7,33 @@
   display: flex;
   flex-direction: column;
 
+  .header {
+    flex: 0 0 auto;
+    height: 32px;
+    width: 100%;
+    padding: 8px 0;
+    background: $gray-1;
+    box-shadow: inset 0px -1px 0px $gray-2;
+    border-radius: 6px 6px 0px 0px;
+    display: flex;
+    align-items: center;
+
+    .header-key {
+      flex: 1 1 auto;
+      font-weight: 600;
+      font-size: 12px;
+      line-height: 14px;
+      margin-left: 12px;
+      text-transform: uppercase;
+    }
+    .header-key:nth-child(1) {
+      width: 40%;
+    }
+    .header-key:nth-child(2) {
+      width: 60%;
+    }
+  }
+
   .data-row {
     flex: 0 0 auto;
     min-height: 40px;
--- a/projects/components/src/list-view/list-view.component.scss
+++ b/projects/components/src/list-view/list-view.component.scss
@@ -1,4 +1,17 @@
+@import 'font';
 @import 'color-palette';
+
+$key-width: 40%;
+$value-width: 60%;
+$margin-left: 12px;
+
+@mixin grid-view {
+  width: 100%;
+  padding: 8px 0;
+  display: grid;
+  grid-template-columns: $key-width $value-width;
+  align-content: center;
+}
 
 .list-view {
   height: 100%;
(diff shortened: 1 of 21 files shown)


projects/components/src/list-view/list-view.component.scss
```diff
@@ -7,6 +7,33 @@
   display: flex;
   flex-direction: column;
 
+  .header {
+    flex: 0 0 auto;
```
anandtiwary, line 11: `flex: 0 0 auto;` you probably don't need this if you are setting a fixed width/height

projects/components/src/list-view/list-view.component.scss
```diff
@@ -7,6 +7,33 @@
...
+    padding: 8px 0;
+    background: $gray-1;
+    box-shadow: inset 0px -1px 0px $gray-2;
+    border-radius: 6px 6px 0px 0px;
+    display: flex;
+    align-items: center;
+
+    .header-key {
+      flex: 1 1 auto;
+      font-weight: 600;
```
anandtiwary, line 23: Is there a mixin that we can use here?

projects/components/src/list-view/list-view.component.scss
```diff
@@ -7,6 +7,33 @@
...
+      font-size: 12px;
+      line-height: 14px;
+      margin-left: 12px;
+      text-transform: uppercase;
+    }
+    .header-key:nth-child(1) {
+      width: 40%;
```
anandtiwary, line 30: This should be in sync with the data row. Why not use css grids?  we can define a fixed column widths that both header and data row can share

projects/components/src/list-view/list-view.component.ts
```diff
@@ -6,6 +6,11 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div class="list-view">
+      <div *ngIf="this.headerKeys && this.headerKeys.length === 2" class="header">
```
anandtiwary, line 9: nit: `class="header-row"`

projects/components/src/list-view/list-view.component.scss
itssharmasandeep: I was looking for it but could not find.
I will check more

projects/components/src/list-view/list-view.component.scss
itssharmasandeep: I can refactor this whole css file using css grid.

projects/components/src/list-view/list-view.component.scss
itssharmasandeep: Okay, I will check for it.

projects/components/src/list-view/list-view.component.ts
```diff
@@ -18,6 +23,9 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
   `
 })
 export class ListViewComponent {
+  @Input()
+  public headerKeys?: [string, string];
```
anandtiwary, line 27: ListViewRecord also has `key` and `value` properties. So, instead of `headerKeys` should we have two properties: `keyHeaderLabel: string` and 'valueHeaderLabel: string`

itssharmasandeep: Yeah we can have that, But instead of making these two different inputs, we can just create and interface  with these two attributes and use that.

anandtiwary: Yeah, like 'headerLabel?: ListHeaderLabel;'

Interface ListHeaderLabel { key: string; value: string;}

itssharmasandeep: I guess,` interface ListHeader { keyLabel: string; valueLabel: string}` this would be better?

projects/components/src/list-view/list-view.component.ts
itssharmasandeep: Fixed!!

projects/components/src/list-view/list-view.component.ts
anandtiwary: Yeah. This is fine too!

projects/components/src/list-view/list-view.component.scss
```diff
@@ -1,46 +1,57 @@
...
+
 .list-view {
   height: 100%;
   width: 100%;
   overflow: auto;
   display: flex;
   flex-direction: column;
 
+  .header-row {
+    @include grid-view();
```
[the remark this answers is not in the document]
itssharmasandeep: Fixed!!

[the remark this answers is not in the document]
anandtiwary: Oh I was thinking more like this 

```
.list-view{
  height: 100%;
  width: 100%;
  overflow: auto;
  width: 100%;
  padding: 8px 0;
  display: grid;
  grid-template-columns: $key-width $value-width;
  align-content: center;
  
  
  .header-key-label{}
  
  .header-value-label{}
  
  .data-key{}
  
  .data-value{} 
}
```

Is this something we can do?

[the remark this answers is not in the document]
itssharmasandeep: But then we'll have to put some styles twice like
1.  Header shadow and background
2. Alternate background for data row and column etc.

projects/components/src/list-view/list-view.component.scss
```diff
@@ -1,46 +1,59 @@
...
   display: flex;
   flex-direction: column;
 
-  .data-row {
-    flex: 0 0 auto;
-    min-height: 40px;
-    width: 100%;
-    word-break: break-word;
-    padding: 8px 0;
+  & > * {
```
arjunlalb, line 23: list-view could have a huge number of children. In such cases, this does us more harm than good. We should use a class name here instead of the wildcard.

arjunlalb (approved):
LGTM.

projects/components/src/list-view/list-view.component.scss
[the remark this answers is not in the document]
itssharmasandeep: if needed for this, we can take this as follow up work.
I am merging changes for now.

Issue #749, opened 2021-04-06
Metrics are showing in a bucket of 15 seconds 


subintp:
Problem:
Dashboard is showing metrics grouped in 15 seconds. This is creating confusion in correlating Hypertrace metrics with other metricc sources. Ideally these metrics should be visualized per second or per miniute. This will help to correlate easily 

Screenshot:
<img width="1125" alt="Screenshot 2021-04-06 at 3 06 33 PM" src="https://user-images.githubusercontent.com/1742357/113691055-b0a3c000-96e9-11eb-8450-d6d11ea73e80.png">

JBAhire:
As per the offline discussion, the main pain point here is confusion because of inconsistency in time range of numerical and visual metrics. Numerical error rate metric here show number of errors per sec whereas every data point on visual metric shows number of error per 15 secs, which is confusing for some of the users. 

We came up with few solutions to address this:
- Add label on time axis clearly stating that this aggregations are on 15 secs
- Add time-range selector we used to have back in the days. That will signify that current time-range is `auto-15s` and user can choose from different time-ranges as per the requirements. 

The second option makes more sense to me in visual as well as use-case perspective and that's what we agreed on. 

Here's the screenshot from older version which shows that: 

![](https://camo.githubusercontent.com/96a3888c5695026b6c5a2e14bb0220ce61de23de78ec971823f1f5ee319a7127/68747470733a2f2f73332e616d617a6f6e6177732e636f6d2f687970657274726163652d646f63732f64617368626f6172642d312e706e67)

So, @anandtiwary and @itssharmasandeep can we please get it done?

cc: @subintp , @kotharironak

PR #807, opened 2021-05-03
fix: enable selectable time intervals for metrics across all screens

JBAhire:
## Description
Resolves https://github.com/hypertrace/hypertrace-ui/issues/749 by making changes as per the discussion here: https://github.com/hypertrace/hypertrace-ui/issues/749#issuecomment-829140138 . 

Enabling `selectable-interval` will show up in UI like this screenshot from earlier version:

image

### Testing
NA

### Checklist:

### DIFF
--- a/projects/observability/src/pages/apis/api-detail/overview/api-overview.dashboard.ts
+++ b/projects/observability/src/pages/apis/api-detail/overview/api-overview.dashboard.ts
@@ -198,7 +198,7 @@ export const apiOverviewDashboard: DashboardDefaultConfiguration = {
           {
             type: 'cartesian-widget',
             title: 'Latency',
-            'selectable-interval': false,
+            'selectable-interval': true,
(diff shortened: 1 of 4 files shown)


anandtiwary (approved):
Did we change it to false recently?

itssharmasandeep (approved):

JBAhire:
> Did we change it to false recently?

Actually, it has been false since long time. That change happened post launch. I think @arjunlalb , @aaron-steinfeld or @skjindal93 might have some additional context here.

skjindal93:
> > Did we change it to false recently?
> 
> Actually, it has been false since long time. That change happened post launch. I think @arjunlalb , @aaron-steinfeld or @skjindal93 might have some additional context here.

We changed it to false, because we felt that the interval selector was overcrowding the cartesians in overview and home dashboards

PR #815, opened 2021-05-03
feat: add navigation handling for inactive entities

jake-bassett:
## Description
This change allows navigation routes to know when an entity is inactive so it can be routed differently if needed.

### DIFF
--- a/projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
+++ b/projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
@@ -33,6 +33,9 @@ export class EntityRendererComponent implements OnChanges {
   public entity?: Entity;
 
   @Input()
+  public inactive: boolean = false;
+
+  @Input()
   public navigable: boolean = true;
 
   @Input()
@@ -59,8 +62,9 @@ export class EntityRendererComponent implements OnChanges {
       this.setIconType();
     }
   }
+
   public onClickNavigate(): void {
-    this.navigable && this.entity && this.entityNavService.navigateToEntity(this.entity);
+    this.navigable && this.entity && this.entityNavService.navigateToEntity(this.entity, this.inactive);
   }
 
   private setName(): void {
--- a/projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-parser.ts
+++ b/projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-parser.ts
@@ -1,14 +1,23 @@
 import { TableCellParser, TableCellParserBase, TableRow } from '@hypertrace/components';
 import { Entity, entityIdKey } from '../../../../graphql/model/schema/entity';
 import { ObservabilityTableCellType } from '../../observability-table-cell-type';
-import { parseEntityFromTableRow } from './entity-table-cell-renderer-util';
+import { isInactiveEntity, parseEntityFromTableRow } from './entity-table-cell-renderer-util';
 
 @TableCellParser({
   type: ObservabilityTableCellType.Entity
 })
 export class EntityTableCellParser extends TableCellParserBase<CellData, CellData, string | undefined> {
   public parseValue(cellData: CellData, row: TableRow): CellData {
-    return parseEntityFromTableRow(cellData, row);
+    const entity = parseEntityFromTableRow(cellData, row);
+
+    if (entity === undefined) {
+      return undefined;
+    }
+
+    return {
+      ...entity,
+      isInactive: isInactiveEntity(row) === true
+    };
   }
 
   public parseFilterValue(cellData: CellData): string | undefined {
@@ -16,4 +25,8 @@ export class EntityTableCellParser extends TableCellParserBase<CellData, CellDat
   }
 }
 
-type CellData = Entity | undefined;
+type CellData = MaybeInactiveEntity | undefined;
+
+export interface MaybeInactiveEntity extends Entity {
+  isInactive: boolean;
+}
--- a/projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer-util.ts
+++ b/projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer-util.ts
@@ -43,6 +43,6 @@ export const isInactiveEntity = (row: TableRow): boolean | undefined => {
 };
 
 const filterMetricAggregations = (row: TableRow): MetricAggregation[] =>
-  Object.values(row).filter(value => isMetricAggregation(value)) as MetricAggregation[];
+  Object.values(row).filter(isMetricAggregation);
 
 const isValidMetricAggregation = (metricAggregation: MetricAggregation): boolean => !isNull(metricAggregation.value);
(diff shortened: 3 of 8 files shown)


projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-parser.ts
```diff
@@ -1,19 +1,32 @@
...
-    return parseEntityFromTableRow(cellData, row);
+    const entity = parseEntityFromTableRow(cellData, row);
+
+    if (entity === undefined) {
+      return undefined;
+    }
+
+    return {
+      ...entity,
+      isInactive: isInactiveEntity(row) === true
```
aaron-steinfeld, line 19: *twitch*

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer.component.ts
```diff
@@ -1,15 +1,19 @@
...
 @Component({
   selector: 'ht-entity-table-cell-renderer',
   styleUrls: ['./entity-table-cell-renderer.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div class="fill-container entity-cell" [ngClass]="{ 'first-column': this.isFirstColumn }">
-      <ht-entity-renderer [entity]="this.value" [navigable]="true"></ht-entity-renderer>
+      <ht-entity-renderer
+        [entity]="this.value"
+        [inactive]="this.value?.isInactive === true"
```
aaron-steinfeld, line 14: Why not read this off the value that's already passed in?

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer-util.ts
```diff
@@ -27,3 +30,22 @@ export const parseEntityFromTableRow = (cell: Entity | undefined, row: Dictionar
 };
 
 const isInteraction = (neighbor: unknown): neighbor is Interaction => typeof neighbor === 'object';
+
+export const isInactiveEntity = (row: TableRow): boolean | undefined => {
+  const maxEndTimeAggregation = row['max(endTime)']; // Ew.
```
aaron-steinfeld, line 35: This seems... ripe for improvement. Unless we change the api to pass this info in (which wouldn't be a bad idea but likely won't happen immediately), a better heuristic for an inactive entity that avoids looking at any particular field which may or may not be requested/exist on all entities would be to check:
1. The entity includes at least one metric AND
2. All metrics fetched are undefined

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer.component.ts
jake-bassett: The `value` isn't actually an `Entity` as the typing would have you believe. Its only the `id` and `type`. I almost changed the `Inputs` to reflect this realization (that threw me off at first too), but thought that would be too much for the PR.

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-parser.ts
jake-bassett: Yeah...

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer-util.ts
jake-bassett: Yeah, this is really really ugly. I felt better putting it in this file which already holds some ugly Entity handling code. At least it is all isolated.

Talked with the backend folks about which heuristic to use here. Checking that `max(endTime)` is a valid number is the most solid way to know for sure.

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer.component.ts
aaron-steinfeld: an entity is just a minimum of id and type, everything else is optional. I meant that isInactive would be available internally, but that MaybeEntity type shouldn't be expected there so another input is fine.

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer-util.ts
aaron-steinfeld: > Talked with the backend folks about which heuristic to use here. Checking that max(endTime) is a valid number is the most solid way to know for sure.

See above - max(endTime) is a specific subset of the general heuristic I described which allows us to remove any coupling to a specific usage. No metric will be available for an inactive entity, so it just limits us to pick one to check.

projects/observability/src/shared/components/table/data-cell/entity/entity-table-cell-renderer-util.ts
```diff
@@ -27,3 +30,18 @@ export const parseEntityFromTableRow = (cell: Entity | undefined, row: Dictionar
...
+
+export const isInactiveEntity = (row: TableRow): boolean | undefined => {
+  const metricAggregations = filterMetricAggregations(row);
+
+  if (metricAggregations.length === 0) {
+    // If an aggregation wasn't fetched, we have no way of knowing if this Entity is inactive.
+    return undefined;
+  }
+
+  return metricAggregations.some(metricAggregation => !isValidMetricAggregation(metricAggregation));
```
aaron-steinfeld, line 42:
```suggestion
  return metricAggregations.every(metricAggregation => !isValidMetricAggregation(metricAggregation));
```

jake-bassett: Hah. That's what I had originally, but figured `some` was sufficient and potentially faster since first match exits. Why do you feel `every` is necessary?

aaron-steinfeld: Any given metric theoretically could be null (although as far as out of the box metrics, I agree that none come to mind today that would be null when others are present). It seems better to me to be more thorough though, even if it's at the expense of a few microseconds to do an extra handful of iterations.

projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
```diff
@@ -59,8 +62,9 @@ export class EntityRendererComponent implements OnChanges {
       this.setIconType();
     }
   }
+
   public onClickNavigate(): void {
-    this.navigable && this.entity && this.entityNavService.navigateToEntity(this.entity);
+    this.navigable && this.entity && this.entityNavService.navigateToEntity(this.entity, this.inactive);
```
anandtiwary, line 67: What does inactive do? why do we need this?

jake-bassett: If an entity is inactive, it means we haven't seen any spans in the time range. Therefore, some of the screens won't show any data available and we may want to navigate to alternative pages instead for those APIs when drilled in to.

anandtiwary: So is it same as under discovery entity? 
The name is not clear honestly. `navigationFunction(entityId, sourceRoute, isInactive)` i guess this alternate url would be defined in the entity map?

aaron-steinfeld (approved):

PR #817, opened 2021-05-04
feat: fetching log records and showing log icon

itssharmasandeep:
## Description
Fetching log records and showing log icon

### Testing
Local testing done.

### Checklist:

Example
Screenshot 2021-05-04 at 4 30 32 PM

### DIFF
--- a/projects/distributed-tracing/src/shared/components/span-detail/span-data.ts
+++ b/projects/distributed-tracing/src/shared/components/span-detail/span-data.ts
@@ -1,4 +1,5 @@
 import { Dictionary } from '@hypertrace/common';
+import { LogEvent } from '../../dashboard/widgets/waterfall/waterfall/waterfall-chart';
 
 export interface SpanData {
   id: string;
@@ -12,4 +13,6 @@ export interface SpanData {
   tags: Dictionary<unknown>;
   requestUrl: string;
   exitCallsBreakup?: Dictionary<string>;
+  startTime?: number;
+  logEvents?: LogEvent[];
 }
--- a/projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
@@ -13,7 +13,7 @@ import {
   TRACE_GQL_REQUEST
 } from '../../../../graphql/request/handlers/traces/trace-graphql-query-handler.service';
 import { MetadataService } from '../../../../services/metadata/metadata.service';
-import { WaterfallData } from '../../../widgets/waterfall/waterfall/waterfall-chart';
+import { LogEvent, WaterfallData } from '../../../widgets/waterfall/waterfall/waterfall-chart';
 import { GraphQlDataSourceModel } from '../graphql-data-source.model';
 
 @Model({
@@ -60,6 +60,14 @@ export class TraceWaterfallDataSourceModel extends GraphQlDataSourceModel<Waterf
     this.specificationBuilder.attributeSpecificationForKey('errorCount')
   ];
 
+  protected readonly logEventSpecifications: Specification[] = [
+    this.specificationBuilder.attributeSpecificationForKey('traceId'),
+    this.specificationBuilder.attributeSpecificationForKey('attributes'),
+    this.specificationBuilder.attributeSpecificationForKey('timestamp'),
+    this.specificationBuilder.attributeSpecificationForKey('spanId'),
+    this.specificationBuilder.attributeSpecificationForKey('summary')
+  ];
+
   public getData(): Observable<WaterfallData[]> {
     return combineLatest([this.getTraceData(), this.getDurationAttribute()]).pipe(
       map(combinedData => this.mapResponseObject(combinedData[0], combinedData[1]))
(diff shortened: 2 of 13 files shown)


projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.ts
```diff
@@ -37,3 +38,12 @@ export interface WaterfallChartState {
   children: WaterfallDataNode[];
   expanded: boolean;
 }
+
+export interface LogEvent {
+  [key: string]: unknown;
+  traceId?: string;
```
arjunlalb, line 44: why are all the fields optional?

itssharmasandeep: I guess, by mistake. Fixed!!

projects/distributed-tracing/src/shared/components/span-detail/span-data.ts
```diff
@@ -12,4 +13,6 @@ export interface SpanData {
   tags: Dictionary<unknown>;
   requestUrl: string;
   exitCallsBreakup?: Dictionary<string>;
+  startTime?: number;
```
jake-bassett, line 16: Surprised `startTime` isn't required. Is there a reason some of these properties optional?

projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.ts
```diff
@@ -23,6 +23,7 @@ export interface WaterfallData {
   responseBody?: string;
   tags: Dictionary<unknown>;
   errorCount: number;
+  logEvents?: LogEvent[];
```
jake-bassett, line 26: Sort of the same question. Should this be optional? I guess in this case maybe we could expect `logEvents` to be undefined if none are associated with the span?

itssharmasandeep: Yeah, this can be the case.

projects/distributed-tracing/src/shared/components/span-detail/span-data.ts
itssharmasandeep: For `startTime` I've fixed. `logEvents` can be `undefined`. Apart from these I do not have much idea. may be same reason for others as well.

arjunlalb (dismissed):
LGTM

projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
```diff
@@ -60,6 +60,14 @@ export class TraceWaterfallDataSourceModel extends GraphQlDataSourceModel<Waterf
     this.specificationBuilder.attributeSpecificationForKey('errorCount')
   ];
 
+  protected readonly logEventSpecifications: Specification[] = [
+    this.specificationBuilder.attributeSpecificationForKey('traceId'),
```
aaron-steinfeld, line 64: nit: do we actually need the spanid and trace id in these? isn't it already available in the span parent?

projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.ts
```diff
@@ -37,3 +38,12 @@ export interface WaterfallChartState {
   children: WaterfallDataNode[];
   expanded: boolean;
 }
+
+export interface LogEvent {
+  [key: string]: unknown;
```
aaron-steinfeld, line 43: what unknown fields does this have? would anything break if you removed this?

projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.ts
aaron-steinfeld: logEvents will never return null, always an empty array - unless we're conditionally requesting it (but I'd argue in this context, we should ensure it's there for simplicity downstream).

projects/distributed-tracing/src/shared/components/span-detail/span-data.ts
```diff
@@ -12,4 +13,6 @@ export interface SpanData {
   tags: Dictionary<unknown>;
   requestUrl: string;
   exitCallsBreakup?: Dictionary<string>;
+  startTime: number;
```
aaron-steinfeld, line 16: We hold timestamps as dates generally - converting them as soon as feasible. In this PR, it looks like we're adding a few different timestamp fields - numbers in one spot, string in another. Can those be converted to dates up front?

projects/distributed-tracing/src/shared/dashboard/data/graphql/waterfall/trace-waterfall-data-source.model.ts
itssharmasandeep: yeah, we don't need that actually.
thank you for pointing that out.

projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.ts
itssharmasandeep: Actually I added this as I was using this as a table row, but lately we did not use that so now removed it.
Fixed!!

projects/distributed-tracing/src/shared/dashboard/widgets/waterfall/waterfall/waterfall-chart.ts
itssharmasandeep: Added extra condition to handle that.
So it is not optional anymore.

aaron-steinfeld (approved):

PR #818, opened 2021-05-04
feat: relative timestamp table cell renderer and log events table in sheet 

itssharmasandeep:
## Description
Log detail widget and log timestamp table cell renderer components

### Testing
Local testing done
### Checklist:
- [x] PR dependency https://github.com/hypertrace/hypertrace-ui/pull/817

Screenshot 2021-05-07 at 7 52 36 PM


projects/observability/src/shared/components/table/data-cell/log-timestamp/log-timestamp-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  TABLE_COLUMN_CONFIG,
+  TABLE_COLUMN_INDEX,
+  TABLE_DATA_PARSER,
+  TABLE_ROW_DATA
+} from '@hypertrace/components';
+import { LogEvent } from '@hypertrace/distributed-tracing';
+import { ObservabilityTableCellType } from '../../observability-table-cell-type';
+
+@Component({
+  selector: 'ht-log-timestamp-table-cell-renderer',
```
aaron-steinfeld, line 20: Can we decouple this from logs? This is just a general relative timestamp renderer, right?

projects/distributed-tracing/src/shared/dashboard/widgets/log-detail/data/log-detail-data-source.model.ts
```diff
@@ -0,0 +1,21 @@
+import { Dictionary } from '@hypertrace/common';
+import { Model, ModelProperty, PLAIN_OBJECT_PROPERTY } from '@hypertrace/hyperdash';
+import { Observable, of } from 'rxjs';
+import { GraphQlDataSourceModel } from '../../../data/graphql/graphql-data-source.model';
+import { LogEvent } from '../../waterfall/waterfall/waterfall-chart';
+
+@Model({
+  type: 'log-detail-data-source'
```
aaron-steinfeld, line 8: Not sure this page really needs to be a widget - I wouldn't actively avoid it, and if the work is already done it's fine - but for future pages, we really need to get out of the habit of making complex full-screen widgets.

arjunlalb: I think it makes sense for this to be a widget so that it can go in as a full page list in the trace level view in Phase 2.

projects/distributed-tracing/src/shared/dashboard/widgets/log-detail/log-detail-widget-renderer.component.ts
```diff
@@ -0,0 +1,41 @@
...
+})
+export class LogDetailWidgetRendererComponent extends WidgetRenderer<LogDetailWidgetModel, Dictionary<unknown>> {
+  public readonly header: ListViewHeader = { keyLabel: 'key', valueLabel: 'value' };
+
+  protected fetchData(): Observable<Dictionary<unknown>> {
+    return this.model.getData();
+  }
+
+  public getLogEventAttributeRecords(attributes: Dictionary<unknown>): ListViewRecord[] {
+    if (Boolean(attributes)) {
```
arjunlalb, line 32: use `isEmpty`

projects/observability/src/shared/components/table/data-cell/log-timestamp/log-timestamp-table-cell-renderer.component.ts
arjunlalb: +1

projects/observability/src/shared/components/table/data-cell/log-timestamp/log-timestamp-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,69 @@
...
+    @Inject(TABLE_DATA_PARSER)
+    parser: TableCellParserBase<string, string, unknown>,
+    @Inject(TABLE_CELL_DATA) cellData: any,
+    @Inject(TABLE_ROW_DATA) rowData: LogEvent
+  ) {
+    super(columnConfig, index, parser, cellData, rowData);
+    this.spanStartTime = rowData.spanStartTime as number;
+    this.timestamp = rowData.timestamp;
+    this.readableDateTime =
+      this.timestamp
```
arjunlalb, line 55: is there a cleaner way to do this? whats the goal here? Why do we have a time string as input here?

projects/distributed-tracing/src/shared/dashboard/widgets/log-detail/data/log-detail-data-source.model.ts
aaron-steinfeld: I'd extend the same argument to those views too though 😉 If I understand the comment/remember the screen, the trace view today is one big widget, and with this it will now have two views - the sequence and the log list, right? When we do that work, I'd make the same argument that the tabs and their contents need not be (but easily could be) dashboards.

projects/observability/src/shared/components/table/data-cell/log-timestamp/log-timestamp-table-cell-renderer.component.ts
aaron-steinfeld: +1 - please use native/our existing utils like date coercer and TImeDuration. We never should be working with the strings directly. This one, for example we could take the difference and use our duration class.

itssharmasandeep: So currently we're getting a timestamp string which is in nano seconds for example `2021-05-04T12:13:55.543456T`

So all this logic is done to show cell value and tooltip in an expected format.

aaron-steinfeld: Nanos shouldn't be relevant at this level - that's why we use the ISO format, so the UI is agnostic to the unit. The timestamp should render the same as all our other timestamps, right? Using the date coercer, then the display date pipe. The new part is the relative offset between the two timestamps (log event time and span start time), which would require taking the difference between the two and using our duration pipe (I'm not 100% sure that duration pipe is up to date for these requirements, but conceptually at least, that's what we should be going for)

projects/observability/src/shared/components/table/data-cell/log-timestamp/log-timestamp-table-cell-renderer.component.ts
itssharmasandeep: Done that!!

projects/observability/src/shared/components/table/data-cell/log-timestamp/log-timestamp-table-cell-renderer.component.ts
itssharmasandeep: I have updated the `new look` for this in description.
Please do have a look!!

projects/components/src/table/cells/data-renderers/relative-timestamp/relative-timestamp-table-cell-renderer.component.test.ts
```diff
@@ -0,0 +1,41 @@
...
+} from '@hypertrace/components';
+import { createComponentFactory } from '@ngneat/spectator/jest';
+import { MockComponent } from 'ng-mocks';
+import { tableCellDataProvider } from '../../test/cell-providers';
+import {
+  RelativeTimestampTableCellRendererComponent,
+  RowData
+} from './relative-timestamp-table-cell-renderer.component';
+
+describe('log timestamp table cell renderer component', () => {
```
anandtiwary, line 15: nit: relative timestamp

projects/components/src/table/cells/data-renderers/relative-timestamp/relative-timestamp-table-cell-renderer.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  TABLE_ROW_DATA
+} from '../../table-cell-injection';
+import { TableCellParserBase } from '../../table-cell-parser-base';
+import { TableCellRenderer } from '../../table-cell-renderer';
+import { TableCellRendererBase } from '../../table-cell-renderer-base';
+import { CoreTableCellParserType } from '../../types/core-table-cell-parser-type';
+import { CoreTableCellRendererType } from '../../types/core-table-cell-renderer-type';
+import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
+
+export interface RowData {
```
anandtiwary, line 18: nit: You can do 
```
export interface RowData extends Dictionary<unknown> {
baseTimestamp: Date | number;
}
```

anandtiwary: Also, can the timestamp be string too?

projects/distributed-tracing/src/shared/dashboard/widgets/log-detail/data/log-detail-data-source.model.ts
anandtiwary: @itssharmasandeep This is for the tab view right? don't we fetch all the details beforehand and show the tab as a normal component?

projects/components/src/table/cells/data-renderers/relative-timestamp/relative-timestamp-table-cell-renderer.component.ts
itssharmasandeep: I guess, `Date` can handle the `string`, I have tested this, works fine

custom type `DateOrNumber` is used as type for celldata as well so I didn't want to repeat `Date | number` so created this type.

`extends Dictionary<unknown> ` part is fixed!!

projects/components/src/table/cells/data-renderers/relative-timestamp/relative-timestamp-table-cell-renderer.component.test.ts
itssharmasandeep: Fixed!!

projects/distributed-tracing/src/shared/dashboard/widgets/log-detail/data/log-detail-data-source.model.ts
itssharmasandeep: Yeah, but since we're using table dashboard, so for child template we need data source, right?
Currently I am not fetching anything.

jake-bassett (commented):
LGTM other than the comments about page vs widget, so I'll defer to others on that part.

itssharmasandeep:
@aaron-steinfeld @anandtiwary @arjunlalb 
is this okay to be merged?

aaron-steinfeld:
> @aaron-steinfeld @anandtiwary @arjunlalb
> is this okay to be merged?

This PR is targeting one of your own branches - so reviews are generally for information only.

Generally, the way we stage features is a series of branches like this, and then when the first is ready we merge it to main, and the second will automatically be updated to target main - so the approvals will come in sequence.

itssharmasandeep:
@aaron-steinfeld ok, got it.
I've updated the PR dependency in the description as well.

See if this can go ahead https://github.com/hypertrace/hypertrace-ui/pull/817 (first in the series)

projects/distributed-tracing/src/shared/dashboard/widgets/log-detail/data/log-detail-data-source.model.ts
aaron-steinfeld: Hmm I think my comment got lost. My original comment was about the table itself - if that needed to be a dashboard (and likewise, everything downstream). The dashboards have really... evolved past how they were intended to be used. Here for example, we could just as easily remove the data source and pass the data directly to the widget - it's not providing any value if it's a custom pass through for a custom widget. (just to clarify, we don't need to change this here, but it's something we should be cognizant of as we build new screens).

As far as this data source goes, can we just reuse `StaticDataSource` since this one's not doing anything but unwrapping which we could do elsewhere? Or if we keep it, it shouldn't extend `GraphQlDataSourceModel` because it's not querying GQL.

anandtiwary: +1 to aaron's point. 

@itssharmasandeep For the sheet/tab view, we don't necessarily need a data source/widget for child template. You can just pass any ng-template in it and it would render. For the page view, we can use the static data source like Aaron described above.

anandtiwary: Shared the details on slack

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  public spanStartTime?: number;
+
+  public readonly header: ListViewHeader = { keyLabel: 'key', valueLabel: 'value' };
+  private readonly dateCoercer: DateCoercer = new DateCoercer();
+
+  public dataSource?: TableDataSource<TableRow>;
+  public columnConfigs: TableColumnConfig[] = [];
+
+  public ngOnInit(): void {
+    this.buildDataSource();
```
anandtiwary, line 64: Probably need to call this inside `ngOnChanges`

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  TableDataResponse,
+  TableDataSource,
+  TableMode,
+  TableRow
+} from '@hypertrace/components';
+import { isEmpty } from 'lodash-es';
+import { Observable, of } from 'rxjs';
+
+export const enum LogEventsTableViewType {
+  Sheet = 'sheet',
```
anandtiwary, line 17: We do need better name for this. Umm. How about `Condensed` for the sheet and `Detailed` for the page? Any other suggestions anyone?

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  public ngOnInit(): void {
+    this.buildDataSource();
+    this.columnConfigs = this.getTableColumnConfigs();
+  }
+
+  public getLogEventAttributeRecords(attributes: Dictionary<unknown>): ListViewRecord[] {
+    if (!isEmpty(attributes)) {
+      return Object.entries(attributes).map((attribute: [string, unknown]) => ({
+        key: attribute[0],
+        value: attribute[1] as string | number
```
anandtiwary, line 72: So we do know the type of value -> `string | number`. Why don't we update the type of attributes and logEvents to reflect this? 

```
  @Input()
  public logEvents: Dictionary<string | number>[] = [];
```

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  public columnConfigs: TableColumnConfig[] = [];
+
+  public ngOnInit(): void {
+    this.buildDataSource();
+    this.columnConfigs = this.getTableColumnConfigs();
+  }
+
+  public getLogEventAttributeRecords(attributes: Dictionary<unknown>): ListViewRecord[] {
+    if (!isEmpty(attributes)) {
+      return Object.entries(attributes).map((attribute: [string, unknown]) => ({
```
anandtiwary, line 70: you can also do 
```
map( ([key, value]) => ({key: key, value: value}))
```

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
itssharmasandeep: Done!!

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
itssharmasandeep: Fixed, thanks for suggestion!!

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
anandtiwary: Still don't see this change though

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.ts
anandtiwary: Still don't see this change though

PR #822, opened 2021-05-04
fix: close popup on destroy

aaron-steinfeld:
## Description
Today, when moving away from a view with a popup (for example, a modal) any open popovers remain hanging open. This change ties their lifecycle to the underlying component, so the popover overlay will be removed if the trigger itself is destroyed.

### Testing
Tested manually

### DIFF
--- a/projects/components/src/popover/popover.component.ts
+++ b/projects/components/src/popover/popover.component.ts
@@ -9,7 +9,6 @@ import {
   OnDestroy,
   Output
 } from '@angular/core';
-import { Subscription } from 'rxjs';
 import { PopoverBackdrop, PopoverPositionType, PopoverRelativePositionLocation } from './popover';
 import { PopoverContentComponent } from './popover-content.component';
 import { PopoverRef } from './popover-ref';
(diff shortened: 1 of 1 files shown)


jake-bassett (approved):

projects/components/src/popover/popover.component.ts
```diff
@@ -76,17 +77,17 @@ export class PopoverComponent implements OnDestroy {
     });
 
     // Closing can happen internal to the Popover for things like closeOnBackdropClick. Let the consumer know.
-    this.subscription = popover.closed$.subscribe(() => this.popoverClose.emit());
+    this.popover.closed$.subscribe(() => this.popoverClose.emit());
```
anandtiwary, line 80: Why are not unsubscribing from this?

aaron-steinfeld: No longer necessary, because the closed observable ends when we invoke close which now will always hope at or before destroy.

anandtiwary (approved):

PR #823, opened 2021-05-05
Table multiselect filters

jake-bassett:
## Description
This change is to make all of the table control dropdowns multiselect instead of single select. 

It also cleans up some messiness around the table control filter types. Using multi-select is actually easier to configure because it no-longer requires an `Unset` option in the dropdown. This simplifies the API for the `select` dropdowns and reduces its possible filter types to just `TableControlOptionType.Filter`. Now only `checkbox` table controls use `TableControlOptionType.Unset` and `TableControlOptionType.Property` types.

https://user-images.githubusercontent.com/45181984/117207822-4134f500-ada9-11eb-8ce1-0135296d830a.mov

### DIFF
--- a/projects/components/src/table/controls/table-controls-api.ts
+++ b/projects/components/src/table/controls/table-controls-api.ts
@@ -59,7 +59,7 @@ export interface TableCheckboxChange {
   option: TableCheckboxControlOption;
 }
 
-export type TableCheckboxControlOption<T extends boolean = boolean> = TableControlOption & {
+export type TableCheckboxControlOption<T = boolean> = TableControlOption & {
   value: T;
 };
 
@@ -79,5 +79,5 @@ export const toInFilter = (tableFilters: TableFilter[]): TableFilter =>
       field: previousValue.field,
       operator: FilterOperator.In,
       value: [...(Array.isArray(previousValue.value) ? previousValue.value : [previousValue.value]), currentValue.value]
-    }
+    };
   });
--- a/projects/components/src/table/controls/table-controls-api.ts
+++ b/projects/components/src/table/controls/table-controls-api.ts
@@ -1,17 +1,52 @@
 import { Dictionary } from '@hypertrace/common';
-import { SelectOption } from '../../select/select-option';
+import { FilterOperator } from '../../filtering/filter/filter-operators';
 import { TableFilter } from '../table-api';
 
+export const enum TableControlOptionType {
+  Filter = 'filter',
+  Property = 'property',
+  Unset = 'unset'
+}
+
+export type TableControlOption = TableUnsetControlOption | TableFilterControlOption | TablePropertyControlOption;
+
+export type TableFilterControlOption = {
+  type: TableControlOptionType.Filter;
+  label: string;
+  metaValue: TableFilter;
+};
+
+export type TableUnsetControlOption = {
+  type: TableControlOptionType.Unset;
+  label: string;
+  metaValue: string;
+};
+
+export type TablePropertyControlOption = {
+  type: TableControlOptionType.Property;
+  label: string;
+  metaValue: Dictionary<unknown>;
+};
+
+/*
+ * Select Control
+ */
+
 export interface SelectControl {
-  placeholder?: string;
-  default?: SelectOption<TableControlOption>;
-  options: SelectOption<TableControlOption>[];
+  placeholder: string;
+  options: TableSelectControlOption[];
 }
 
 export interface SelectChange {
   select: SelectControl;
-  value: TableControlOption;
+  values: TableSelectControlOption[];
 }
+
+export type TableSelectControlOption = TableFilterControlOption;
+
+/*
+ * Checkbox Control
+ */
 
 export interface CheckboxControl {
   label: string;
@@ -21,38 +56,22 @@ export interface CheckboxControl {
 
 export interface CheckboxChange {
   checkbox: CheckboxControl;
-  option: TableControlOption<boolean>;
+  option: TableCheckboxControlOption;
 }
 
-export const enum TableControlOptionType {
-  Filter = 'filter',
-  Property = 'property',
-  UnsetFilter = 'unset-filter'
-}
-
-export type TableControlOption<T = unknown> =
-  | TableUnsetFilterControlOption<T>
-  | TableFilterControlOption<T>
-  | TablePropertyControlOption<T>;
-
-interface TableControlOptionBase<T> {
-  value?: T;
-}
-export interface TableUnsetFilterControlOption<T = unknown> extends TableControlOptionBase<T> {
-  type: TableControlOptionType.UnsetFilter;
-  metaValue: string;
-}
-
-export interface TableFilterControlOption<T = unknown> extends TableControlOptionBase<T> {
-  type: TableControlOptionType.Filter;
-  metaValue: TableFilter;
-}
-
-export interface TablePropertyControlOption<T = unknown> extends TableControlOptionBase<T> {
-  type: TableControlOptionType.Property;
-  metaValue: Dictionary<unknown>;
-}
-
-export type TableCheckboxControlOption<T extends boolean> = TableControlOption<T> & { label: string };
+export type TableCheckboxControlOption<T extends boolean = boolean> = TableControlOption & {
+  value: T;
+};
 
 export type TableCheckboxOptions = [TableCheckboxControlOption<true>, TableCheckboxControlOption<false>];
+
+/*
+ * Util
+ */
+
+export const toInFilter = (tableFilters: TableFilter[]): TableFilter =>
+  tableFilters.reduce((previousValue, currentValue) => ({
+    field: previousValue.field,
+    operator: FilterOperator.In,
+    value: [...(Array.isArray(previousValue.value) ? previousValue.value : [previousValue.value]), currentValue.value]
+  }));
--- a/projects/components/src/table/controls/table-controls-api.ts
+++ b/projects/components/src/table/controls/table-controls-api.ts
@@ -70,8 +70,14 @@ export type TableCheckboxOptions = [TableCheckboxControlOption<true>, TableCheck
  */
 
 export const toInFilter = (tableFilters: TableFilter[]): TableFilter =>
-  tableFilters.reduce((previousValue, currentValue) => ({
-    field: previousValue.field,
-    operator: FilterOperator.In,
-    value: [...(Array.isArray(previousValue.value) ? previousValue.value : [previousValue.value]), currentValue.value]
-  }));
+  tableFilters.reduce((previousValue, currentValue) => {
+    if (currentValue.operator !== FilterOperator.Equals || previousValue.field !== currentValue.field) {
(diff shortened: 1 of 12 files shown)


projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,58 +1,77 @@
...
-export interface TablePropertyControlOption<T = unknown> extends TableControlOptionBase<T> {
-  type: TableControlOptionType.Property;
-  metaValue: Dictionary<unknown>;
+export interface CheckboxChange {
+  checkbox: CheckboxControl;
+  option: TableCheckboxControlOption;
 }
 
-export type TableCheckboxControlOption<T extends boolean> = TableControlOption<T> & { label: string };
+export type TableCheckboxControlOption<T extends boolean = boolean> = TableControlOption & {
```
anandtiwary, line 62: T extends boolean, a primitive? why can't it be just a boolean?

anandtiwary: ok so this is the usage. We can just keep it as boolean right?
```
export type TableCheckboxOptions = [TableCheckboxControlOption<true>, TableCheckboxControlOption<false>]
```

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,58 +1,77 @@
...
 
 export type TableCheckboxOptions = [TableCheckboxControlOption<true>, TableCheckboxControlOption<false>];
+
+/*
+ * Util
+ */
+
+export const toInFilter = (tableFilters: TableFilter[]): TableFilter =>
+  tableFilters.reduce((previousValue, currentValue) => ({
+    field: previousValue.field,
```
anandtiwary, line 74: what if the fields in the tableFilter are different?. In this reduce, we are just considering the last field as field for the final In filter.

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,58 +1,77 @@
...
-  | TableUnsetFilterControlOption<T>
-  | TableFilterControlOption<T>
-  | TablePropertyControlOption<T>;
+/*
+ * Select Control
+ */
 
-interface TableControlOptionBase<T> {
-  value?: T;
+export interface SelectControl {
```
anandtiwary, line 35: This should be prefixed with 'Table'

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,58 +1,77 @@
...
+import { FilterOperator } from '../../filtering/filter/filter-operators';
 import { TableFilter } from '../table-api';
 
-export interface SelectControl {
-  placeholder?: string;
-  default?: SelectOption<TableControlOption>;
-  options: SelectOption<TableControlOption>[];
+export const enum TableControlOptionType {
+  Filter = 'filter',
+  Property = 'property',
```
anandtiwary, line 7: What is a property control option? any visual example?

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,58 +1,77 @@
...
+
+export interface SelectChange {
+  select: SelectControl;
+  values: TableSelectControlOption[];
 }
 
-export interface TableFilterControlOption<T = unknown> extends TableControlOptionBase<T> {
-  type: TableControlOptionType.Filter;
-  metaValue: TableFilter;
+export type TableSelectControlOption = TableFilterControlOption;
```
anandtiwary, line 45: do we really need so many different types? Like below where you specify metaValue type

```
export interface TableControlOption<T> {
  type: TableControlOptionType;
  label: string;
  metaValue: T;
}
}
```

projects/components/src/table/controls/table-controls-api.ts
```diff
@@ -1,58 +1,77 @@
...
+  metaValue: TableFilter;
 }
 
-export interface CheckboxChange {
-  checkbox: CheckboxControl;
-  option: TableControlOption<boolean>;
+export interface TableUnsetControlOption {
+  type: TableControlOptionType.Unset;
+  label: string;
+  metaValue: string;
```
anandtiwary, line 22: what is meta value supposed to do? How is it different than 'value' ?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-control-checkbox-option.model.ts
```diff
@@ -31,7 +32,10 @@ export class TableWidgetControlCheckboxOptionModel extends TableWidgetControlMod
     );
   }
 
-  private isValidCheckboxControlOption(options: TableControlOption[]): options is TableCheckboxOptions {
-    return options.length === 2 && options.every(option => typeof option.value === 'boolean');
+  private isValidCheckboxControlOptions(options: TableControlOption[]): options is TableCheckboxOptions {
+    return (
+      options.length === 2 &&
+      options.every(option => hasOwnProperty(option, 'value') && typeof option.value === 'boolean')
```
anandtiwary, line 38: I am yet to understand the inheritance in the table control options. value property seems to be defined at only one place. We can also use `in` here, i think

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -277,21 +272,15 @@ export class TableWidgetRendererComponent
...
-        this.selectFilterSubject.next(this.mergeFilters(changed.value.metaValue));
-        break;
-      case TableControlOptionType.Property:
-        this.queryPropertiesSubject.next(this.mergeQueryProperties(changed.value.metaValue));
-        break;
-      default:
-        assertUnreachable(changed.value);
+    if (changed.values.length === 0) {
+      // tslint:disable-next-line: no-void-expression
+      return this.selectFilterSubject.next(this.removeFilters(changed.select.options[0].metaValue.field));
```
anandtiwary, line 277: we don't need the return here right?

jake-bassett: Don't _need_ to, but returning makes the code a little cleaner because we don't need an `else` block or a separate  `return;` statement within the `if` block. I could eliminate the `return` below, I suppose with the same results. It just looked odd.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-control-checkbox-option.model.ts
jake-bassett: Yeah, they are confusing. Fortunately it got a little bit cleaner with this PR.

Correct, that `value` is only for checkbox control now, since it has to hold a `true` or `false` while the select control doesn't. Checkbox controls are actually the oddball because they have to hold more ways of modifying a query. 

Both select and checkbox controls have a `metaValue`. This is what contains the information about what to use in the query. It is most typically a `Filter` (see the `metaValue` for `TableFilterControlOption`), containing `key`, `operator`, and `value`. This is the `metaValue` that all select controls use now.

There are two other cases with query info, but they are now only used for checkbox controls. In addition to the 'Filter' 'metaValue`, there is an `Unset` and a `Property` `metaValue` (see the `metaValue` for `TableUnsetControlOption` and `TablePropertyControlOption`). These are also used to modify the query, but in a different way.

The `Unset` `metaValue` is used to hold a `string` that identifies which filter to _remove_ from our filter list. The `Property` `metaValue` is a `Dictionary` used to add request properties to the query. Currently, our only example of this is `includeInactive: true/false` that we add to the request for APIs.

So, the reason that checkbox has a `value` and not the select dropdown is that checkbox has this additional user specified `true/false` in addition to the `value` that is buried in the `metaValue` for all controls.

Not sure if that explains it or makes it more confusing, but happy to expand on anything.

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: Yeah, this is my issue with this too. 

We could validate here or I can break up the args to ensure we only are passed one `field` and `operator`, but I felt it was overkill and traded a little bit of confusion (and a potential for a bug if used incorrectly) for ease of use. 

I can add some validation easy enough, so I will. I can revisit other options if we feel strongly.

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: Yeah, also strange, but the `extends` is required for us to do line 66.

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: I didn't add them just for fun! 😂

But valid question. I explain a little more in [this](https://github.com/hypertrace/hypertrace-ui/pull/823/files#r626981848) comment.

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: Please see [this](https://github.com/hypertrace/hypertrace-ui/pull/823/files#r626981848) comment below.

projects/components/src/table/controls/table-controls-api.ts
jake-bassett: Please see [this](https://github.com/hypertrace/hypertrace-ui/pull/823/files#r626981848) comment below.

projects/components/src/table/controls/table-controls-api.ts
anandtiwary: if we keep boolean why wouldn't line 66 work?

jake-bassett: Hmm. Trying again, removing `extends` actually works now that I am setting a default. I wasn't before and only recently added it. Update incoming.

anandtiwary: cool

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
anandtiwary: I honestly find this more confusing. If we can use a separate `return` i think that's more clear.

jake-bassett: This is the strangest thing to hold up a PR for.

jake-bassett: I'll change it in my local.

anandtiwary (approved):

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
anandtiwary: It is a strong nit from my end

jake-bassett: I'm going to change it. Still silly to hold up a PR over this.

PR #843, opened 2021-05-12
fix: style fixes

itssharmasandeep:
## Description
Style fixes

### DIFF
--- a/projects/components/src/list-view/list-view.component.scss
+++ b/projects/components/src/list-view/list-view.component.scss
@@ -34,7 +34,7 @@ $horizontal-offset: 12px;
     .header-key-label,
     .header-value-label {
       @include overline;
-      padding: 0px $horizontal-offset;
+      padding: 0 $horizontal-offset;
       width: 100%;
     }
   }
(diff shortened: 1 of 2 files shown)


projects/components/src/list-view/list-view.component.scss
```diff
@@ -34,7 +34,7 @@ $margin-left: 12px;
     .header-key-label,
     .header-value-label {
       @include overline;
-      margin-left: $margin-left;
+      padding: 0px $horizontal-offset;
```
arjunlalb, line 37: make these `padding: 0 $horizontal-offset`

0 doesn't need units.

projects/distributed-tracing/src/shared/components/log-events/log-events-table.component.scss
```diff
@@ -1,7 +1,13 @@
 @import 'font';
 
 .log-events-table {
-  margin-top: 25px;
+  padding-top: 25px;
```
arjunlalb, line 4: we use spacing in multiple of 4. what does the mock say here? perhaps 24 ?

itssharmasandeep: Yeah,Fixed!!

projects/components/src/list-view/list-view.component.scss
itssharmasandeep: yeah, Fixed!!

arjunlalb (approved):

PR #844, opened 2021-05-12
feat: allow user to configure default string and separator for dislay…

jake-bassett:
This change allows `displayStringEnum` to have its default value and separator configured through args.

### DIFF
--- a/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
+++ b/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
@@ -5,7 +5,7 @@ import { displayStringEnum } from './display-string-enum';
   name: 'htDisplayStringEnum'
 })
 export class DisplayStringEnumPipe implements PipeTransform {
-  public transform(value?: string, unknown: string = '-', separator: string = ' '): string {
-    return displayStringEnum(value, unknown, separator);
+  public transform(value?: string, defaultValue: string = '-', separator: string = ' '): string {
+    return displayStringEnum(value, defaultValue, separator);
   }
 }
--- a/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
+++ b/projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
@@ -5,7 +5,7 @@ import { displayStringEnum } from './display-string-enum';
   name: 'htDisplayStringEnum'
 })
 export class DisplayStringEnumPipe implements PipeTransform {
-  public transform(value: string): string {
-    return displayStringEnum(value);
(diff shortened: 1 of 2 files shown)


projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
```diff
@@ -5,7 +5,7 @@ import { displayStringEnum } from './display-string-enum';
   name: 'htDisplayStringEnum'
 })
 export class DisplayStringEnumPipe implements PipeTransform {
-  public transform(value: string): string {
-    return displayStringEnum(value);
+  public transform(value?: string, unknown: string = '-', separator: string = ' '): string {
```
aaron-steinfeld, line 8: variable naming: `unknown` shouldn't be used as it's a type - is that the `defaulttValue` or similar?

projects/common/src/utilities/formatters/enum/display-string-enum.ts
```diff
@@ -1,13 +1,15 @@
...
   }
 
   // This removes dashes and underscores and gives all words initial caps
   const startCased = startCase(provided);
 
+  const startCasedSeparated = startCased.replace(' ', separator);
+
   // We only want the first word capitalized.
-  return `${startCased[0]}${startCased.substr(1).toLowerCase()}`;
+  return `${startCasedSeparated[0]}${startCasedSeparated.substr(1).toLowerCase()}`;
```
aaron-steinfeld, line 14: I feel like I commented on this at the time but who knows:
```suggestion
  return capitalize(startCasedSeparated);
```

jake-bassett: You did, but that doesn't work for certain use cases and I don't feel the alternative you provided is really any cleaner than this.

https://github.com/hypertrace/hypertrace-ui/pull/752/files#r609080471

projects/common/src/utilities/formatters/enum/display-string-enum.pipe.ts
jake-bassett: Fixed.

projects/common/src/utilities/formatters/enum/display-string-enum.ts
aaron-steinfeld: I'm confused - you said in that linked comment you agreed that `startCase(lowerCase(provided))` is cleaner, but it looks like it got merged without making that change. What use case isn't supported?

Here, I was just suggesting capitalize which is a 1:1 replacement of the explicit impl on line 14, but using
 ```ts 
 return capitalize(lowerCase(provided)).replace(' ', separator);`
 ```
 to replace lines 8-14 I think is even better.

aaron-steinfeld (approved):

PR #856, opened 2021-05-18
Open in new tab

alok-traceable:
## Description
- Creating component Open In a new tab which basically wraps link component and icon component.
- Adding support of query params in External URL navigator.
- Adding utility to convert string dictionary to URL query param string and vice versa.

### Testing
- Unit test cases

### Checklist:


projects/common/src/navigation/navigation.service.ts
```diff
@@ -89,7 +89,8 @@ export class NavigationService {
           }
         ],
         extras: {
-          skipLocationChange: true // Don't bother showing the updated location, we're going external anyway
+          skipLocationChange: true, // Don't bother showing the updated location, we're going external anyway
+          queryParams: params.queryParams
```
aaron-steinfeld, line 93: Is it possible to encode these into the url we send at this level, so we don't have to worry about stitching together the url downstream?

projects/common/src/utilities/url/url-utilities.ts
```diff
@@ -0,0 +1,26 @@
+import { Dictionary, isEmpty } from "lodash"
+
+export const getQueryParamStringFromObject = (params: Dictionary<string | number>): string => {
```
aaron-steinfeld, line 3: I feel like angular already has this type of logic available to us for reuse - `UrlTree.toString()`, perhaps?

projects/common/src/time/time-range.service.ts
```diff
@@ -30,15 +32,21 @@ export class TimeRangeService {
...
-    const timeRangeParam = `${TimeRangeService.TIME_RANGE_QUERY_PARAM}=${timeRangeParamValue}`;
+    const timeRangeParam = getQueryParamStringFromObject(this.getTimeRangeParams());
     const timeRange = this.getCurrentTimeRange();
     const fixedTimeRange: FixedTimeRange = TimeRangeService.toFixedTimeRange(timeRange.startTime, timeRange.endTime);
     const newTimeRangeParam = `${TimeRangeService.TIME_RANGE_QUERY_PARAM}=${fixedTimeRange.toUrlString()}`;
 
     return this.navigationService.getAbsoluteCurrentUrl().replace(timeRangeParam, newTimeRangeParam);
   }
 
+  public getTimeRangeParams(): Dictionary<string> {
```
aaron-steinfeld, line 43: Does this need to be public?

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
```diff
@@ -1,46 +1,20 @@
...
-      ></ht-button>
+    <div *ngIf="this.navigationPath" class="open-in-new-tab" htTooltip="Open in a new tab">
+      <ht-link [paramsOrUrl]="this.paramsOrUrl">
+        <ht-icon icon="${IconType.OpenInNewTab}" [size]="this.iconSize" ></ht-icon>
+      </ht-link>
     </div>
   `
 })
-export class OpenInNewTabComponent {
+export class OpenInNewTabComponent extends LinkComponent {
```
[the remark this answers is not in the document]
anandtiwary: +1. 

We are already composing ht-link in the template. I think you can just re-declare required inputs in this component `paramsOrUrl, navigationPath` and pass it on to ht-link . 

One question: In order to open in new tab, this component should internally change the input `paramsOrUrl` in to a navigation param of External type, right?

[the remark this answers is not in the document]
anandtiwary: Since this component says "open in new tab", it should be this component's responsibility to always open in new tab, even if the input nav param is of the internal type.

projects/common/src/navigation/navigation.service.ts
arjunlalb: We'll have to stitch it up to url somewhere. Isn't it cleaner to do it here so that the signature for External and Internal navigation params remain clear and uniform? Otherwise, one kind of navigation params will take queryParams as object and another will expect it as part of URL.

aaron-steinfeld: Yep, sorry I may have been unclear here (which is downstream and which is upstream?!) - I do want it here so callers aren't concerned (which is how it's maintained here), but I was hoping the url we sent to the external nav component would be complete. Since it has already joined all the other pieces like protocol/host/path params, keeping it as a one url rather than treating query params special felt cleaner, letting the nav service deal with the conversion of various inputs to one final url.

projects/common/src/utilities/url/url-utilities.ts
```diff
@@ -0,0 +1,29 @@
+import { Dictionary, isEmpty } from 'lodash';
+
+export const getQueryParamStringFromObject = (params: Dictionary<string | number>): string => {
+  try {
```
arjunlalb, line 4: why do we have a try catch here?

projects/common/src/utilities/url/url-utilities.ts
```diff
@@ -0,0 +1,29 @@
...
+      .join('&');
+
+    return isEmpty(paramString) ? '' : paramString;
+  } catch (error) {
+    return ``;
+  }
+};
+
+export const getQueryParamObjectFromString = (query: string): Dictionary<string> => {
+  try {
```
arjunlalb, line 17: we don't have a usecase for this method today. you can remove this.

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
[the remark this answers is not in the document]
arjunlalb: I recommended extending the LinkComponent here specifically because this is essentially LinkComponent itself, but with an icon instead of any content inside this. In that case, isn't it better to actually maintain the dependency via inheritance instead of duplicating the interfaces and the logic? @aaron-steinfeld @anandtiwary

projects/common/src/utilities/url/url-utilities.ts
arjunlalb: +1 to this. I looked this up after our conversation @alok-traceable .

We can avoid the custom logic and replace it with

`new HttpParams({ fromObject: queryParams }).toString()`

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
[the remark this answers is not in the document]
anandtiwary: Extending an angular component like this may have many side effects. 
- The main issue I see is the mismatch within the template. As you can see here, the template itself is showing a `has-a` relationship with ht-link. There is no way to have an 'is-a' relationship in the template. So imo the backing angular component should reflect the same.

- If one of the property of the base component changes then we would also have to update all occurrence of child component where this property is used. This also goes for tests. 

In general, `Has-a` relationship is more loosely coupled than `Is-a` relationship. Hence my preference.

projects/common/src/utilities/url/url-utilities.ts
alok-traceable: removed

projects/common/src/utilities/url/url-utilities.ts
alok-traceable: removed

package.json
```diff
@@ -60,6 +60,7 @@
     "graphql": "^15.5.0",
     "graphql-tag": "^2.12.3",
     "iso8601-duration": "^1.3.0",
+    "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
alok-traceable: removed

projects/common/src/time/time-range.service.ts
alok-traceable: We need the time range param to generate the new tab URL

projects/common/src/utilities/url/url-utilities.ts
alok-traceable: Done suggested change on the util.

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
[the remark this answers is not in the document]
alok-traceable: Open in a new tab component is no longer extending `LinkComponent`. Also, it enforces paramsOrUrl to be of type `ExternalNavigationParamsNew`

projects/common/src/utilities/url/url-utilities.ts
```diff
@@ -0,0 +1,26 @@
...
+			.map(([key, value]) => value !== undefined ? `${key}=${value}` : '')
+			.filter(item => item)
+			.join('&');
+		return isEmpty(paramString) ? '' : paramString;
+	} catch (error) {
+		return ``;
+	}
+};
+
+export const getQueryParamObjectFromString = (string: string): Dictionary<string> => {
```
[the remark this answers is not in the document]
alok-traceable: It is not used anywhere else, I have removed it.

projects/common/src/utilities/url/url-utilities.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { Params } from '@angular/router';
+import {
+  getQueryParamStringFromObject,
+  getQueryParamObjectFromString
+} from './url-utilities';
+
+describe('getQueryParamStringFromObject', () => {
```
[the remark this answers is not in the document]
alok-traceable: Renamed the files and also replaced `it` with `test`

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
```diff
@@ -1,46 +1,23 @@
...
+      <ht-link [paramsOrUrl]="this.paramsOrUrl">
+        <ht-icon icon="${IconType.OpenInNewTab}" [size]="this.iconSize"></ht-icon>
+      </ht-link>
     </div>
   `
 })
 export class OpenInNewTabComponent {
   @Input()
-  public size?: ButtonSize = ButtonSize.Small;
+  public paramsOrUrl?: ExternalNavigationParamsNew | string;
```
aaron-steinfeld, line 19: Do we still have the same default behavior of the current url?

projects/common/src/navigation/navigation.service.ts
```diff
@@ -80,11 +82,16 @@ export class NavigationService {
 
     if (params.navType === NavigationParamsType.External) {
       // External
+      const queryParamString = getQueryParamStringFromObject(params.queryParams ?? {});
```
aaron-steinfeld, line 85: It might be better to let angular do any url building for us (such as url tree). This would break if we passed it a a url that contained query params, since it wouldn't merge. Say `http://website.com?param1=value` and a second param via query params `{ param2=value }`, we'd get `http://website.com?param1=value?param2=value` instead of `http://website.com?param1=value&param2=value`.

I think you can combine `router.parseUrl`, `router.serializeUrl` and maybe `router.createUrlTree` to get all the right behavior here.

projects/common/src/time/time-range.service.ts
```diff
@@ -30,15 +32,22 @@ export class TimeRangeService {
...
-    const timeRangeParam = `${TimeRangeService.TIME_RANGE_QUERY_PARAM}=${timeRangeParamValue}`;
+    const timeRangeParam = getQueryParamStringFromObject(this.getTimeRangeParams());
     const timeRange = this.getCurrentTimeRange();
     const fixedTimeRange: FixedTimeRange = TimeRangeService.toFixedTimeRange(timeRange.startTime, timeRange.endTime);
     const newTimeRangeParam = `${TimeRangeService.TIME_RANGE_QUERY_PARAM}=${fixedTimeRange.toUrlString()}`;
 
     return this.navigationService.getAbsoluteCurrentUrl().replace(timeRangeParam, newTimeRangeParam);
   }
 
+  public getTimeRangeParams(): Params {
```
aaron-steinfeld, line 43: So the fuller picture is coming together here for me - I'm guessing this was left public to later use as query params in an open in new tab case that should maintain the time range, and that's also why queryParams handling was added in nav service. The problem here is we do not want callers to be concerned about this. Time range is not the only global param that needs to be preserved, and we don't want callers to even have to be aware of global params by design.

My suggestion would be to 
- revert the changes here (or at least make this method private)
- update `NavigationService.buildNavigationParams` to use `buildQueryParam` for relative external urls, similar to what we do in `navigateWithinApp`
- remove queryParams from `ExternalNavigationParamsNew` - it's never used today and shouldn't be for this use case, so no reason to build logic for it

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
anandtiwary, line 19: - Because this component is named as `OpenInNewTab`, i feel it should be able to take both`InApp` and `External` params or a url and build a local `External` Param object. This would mean that no matter what url or params you pass, this component would always open it in a new tab. I think we can build the local param inside `ngOnchanges` with reasonable defaults for missing properties.
- 
- Could we please rename `ExternalNavigationParamsNew` to `ExternalNavigationParams` ?

anandtiwary: thoughts? cc @aaron-steinfeld

aaron-steinfeld: Agree with the rename - how did that even get in?!

Also, yes - I think ideally the open in new tab component should be able to support and convert inApp params to external ones, but I personally don't feel it needs to hold up this PR, since we can add it later with fully backwards compatible support as it's a union type already.

anandtiwary: I missed it when i was migrating to the new interface. But you approved the PR, so....

anandtiwary: Sure we can make that change separately

projects/common/src/navigation/navigation.service.ts
```diff
@@ -73,6 +73,22 @@ export class NavigationService {
     return this.currentParamMap.getAll(parameterName);
   }
 
+  public constructExternalUrl(urlString: string): string {
```
aaron-steinfeld, line 76: thanks! looks good - with these changes, do we still need the changes in time range service or the new url util?

aaron-steinfeld (commented):
lgtm other than those two leftovers

projects/common/src/navigation/navigation.service.ts
alok-traceable: nope, I already reverted those changes.

projects/common/src/utilities/url/url-utilities.ts
```diff
@@ -0,0 +1,5 @@
+import { HttpParams } from '@angular/common/http';
+import { Params } from '@angular/router';
+
+export const getQueryParamStringFromObject = (params: Params): string =>
```
anandtiwary, line 4: Is this needed or should we use the below function directly?

projects/common/src/navigation/navigation.service.ts
aaron-steinfeld: oh right, the TR ones are gone - one of the url utils looks like it's still there being unused though

projects/common/src/utilities/url/url-utilities.ts
alok-traceable: It is better to have a separate function, firstly consumers didn't have to write a new instance every time, and secondly if later we want to do standardize queryParams key-value pair we need to change this function only.

aaron-steinfeld (approved):
Thanks @alok-traceable

anandtiwary (approved):

PR #860, opened 2021-05-19
fix: combo-box dropdown overflow-x

jake-bassett:
## Description
Remove horizontal scrollbar from combo-box dropdown.

### DIFF
--- a/projects/components/src/combo-box/combo-box.component.scss
+++ b/projects/components/src/combo-box/combo-box.component.scss
@@ -108,6 +108,7 @@
 
   .option-list {
     overflow-y: auto;
+    overflow-x: hidden;
     height: 100%;
   }
 


projects/components/src/combo-box/combo-box.component.scss
```diff
@@ -108,6 +108,7 @@
 
   .option-list {
     overflow-y: auto;
+    overflow-x: hidden;
```
anandtiwary, line 111: So it will clip it? Should we put ellipsis here?

anandtiwary (approved):

projects/components/src/combo-box/combo-box.component.scss
jake-bassett: It doesn't clip in my testing. This just fixes that few pixel overflow that adds a mostly useless scrollbar.

PR #865, opened 2021-05-20
feat: adding count label on multiselect component

alok-traceable:
## Description
Current behaviour of Multiselect component label is <firstSelectedItem> and <count> more. 
This PR contains changes which will Instead of showing "and <count>" will show "+<count>" tag with a grey background and border radius.

### Testing
Unit tests to verify the existence of the tag and its value.

### Checklist:


projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -54,11 +54,20 @@
...
         padding-left: 8px;
+        width: calc(100% - 32px - 52px);
+        flex-grow: 1;
+      }
+
+      .trigger-more-items {
+        padding: 0 6px;
+        background-color: $gray-2;
+        border-radius: 4px;
+        margin-left: 5px;
```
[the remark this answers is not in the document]
alok-traceable: Updated

projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -54,11 +54,22 @@
       display: flex;
       width: 100%;
       align-items: center;
-      padding-left: 8px;
 
       .trigger-label {
         @include body-2-medium();
         padding-left: 8px;
+        width: calc(100% - 32px - 38px);
```
anandtiwary, line 61: Is there a way to do this without calc and hardcoded widths?  What if we use `flex: 1 1 auto` ?

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -50,6 +50,9 @@ import { MultiSelectJustify } from './multi-select-justify';
             <ht-icon *ngIf="this.icon" [icon]="this.icon" [size]="this.iconSize"> </ht-icon>
             <div *ngIf="!this.isIconOnlyMode()" class="trigger-label-container">
               <ht-label class="trigger-label" [label]="this.triggerLabel"></ht-label>
+              <span *ngIf="this.selectedItemsCount > 1" class="trigger-more-items"
+                >+{{ this.selectedItemsCount - 1 }}</span
```
anandtiwary, line 54: @arjunlalb don't we have a component for this view?

arjunlalb: There was a ticket hanging around for this. Will dig that up.

projects/components/src/multi-select/multi-select.component.scss
alok-traceable: Trigger label will always overflow container unless it has some fixed or calculated width. Without it won't show ellipses.
I have tried `flex: 1 1 auto` but it is not solving the issue.

It looks hardcoded but actually, it is assuming only 2 simple things
1. trigger label count is less than 999.
2. Icon on the right has a width of 32px.

and for the last part, it is being taken care of by `flex-grow: 1` on trigger-label

projects/components/src/multi-select/multi-select.component.ts
alok-traceable: I will create another PR for it.

projects/components/src/multi-select/multi-select.component.scss
anandtiwary: okay. its fine.

projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -268,10 +280,14 @@ describe('Multi Select Component', () => {
...
     spectator.tick();
     const selectedCheckboxElements = spectator.queryAll('ht-checkbox', { root: true });
+    expect(spectator.query('.trigger-more-items')).toExist();
     expect(selectedCheckboxElements.length).toBe(6);
 
     expect(onChange).toHaveBeenCalledWith(selectionOptions.map(option => option.value));
-    expect(spectator.query(LabelComponent)?.label).toEqual('first and 5 more');
+    expect(spectator.query(LabelComponent)?.label).toEqual('first');
+
+    expect(spectator.component.selectedItemsCount).toBe(6);
```
anandtiwary, line 289: We should test this from Dom. Example:
```
expect(spectator.query('.trigger-more-items')).toHaveText('6')
```

alok-traceable: updated

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -236,10 +240,13 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
     );
     if (selectedItems === undefined || selectedItems.length === 0) {
       this.triggerLabel = this.placeholder;
+      this.selectedItemsCount = 0;
```
aaron-steinfeld, line 243: nit: can move this assignment out of the condition then collapse the conditional

```ts
this.selectedItemsCount = selectedItems?.length ?? 0;
this.triggerLabel = this.selectedItemsCount === 0 ? this.placeholder : selectedItems[0].label;

File: projects/components/src/multi-select/multi-select.component.scss [hunk 6]
```diff
@@ -54,11 +54,22 @@
...
+      }
+
+      .trigger-more-items {
+        padding: 0 4px;
+        background-color: $gray-2;
+        border-radius: 4px;
+        margin-left: 8px;
+        max-width: 38px;
+        font-size: 12px;
+        font-weight: 600;
```
aaron-steinfeld, line 72: can we pull this font from our shared font styles? We generally want all fonts to be defined in font.scss so we're not defining them adhoc in every style sheet.

projects/components/src/multi-select/multi-select.component.ts
alok-traceable: updated

projects/components/src/multi-select/multi-select.component.scss
alok-traceable: updated

aaron-steinfeld (approved):

PR #875, opened 2021-05-24
fix: changing card hover behaviour

alok-traceable:
## Description
Global changes on card component hover behavior.
- Removing grey background on hover
- Changes on box-shadow on hover
- Removing change in border color on hover

### Testing
Verified on the local environment. Mostly all changes are on stylings.

### Checklist:


projects/assets-library/assets/styles/_interaction.scss
```diff
@@ -9,12 +9,16 @@
 
 @mixin card-hover {
   &:hover {
-    @include dropdown-box-shadow();
+    @include box-shadow-2dp();
     cursor: pointer;
   }
 }
 
 @mixin filter-box-shadow {
```
aaron-steinfeld, line 17: nit: why the 2 separate mixins defined the same?

aaron-steinfeld (approved):

anandtiwary:
@alok-traceable Could you please share a before and after screenshot?

projects/observability/src/shared/components/card-list/card-list.component.scss
```diff
@@ -33,21 +34,13 @@
       }
 
       &:hover {
-        background: $blue-2;
```
anandtiwary, line 36: We don't want to show a darker blue when user hover on selected card?

alok-traceable: Need to remove these on hover on card
1. border
2. background color change

To maintain consistency on hover elements, selected and un-selected cards should follow the same behavior.
I can revert this change if required.

PR #876, opened 2021-05-24
feat: table cell no-op parser update

palbizu:
## Description
Table Cell NoOp was updated to allow single data in the parser

### Checklist:

### Documentation


projects/components/src/table/cells/data-parsers/table-cell-no-op-parser.ts
```diff
@@ -11,7 +11,7 @@ export class TableCellNoOpParser<T = unknown> extends TableCellParserBase<T, T,
   }
 
   public parseFilterValue(cellData: T | { filterValue?: string }): string | undefined {
-    if (typeof cellData === 'object' && cellData !== null && 'filterValue' in cellData) {
+    if ('filterValue' in cellData) {
```
aaron-steinfeld, line 14: the previous conditions should remain - the lint rule either needs to be ignored or that type fixed - it wasn't quite right. This would throw if cellData were null, for example

aaron-steinfeld (approved):

PR #884, opened 2021-05-27
feat: show selected item's icon in trigger label if applicable

arjunlalb:
## Description
show selected item's icon in trigger label if applicable

### Testing
Manually verified. UT added.

### Checklist:

### DIFF
--- a/projects/components/src/select/select.component.test.ts
+++ b/projects/components/src/select/select.component.test.ts
@@ -2,7 +2,7 @@ import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { fakeAsync, flush } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import { IconLibraryTestingModule, IconType } from '@hypertrace/assets-library';
-import { MemoizeModule, NavigationService } from '@hypertrace/common';
+import { NavigationService } from '@hypertrace/common';
 import { IconComponent } from '@hypertrace/components';
 import { createHostFactory, mockProvider, SpectatorHost } from '@ngneat/spectator/jest';
 import { MockComponent } from 'ng-mocks';
@@ -15,7 +15,7 @@ import { SelectModule } from './select.module';
 describe('Select Component', () => {
   const hostFactory = createHostFactory<SelectComponent<string>>({
     component: SelectComponent,
-    imports: [SelectModule, HttpClientTestingModule, IconLibraryTestingModule, MemoizeModule],
+    imports: [SelectModule, HttpClientTestingModule, IconLibraryTestingModule],
     declareComponent: false,
     declarations: [MockComponent(IconComponent)],
     providers: [
@@ -175,7 +175,7 @@ describe('Select Component', () => {
     expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.icon).toBe(
       'select-level-icon'
     );
-    expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.color).toBe(undefined);
+    expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.color).toBe(null);
 
     // Selection with no icon -> no icon and no color
     spectator.click('.trigger-content');
--- a/projects/components/src/select/select.component.test.ts
+++ b/projects/components/src/select/select.component.test.ts
@@ -175,6 +175,7 @@ describe('Select Component', () => {
     expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.icon).toBe(
       'select-level-icon'
     );
+    // tslint:disable-next-line:no-null-keyword
     expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.color).toBe(null);
 
     // Selection with no icon -> no icon and no color
--- a/projects/components/src/select/select.component.test.ts
+++ b/projects/components/src/select/select.component.test.ts
@@ -178,12 +178,15 @@ describe('Select Component', () => {
     // tslint:disable-next-line:no-null-keyword
     expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.color).toBe(null);
 
-    // Selection with no icon -> no icon and no color
+    // Selection with no icon -> default icon and no color
     spectator.click('.trigger-content');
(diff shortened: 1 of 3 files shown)


projects/components/src/select/select.component.ts
```diff
@@ -204,6 +210,27 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
...
 
+  public getPrefixIcon(
+    selectedOption: SelectOptionComponent<V> | undefined,
+    triggerIcon: string | undefined
+  ): string | undefined {
+    if (selectedOption !== undefined && selectedOption?.icon !== undefined) {
+      return selectedOption.icon;
+    }
+
+    if (selectedOption !== undefined) {
```
anandtiwary, line 221: if selected option is defined then we are returning undefined for the icon. I didn't understand this.

did you mean `selectedOption=== undefined` ?

projects/components/src/select/select.component.ts
```diff
@@ -55,7 +55,13 @@ import { SelectSize } from './select-size';
               class="trigger-content menu-with-border"
               [ngClass]="[this.justifyClass]"
             >
-              <ht-icon *ngIf="this.icon" class="trigger-prefix-icon" [icon]="this.icon" [size]="this.iconSize">
+              <ht-icon
+                *ngIf="this.getPrefixIcon | htMemoize: selected:this.icon"
```
aaron-steinfeld, line 59: simple conditional logic inside probably doesn't need a memoize (also, this.icon is accessible inside the function, anyway). Reserve the confusing syntax for heavy logic or anywhere that returns a new reference each time.

projects/components/src/select/select.component.ts
```diff
@@ -204,6 +210,27 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
     }
   }
 
+  public getPrefixIcon(
+    selectedOption: SelectOptionComponent<V> | undefined,
+    triggerIcon: string | undefined
+  ): string | undefined {
+    if (selectedOption !== undefined && selectedOption?.icon !== undefined) {
```
aaron-steinfeld, line 217: So we're favoring the selections icon in the trigger over the trigger's icon? that doesn't seem right to me, but I guess I'd also find it strange to use a trigger icon and individual icons together

aaron-steinfeld: (also, unnecessary first condition)
```suggestion
    if (selectedOption?.icon !== undefined) {
```

projects/components/src/select/select.component.ts
aaron-steinfeld: it's saying hide the trigger icon if anything is selected which matches the test logic. honestly not sure how we expect it to be used, but the logic can convey the current intent better IMO by collapsing the whole function to:
```ts
  return selectedOption ? selectedOption.icon : triggerIcon;
```

projects/components/src/select/select.component.ts
```diff
@@ -204,6 +210,27 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
...
+      return undefined;
+    }
+
+    return triggerIcon;
+  }
+
+  public getPrefixIconColor(selectedOption: SelectOptionComponent<V> | undefined): string | undefined {
+    return selectedOption !== undefined && selectedOption?.iconColor !== undefined
+      ? selectedOption.iconColor
+      : undefined;
```
aaron-steinfeld, line 231:
```suggestion
    return selectedOption?.iconColor;
```

projects/components/src/select/select.component.ts
arjunlalb: if there's a selection, we should show the selection's icon or nothing.
If there's no selection, show the trigger icon. That's the intent.

It doesn't make sense to show the regular trigger icon along with a selection. that would be confusing.

we already show selected label in the trigger the same way. This is extending it to the icon as well. 

Fixed the logic.

projects/components/src/select/select.component.ts
arjunlalb: fixed this.

projects/components/src/select/select.component.ts
arjunlalb: I think my brain wasn't functioning last night when I wrote the logic for this. 🤔

projects/components/src/select/select.component.ts
arjunlalb: simplified

projects/components/src/select/select.component.ts
arjunlalb: fixed.

projects/components/src/select/select.component.ts
aaron-steinfeld: Using the selected icon if defined makes sense, but not using the component wide icon if the selections aren't individually... iconized breaks the original intent. I think it's just two different styles:
image

in this one, that icon just is meant to apply to every option, but not appear as a duplicate in the option list.

projects/components/src/select/select.component.test.ts
```diff
@@ -146,12 +150,56 @@ describe('Select Component', () => {
...
+        hostProps: {
+          options: selectionOptions,
+          icon: 'select-level-icon'
+        }
+      }
+    );
+    spectator.tick();
+
+    // No selection -> select component level icon and no color
+    expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.icon).toBe(
```
aaron-steinfeld, line 175: is there a reason to use debug element querying rather than spectator querying ? is it because spectator doesn't give you access to the component on root queries?

projects/components/src/select/select.component.ts
arjunlalb: updated

projects/components/src/select/select.component.test.ts
```diff
@@ -146,12 +150,56 @@ describe('Select Component', () => {
...
+      'select-level-icon'
+    );
+    // tslint:disable-next-line:no-null-keyword
+    expect(spectator.debugElement.query(By.css('.trigger-prefix-icon')).componentInstance.color).toBe(null);
```
aaron-steinfeld, line 179: nit: `toBeNull()` (avoids the tslint disable)

aaron-steinfeld (approved):

projects/components/src/select/select.component.test.ts
arjunlalb: yep. wanted to access component to verify the contents of it.

PR #893, opened 2021-06-01
chore: allow ht link to support null from async pipe

aaron-steinfeld:
## Description
When used with async pipe, link would error on initial null state.

### DIFF
--- a/projects/components/src/link/link.component.ts
+++ b/projects/components/src/link/link.component.ts
@@ -1,6 +1,7 @@
 import { ChangeDetectionStrategy, Component, Input, OnChanges } from '@angular/core';
 import { NavigationExtras } from '@angular/router';
 import { NavigationParams, NavigationPath, NavigationService } from '@hypertrace/common';
+import { isNil } from 'lodash-es';
 
 @Component({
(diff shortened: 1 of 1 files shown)


projects/components/src/link/link.component.ts
```diff
@@ -38,7 +39,10 @@ export class LinkComponent implements OnChanges {
   }
 
   private setNavigationParams(): void {
-    if (this.paramsOrUrl !== undefined) {
+    if (isNil(this.paramsOrUrl)) {
+      this.navigationPath = undefined;
```
anandtiwary, line 43: Duplicate?

aaron-steinfeld: dammit.

projects/components/src/link/link.component.ts
```diff
@@ -38,7 +39,10 @@ export class LinkComponent implements OnChanges {
...
+      this.navigationPath = undefined;
```
aaron-steinfeld, line 44:
```suggestion
      this.navigationOptions = undefined;
```

anandtiwary (approved):

PR #902, opened 2021-06-04
feat: support for overriding color for a data point using a method

arjunlalb:
## Description
support for overriding color for a data point using a method

### Testing
Manually tested. UTs added.

### Checklist:


projects/observability/src/shared/components/cartesian/d3/data/series/cartesian-column.ts
```diff
@@ -34,14 +34,18 @@ export class CartesianColumn<TData> extends CartesianSeries<TData> {
...
+      context.strokeStyle = this.getColorForDatum(datum);
+      context.fillStyle = this.getColorForDatum(datum);
       context.fill();
       context.stroke();
       context.resetTransform();
     });
   }
 
+  private getColorForDatum(datum: TData): string {
+    return this.series.getColor ? this.series.getColor(datum) : this.series.color;
```
aaron-steinfeld, line 46: this can shorten with optional function access, but a broader question - wouldn't it make more sense to support specifying a color on the data object rather than a lookup function? that is: `{x, y, color}`

arjunlalb: it is possible, but the actual contents of the series is generic. Doing it that way would be much more intrusive and may have a broader impact across cartesians.

arjunlalb: updated to use optional function access.

aaron-steinfeld (approved):

PR #903, opened 2021-06-04
fix: word wrap update for confirmation modal content

arjunlalb:
## Description
word wrap update for confirmation modal content

### DIFF
--- a/projects/components/src/confirmation/confirmation-modal.component.scss
+++ b/projects/components/src/confirmation/confirmation-modal.component.scss
@@ -14,7 +14,7 @@
   .description {
     @include body-1-regular($gray-7);
     text-align: center;
-    word-break: break-all;
+    word-break: break-word;
   }
 
   .controls {


projects/components/src/confirmation/confirmation-modal.component.scss
```diff
@@ -14,7 +14,7 @@
   .description {
     @include body-1-regular($gray-7);
     text-align: center;
-    word-break: break-all;
+    word-break: break-word;
```
anandtiwary, line 17: Any side effects of this change? Possibly on API DNA - mark sensitive data side? 

cc @jake-bassett

arjunlalb: I think it should be fine. Tested at few places. Looked alright. We are just going from breaking mid-word to a more readable word break.

anandtiwary: Fine if you tested it.

anandtiwary (approved):

PR #920, opened 2021-06-09
feat: add filled check circle icon and larger icon size

jake-bassett:
## Description
New icon for a filled checked circle and a larger icon size.


projects/components/src/icon/icon-size.ts
```diff
@@ -4,6 +4,7 @@ export const enum IconSize {
   Medium = 'medium',
   Large = 'large',
   ExtraLarge = 'extra-large',
+  ExtraLarger = 'extra-larger',
```
anandtiwary, line 7: We need better names

anandtiwary (approved):

anandtiwary:
Should we switch to L, XL, XXL etc ?

projects/components/src/icon/icon-size.ts
jake-bassett: Oh, and here I was hoping to squeeze in an `ExtraMoreLargerest` at some point.

jake-bassett:
> Should we switch to L, XL, XXL etc ?

Yeah, if we add any more we should move to something else. It's not too bad yet. Dare I say, even a little fun. Any more though and it will be too much.

PR #923, opened 2021-06-10
Fix to retain table control filters on refresh

jake-bassett:
## Description
This fixes a bug where when a table filter is applied and a user uses the internal refresh in the top bar, the table controls get cleared and there is no indication to the user that the filter is still applied. With this change, we now look at the currently applied filters and reapply them to the table controls upon refresh.


itssharmasandeep (approved):
Looks good to me!!

PR #924, opened 2021-06-11
fix: svg content can overflow container

alok-traceable:
## Description
Text is cropping on an SVG graph because the container is overflow hidden and SVG is also overflow hidden on this particular graph. To solve this we have 3 options 
- Make text shorter. (Not recommended as text is already very short and reducing it even further may make it confusing)
- Reduce font size. (Not recommended as the font is already very small and reducing it even further may make it unreadable)
- Allow SVG to overflow and control it from the consumer container.

### Testing
Changes verified locally on all affected areas. This PR contains CSS changes only, no new test cases are added or updated

### Checklist:

### DIFF
--- a/projects/components/src/titled-content/titled-content.component.scss
+++ b/projects/components/src/titled-content/titled-content.component.scss
@@ -35,7 +35,6 @@
     height: 100%;
     width: 100%;
     flex: 1 1 auto;
-    overflow: hidden;
   }
 
   .footer {
--- a/projects/observability/src/shared/components/radar/radar-chart.scss
+++ b/projects/observability/src/shared/components/radar/radar-chart.scss
@@ -6,6 +6,10 @@
   height: 100%;
 
   ::ng-deep {
+    svg {
+      overflow: visible;
+    }
+
     .axis {
       line {
         stroke: $gray-1;


projects/components/src/titled-content/titled-content.component.scss
```diff
@@ -35,7 +35,6 @@
     height: 100%;
     width: 100%;
     flex: 1 1 auto;
-    overflow: hidden;
```
aaron-steinfeld, line 38: the  radar change is fine, but this one scares me - the titled content is used all over, by most widgets, and we don't want to allow any of them to overflow. Is there any way we can keep this change local to radar?

aaron-steinfeld: The other options not mentioned in your summary would be to improve the text wrapping, or introduce something like ellipsis (don't have the offending widget available at the moment, not sure if that would make sense there) or shrink the radar itself so the text can fit.

arjunlalb: +1 on this. 

This change on titled-content is very risky.

alok-traceable: Okay, I will see if there are any other workarounds for this. 

My concern is `overflow: hidden`
If something is overflowing the parent container overflow hidden is just going to crop the content which is also not a good solution. counterexamples would be cropped titles, units (ms, MB), graphs, borders, etc.
It would be better if we define minimum and maximum supported widths for UI and just make sure it is not breaking in that region by adjusting font sizes and spacings.

aaron-steinfeld: > If something is overflowing the parent container overflow hidden is just going to crop the content which is also not a good solution.

Agree - the role of this container is basically for a widget on a dashboard. We don't want to allow widgets to either overlap each other or scroll internally, but that doesn't mean we want them to be clipped - instead, it's on each particular widget to handle how to lay itself out in the space given. In this case, the radar is trying to use more space than it was given presumably due to some incorrect measurements (maybe it's using a fixed size for the text spacing for example, rather than adding the text then measuring the remaining room; or rather than giving the text its own overflow behavior).

PR #936, opened 2021-06-17
fix: setting default never for data$

anandtiwary:
## Description
fix: setting default never for data$

This will allow us to set dynamic loading message

### Testing

### Checklist:

### Documentation


projects/components/src/spinner/spinner.component.ts
```diff
@@ -52,9 +52,13 @@ export class SpinnerComponent implements OnChanges {
 
   public state$?: Observable<SpinnerAsyncState>;
 
+  public ngOnInit(): void {
```
aaron-steinfeld, line 55: onInit shouldn't be necessary on top of the onChanges. In fact, these days I find onInit is rarely what we want, it just distract from real workhorse, onChanges. I'd consider it maybe not a red flag, but a yellow one.

Specifically in this case as long as data$ is assigned, line 61 would have already run before we got here. This would only be useful in the case where data$ is never assigned (which isn't really a use case, but if you want to default state$ to NEVER on line 53, that's fine, avoids making it optional too).

anandtiwary: If you don't assign anything to data then like 61 wouldn't execute. Bare minimum, i will have to assign `undefined`. 

In my use case, I am not assigning anything to data$. So Spinner will always be in `loading` state and on the parent side we will just keep updating the loading message like 'Received 10 events'   ->  `Received 20 events`

aaron-steinfeld: You could just assign NEVER and it would work for your special case, right? why build it into the component?

anandtiwary: I could. But doesn't it read better? Also, shouldn't the spinner support undefined data$ case?

aaron-steinfeld: I could go either way, doesn't feel particularly better to me. That said, if you prefer that my initial suggestion was to default it to NEVER anyway - just via assignment rather than onInit.

anandtiwary: Just assigning to Never wouldn't work for the result i want :). I will have to do something like 
```
 public state$: Observable<SpinnerAsyncState> = this.mapObservableState(NEVER);
 ```
 
 Would this work?

anandtiwary: Actually, let me confirm it.

aaron-steinfeld: Yep or just `public state$: Observable<SpinnerAsyncState> = of(SpinnerAsyncState.Loading);`

anandtiwary: done

aaron-steinfeld (approved):

projects/components/src/spinner/spinner.component.ts
anandtiwary: Thanks. Don't you have more code reviews in another 7 hours?

aaron-steinfeld: more like 5

aaron-steinfeld (approved):

PR #950, opened 2021-06-28
fix: fix misc issues with gauge and donuts

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation


projects/observability/src/shared/components/donut/donut-builder.service.ts
```diff
@@ -126,19 +126,7 @@ export class DonutBuilderService extends D3VisualizationBuilderService<
   }
 
   protected decorateDimensions(calculatedDimensions: ChartDimensions): DonutDimensions {
-    let diameter = Math.min(calculatedDimensions.visualizationWidth, calculatedDimensions.visualizationHeight);
```
anandtiwary, line 129: @skjindal93 @aaron-steinfeld Do you guys remember why we need this logic? It seems to reduce the size of donut by a  lot. It creates problem with smaller container size. 

After I removed it, I didn't notice any issues.

skjindal93: Basically, we wanted the smaller dimension to decide the diameter for the donut. If the screen has more width than the height, the height determines the diameter and vice versa

If there is a way to automatically fit in the donut, irrespective of different screen sizes, then that would be perfect

aaron-steinfeld: Yeah, this logic was for sizing, based on the original donut mocks way back. The comments explain it, but it's mostly about legend, padding and fitting a square visualization into potentially rectangular bounding boxes like @skjindal93 mentioned.

anandtiwary: okay, so with the new changes, the Donut is still working fine. It is however making a better use of the space and appearing bigger. Apart from that, I am not seeing any failure in terms of rendering. We might want to reduce the size of the dashboard containers though so that it can match previous size.

aaron-steinfeld: I'm a little hesitant about this change, mainly wrt the legend. Two things are being changed here.
1. We're not assigning legend width (why not? are we not using a legend?). Does legend still work if rendered below? to the side?
2. We're not using the configured padding. This is fine if that's the intended visual change, but if so, any supporting code like that constant should be removed. That said, I'd guess that would show up as the legend sitting right next to the chart with 0 gap, which will look broken too.

The logic looks like it's doing these things regardless of legend position (If I vaguely recall, I think when written we always showed right legends) so that should be fixed, but this reads to me like we changed an assumption of a legend into an assumption of no legend, which is equally wrong.

anandtiwary: we are already using Flex in container. So css is automatically assigning them correct width and position. The padding also appears to be fine visually.

projects/observability/src/shared/components/donut/donut-builder.service.ts
```diff
@@ -38,8 +38,8 @@ export class DonutBuilderService extends D3VisualizationBuilderService<
   private static readonly DONUT_VALUE_CLASS: string = 'donut-value';
   private static readonly DONUT_ARC_CLASS: string = 'donut-arc';
 
-  private static readonly DONUT_PADDING_PX: number = 60;
-  private static readonly LEGEND_SIZE_MULTIPLE: number = 1.5;
+  public static readonly DONUT_PADDING_PX: number = 10;
+  public static readonly LEGEND_SIZE_MULTIPLE: number = 1.5;
```
aaron-steinfeld, line 42: can you remove the unused ones and revert any still used to private

aaron-steinfeld (commented):
lgtm other than small cleanup. let's see what happens!

.prettierignore
```diff
@@ -15,3 +15,4 @@ browserslist
 LICENSE*
 conf/
 coverage/
+test-results/
```
jake-bassett, line 18: Thank you! Been meaning to do this for... months.

jake-bassett (approved):

PR #963, opened 2021-07-01
Adding new style button

palbizu:
## Description

- Add a new style called 'Plain Text' similar to 'Text' but without padding. 

### Testing
Visual testing. 

- 'Text' style with padding

image

- 'Plain Text' style with no padding

image

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/button/button.component.ts
+++ b/projects/components/src/button/button.component.ts
@@ -17,7 +17,7 @@ import { ButtonRole, ButtonSize, ButtonStyle } from './button';
           [label]="this.label"
           [size]="this.getIconSizeClass()"
           class="icon leading"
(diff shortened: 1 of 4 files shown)


projects/components/src/button/button.component.ts
```diff
@@ -45,6 +47,9 @@ export class ButtonComponent {
   @Input()
   public trailingIcon?: boolean;
 
+  @Input()
+  public flippingIcon?: boolean;
```
anandtiwary, line 51: can we use a different icon instead ?

palbizu: yes, removed it.

anandtiwary (approved):
Not sure if we really need a no padding version for Text display mode. But it is fine for now. We can come back to it later if required.

PR #979, opened 2021-07-08
fix: support disabling options in multi-select

arjunlalb:
## Description
support disabling options in multi-select

### Testing
Manually verified. UTs added.

### Checklist:
- [x I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/projects/components/src/multi-select/multi-select.component.scss
+++ b/projects/components/src/multi-select/multi-select.component.scss
@@ -92,7 +92,7 @@
   flex-direction: column;
 
   .divider {
-    padding: 0px 16px 4px 16px;
+    padding: 0 16px 4px 16px;
   }
 
   .multi-select-options {
@@ -107,8 +107,13 @@
     font-size: 15px;
     align-items: center;
 
+    &.disabled {
+      color: $gray-5;
+      cursor: not-allowed;
+    }
+
     .checkbox {
-      margin: 0px;
+      margin: 0;
       color: $gray-3;
 
       ::ng-deep .mat-ripple {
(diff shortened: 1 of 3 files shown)


projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -92,7 +92,7 @@
   flex-direction: column;
 
   .divider {
-    padding: 0px 16px 4px 16px;
+    padding: 0 16px 4px 16px;
```
anandtiwary, line 95: why remove px ?

projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -107,8 +107,13 @@
...
     align-items: center;
 
+    &.disabled {
+      color: $gray-5;
+      cursor: not-allowed;
+    }
+
     .checkbox {
-      margin: 0px;
+      margin: 0;
```
anandtiwary, line 116: same question here

anandtiwary (commented):
Looks good apart from the one question

projects/components/src/multi-select/multi-select.component.scss
arjunlalb: zeroes don't need px :)

anandtiwary: so this is just a nit change?

anandtiwary (approved):

projects/components/src/multi-select/multi-select.component.scss
arjunlalb: yea. this is just a minor cleanup since I was in the area :)

PR #1013, opened 2021-07-20
feat: add esnext.string to ts libs

jake-bassett:
## Description
Added esnext.string to tsconfig libs.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -16,7 +16,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "downlevelIteration": true,
-    "lib": ["es2015", "es2016", "es2017", "esnext.array", "esnext.asynciterable", "dom"],
+    "lib": ["es2015", "es2016", "es2017", "esnext.string","esnext.array", "esnext.asynciterable", "dom"],
     "paths": {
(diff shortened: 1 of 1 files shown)


tsconfig.base.json
```diff
@@ -16,7 +16,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "downlevelIteration": true,
-    "lib": ["es2015", "es2016", "es2017", "esnext.array", "esnext.asynciterable", "dom"],
+    "lib": ["es2015", "es2016", "es2017", "esnext.string","esnext.array", "esnext.asynciterable", "dom"],
```
aaron-steinfeld, line 19: did you verify that we polyfill these?

jake-bassett: Not _everything_, but I tested `replaceAll`. What else should I look for?

aaron-steinfeld: we basically need to make sure that it works in any browser we support (not just local) - so either it's been implemented in all that we support, or safer - that core-js polyfills replaceall (I think that's all that comes in here, right?)

jake-bassett: Ah, that I did check.

We are good on all modern browsers. 

https://caniuse.com/?search=replaceAll

aaron-steinfeld (approved):

PR #1019, opened 2021-07-23
feat: add option to open link in new tab

jake-bassett:
## Description
This provides an option to open links in a new tab.

### DIFF
--- a/projects/components/src/link/link.component.ts
+++ b/projects/components/src/link/link.component.ts
@@ -12,7 +12,7 @@ import { EMPTY, Observable } from 'rxjs';
     <a
       *htLetAsync="this.navData$ as navData"
       class="ht-link"
-      target="{{this.openInNewTab ? '_blank' : ''}}"
+      target="{{ this.openInNewTab ? '_blank' : '' }}"
       rel="noopener noreferrer"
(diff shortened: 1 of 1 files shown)


palbizu (approved):

projects/components/src/link/link.component.ts
```diff
@@ -12,6 +12,8 @@ import { EMPTY, Observable } from 'rxjs';
     <a
       *htLetAsync="this.navData$ as navData"
       class="ht-link"
+      target="{{ this.openInNewTab ? '_blank' : '' }}"
```
aaron-steinfeld, line 15: how's this differ from the existing way of doing this?

This component takes an ExternalNavigationParams as an input, which includes window handling.

jake-bassett: Hmm. I thought it suspicious that we didn't have this already. I though that might be how, but it didn't seem to work when I was playing with it. I'll circle back and try again.

anandtiwary: Yep.

anandtiwary: ExternalNavigationParams option should support this.

PR #1030, opened 2021-07-26
feat: time range uses default when unavailable

jake-bassett:
## Description
Calling `getCurrentTimeRange()` before initialization now provides the default time range instead of throwing an error.


palbizu (approved):

aaron-steinfeld (changes requested):
We should fix any usages to not hit this error, rather than hiding it.

PR #1024, opened 2021-07-26
fix: hide navigation header if there is nothing underneath

itssharmasandeep:
## Description
Please include a summary of the change, motivation and context.

### Testing
Local testing done.

### Checklist:
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #1035, opened 2021-07-28
fix: hide navigation header if nothing underneath

itssharmasandeep:
## Description
hide navigation header if there is nothing underneath

### Testing
Local Testing done.

### Checklist:

There was some problem with https://github.com/hypertrace/hypertrace-ui/pull/1024/ so created this new one.


jake-bassett (approved):

projects/components/src/navigation/nav-item/nav-item.component.ts
```diff
@@ -11,7 +11,7 @@ import { NavItemLinkConfig } from '../navigation-list.component';
   template: `
     <ht-link *ngIf="this.config" [paramsOrUrl]="buildNavigationParam | htMemoize: this.config">
       <div
-        *htIfFeature="this.config.features | htFeature as featureState"
+        *htIfFeature="this.config.featureState$ | async as featureState"
```
anandtiwary, line 14: ooh.. @itssharmasandeep I think we still need to apply `htFeature` here

itssharmasandeep: Nope

itssharmasandeep: `htFeature` takes the `features` which are `string[]` and convert this to `featureState`
I am converting this to `featureState` in the service itself and pass here

PR #1046, opened 2021-08-02
feat: Adding ticks count

palbizu:
## Description
Adding ticks count in Cartesian Chart Component

### Testing
Visual Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/components/cartesian/chart.ts
+++ b/projects/observability/src/shared/components/cartesian/chart.ts
@@ -125,7 +125,7 @@ export interface Axis {
   max?: number;
 
   /**
-   * Determine the tick count or the interval of time between each tick.
+   * Determine the tick count labels (number or time interval)
    */
   tickCount?: number | TimeInterval;
 }
--- a/projects/observability/src/shared/components/cartesian/chart.ts
+++ b/projects/observability/src/shared/components/cartesian/chart.ts
@@ -1,4 +1,5 @@
 import { TimeRange } from '@hypertrace/common';
+import { TimeInterval } from 'd3-time';
 import { LegendPosition } from '../legend/legend.component';
 import { ChartTooltipRef } from '../utils/chart-tooltip/chart-tooltip-popover';
 import { ChartEvent, ChartEventListener } from './chart-interactivty';
@@ -124,9 +125,9 @@ export interface Axis {
   max?: number;
 
   /**
-   * Determine the distance in days between each tick.
+   * Determine the tick count or the interval of time between each tick.
    */
-  tickDaysDistance?: number;
+  tickCount?: number | TimeInterval;
 }
 
 export interface AxisCrosshair {
--- a/projects/observability/src/shared/components/cartesian/chart.ts
+++ b/projects/observability/src/shared/components/cartesian/chart.ts
@@ -122,6 +122,11 @@ export interface Axis {
    * Maximum value of the axis. If unset, defaults to maximum value of provided data.
    */
   max?: number;
+
+  /**
+   * Determine the distance in days between each tick.
+   */
+  tickDaysDistance?: number;
 }
 
 export interface AxisCrosshair {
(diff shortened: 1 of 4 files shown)


projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis.ts
```diff
@@ -1,5 +1,6 @@
 import { Axis as D3Axis, axisBottom, AxisDomain, axisLeft, axisRight, AxisScale, axisTop } from 'd3-axis';
 import { BaseType, select, Selection } from 'd3-selection';
+import * as d3 from 'd3-time';
```
anandtiwary, line 3: A '*' import like this would import all modules from d3 and increase the bundle size. We can just do `import {timeDay} from 'd3-time'`

palbizu: Yes, fixed

projects/observability/src/shared/components/cartesian/chart.ts
```diff
@@ -122,6 +122,11 @@ export interface Axis {
    * Maximum value of the axis. If unset, defaults to maximum value of provided data.
    */
   max?: number;
+
+  /**
+   * Determine the distance in days between each tick.
+   */
+  tickDaysDistance?: number;
```
anandtiwary, line 129: hmm, so does this depend on the range of time in data points? What if all data points are within the last hour and `tickDaysDistance `is set to 4? How many ticks would it end up generating?

projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis.ts
```diff
@@ -235,4 +240,4 @@ export class CartesianAxis<TData = {}> {
   }
 }
 
-type DefaultedAxisConfig = Axis & Omit<Required<Axis>, 'scale' | 'crosshair' | 'min' | 'max'>;
+type DefaultedAxisConfig = Axis & Omit<Required<Axis>, 'scale' | 'crosshair' | 'min' | 'max' | 'tickDaysDistance'>;
```
jake-bassett, line 243: This naming seems very specific to the use case and not what I would expect in a generic component.

projects/observability/src/shared/components/cartesian/chart.ts
palbizu: I understand that this logic only works with periods of time relative to days, to adjust a change of that type you should go back to the variable of type TimeInterval and pass the specific function directly to it.

projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis.ts
palbizu: I will change it for a less specific naming/data

projects/observability/src/shared/components/cartesian/chart.ts
```diff
@@ -122,6 +123,11 @@ export interface Axis {
    * Maximum value of the axis. If unset, defaults to maximum value of provided data.
    */
   max?: number;
+
+  /**
+   * Determine the tick count or the interval of time between each tick.
```
jake-bassett, line 128: The name is good, but the second part of this description seems off to me. What is meant by interval of time?

Isn't this property just setting how many ticks to see on the axis? It shouldn't really matter what values or even meaning (like time) the x-axis has, or am I misunderstanding?

Additionally, mentioning "time" is another more use case specific thing. While most cartesians in the product use a time based X-axis, there are some that could be other non-time values. For instance, we could do a group-by in Explorer, then the x-axis represents those buckets of metrics, rather than a time range.

palbizu: Done.

jake-bassett (approved):

projects/observability/src/shared/components/cartesian/chart.ts
```diff
@@ -122,6 +123,11 @@ export interface Axis {
    * Maximum value of the axis. If unset, defaults to maximum value of provided data.
    */
   max?: number;
+
+  /**
+   * Determine the tick count labels (number or time interval)
+   */
+  tickCount?: number | TimeInterval;
```
aaron-steinfeld, line 130: Why would this be a d3 time interval? d3 is not even part of the public api of chart

aaron-steinfeld (changes requested):

PR #1051, opened 2021-08-06
refactor: splitting request response in its own query

anandtiwary:
## Description

Need to update test and fix any lint

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/distributed-tracing/src/shared/components/span-detail/span-detail.component.ts
+++ b/projects/distributed-tracing/src/shared/components/span-detail/span-detail.component.ts
@@ -1,6 +1,7 @@
 import { ChangeDetectionStrategy, Component, EventEmitter, Input, OnChanges, Output } from '@angular/core';
 import { TypedSimpleChanges } from '@hypertrace/common';
 import { isEmpty } from 'lodash-es';
+import { combineLatest } from 'rxjs';
 import { SpanData } from './span-data';
 import { SpanDetailLayoutStyle } from './span-detail-layout-style';
 import { SpanDetailTab } from './span-detail-tab';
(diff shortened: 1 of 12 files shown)


projects/distributed-tracing/src/shared/components/span-detail/span-detail.component.ts
```diff
@@ -78,8 +78,8 @@ export class SpanDetailComponent implements OnChanges {
 
   public ngOnChanges(changes: TypedSimpleChanges<this>): void {
     if (changes.spanData) {
-      this.showRequestTab = !isEmpty(this.spanData?.requestHeaders) || !isEmpty(this.spanData?.requestBody);
-      this.showResponseTab = !isEmpty(this.spanData?.responseHeaders) || !isEmpty(this.spanData?.responseBody);
+      this.showRequestTab = !isEmpty(this.spanData?.requestHeaders) || !isEmpty(this.spanData?.requestBody$);
```
aaron-steinfeld, line 81: this isEmpty check isn't meaningful on an observable. If we want to conditionally show the tab, then we need to fetch the data (could make this async), but that seems like it defeats the purpose of making it lazy.

anandtiwary: This field is not available in hypertrace. Hence this logic.

jake-bassett: This check seems to be what is causing a bug in waterfall. That check on the observable is always `true` since its an object, so the tabs are always showing.

anandtiwary: Oh yeah. Good review, Aaron!

anandtiwary: Should we convert the showRequestTab into an observable so that it can consume RequestBody$ correctly?

jake-bassett: Any reason why this won't work?

Screen Shot 2021-08-06 at 1 57 50 PM

anandtiwary: You may have to trigger change detection, idk. 

But this got me thinking. Would we ever have a case when request header is not present but requestBody will be? I would think not. So, imo we can just check presence of requestHeaders. and we wont need observables or subscription

jake-bassett: This seems to work in my testing so far. Going to run with it.

anandtiwary: @jake-bassett please check my latest comment

jake-bassett: Ah, just saw your last couple comments.

I agree with your assumptions about not having a body without headers. That is probably safe... buuuuut... I have tested the above code some more and am fairly confident this works for the case of a body with no headers. Not that we will likely ever run into it, but I see no harm in accounting for it if we already have the code.

aaron-steinfeld: It's a double whammy - it both brings in a code subscribe rather than in the template (nooo!) but more relevant here, it makes the query eager, which I thought is exactly what this whole change is to avoid, I thought? We shouldn't be fetching until the tab is clicked on - the hiding logic being no longer relevant. If we think that there are scenarios where we won't even query this data (e.g. the attribute isn't present), then we need to make the whole property optional so we don't have to subscribe to it to find that out).

aaron-steinfeld (changes requested):
marking it as request changes so we don't accidentally merge this.

anandtiwary:
This was a temporary change. For some reason I don't remember, we did not merge this change in UI. I am going to close this for now. Will reopen it if required.

PR #1052, opened 2021-08-09
268 p99 sort

jyothishjose6190:
## Description
Made dashboard configs runtime configurable.

- **on a feature**: Created a new config service that will read from config.json while the app initializes. 

### Testing
All tests are passed 

### Checklist:

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.test.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.test.ts
@@ -1,3 +1,4 @@
+import { HttpClientModule } from '@angular/common/http';
 import {
   Filter,
   FilterAttributeType,
@@ -20,7 +21,7 @@ import { NavigableDashboardComponent } from './navigable-dashboard.component';
 describe('Navigable dashboard component', () => {
   const hostFactory = createHostFactory({
     component: NavigableDashboardComponent,
-    imports: [LoadAsyncModule],
+    imports: [LoadAsyncModule, HttpClientModule],
     declarations: [MockComponent(ApplicationAwareDashboardComponent), MockComponent(FilterBarComponent)],
     providers: [
       mockProvider(MetadataService, {
--- a/projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
@@ -4,12 +4,15 @@ import { Filter } from '@hypertrace/components';
 import { DashboardPersistenceService } from '@hypertrace/dashboards';
 import { Dashboard, ModelJson } from '@hypertrace/hyperdash';
 import { EMPTY, Observable, of } from 'rxjs';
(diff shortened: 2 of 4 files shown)


aaron-steinfeld (changes requested):
I'm guessing you're trying to load a different dashboard config - let's have a design discussion first, please. A couple points:
- Similar to the discussion on the configurable time ranges, _anything_ we do on the UI side is equivalent to a fork - since loading from a config file implies that your config file is part of the image, so a custom image has been built. 
- dashboard persistence service is the placeholder to support loading a dashboard from config, no need to introduce a new one.

jyothishjose6190:
> I'm guessing you're trying to load a different dashboard config - let's have a design discussion first, please. A couple points:
> 
> * Similar to the discussion on the configurable time ranges, _anything_ we do on the UI side is equivalent to a fork - since loading from a config file implies that your config file is part of the image, so a custom image has been built.
> * dashboard persistence service is the placeholder to support loading a dashboard from config, no need to introduce a new one.
@aaron-steinfeld  Sorry I thought you are part of hypertrace slack channel. There We had a discussion from the beginning of the work.  The final conclusion was to make a fork or make it configurable.

aaron-steinfeld:
> > I'm guessing you're trying to load a different dashboard config - let's have a design discussion first, please. A couple points:
> > 
> > * Similar to the discussion on the configurable time ranges, _anything_ we do on the UI side is equivalent to a fork - since loading from a config file implies that your config file is part of the image, so a custom image has been built.
> > * dashboard persistence service is the placeholder to support loading a dashboard from config, no need to introduce a new one.
> >   @aaron-steinfeld  Sorry I thought you are part of hypertrace slack channel. There We had a discussion from the beginning of the work.  The final conclusion was to make a fork or make it configurable.

Not sure which channel it is, but feel free to invite me. Making it configurable is great (as well as the long term direction), but configurable only at build time is equivalent to a fork - we'd need it to be configurable at least deploy time (see the discussion on time ranges for details here).

jyothishjose6190:
_At deploy time, when spinning up a docker image (either via helm or directly via docker/compose). We do this for example via helm values files for other services, but since the UI is static content, we don't really do any configuration in this way today. That said, it's technically possible if we either do some scripting on startup or load config files into the browser and apply them. I'm pretty hesitant to go down either path though, to be honest. This option would also affect all tenants of the deployment._  This is the comment from time range. Here what I am doing is wrote some script to load a config at startup. if it is not there then will load the default one in the dashbaord persistence service.

aaron-steinfeld:
> _At deploy time, when spinning up a docker image (either via helm or directly via docker/compose). We do this for example via helm values files for other services, but since the UI is static content, we don't really do any configuration in this way today. That said, it's technically possible if we either do some scripting on startup or load config files into the browser and apply them. I'm pretty hesitant to go down either path though, to be honest. This option would also affect all tenants of the deployment._  This is the comment from time range. Here what I am doing is wrote some script to load a config at startup. if it is not there then will load the default one in the dashboard persistence service.

what script? that script would need to be part of the image, too - basically, converting helm values into files in the container. Still don't think that's the right approach, though for the same reasons discussed on that other PR.

Another consideration - in both approaches (ui asset file or gql config), the application is blocked from loading until a remote call fetches the dashboard (even if not configured to use a config file), so we're slowing down our load times.

PR #1053, opened 2021-08-09
fix: donut center minimum font size

alok-traceable:
## Description
Change to have a minimum font size of the title text at the donut center, else it can go to unreadable font sizes.

### Testing
Tested locally not breaking anything.

### Checklist:


projects/observability/src/shared/components/donut/donut-builder.service.ts
```diff
@@ -74,7 +75,10 @@ export class DonutBuilderService extends D3VisualizationBuilderService<
       .attr('transform', `translate(0,-${dimensions.donutInnerRadius / 2})`)
       .attr(
         'font-size',
-        Math.min(Math.floor(dimensions.donutInnerRadius / 8), DonutBuilderService.MAX_FONT_SIZE_FOR_TITLE)
```
itssharmasandeep, line 77: Code wise looks fine to me.
It is just we have to see what could be the min size for donut and for that case does this title text overflow?
cc: @jake-bassett @anandtiwary

anandtiwary: fair point. Is there a min size for the donut too? If yes, then deciding a min font size would be easier.

jake-bassett: We necessarily have a known minimum size for the donut itself, but for our screensizes we try to test down to 1366-by-768. 

As a matter of what to optimize for, I would be more concerned about the text spilling out than it getting too small. I think the overflow would look more broken than the text being illegible.

alok-traceable: So, there would be 2 aspects of this, 

1. Donut size, with smaller donut size than this, ideally we shouldn't put anything in between. If it is required then we can just put "value with a tooltip" and the font size of the value is still calculated so these PR doesn't affect that logic.
2. Font size of the title, the most important thing here should be whether the feature we are putting is usable or not, with a smaller (calculated) font size even though it looks good but it can't be read without a lens. The title is always static if it is overflowing we should be changing the donut size to make it useable and not reduce the font size to make it look good. IMO usability is always prefered over visuals.

alok-traceable: I won't recommend hiding the center content dynamically, this would be adding a factor to unpredictability also this would take away the freedom of putting only value at the center without a title. 
Anyway, it is a static title and while developing we can always make sure it is not overflowing in any of the supported screen widths and given static title length.

jake-bassett (approved):

PR #1060, opened 2021-08-10
feat: adding grey stroke in option

palbizu:
## Description
Adding border-bottom line in selected option.

### Testing
Visual testing

image

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/select/select.component.scss
+++ b/projects/components/src/select/select.component.scss
@@ -158,34 +158,45 @@
   .select-option {
     display: flex;
     height: 34px;
-    padding: 4px 16px;
+    padding: 0px 16px;
     cursor: pointer;
     font-size: 14px;
-    align-items: center;
-    justify-content: space-between;
-    white-space: nowrap;
 
     &.disabled {
       color: $gray-5;
       cursor: not-allowed;
     }
 
-    .select-option-info {
+    .select-option-content {
+      width: 100%;
       display: flex;
-      flex-direction: row;
       align-items: center;
+      justify-content: space-between;
+      white-space: nowrap;
+      padding: 4px inherit; 
 
-      & > * {
-        margin-right: 8px;
+      &.selected-option {
+        border-bottom: 1px solid $gray-2;
+        padding: 0px inherit;
       }
-    }
 
-    .label {
-      padding-right: 8px;
-    }
-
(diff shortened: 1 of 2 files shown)


jake-bassett:
@palbizu do you have an updated screenshot? The gray stroke looks right if both white and blue are selected, but it looks like white is not selected.

Also are we adding the config options for gray border to the white square in this PR or is that separate?

palbizu:
> @palbizu do you have an updated screenshot? The gray stroke looks right if both white and blue are selected, but it looks like white is not selected.
> 
> Also are we adding the config options for gray border to the white square in this PR or is that separate?

The gray border will be added in a new PR. I will attach all possible options selected

jake-bassett (changes requested):
Talked with @palbizu offline, but setting this to request changes.

The ask here is that the gray line divides all of the checked options from the unchecked options. The checked options should float to the top (in the original order) above the line. The unchecked options should sink to the bottom (in the original order) below the line.

palbizu:
> Talked with @palbizu offline, but setting this to request changes.
> 
> The ask here is that the gray line divides all of the checked options from the unchecked options. The checked options should float to the top (in the original order) above the line. The unchecked options should sink to the bottom (in the original order) below the line.

Adding a line in the last selected element. I think that reordering of options should be a responsability of the parent component.

anandtiwary:
Is this ready for another round of review?

projects/components/src/select/select.component.scss
```diff
@@ -158,34 +158,45 @@
   .select-option {
     display: flex;
     height: 34px;
-    padding: 4px 16px;
+    padding: 0px 16px;
```
arjunlalb, line 161: avoid `px` for zero values

projects/components/src/select/select.component.scss
```diff
@@ -158,34 +158,45 @@
...
       align-items: center;
+      justify-content: space-between;
+      white-space: nowrap;
+      padding: 4px inherit;
 
-      & > * {
-        margin-right: 8px;
+      &.selected-option:last-child {
+        border-bottom: 1px solid $gray-2;
+        padding: 0px inherit;
```
arjunlalb, line 180: here as well

PR #1071, opened 2021-08-17
Initial draft for new loaders using gifs

alok-traceable:
## Description
New gif loader.

### Checklist:


anandtiwary:
Discussed this in our call today. Overall a good change.

Few Action items:
- Rename `size` to `type`
- Rename Large, Medium, Small to better names like Page, widget etc. We can look into how UX team is planning to use them and then decide the appropriate name
- Add a similar `type` property to `htLoadAsync` directive and propagate it to the loader wrapper and the loader component. Set an appropriate default value that would take care of majority of cases.
- Then go and set specific value to any loader which needs a different loader than the default one.

alok-traceable:
> Discussed this in our call today. Overall a good change.
> 
> Few Action items:
> 
> * Rename `size` to `type`
> * Rename Large, Medium, Small to better names like Page, widget etc. We can look into how UX team is planning to use them and then decide the appropriate name
> * Add a similar `type` property to `htLoadAsync` directive and propagate it to the loader wrapper and the loader component. Set an appropriate default value that would take care of majority of cases.
> * Then go and set specific value to any loader which needs a different loader than the default one.

updated

projects/assets-library/src/images/image-type.ts
```diff
@@ -0,0 +1,12 @@
+export const enum ImagesAssetPath {
+  ErrorPage = 'assets/images/error-page.svg',
```
anandtiwary, line 2: I will raise the gif scalability issue with Michael today.

projects/components/src/load-async/load-async.directive.ts
```diff
@@ -65,6 +60,23 @@ export class LoadAsyncDirective implements OnChanges, OnDestroy {
   private buildWrapperView(): ComponentRef<LoadAsyncWrapperComponent> {
     const componentFactory = this.componentFactoryResolver.resolveComponentFactory(LoadAsyncWrapperComponent);
 
+    // Second param for structural directive is undefined until this.data$ is defined
+    // So putting this assignment in constuctor will not work
+    // This will execute only once as this method is called only
+    this.wrapperInjector = Injector.create({
```
anandtiwary, line 66: We don't need to create a new injector. It would be problematic and also it wouldn't work since right now we are only creating a new instance of component when data$ reference itself change. New emits on data$ is automatically passed via the state$ property.

What I would recommend it to modify line 45 instead and add loaderType$ as a property in `LoadAsyncWrapperParameters`. 

Create a `ReplaySubject` for loaderType -> 
```
private readonly loaderTypeSubject: BehaviourSubject<LoaderType> = new BehaviourSubject(<default-value>);
private readonly loaderType$: Observable<LoaderType>= this.loaderTypeSubject.asObservable();
```

Then use add loaderType$ on like 46. 

and add a separate ngOnChanges trigger for loaderType, that would just emit on this subject.

projects/assets-library/src/images/image-type.ts
```diff
@@ -0,0 +1,12 @@
...
+  LoaderSmallCircle = 'assets/images/loader-small.gif',
+  LoaderLargeSquare = 'assets/images/loader-large.gif',
+  LoaderHorizontal = 'assets/images/loader-horizontal.gif'
+}
+
+export const enum LoaderTypes {
```
anandtiwary, line 8: LoaderType

projects/assets-library/src/images/image-type.ts
```diff
@@ -0,0 +1,12 @@
...
+  SmallCircle = 'small-circle',
+  Horizontal = 'horizontal',
+  LargeSquare = 'large-square'
```
anandtiwary, line 11: Spoke to Michael. 

Let's use smallCircle directly in Spinner. We can remove it from here

Horizontal can be renamed to `ExpandableRow`. LargeSquare can be renamed to 'Page'.

anandtiwary: Also, we would need to bring back the old svg icon approach for other skeleton loaders.

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,15 +1,38 @@
...
+import { ImagesAssetPath, LoaderTypes } from '@hypertrace/assets-library';
+import { assertUnreachable } from '@hypertrace/common';
 
 @Component({
   selector: 'ht-loader',
   styleUrls: ['./loader.component.scss'],
   template: `
     <div class="ht-loader">
-      <ht-icon icon="${IconType.Loading}" size="${IconSize.Large}"></ht-icon>
+      <img [ngClass]="[this.type]" [src]="this.imagePath" />
```
anandtiwary, line 10: So, as per my other comment, Please have both ht-icon and img here. If loader type is RowExpandable or Page then we use Img tag. If not, we can use the existing ht-icon. So, let's ng-if on this condition.

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,15 +1,38 @@
...
 })
-export class LoaderComponent {}
+export class LoaderComponent implements OnChanges {
+  @Input()
+  public type: LoaderTypes = LoaderTypes.Horizontal;
+  public imagePath: ImagesAssetPath = ImagesAssetPath.LoaderHorizontal;
+
+  public ngOnChanges(): void {
+    // Unfortunatly passing undefined is overwriting default value
+    this.type = this.type ?? LoaderTypes.Horizontal;
```
anandtiwary, line 22: We shouldn't be modifying the input property. From where, is it getting the undefined value?

projects/components/src/load-async/loader/loader.component.ts
alok-traceable: https://github.com/hypertrace/hypertrace-ui/pull/1071#discussion_r697169599

projects/components/src/load-async/loader/loader.component.ts
alok-traceable: updated

projects/assets-library/src/images/image-type.ts
alok-traceable: updated

projects/components/src/load-async/load-async.directive.ts
alok-traceable: updated

alok-traceable:
I will continue on writing test cases, is there anything else pending on this PR? 
ping: @anandtiwary

projects/components/src/load-async/load-async.directive.ts
```diff
@@ -23,8 +24,11 @@ import {
 export class LoadAsyncDirective implements OnChanges, OnDestroy {
   @Input('htLoadAsync')
   public data$?: Observable<unknown>;
+  @Input()
+  public htLoadAsyncLoaderType!: LoaderType;
```
anandtiwary, line 28: this should be optional. You can assign a default value here

projects/components/src/load-async/load-async.directive.ts
```diff
@@ -23,8 +24,11 @@ import {
...
+
   private readonly wrapperParamsSubject: ReplaySubject<LoadAsyncWrapperParameters> = new ReplaySubject(1);
-  private readonly wrapperInjector: Injector;
+  private wrapperInjector!: Injector;
```
anandtiwary, line 31: any reason for removing the readonly?

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,15 +1,36 @@
...
+      <img [ngClass]="[this.type]" [src]="this.imagePath" />
     </div>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class LoaderComponent {}
+export class LoaderComponent implements OnChanges {
+  @Input()
+  public type: LoaderType = LoaderType.ExpandableRow;
+  public imagePath: ImagesAssetPath = ImagesAssetPath.LoaderExpandableRow;
```
anandtiwary, line 18: Is formatting right? Can add a new line after L17

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,15 +1,36 @@
...
-      <ht-icon icon="${IconType.Loading}" size="${IconSize.Large}"></ht-icon>
+      <img [ngClass]="[this.type]" [src]="this.imagePath" />
     </div>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class LoaderComponent {}
+export class LoaderComponent implements OnChanges {
+  @Input()
+  public type: LoaderType = LoaderType.ExpandableRow;
```
anandtiwary, line 17: Should this be the default?

projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
```diff
@@ -39,9 +40,13 @@ export class LoadAsyncWrapperComponent {
 
   public content?: TemplateRef<LoadAsyncContext>;
 
+  private readonly loaderTypeSubject: BehaviorSubject<LoaderType> = new BehaviorSubject<LoaderType>(LoaderType.Spinner);
```
anandtiwary, line 43: We are setting different default options at multiple places. It would be useful if we set it at only one place

projects/components/src/load-async/load-async.directive.ts
alok-traceable: Screen Shot 2021-09-06 at 8 49 47 PM
its old code

projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
alok-traceable: then, I need to put undefined as the initial value.

projects/components/src/load-async/loader/loader.component.ts
alok-traceable: updated

projects/components/src/load-async/loader/loader.component.ts
alok-traceable: done

projects/components/src/load-async/load-async.directive.ts
alok-traceable: updated

projects/components/src/load-async/load-async.directive.ts
alok-traceable: updated

projects/components/src/load-async/load-async.directive.ts
```diff
@@ -23,8 +24,11 @@ import {
 export class LoadAsyncDirective implements OnChanges, OnDestroy {
   @Input('htLoadAsync')
   public data$?: Observable<unknown>;
+  @Input()
+  public htLoadAsyncLoaderType: LoaderType = LoaderType.Spinner;
+
   private readonly wrapperParamsSubject: ReplaySubject<LoadAsyncWrapperParameters> = new ReplaySubject(1);
-  private readonly wrapperInjector: Injector;
+  private readonly wrapperInjector!: Injector;
```
anandtiwary, line 31: Is this required?

projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
```diff
@@ -1,20 +1,21 @@
...
 );
-
 @Component({
   selector: 'ht-load-async-wrapper',
   template: `
     <div *ngIf="this.state$ | async as state" class="fill-container" [ngSwitch]="state.type">
       <ng-container *ngSwitchCase="'${LoadAsyncStateType.Loading}'">
-        <ht-loader></ht-loader>
+        <ng-container *ngIf="this.loaderType$ | async as loaderType">
+          <ht-loader [type]="loaderType"></ht-loader>
```
anandtiwary, line 17: ng-container is not needed. You can use `async` pipe directly with ht-loader

projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
```diff
@@ -39,9 +40,15 @@ export class LoadAsyncWrapperComponent {
...
 
+  private readonly loaderTypeSubject: BehaviorSubject<LoaderType | undefined> = new BehaviorSubject<
+    LoaderType | undefined
+  >(undefined);
+  public readonly loaderType$: Observable<LoaderType | undefined> = this.loaderTypeSubject.asObservable();
+
   public constructor(@Inject(ASYNC_WRAPPER_PARAMETERS$) parameters$: Observable<LoadAsyncWrapperParameters>) {
     this.state$ = parameters$.pipe(
       tap(params => (this.content = params.content)),
+      tap(params => this.loaderTypeSubject.next(params.loaderType)),
```
anandtiwary, line 51: do we gain anything by storing it as observable? If we store the `loaderType` similar to `content`, would it cause any issues?

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,15 +1,37 @@
...
-      <ht-icon icon="${IconType.Loading}" size="${IconSize.Large}"></ht-icon>
+      <img [ngClass]="[this.type]" [src]="this.imagePath" />
     </div>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class LoaderComponent {}
+export class LoaderComponent implements OnChanges {
+  @Input()
+  public type!: LoaderType;
```
anandtiwary, line 17: This property can be optional right?

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,15 +1,37 @@
...
+  private getImagePathFromType(): ImagesAssetPath {
+    switch (this.type) {
+      case LoaderType.ExpandableRow:
+        return ImagesAssetPath.LoaderExpandableRow;
+      case LoaderType.Page:
+        return ImagesAssetPath.LoaderPage;
+      case LoaderType.Spinner:
+        return ImagesAssetPath.LoaderSpinner;
+      default:
+        return assertUnreachable(this.type);
```
anandtiwary, line 34: may be we can leave `type` as optional and in default section here, we can assign the default image path instead of throwing exception

alok-traceable: Actually, No. When a consumer is passing a wrong value it should throw an error, on the other hand, if the consumer is not passing a value it should show default value.

projects/components/src/load-async/loader/loader.component.test.ts
```diff
@@ -0,0 +1,49 @@
+import { CommonModule } from '@angular/common';
+import { ImagesAssetPath, LoaderType } from '@hypertrace/assets-library';
+import { createHostFactory, SpectatorHost } from '@ngneat/spectator/jest';
+import { LoaderComponent } from './loader.component';
+
+describe('Link component', () => {
```
anandtiwary, line 6: name

projects/components/src/load-async/load-async.directive.ts
```diff
@@ -23,8 +24,11 @@ import {
 export class LoadAsyncDirective implements OnChanges, OnDestroy {
   @Input('htLoadAsync')
   public data$?: Observable<unknown>;
+  @Input()
```
anandtiwary, line 27:
```
@Input('htLoadAsyncLoaderType')
public loaderType: LoaderType;
```

This should work

projects/components/src/load-async/load-async.directive.ts
```diff
@@ -49,6 +53,7 @@ export class LoadAsyncDirective implements OnChanges, OnDestroy {
       this.wrapperView = this.wrapperView || this.buildWrapperView();
       this.wrapperParamsSubject.next({
         state$: this.loadAsyncService.mapObservableState(this.data$),
+        loaderType: this.htLoadAsyncLoaderType,
```
anandtiwary, line 56: So basically, if htLoadAsyncLoaderType changes, then this if block will not be triggered? So, how would you handle the case when loaderType changes after init?


Why don't you send `loaderType` to `mapObservableState` as another param? then this function can add this loaderType in ```interface LoadingAsyncState {
  type: LoadAsyncStateType.Loading;
}```
Also, update L52 to also include change on LoaderType

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,15 +1,37 @@
...
     </div>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class LoaderComponent {}
+export class LoaderComponent implements OnChanges {
+  @Input()
+  public type: LoaderType = LoaderType.Spinner;
+
+  public imagePath: ImagesAssetPath = ImagesAssetPath.LoaderSpinner;
```
anandtiwary, line 19: If we keep it unassigned, would it get the same value from `getImagePathFromType` call inside ngOnChanges?

projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
```diff
@@ -39,9 +40,15 @@ export class LoadAsyncWrapperComponent {
...
       switchMap(parameter => parameter.state$),
```
anandtiwary, line 52: As per prev comments, we can add loadertype as part of LoadingState. We wouldn't need to store another subject here. We can reference it from the state object directly.

projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
anandtiwary: loaderType here will be defined since we are setting default value.

anandtiwary:
@alok-traceable I am guessing you are still working on this, right?

PR #1073, opened 2021-08-18
chore: upgrade to angular 12

itssharmasandeep:


.npmrc
```diff
@@ -1 +1,2 @@
 @hypertrace:registry=https://hypertrace.jfrog.io/artifactory/api/npm/npm/
+legacy-peer-deps=true
```
itssharmasandeep, line 2: This change is for npm 7.
Please to get more context https://github.com/angular/angular-cli/issues/19957

aaron-steinfeld: We shouldn't need this - the legacy peer dep behavior is warning instead of errors, but we should be resolving any peer dependency issues - we want neither warning nor errors.

anandtiwary: Also, we can stay on npm 6 for now.

```
 ~/trc/traceable-ui/hypertrace-ui   ng-12  ✔  npm install
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
npm WARN deprecated opn@6.0.0: The package has been renamed to `open`
```

package.json
```diff
@@ -9,12 +9,12 @@
   "scripts": {
     "postinstall": "ngcc --tsconfig './tsconfig.app.json' --use-program-dependencies && ngcc --source 'node_modules/@angular'",
     "ng": "ng",
-    "start": "ng serve",
-    "build": "ng build --prod",
+    "start": "NG_PERSISTENT_BUILD_CACHE=1 ng serve",
+    "build": "NG_PERSISTENT_BUILD_CACHE=1 ng build --configuration production",
```
anandtiwary, line 13: for local builds, we are not enabling prod config yet, right?

angular.json
```diff
@@ -46,7 +45,13 @@
               "includePaths": ["./projects/assets-library/assets/styles", "./node_modules"]
             },
             "scripts": [],
-            "allowedCommonJsDependencies": ["graphql-tag", "zen-observable"]
+            "allowedCommonJsDependencies": ["graphql-tag", "zen-observable"],
```
anandtiwary, line 48:
```
Option "extractCss" is deprecated: Deprecated since version 11.0. No longer required to disable CSS extraction for HMR.
```
I see this warning.

.npmrc
aaron-steinfeld: unless there's a reason to stay on 6, i'd vote to move to 7 as our default - it's been out for a while and is the default one for current node (16, which is an LTS version).

anandtiwary: We need to collectively move to npm 7 so that the lockfile version problem doesn't come up.

Also, last time the problem came from dependabot. It is running on npm 6 and ends up creating a lock file of version 1. I had to downgrade to 6 due to this.

anandtiwary:
@itssharmasandeep before merging, please verify ht-ui end to end with real data. Look for any weirdness or any console error.

.npmrc
aaron-steinfeld: We don't really need to all upgrade or be in sync - the lockfile can bounce back and forth as we see here. Personally, I've been on npm since maybe march? not sure exactly when, but generally keep my versions up to date.

aaron-steinfeld: dependabot looks like it supports it at least: https://github.blog/changelog/2021-02-08-dependabot-npm-v7-support/

aaron-steinfeld: And it looks like we've been using it (this repo is using a v2 lockfile): https://raw.githubusercontent.com/hypertrace/hypertrace-ui/main/package-lock.json

anandtiwary: Oh cool. that works. I think earlier we would end up creating new lock files with newer versions so it was annoying and decided to move back to 6.

itssharmasandeep: Removed!!

package.json
itssharmasandeep: Removed

angular.json
itssharmasandeep: where?

anandtiwary: console. when you do ng build or serve

package.json
```diff
@@ -10,7 +10,7 @@
     "postinstall": "ngcc --tsconfig './tsconfig.app.json' --use-program-dependencies && ngcc --source 'node_modules/@angular'",
     "ng": "ng",
     "start": "NG_PERSISTENT_BUILD_CACHE=1 ng serve",
-    "build": "NG_PERSISTENT_BUILD_CACHE=1 ng build --configuration production",
+    "build": "ng build --configuration production",
```
anandtiwary, line 13: why we remove this? I meant to remove the config prod thing

aaron-steinfeld: this has always been prod - how often do we build (rather than server) a dev build?

anandtiwary:
not seeing any errors on npm 6 locally.

aaron-steinfeld:
> not seeing any errors on npm 6 locally.

what version? I was able to locally repro those build errors with 
```
npx npm@6.14.14 ci && npm run lint
```

anandtiwary:
npm: 6.14.11

node: v15.3.0

anandtiwary:
> > not seeing any errors on npm 6 locally.
> 
> what version? I was able to locally repro those build errors with
> 
> ```
> npx npm@6.14.14 ci && npm run lint
> ```

even with this, i am not seeing any error at install time. What node version are you in?

The only error i see is during lint

```
Odd numbered Node.js versions will not enter LTS status and should not be used for production. For more information, please see https://nodejs.org/en/about/releases/.
TSLint's support is discontinued and we're deprecating its support in Angular CLI.
To opt-in using the community driven ESLint builder, see: https://github.com/angular-eslint/angular-eslint#migrating-an-angular-cli-project-from-codelyzer-and-tslint.
Linting "hypertrace-ui"...
The 'deprecation' rule threw an error in '/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/projects/common/src/utilities/rxjs/rxjs-utils.ts':
TypeError: message.trim is not a function
    at Function.Rule.FAILURE_STRING (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:31:81)
    at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:58:53)
    at visitNode (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:28995:24)
    at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29134:24)
    at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:70:23)
    at visitNodes (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29004:30)
    at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29128:21)
    at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:70:23)
    at visitNode (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:28995:24)
    at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29060:21)
The 'deprecation' rule threw an error in '/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/projects/common/src/navigation/navigation.service.ts':
TypeError: message.trim is not a function
```

aaron-steinfeld:
> > > not seeing any errors on npm 6 locally.
> > 
> > 
> > what version? I was able to locally repro those build errors with
> > ```
> > npx npm@6.14.14 ci && npm run lint
> > ```
> 
> even with this, i am not seeing any error at install time. What node version are you in?
> 
> The only error i see is during lint
> 
> ```
> Odd numbered Node.js versions will not enter LTS status and should not be used for production. For more information, please see https://nodejs.org/en/about/releases/.
> TSLint's support is discontinued and we're deprecating its support in Angular CLI.
> To opt-in using the community driven ESLint builder, see: https://github.com/angular-eslint/angular-eslint#migrating-an-angular-cli-project-from-codelyzer-and-tslint.
> Linting "hypertrace-ui"...
> The 'deprecation' rule threw an error in '/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/projects/common/src/utilities/rxjs/rxjs-utils.ts':
> TypeError: message.trim is not a function
>     at Function.Rule.FAILURE_STRING (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:31:81)
>     at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:58:53)
>     at visitNode (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:28995:24)
>     at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29134:24)
>     at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:70:23)
>     at visitNodes (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29004:30)
>     at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29128:21)
>     at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:70:23)
>     at visitNode (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:28995:24)
>     at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29060:21)
> The 'deprecation' rule threw an error in '/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/projects/common/src/navigation/navigation.service.ts':
> TypeError: message.trim is not a function
> ```

I guess it could be the node resolution, but that'd surprise me. I'm able to consistently repro this when running against this branch - and I use node 16.6.1. 

The error is never at install, it's always at lint - it's basically an artifact, I believe, of how the two versions of npm respect peer dependencies.

aaron-steinfeld:
Also, it's continuing to fail because we're not actually doing an install here - my oversight, we should update the cache key's prefix (across all jobs), so we don't pull the old result (with the same lock) from cache and instead regenerate.

itssharmasandeep:
I updated the cached key a bit and checks worked fine, I just changed `node` to `node-modules` in the key

itssharmasandeep:
> Also, it's continuing to fail because we're not actually doing an install here (at least, that's the first issue) - my oversight, we should update the cache key's prefix (across all jobs), so we don't pull the old result (with the same lock) from cache and instead regenerate.

I think making every cache key different in different jobs would not make that big of a difference, I think changing it to start of the workflow inside npm-install-if-needed would be enough so that it did a new npm install, although I am not sure why did not do that since we changed the `package-lock`

itssharmasandeep:
> > > not seeing any errors on npm 6 locally.
> > 
> > 
> > what version? I was able to locally repro those build errors with
> > ```
> > npx npm@6.14.14 ci && npm run lint
> > ```
> 
> even with this, i am not seeing any error at install time. What node version are you in?
> 
> The only error i see is during lint
> 
> ```
> Odd numbered Node.js versions will not enter LTS status and should not be used for production. For more information, please see https://nodejs.org/en/about/releases/.
> TSLint's support is discontinued and we're deprecating its support in Angular CLI.
> To opt-in using the community driven ESLint builder, see: https://github.com/angular-eslint/angular-eslint#migrating-an-angular-cli-project-from-codelyzer-and-tslint.
> Linting "hypertrace-ui"...
> The 'deprecation' rule threw an error in '/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/projects/common/src/utilities/rxjs/rxjs-utils.ts':
> TypeError: message.trim is not a function
>     at Function.Rule.FAILURE_STRING (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:31:81)
>     at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:58:53)
>     at visitNode (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:28995:24)
>     at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29134:24)
>     at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:70:23)
>     at visitNodes (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29004:30)
>     at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29128:21)
>     at cb (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/tslint/lib/rules/deprecationRule.js:70:23)
>     at visitNode (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:28995:24)
>     at Object.forEachChild (/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/node_modules/typescript/lib/typescript.js:29060:21)
> The 'deprecation' rule threw an error in '/Users/anand.tiwary/trc/traceable-ui/hypertrace-ui/projects/common/src/navigation/navigation.service.ts':
> TypeError: message.trim is not a function
> ```

This error might be related to deprecation of `TSLint` but not sure

aaron-steinfeld:
> I think making every cache key different in different jobs would not make that big of a difference, I think changing it to start of the workflow inside npm-install-if-needed would be enough so that it did a new npm ci, although I am not sure why did not do that since we changed the package-lock

The lock changed, and it did do a new install - see the last commit where it actually changed ( https://github.com/hypertrace/hypertrace-ui/runs/3372942782 )
After that, we updated the npm version which is not accounted for in the cache key, so it never reran. The cache keys need to match across all jobs - couldn't just update npm-install-if-needed, else the other jobs would be reading a different version of node_modules than the workflow has set up.

> This error might be related to deprecation of TSLint but not sure

More or less - the issue is that the format the parser returns for a deprecation has changed I believe, but the tslint rules are no longer being maintained so the newer typescript is breaking them. I'd say let's either ignore it since we're about to migrate from tslint or remove the deprecation rule (or both!).

.github/workflows/build-and-test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: node_modules
-          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('package-lock.json') }}
```
aaron-steinfeld, line 24: this needs to be updated in the publish job, too.

itssharmasandeep: I saw that and I think we're doing something different in publish job

1. `restoreKeys` contain `package.json` instead of `package-lock.json` in the key value and I think it would always be different right?
2. we're not checking `cache-hit` and every time we're doing `npm ci` so cache part seems unused here.

itssharmasandeep:
```
name: Publish artifacts
on:
  # Will only run when release is published.
  release:
    types:
      - created
  workflow_dispatch:

jobs:
  publish-docker:
    runs-on: ubuntu-20.04
    steps:
      # Set fetch-depth: 0 to fetch commit history and tags for use in version calculation
      - name: Check out code
        uses: actions/checkout@v2.3.4
        with:
          fetch-depth: 0

      - name: Cache node modules
        uses: actions/cache@v2
        with:
          path: ~/.npm
          key: ${{ runner.os }}-node-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}
          restore-keys: |
            ${{ runner.os }}-node-${{ hashFiles('package.json') }}-
            ${{ runner.os }}-node-

      - name: NPM Install
        run: npm ci

      - name: Build
        run: npm run build:ci
```

aaron-steinfeld: whoops - yeah, that one is different but shouldn't be. It will still work, just off fresh node modules each time - so we can come back to that as tech debt.

itssharmasandeep: should we go ahead and merge this PR?

itssharmasandeep:
I have tested this with `npm ci` , doesn't give any error and build and tested the `ht-ui` with some data, I think it is working fine, no console errors.

angular.json
```diff
@@ -64,7 +64,6 @@
               "optimization": true,
               "outputHashing": "all",
               "sourceMap": false,
-              "extractCss": true,
```
anandtiwary, line 62: Yep. This would fix that warning

projects/assets-library/assets/styles/_font.scss
```diff
@@ -1,3 +1,4 @@
+@use '~@angular/material' as mat;
```
anandtiwary, line 1: We need this change in all scss. @import -> @use

anandtiwary: But this can happen separately.

aaron-steinfeld (approved):

projects/assets-library/assets/styles/_font.scss
```diff
@@ -329,3 +330,7 @@ $font-family: 'Work Sans', sans-serif;
 @mixin axis-title($color: $gray-9) {
   @include chart-small-regular;
 }
+
+$mat-typography: mat.define-typography-config(
```
anandtiwary, line 334: Did this come on its own with migration

anandtiwary (approved):

projects/assets-library/assets/styles/_font.scss
aaron-steinfeld: I suggested this from reading the docs - looks like they have some form of migration for it, but not sure exactly what conditions trigger it or if we just skipped it: https://github.com/angular/components/pull/22698

Issue #1072, opened 2021-08-17
UX improvement for services and endpoint search


akashsharma95:
## Use Case
Improve UX for services and endpoint search.

## Proposal
Once the service panel is loaded and all the services are listed by querying graphql endpoint. Cache this services list on browser and do text search on that.

![image](https://user-images.githubusercontent.com/1501547/129686256-3a908f59-423f-4f45-a5f8-7f00141aa62f.png)

jyothishjose6190:
@aaron-steinfeld @anandtiwary  @itssharmasandeep  Any particular reason for current implementation?

aaron-steinfeld:
To support server side paging, where we have no guarantee that we have the full result set in the browser. In this particular case, it may be a small enough result set that only a single page is populated, but in other places there could be thousands or even millions of results that we don't want to or can't feasible fetch at once, but still support search across the full result set. In other words, the server version is the more general case. Agree that keeping this on the client side would improve the user experience in scenarios with small data sets, and we can look to add support for that, either via config, or even better, making it dynamic based on the result set size (since we have enough info to know if the full set is loaded).

I think we did something similar for multiselect (supporting either client side or server side searching).

anandtiwary:
I like the idea of choosing client side search when we know we have loaded the full set.

jyothishjose6190:
Currently, we are triggering graphql api for all keystrokes, can we limit the api calls, by triggering only when the user stops typing for a time period (eg 2 second)

aaron-steinfeld:
> Currently, we are triggering graphql api for all keystrokes, can we limit the api calls, by triggering only when the user stops typing for a time period (eg 2 second)

That debounce behavior should already be in there, will check if that's broken.

PR #1074, opened 2021-08-18
Increase default debounce

aaron-steinfeld:
## Description
Update table search debounce from 200ms to 400ms to reduce api calls.

### Testing
Manual, updated UT

### Checklist:

https://github.com/hypertrace/hypertrace-ui/issues/1072

### DIFF
--- a/projects/components/src/table/controls/table-controls.component.ts
+++ b/projects/components/src/table/controls/table-controls.component.ts
@@ -9,9 +9,7 @@ import {
   Output
 } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
-import { SubscriptionLifecycle, TypedSimpleChanges } from '@hypertrace/common';
-import { Subject } from 'rxjs';
-import { debounceTime } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


jake-bassett (approved):

projects/components/src/table/controls/table-controls.component.ts
```diff
@@ -161,7 +161,7 @@ export class TableControlsComponent implements OnChanges {
     this.checkboxDiffer = this.differFactory.find([]).create();
 
     this.subscriptionLifecycle.add(
-      this.searchDebounceSubject.pipe(debounceTime(200)).subscribe(text => this.searchChange.emit(text))
```
anandtiwary, line 164: We can really use the ootb debounceTime property of search box component. Additional subscription here is not required in this table.

Please update the time in search box component as well if this looks good,

aaron-steinfeld: ah yeah I forgot we added that - switched over to using that one (searchbox doesn't have a built in time, it's not debouncing by default)

jake-bassett (approved):

PR #1077, opened 2021-08-19
chore: set up node 16 in workflows

itssharmasandeep:


.github/workflows/build-and-test.yml
```diff
@@ -11,6 +11,11 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
 
+      - name: Set up Node 16
```
anandtiwary, line 14: Looks fine to me. Or should we use a different build image @aaron-steinfeld?

aaron-steinfeld: ubuntu 20.04 is pretty much their newest, I believe https://github.com/actions/virtual-environments#available-environments - and it uses `Npm 6.14.14`, `Node 14.17.5`

aaron-steinfeld: (I'm assuming updating node in this way updates npm too)

anandtiwary: Yep. it should install npm

anandtiwary (approved):

.github/workflows/build-and-test.yml
anandtiwary: going to merge this.

PR #1078, opened 2021-08-19
fix: support anything coercer supports for date parsing

aaron-steinfeld:
## Description
Dates coming through as strings, even though we have the tools to parse them, are not supported today in the table parser or the date display pipe. This adds that support.

### Testing
Manual, added UTs

### DIFF
--- a/projects/common/src/utilities/formatters/date/display-date.pipe.ts
+++ b/projects/common/src/utilities/formatters/date/display-date.pipe.ts
@@ -1,11 +1,14 @@
 import { Pipe, PipeTransform } from '@angular/core';
+import { isNil } from 'lodash-es';
+import { DateCoercer } from '../../coercers/date-coercer';
 import { DateFormatOptions, DateFormatter } from './date-formatter';
 
 @Pipe({
   name: 'htDisplayDate'
 })
 export class DisplayDatePipe implements PipeTransform {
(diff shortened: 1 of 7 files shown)


projects/common/src/utilities/formatters/date/display-date.pipe.ts
```diff
@@ -1,11 +1,14 @@
...
 
 @Pipe({
   name: 'htDisplayDate'
 })
 export class DisplayDatePipe implements PipeTransform {
-  public transform(value?: Date | number | null, options: DateFormatOptions = {}): string {
-    return value !== null ? new DateFormatter(options).format(value) : '-';
+  private readonly dateCoercer: DateCoercer = new DateCoercer();
+  public transform(value?: string | Date | number | null, options: DateFormatOptions = {}): string {
+    return isNil(value) ? '-' : new DateFormatter(options).format(this.dateCoercer.coerce(value));
```
anandtiwary, line 12: what if `this.dateCoercer.coerce(value)` returns undefined? Should we do this before the calling isNil  so that we can assign '-'?

anandtiwary: Should the DateFormatter just depend on dateCoercer?

aaron-steinfeld: addressed

anandtiwary (approved):

Issue #1075, opened 2021-08-19
Tag search from waterfall graph


jyothishjose6190:
On the trace waterfall graph, i should have search links for each attribute/tag to take me to the explore screen

aaron-steinfeld:
+1 - we've talked about doing this for a while but haven't gotten around to it. we should put it on the roadmap.

JBAhire:
@jyothishjose6190 , @aaron-steinfeld , already prioritized this internally. @itssharmasandeep , can you please work with @jyothishjose6190 on this?

PR #1079, opened 2021-08-23
feat: tag search or filter in explorer

itssharmasandeep:
https://user-images.githubusercontent.com/78212931/131332742-f8e901d9-5fbe-4f92-a3d4-0578bb72d19f.mov


itssharmasandeep:
@aaron-steinfeld @anandtiwary this is related to https://github.com/hypertrace/hypertrace-ui/issues/1075
Please have a look.

projects/components/src/list-view/list-view.component.ts
```diff
@@ -31,6 +46,25 @@ export class ListViewComponent {
...
+  @Input()
+  public actionType?: ListViewActionType;
+
+  private readonly scopeQueryParam: string = 'endpoint-traces';
+  private readonly filterBuilder: StringMapFilterBuilder = new StringMapFilterBuilder();
+
+  public constructor(private readonly navigationService: NavigationService) {}
+
+  public searchAttributes(key: string, value: string): void {
+    this.navigationService.navigate({
```
aaron-steinfeld, line 59: this should use the logic from explorer service as mentioned on the other PR.

itssharmasandeep:
> Why I did not use ExploreFilterLinkComponent or ExploreService?

So the reason behind that is, `list-view component` is part of `@hypertrace/components` and `Explore` related things are of `@hypertrace/observability`, since `observability` is already dependent on `components`, do we want `components` to be dependent on `observability`? I'm sceptical about this.

Suppose if go with the approach of doing that then my suggestion would be to extend that component to take different `icons` and take `data` as well, instead of taking generated Url only.

aaron-steinfeld:
> > Why I did not use ExploreFilterLinkComponent or ExploreService?
> 
> So the reason behind that is, `list-view component` is part of `@hypertrace/components` and `Explore` related things are of `@hypertrace/observability`, since `observability` is already dependent on `components`, do we want `components` to be dependent on `observability`? I'm sceptical about this.
> 
> Suppose if go with the approach of doing that then my suggestion would be to extend that component to take different `icons` and take `data` as well, instead of taking generated Url only.
> 
> cc: @aaron-steinfeld

We certainly don't want the circular dependency, but my suggestion of using contents or templates passed in account for that. That's generally how you get component libraries to be flexible - rather than encoding behavior in the common components, you allow the flexibility for the parent to do it. For example, the table works this way - defined commonly with many specific renderers the table itself doesn't know about. So using it could look like (pseudo):
```html
<list>
   <ng-container *listKeyRenderer="let key">
     {{key}}
   </ng-container>

   <ng-container *listValueRenderer="let value">
      <my-fancy-thing-here></my-fancy-thing-here>
    </ng-container>
 <list>
```

itssharmasandeep:
> > > Why I did not use ExploreFilterLinkComponent or ExploreService?
> > 
> > 
> > So the reason behind that is, `list-view component` is part of `@hypertrace/components` and `Explore` related things are of `@hypertrace/observability`, since `observability` is already dependent on `components`, do we want `components` to be dependent on `observability`? I'm sceptical about this.
> > Suppose if go with the approach of doing that then my suggestion would be to extend that component to take different `icons` and take `data` as well, instead of taking generated Url only.
> > cc: @aaron-steinfeld
> 
> We certainly don't want the circular dependency, but my suggestion of using contents or templates passed in account for that. That's generally how you get component libraries to be flexible - rather than encoding behavior in the common components, you allow the flexibility for the parent to do it. For example, the table works this way - defined commonly with many specific renderers the table itself doesn't know about. So using it could look like (pseudo):
> 
> ```
> <list>
>    <ng-container *listKeyRenderer="let key">
>      {{key}}
>    </ng-container>
> 
>    <ng-container *listValueRenderer="let value">
>       <my-fancy-thing-here></my-fancy-thing-here>
>     </ng-container>
>  <list>
> ```

Okay, I got it. I am implementing the content to pass from parent, now the same thing is happening. This `explore-filter-link.component` is in the observability but our parent sits in `distributed-tracing` and importing the module for `ExploreFilterLinkComponent` (from `@hypertrace/observability`) is failing in `distributed-tracing`

`Uncaught TypeError: Cannot read property 'ExploreFilterLinkModule' of undefined
    at Module.ExploreFilterLinkModule (explorer.module.ts:19)
    at Module.24640 (span-tags-detail.module.ts:9)
    at __webpack_require__ (bootstrap:19)
    at Module.33735 (span-detail.component.ts:57)
    at __webpack_require__ (bootstrap:19)
    at Module.84056 (span-detail-widget.model.ts:10)
    at __webpack_require__ (bootstrap:19)
    at Module.22478 (trace-detail-widget.module.ts:12)
    at __webpack_require__ (bootstrap:19)
    at Module.40099 (tracing-dashboard-properties.module.ts:7)`

itssharmasandeep:
@aaron-steinfeld I have changed the approach and used the existing features as well as the approach of passing template.
Please review this again.

projects/components/src/filtering/filter/builder/types/abstract-filter-builder.ts
```diff
@@ -36,6 +36,10 @@ export abstract class AbstractFilterBuilder<TValue> {
   }
 
   protected buildUrlFilterString(attribute: FilterAttribute, operator: FilterOperator, value: TValue): string {
-    return encodeURIComponent(`${attribute.name}${toUrlFilterOperator(operator)}${this.buildValueString(value)}`);
+    return this.buildEncodedUrlFilterString(attribute.name, operator, value);
+  }
+
+  public buildEncodedUrlFilterString(attributeName: string, operator: FilterOperator, value: TValue): string {
```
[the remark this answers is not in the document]
itssharmasandeep: Fixed

projects/components/src/list-view/list-view.component.ts
```diff
@@ -21,6 +26,16 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
         <div class="value">
           <span>{{ record.value }}</span>
         </div>
+        <div *ngIf="this.actionType" class="action-item">
+          <ht-icon
+            *ngIf="this.actionType === '${ListViewActionType.AttributeSearch}'"
+            class="attribute-search icon"
+            icon="${IconType.ChevronRight}"
+            size="${IconSize.Small}"
+            htTooltip="Search in Explorer"
```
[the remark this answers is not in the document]
itssharmasandeep: Used the same

projects/components/src/list-view/list-view.component.ts
itssharmasandeep: Used the same

projects/components/src/list-view/list-view.component.ts
```diff
@@ -31,6 +33,9 @@ export class ListViewComponent {
 
   @Input()
   public records?: ListViewRecord[];
+
+  @Input()
+  public action?: TemplateRef<unknown>;
```
aaron-steinfeld, line 38: We're getting close. This should work, but my suspicion is we'd get a lot more flexibility out of making this renderer the whole value renderer rather than always printing out the value as plaintext the following it with an optional additional renderer. A full value renderer can handle any use case this could, but also could do things like making the actual value itself the link - and IMO have a much more intuitive api.

The difference is code like:
```html
        <ht-list-view [records]="tagRecords" [action]="tagAction" data-sensitive-pii></ht-list-view>
        <ng-template #tagAction let-tag>
          <ht-explore-filter-link
            *ngIf="this.getExploreNavigationParams(tag) | async as exploreNavigationParams"
            [paramsOrUrl]="exploreNavigationParams"
            htTooltip="See traces in Explorer"
          >
          </ht-explore-filter-link
        ></ng-template>
```
vs code like (this would render a bit differently, depends exactly where we want value - but it shows the flexibility).
```html
        <ht-list-view [records]="tagRecords" data-sensitive-pii>
           <div *htListViewValueRenderer="let value">
              <ht-explore-filter-link 
                *ngIf="this.getExploreNavigationParams(tag) | async as exploreNavigationParams"
                [paramsOrUrl]="exploreNavigationParams"
                htTooltip="See traces in Explorer">
                {{ value }}
             </ht-explore-filter-link>
           </div>
       </ht-list-view>

Comment (itssharmasandeep) [reply projects/components/src/list-view/list-view.component.ts:38]: So, this `action` feature, is on a row level, not on the `value` level. So this design of having `listviewvaluerenderer`, I am not so sure about this. `tagRecords` already contains the value. This would have made more impact if we were going to have, different type of renderers for the both the columns (for key and value) like we have for the table. As far as I understand, list view has a sole purpose of showing the text value, right?

Comment (aaron-steinfeld) [reply projects/components/src/list-view/list-view.component.ts:38]: So the whole purpose of this change is to change the renderer for the value column to go from just plain text to show a link after the plain text, right? That's basically a custom renderer for the value column, but we've special cased it to assume any custom renderer would want to be in addition to the plain text renderer, which may solves this use case but becomes redundant if we ever have to solve the general case. Imagine I just wanted to pretty print the values rather than using them as is (that is, format numbers, dates, etc.):

I could do something like
```html
        <ht-list-view [records]="tagRecords" data-sensitive-pii>
           <div *htListViewValueRenderer="let value">
               {{ value | prettyPrint }}
           </div>
       </ht-list-view>
```

> So, this action feature, is on a row level, not on the value level. So this design of having listviewvaluerenderer, I am not so sure about this. tagRecords already contains the value.

Either one kind of works the same. In either case, the list view takes in a template, instantiates it multiple times, and each time gives it some data. We can choose to give both the key and value, or just the value. I started with just the value as the implicit binding in my example, since that matched what you did with the action renderer and is extendable later with no changes to the api ( (e.g. `let-key="key" let-value="value" let-value-again`) 

> This would have made more impact if we were going to have, different type of renderers for the both the columns (for key and value) like we have for the table. As far as I understand, list view has a sole purpose of showing the text value, right?

I left the key renderer out of this because we didn't need it, but we could eventually add it. So, given the new requirement being added here to customize the values being rendered, the main differences with table are really around 1. interaction - paging, sorting, that kind of stuff and 2. having exactly two, known columns. So with table, we configure the columns and part of that is specifying a renderer for each. Here, we can shortcut that since we only need to (optionally) pass in a renderer.

This is how a lot of the library components out there work - with some variation depending on use case. For material table, a cell is defined in a similar way - the main difference is that there's a generic cellDef directive instead of a specific one which is associated with a column not by the directive being used, but by its location in a column def. Since we don't need headers or column defs, ours simplifies to just two named cell defs (only one of which we need right now).
```html
  <ng-container matColumnDef="position">
    <th mat-header-cell *matHeaderCellDef> No. </th>
    <td mat-cell *matCellDef="let element"> {{element.position}} </td>
  </ng-container>
```

projects/observability/src/shared/components/span-detail/span-detail.component.test.ts
```diff
@@ -14,7 +15,7 @@ describe('Span detail component', () => {
     component: SpanDetailComponent,
     imports: [SpanDetailModule, HttpClientTestingModule, IconLibraryTestingModule],
     declareComponent: false,
-    providers: [mockProvider(NavigationService)]
+    providers: [mockProvider(NavigationService), mockProvider(ExplorerService)]
```
aaron-steinfeld, line 18: This test is failing due to its transitive dependencies on router, due to the new usage of a link component. It's tech debt that this test isn't shallow, so let's fix that :)

```diff
diff --git a/projects/observability/src/shared/components/span-detail/span-detail.component.test.ts b/projects/observability/src/shared/components/span-detail/span-detail.component.test.ts
index a169698..22b8fb7 100644
--- a/projects/observability/src/shared/components/span-detail/span-detail.component.test.ts
+++ b/projects/observability/src/shared/components/span-detail/span-detail.component.test.ts
@@ -1,21 +1,14 @@
-import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { fakeAsync } from '@angular/core/testing';
-import { IconLibraryTestingModule } from '@hypertrace/assets-library';
-import { NavigationService } from '@hypertrace/common';
-import { createHostFactory, mockProvider, Spectator } from '@ngneat/spectator/jest';
-import { ExplorerService } from '../../../pages/explorer/explorer-service';
+import { createHostFactory, Spectator } from '@ngneat/spectator/jest';
 import { SpanData } from './span-data';
 import { SpanDetailComponent } from './span-detail.component';
-import { SpanDetailModule } from './span-detail.module';
 
 describe('Span detail component', () => {
   let spectator: Spectator<SpanDetailComponent>;
 
   const createHost = createHostFactory({
     component: SpanDetailComponent,
-    imports: [SpanDetailModule, HttpClientTestingModule, IconLibraryTestingModule],
-    declareComponent: false,
-    providers: [mockProvider(NavigationService), mockProvider(ExplorerService)]
+    shallow: true
   });
 
   test('should display child components', () => {
```

itssharmasandeep: Updated

projects/components/src/list-view/list-view.component.ts
```diff
@@ -19,8 +19,16 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
...
         <div class="value">
-          <span>{{ record.value }}</span>
+          <ng-container
+            *ngTemplateOutlet="
+              this.valueRenderer ? this.valueRenderer!.getTemplateRef() : defaultValueRenderer;
+              context: { $implicit: record }
+            "
+          ></ng-container>
         </div>
+        <ng-template #defaultValueRenderer let-record
```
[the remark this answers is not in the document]
itssharmasandeep: Updated

aaron-steinfeld (approved):

PR #1085, opened 2021-08-23
feat: Updating Jest to v.27

palbizu:
## Description
Upgrading jest version in progress.

### Testing
Manual.

### Checklist:

### Documentation


jest.config.js
```diff
@@ -1,3 +1,5 @@
+require('jest-preset-angular/ngcc-jest-processor');
```
anandtiwary, line 1: why is this only needed in this file but not in debug.config?

jest.config.js
```diff
@@ -6,9 +8,10 @@ module.exports = {
     'ts-jest': {
       tsconfig: '<rootDir>/tsconfig.spec.json',
       stringifyContentPathRegex: '\\.html$',
-      astTransformers: {
-        before: ['jest-preset-angular/build/InlineFilesTransformer', 'jest-preset-angular/build/StripStylesTransformer']
-      }
+      astTransformers: [
+        'jest-preset-angular/build/InlineFilesTransformer',
```
anandtiwary, line 12: This one is opposite of what you have done for debug config. Why so?

jest.config.js
palbizu: Imported in debug.config.

jest.config.js
palbizu: I was doing some tests because it wasn't working but I already fixed this part.

anandtiwary: okay

anandtiwary:
Let me know when it is ready for review again?

jest.config.js
```diff
@@ -1,14 +1,18 @@
+require('jest-preset-angular/ngcc-jest-processor');
+
 process.env.TZ = 'UTC'; // Tests should always run in UTC, no time zone dependencies
 
 module.exports = {
   rootDir: '.',
+  testRunner: 'jest-jasmine2',
```
anandtiwary, line 7: What does this do?

palbizu: Jest v.27 doc: 'we are changing the default test environment from "jsdom" to "node". If you are affected by this change because you use DOM APIs and do not have the test environment explicitly configured, you should be receiving an error when e.g. document is accessed, and you can configure "testEnvironment": "jsdom" or use per-file environment configuration to resolve this.' The tests were not running properly and after add this config started to work.

jest.config.js
```diff
@@ -1,14 +1,18 @@
...
   globals: {
     'ts-jest': {
       tsconfig: '<rootDir>/tsconfig.spec.json',
       stringifyContentPathRegex: '\\.html$',
-      astTransformers: {
-        before: ['jest-preset-angular/build/InlineFilesTransformer', 'jest-preset-angular/build/StripStylesTransformer']
-      }
+      astTransformers: [
+        'jest-preset-angular/build/InlineFilesTransformer',
+        'jest-preset-angular/build/StripStylesTransformer'
```
anandtiwary, line 14: Are these no longer needed?

jest.config.js
aaron-steinfeld: > After years of large-scale use at Facebook and of course in Jest itself, as well as recent adoption in create-react-app, we are now confident that jest-circus is highly compatible with jest-jasmine2 and should work in most environments with little to no migration work. There may be minor differences in execution order and strictness, but we expect no major upgrade difficulties other than for code relying on Jasmine-specific APIs such as jasmine.getEnv(). If you rely extensively on such APIs, you can opt back in to the Jasmine-based test runner by configuring "testRunner": "jest-jasmine2".

We don't use jasmine-specific apis as far as I'm aware, so we shouldn't need this


> Running tests in a JSDOM environment incurs a significant performance overhead. Because this was the default behavior of Jest unless otherwise configured up until now, users who are writing Node apps, for example, may not even know they are given an expensive DOM environment that they do not even need.
For this reason, we are changing the default test environment from "jsdom" to "node". If you are affected by this change because you use DOM APIs and do not have the test environment explicitly configured, you should be receiving an error when e.g. document is accessed, and you can configure "testEnvironment": "jsdom" or use per-file environment configuration to resolve this.


As far as jsdom, I wonder if we can get away with node for a lot of them to help perf - we don't use the dom directly frequently. It's possible that's not practical due to transitive usages, in which case we can switch the testEnvironment back over, but if it is and we just need jsdom in a handful of spots, that'd be really cool.

aaron-steinfeld: (I'd guess that the environment is specified by the runner, so using the legacy runner is giving us the legacy environment, which is the part we actually require)

anandtiwary: So, let's remove the testRunner (that is, we will use the default) and add `"testEnvironment": "jsdom"` for now. We will try migrating to node separately.

jest.config.debug.js
```diff
@@ -1,15 +1,14 @@
...
 
 module.exports = {
   preset: 'jest-preset-angular',
+  testRunner: 'jest-jasmine2',
   globals: {
     'ts-jest': {
       tsConfig: '<rootDir>/tsconfig.spec.json',
-      stringifyContentPathRegex: '\\.html$',
-      astTransformers: [
-        'jest-preset-angular/build/InlineFilesTransformer',
```
anandtiwary, line 10: Why are these removed from debug config?

palbizu:
Moved to PR https://github.com/hypertrace/hypertrace-ui/pull/1103

PR #1087, opened 2021-08-25
feat: custom modal dimension

itssharmasandeep:
## Description
This will give the flexibility to use custom dimensions for modal.
 width/height can be set as expression as well -` calc(100vw - 100px)`


projects/components/src/modal/modal.ts
```diff
@@ -2,8 +2,9 @@ import { InjectionToken, TemplateRef, Type } from '@angular/core';
 import { Observable } from 'rxjs';
 
 export interface ModalConfig<TData = unknown> {
-  size: ModalSize;
   content: TemplateRef<unknown> | Type<unknown>;
+  size?: ModalSize;
```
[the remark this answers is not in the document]
anandtiwary: Not sure if we need to make these many changes. Can we follow the approach in with Sheet module. It has a responsive sheet size enum, can we build something similar here?

```
  &.sheet-size-responsive-extra-large {
    width: 100%;
  }
```

check `hypertrace-ui/projects/components/src/overlay/sheet/sheet-overlay.component.ts`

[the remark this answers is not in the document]
anandtiwary: ofcourse, above class would have height: 100% as well and we would be setting 100% for width and height in popover trigger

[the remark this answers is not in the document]
itssharmasandeep: > it should only be size OR dimension right? can we put them into the same parameter (alternatively, using a union - but seems like overkill). Basically just `size: ModalSize | ModalDimension`

For this, is there any way for us to check, value is type of `ModalSize` or `ModalDimension`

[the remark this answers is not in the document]
anandtiwary: Yeah, but can we do something similar and let the content size drive the modal size? Meaning, we shouldn't need to pass width/height as parameters, but can the modal adjust its size based on the content dom dimensions?

projects/components/src/modal/modal-container.component.ts
```diff
@@ -85,6 +87,24 @@ export class ModalContainerComponent {
     this.visible = false;
     this.popoverRef.close();
   }
+
+  private isModalDimension(size: unknown): size is ModalDimension {
+    // tslint:disable-next-line:strict-type-predicates
+    return typeof size === 'object' && size !== null && (size as ModalDimension).width !== undefined;
```
anandtiwary, line 93: you should probably use 'in' operator here

projects/components/src/modal/modal.ts
```diff
@@ -2,8 +2,8 @@ import { InjectionToken, TemplateRef, Type } from '@angular/core';
 import { Observable } from 'rxjs';
 
 export interface ModalConfig<TData = unknown> {
-  size: ModalSize;
   content: TemplateRef<unknown> | Type<unknown>;
+  size: ModalSize | ModalDimension;
```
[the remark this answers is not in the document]
anandtiwary: fine. I don't have strong feelings. I find it weird when we put type based logic in the code. IMO, keeping separate properties makes it cleaner from API perspective. But it is a preference

projects/components/src/modal/modal.ts
```diff
@@ -19,6 +19,33 @@ export const enum ModalSize {
   MediumWide = 'medium-wide'
 }
 
+export interface ModalDimension {
+  // Number => without unit (considered px) and String => with units (expression included)
+  width: number | string;
```
anandtiwary, line 24: What if I pass `23sdgsg` ?

itssharmasandeep: I mean, it won't work at that time. And I think this should be the case. Like suppose if we do this in the css (putting wrong value for width), then it doesn't work. So It is developer's responsibility to put correct value and I think developer can handle this.

aaron-steinfeld: If you want to get really fancy, TS supports this kind of thing now. I'd also be fine just taking in any string, we don't validate to nearly this level elseewhere.

```ts
type Unit = "px" | "em" | "fr" | "%"; // probably some others too

type SizeWithUnit = `${number}${Unit}`;
```

aaron-steinfeld: in action: 
image

itssharmasandeep: Yeah, we could do this, but my intention of using string, to use these kind of expression as well, not only simple units - `calc(100vh - 100px)`

aaron-steinfeld:
```ts
type SizeExpression = SizeWithUnit | `calc(${SizeWithUnit} - ${SizeWithUnit})`
```

Typescript type system is turing complete 😉 

But seriously, yes I agree, overkill.

itssharmasandeep: Yeah!!!

anandtiwary: I will let you guys decide :D

projects/components/src/modal/modal-container.component.ts
itssharmasandeep: done

projects/components/src/modal/modal.ts
itssharmasandeep: According to me, It is good enough and as @aaron-steinfeld  said doing this, is an overkill.
I think we can merge this.

anandtiwary (approved):

PR #1098, opened 2021-08-31
Trace drilldown  to explorer screen

jyothishjose6190:
https://user-images.githubusercontent.com/86998924/134654520-0bc27333-94a3-4683-b3cf-a081f815cf14.mov

Enables the user to click on a trace ID and view the explorer screen with filters applied

- **on a feature**: User Can click on trace ID in trace view,  full trace view and it will take the user to the explorer screen with the trace ID filter applied

### Testing
All tests passed

### Checklist:
- [x ] My changes generate no new warnings


projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts
```diff
@@ -31,12 +33,19 @@ import { ApiTraceDetails, ApiTraceDetailService } from './api-trace-detail.servi
...
-            class="summary-value"
-            icon="${IconType.TraceId}"
-            label="Trace ID"
-            [value]="traceDetails.traceId"
-          ></ht-summary-value>
+
+          <ng-container
+            *ngIf="this.getExploreNavigationParams(traceDetails.traceId) | async as exploreNavigationParams"
+          >
+            <ht-link class="link" [paramsOrUrl]="exploreNavigationParams">
```
aaron-steinfeld, line 40: Could you add a screenshot of what this looks like after the change? Specifically, I'm looking for what this looks like vs what it would look like to use `ExploreFilterLinkComponent` instead of htLink directly.

projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts
```diff
@@ -31,12 +33,19 @@ import { ApiTraceDetails, ApiTraceDetailService } from './api-trace-detail.servi
...
           ></ht-summary-value>
-          <ht-summary-value
-            class="summary-value"
-            icon="${IconType.TraceId}"
-            label="Trace ID"
-            [value]="traceDetails.traceId"
-          ></ht-summary-value>
+
+          <ng-container
+            *ngIf="this.getExploreNavigationParams(traceDetails.traceId) | async as exploreNavigationParams"
```
aaron-steinfeld, line 38: 1 nit, 1 major here:
nit - the ngIf can move down onto the link, the extra container is unnecessary. 
major - calling a function to return an observable will return a new observable each time - so every cycle will be treated as a change from the async pipe's perspective. Instead, you can either assign the observable to a variable and reference that here, or use `htMemoize` pipe before the async one. My guess is the former will be clearer rather than chaining pipes.

aaron-steinfeld (commented):
only commented on one, but they apply in both places

projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts
jyothishjose6190: Screenshot 2021-09-01 at 3 42 12 PM
while hovering we can see cursor pointer and tooltip , otherwise no visual changes

jyothishjose6190: https://user-images.githubusercontent.com/86998924/131654660-0d072b70-a784-4b0a-89cd-b6da3f981a24.mov

please check uploaded a small video

aaron-steinfeld: got it, thanks - I think for consistency, it would make since to use the filter link here. It also gives a visual indicator, without requiring a hover, that it's actionable.

Something like:
```html
<ht-explore-filter-link [paramsOrUrl]="this.getExploreNavigationParams(traceDetails.traceId) | async"
  <ht-summary-value
    class="summary-value"
    icon="${IconType.TraceId}"
    label="Trace ID"
    [value]="traceDetails.traceId">
  </ht-summary-value>
</ht-explore-filter-link>

Comment (anandtiwary) [reply projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts:40]: Yep, I thought we recommended `ht-explore-filter-link` only. It is same as ht-link but it brings in an icon at the end that indicates the user what type of link it is.

File: projects/observability/src/pages/trace-detail/trace-detail.page.component.ts [hunk 3]
```diff
@@ -30,12 +32,16 @@ import { TraceDetails, TraceDetailService } from './trace-detail.service';
             icon="${IconType.Time}"
             [value]="traceDetails.timeString"
           ></ht-summary-value>
-          <ht-summary-value
-            class="summary-value"
-            icon="${IconType.TraceId}"
-            label="Trace ID"
-            [value]="traceDetails.id"
-          ></ht-summary-value>
+          <ng-container *ngIf="this.getExploreNavigationParams(traceDetails.id) | async as exploreNavigationParams">
```
anandtiwary, line 35: Same changes here

projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts
aaron-steinfeld: Both this comment and the one about the link component came up in the PR to add this support on span attributes, so you could use that PR as a reference too (although there, the decision was made for styling reasons to keep the filter link and value rendered separately - not sure if the same applies here):

https://github.com/hypertrace/hypertrace-ui/pull/1079/files#diff-8912ac915c8ff140908e47b3997b34e34b6eb5a0defd418b07953b86980ba4c5R19-R24

jyothishjose6190:
@aaron-steinfeld @anandtiwary  made changes. please review

projects/observability/src/pages/trace-detail/trace-detail.page.component.ts
```diff
@@ -82,4 +92,9 @@ export class TraceDetailPageComponent {
   public onClickBack(): void {
     this.navigationService.navigateBack();
   }
+
+  public getExplorerNavigationParams = (traceDetails: ApiTraceDetails | undefined): Observable<NavigationParams> =>
```
aaron-steinfeld, line 96: This is a full trace, but also doesn't appear like it can ever be called with undefined.
```suggestion
  public getExplorerNavigationParams = (traceDetails: TraceDetails): Observable<NavigationParams> =>
```

projects/observability/src/pages/trace-detail/trace-detail.page.module.ts
```diff
@@ -47,10 +49,13 @@ const ROUTE_CONFIG: HtRoute[] = [
   imports: [
     RouterModule.forChild(ROUTE_CONFIG),
     CommonModule,
+    ExploreFilterLinkModule,
     LabelModule,
     TracingDashboardModule,
     IconModule,
     SummaryValueModule,
+    LinkModule,
```
aaron-steinfeld, line 57: link no longer used

projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts
```diff
@@ -86,4 +96,9 @@ export class ApiTraceDetailPageComponent {
   public navigateToFullTrace(traceId: string, startTime: string): void {
     this.navigationService.navigateWithinApp(['/trace', traceId, { startTime: startTime }]);
   }
+
+  public getExplorerNavigationParams = (traceDetails: ApiTraceDetails | undefined): Observable<NavigationParams> =>
```
aaron-steinfeld, line 100:
```suggestion
  public getExplorerNavigationParams = (traceDetails: ApiTraceDetails): Observable<NavigationParams> =>
```

projects/observability/src/shared/dashboard/widgets/waterfall/waterfall-widget-renderer.component.ts
```diff
@@ -67,12 +70,18 @@ import { MarkerSelection, WaterfallChartComponent } from './waterfall/waterfall-
...
-          <ht-summary-value
-            data-sensitive-pii
-            icon="${IconType.SpanId}"
-            label="Span ID"
-            [value]="this.selectedData!.id"
-          ></ht-summary-value>
+          <ht-explore-filter-link
+            class="filter-link"
+            [paramsOrUrl]="getExploreNavigationParams | htMemoize: this.selectedData | async"
+            htTooltip="See traces in Explorer"
```
aaron-steinfeld, line 76: it looks like this is going to see spans, not traces

projects/observability/src/pages/api-trace-detail/api-trace-detail.page.component.ts
anandtiwary: +1

projects/observability/src/pages/trace-detail/trace-detail.page.component.ts
anandtiwary: +1

projects/observability/src/pages/trace-detail/trace-detail.page.module.ts
anandtiwary: please remove it

projects/observability/src/shared/dashboard/widgets/waterfall/waterfall-widget-renderer.component.test.ts
```diff
@@ -14,9 +17,13 @@ describe('Waterfall widget renderer component', () => {
...
-    imports: [LoadAsyncModule],
-    providers: [...mockDashboardWidgetProviders(mockModel), mockProvider(OverlayService)],
-    declarations: [MockComponent(WaterfallChartComponent)]
+    imports: [LoadAsyncModule, MemoizeModule],
+    providers: [
+      ...mockDashboardWidgetProviders(mockModel),
+      mockProvider(OverlayService),
+      mockProvider(ExplorerService)
+    ],
+    declarations: [MockComponent(WaterfallChartComponent), MockComponent(ExploreFilterLinkComponent)]
```
anandtiwary, line 26: Can we add some tests for this? (and in the other components)

projects/observability/src/shared/dashboard/widgets/waterfall/waterfall-widget-renderer.component.ts
anandtiwary: @jyothishjose6190 Please update this.

jyothishjose6190:
@anandtiwary @aaron-steinfeld  please review this

itssharmasandeep (changes requested):
Instead of doing this
```
<ht-explore-filter-link
class="filter-link"
[paramsOrUrl]="getExplorerNavigationParams | htMemoize: traceDetails | async"
htTooltip="See traces in Explorer"
 >
  <ht-summary-value
    class="summary-value"
    icon="${IconType.TraceId}"
    label="Trace ID"
    [value]="traceDetails.traceId"
  ></ht-summary-value>
 </ht-explore-filter-link>
```

Please do this similar to this at all the places, with appropriate CSS
```
<div class="filterable-summary-value">
  <ht-summary-value
  class="summary-value"
  icon="${IconType.TraceId}"
  label="Trace ID"
  [value]="traceDetails.traceId"
  ></ht-summary-value>
  <ht-explore-filter-link
    class="filter-link"
    [paramsOrUrl]="getExplorerNavigationParams | htMemoize: traceDetails | async"
    htTooltip="See traces in Explorer"
  >
  </ht-explore-filter-link>
</div>
```

The first approach is creating some issues (like 2 tooltips for spans etc and it also doesn't look good). I think the behaviour should be similar like other place where we click on the filter button and then we filter things.

aaron-steinfeld:
> Instead of doing this
> 
> ```
> <ht-explore-filter-link
> class="filter-link"
> [paramsOrUrl]="getExplorerNavigationParams | htMemoize: traceDetails | async"
> htTooltip="See traces in Explorer"
>  >
>   <ht-summary-value
>     class="summary-value"
>     icon="${IconType.TraceId}"
>     label="Trace ID"
>     [value]="traceDetails.traceId"
>   ></ht-summary-value>
>  </ht-explore-filter-link>
> ```
> 
> Please do this similar to this at all the places, with appropriate CSS
> 
> ```
> <div class="filterable-summary-value">
>   <ht-summary-value
>   class="summary-value"
>   icon="${IconType.TraceId}"
>   label="Trace ID"
>   [value]="traceDetails.traceId"
>   ></ht-summary-value>
>   <ht-explore-filter-link
>     class="filter-link"
>     [paramsOrUrl]="getExplorerNavigationParams | htMemoize: traceDetails | async"
>     htTooltip="See traces in Explorer"
>   >
>   </ht-explore-filter-link>
> </div>
> ```
> 
> The first approach is creating some issues (like 2 tooltips for spans etc and it also doesn't look good). I think the behaviour should be similar like other place where we click on the filter button and then we filter things.

So conceptually, the first approach is better. The whole idea of a filter link component accepting content is exactly this. If it doesn't look good or has duplicate tooltips (is this because the component already has a tooltip and we're adding a second one here?), I'd say we should fix that.

itssharmasandeep:
> So conceptually, the first approach is better. The whole idea of a filter link component accepting content is exactly this. If it doesn't look good or has duplicate tooltips (is this because the component already has a tooltip and we're adding a second one here?), I'd say we should fix that.

But the problem I see with the current approach is, we have `<ng-content>` inside the `anchor` tag, which I think should not have. So problem it creating is we are able to click `traceID/spanID` as well with the filter icon (which I see as not a good UX), either we remove the filter icon if `traceID` is clickable or we make traceID non clickable.

IMO, The filter icon should only be responsible to click and filter, rather than whole component(content inside it)

jyothishjose6190:
> > So conceptually, the first approach is better. The whole idea of a filter link component accepting content is exactly this. If it doesn't look good or has duplicate tooltips (is this because the component already has a tooltip and we're adding a second one here?), I'd say we should fix that.
> 
> But the problem I see with the current approach is, we have `<ng-content>` inside the `anchor` tag, which I think should not have. So problem it creating is we are able to click `traceID/spanID` as well with the filter icon (which I see as not a good UX), either we remove the filter icon if `traceID` is clickable or we make traceID non clickable.
> 
> IMO, The filter icon should only be responsible to click and filter, rather than whole component(content inside it)

Also on other screens, we have the same functionality in other screens also. click is only for the filter icon. 
Screenshot 2021-09-23 at 12 31 18 PM

aaron-steinfeld:
> Also on other screens, we have the same functionality. Clicking on the filter icon alone will apply the filter. I think the second approach suggested by @itssharmasandeep, matches the implementations in other screens

So purely talking from a UX perspective, the behavior I'd expect for the code above (`<filter-link><content/><filter-link>`) is to render only the content, show up the filter icon next to it when the content is hovered, and only support clicking on the filter icon. Based on the screenshot, that sounds like what we're doing on that table.

If that's not how the filter link component behaves today, we should probably change that but it doesn't need to be part of this PR. The main goal is consistency in UX, consistency in code can follow after.

itssharmasandeep:
> So purely talking from a UX perspective, the behavior I'd expect for the code above (<filter-link><content/><filter-link>) is to render only the content, show up the filter icon next to it when the content is hovered, and only support clicking on the filter icon. Based on the screenshot, that sounds like what we're doing on that table.
If that's not how the filter link component behaves today, we should probably change that but it doesn't need to be part of this PR. The main goal is consistency in UX, consistency in code can follow after.

I completely agree with you on this, when I was working on tag search then I implemented this hover effect (but not a part of filter link component) and it is not there inside it currently.
If we don't want to change filter link component here, then I think my approach would be better as it would give us consistency that we want right now, we can take filter link component change in a follow up PR.

aaron-steinfeld:
> If we don't want to change filter link component here, then I think my approach would be better as it would give us consistency that we want right now, we can take filter link component change in a follow up PR.

Yep, either way works for me. Doing it in this PR is probably quicker, but is a scope increase - but given its only used in one other spot that I see, not a significant one.

> when I was working on tag search then I implemented this hover effect (but not a part of filter link component) and it is not there inside it currently.

I could not for the life of me figure out how that was working... turns out I haven't pulled in weeks and was comparing a UI with your changes and code without :(

jyothishjose6190:
@aaron-steinfeld @itssharmasandeep please review

itssharmasandeep (approved):

itssharmasandeep (approved):
It looks good to me, works fine as well, as expected.
@jyothishjose6190 let @aaron-steinfeld also review then we can merge this

aaron-steinfeld (approved):

PR #1100, opened 2021-08-31
User telemetry

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation


anandtiwary:
@aaron-steinfeld This is a very early change for Telemetry. But I want to get a feedback early. Please take a look. 

This change is working and I am able to load and send data to two telemetry platforms (via their autotrack feature)

We will have to store snippets and type d.ts in some cases. I honestly don't like keeping them with ht-ui code. The only alternative I see is for this package to live in its own repo like hyperdash.

Usage:
```
    UserTelemetryModule.withProviders([
      {
        telemetryProvider: UserTelemetryType.FullStory,
        telemetryProviderConfig: { orgId: environment.fullStoryOrgId },
        enablePageTracking: true,
        enableEventTracking: true,
        enableErrorTracking: true
      },
      {
        telemetryProvider: UserTelemetryType.FreshPaint,
        telemetryProviderConfig: { orgId: '1d2f4536-0f07-4564-a5e9-8ed11539c07a' },
        enablePageTracking: true,
        enableEventTracking: true,
        enableErrorTracking: true
      }
    ])
  ]
```

aaron-steinfeld (commented):
ignored specific impls for this review

aaron-steinfeld:
> We will have to store snippets and type d.ts in some cases. I honestly don't like keeping them with ht-ui code. The only alternative I see is for this package to live in its own repo like hyperdash.

We could move them into their own repo, but it doesn't really gain anything if ht still has to depend on it. We already have a mini-mono repo for HT, so the only reason to move some dependency out is if we want it consumable outside of the HT application (which we may eventually, but don't need right now).

> Usage: This api would change as per our latest discussion.

Added a number of comments on the API.

projects/common/src/telemetry/user-telemetry.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
+import { UserTraits } from './telemetry';
+import { UserTelemetryInternalService } from './user-telemetry-internal.service';
+
+@Injectable({ providedIn: 'root' })
+export class UserTelemetryService2 {
```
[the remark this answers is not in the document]
anandtiwary: yep. temp

projects/common/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,43 @@
+import { InjectionToken } from '@angular/core';
+import { Dictionary } from './../utilities/types/types';
+
+export interface UserTelemetry {
+  readonly config: UserTelemetryConfig;
+  readonly initialized: boolean;
+  initialize(): void;
```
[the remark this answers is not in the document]
anandtiwary: we can. I was thinking they can do a no-op. But optional makes it cleaner

projects/common/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,43 @@
...
+  identify(userTraits: UserTraits): void;
+  trackEvent(name: string, eventData: Dictionary<unknown>): void;
+  trackPage(url: string): void;
+  shutdown(): void;
+}
+
+export interface UserTraits extends Dictionary<unknown> {
```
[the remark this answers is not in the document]
anandtiwary: They can be optional.

projects/common/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,43 @@
...
+  enablePageTracking: boolean;
+  enableEventTracking: boolean;
+  enableErrorTracking: boolean;
+}
+
+export interface TelemetryProviderConfig {
+  orgId: string;
+}
+
+export const enum UserTelemetryType {
```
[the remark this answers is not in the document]
anandtiwary: That makes this code more useful. I understand that pluggable model is good and I want to support that. But OOTB would make it easier for multiple consumers, in this case hypertrace and traceable.

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,70 @@
...
+  public identify(userTraits: UserTraits): void {
+    this.initializedTelemetryProviders.forEach(telemetryProvider => telemetryProvider.identify(userTraits));
+  }
+
+  public shutdown(): void {
+    this.initializedTelemetryProviders.forEach(telemetryProvider => telemetryProvider.shutdown());
+  }
+
+  public trackEvent(_name: string, _data: Dictionary<unknown>): void {
+    this.initializedTelemetryProviders.forEach(telemetryProvider => telemetryProvider.shutdown());
```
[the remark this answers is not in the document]
anandtiwary: Yeah, those i need to fix. I wanted a feedback just on the overall approach. Specifics still need to be improved.

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,70 @@
...
+        return new FreshPaintTelemetry(config);
+
+      case UserTelemetryType.FullStory:
+      default:
+        return new FullStoryTelemetry(config);
+    }
+  }
+
+  private maybeSetupPageTracking(): void {
+    const enablePageTracking = this.initializedTelemetryProviders.some(provider => provider.config.enablePageTracking);
```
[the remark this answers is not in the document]
anandtiwary: I was initially thinking of putting the checks inside, but I agree better to check it upfront in this service before calling the methods

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
aaron-steinfeld: There's a difference between having support for certain ones out of the box vs having the common code know about them - that's what I was trying to get at here. Having OOTB support is fine, but if we take in injection tokens instead of well known strings, the common code doesn't need to know about any specific provider - which we'll need anyway to support providing custom ones.


So basically:
```ts
    UserTelemetryModule.withProviders([
      {
        telemetryProvider: FullStoryTelemetryProvider, // Instead of: UserTelemetryType.FullStory,
        telemetryProviderConfig: { orgId: environment.fullStoryOrgId },
        enablePageTracking: true,
        enableEventTracking: true,
        enableErrorTracking: true
      }
    ]);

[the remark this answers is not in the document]
Comment (anandtiwary) [reply projects/common/src/telemetry/telemetry.ts:33]: So what is provider in this case? Is it a token or a class implementation object?

File: projects/common/src/telemetry/telemetry.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (anandtiwary) [reply projects/common/src/telemetry/telemetry.ts:7]: one side effect of interface is that required properties like config and initialized have to be public. This increases the coupling.

File: projects/common/src/telemetry/telemetry.ts [hunk 7]
```diff
@@ -0,0 +1,43 @@
...
+
+export interface UserTraits extends Dictionary<unknown> {
+  email: string;
+  companyName: string;
+  name: string;
+  displayName: string;
+}
+
+export interface UserTelemetryConfig<TConfig extends TelemetryProviderConfig = TelemetryProviderConfig> {
+  telemetryProvider: UserTelemetryType | string;
```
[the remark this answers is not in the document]
anandtiwary: how would the token approach work here for custom impl ? With injection token, sure I can use it for defining a custom module. But registration is happening at module load time only. So, I can just use that to register a given provider with the provider key and then do a look up when a config with matching key is passed. I won't be injecting it in a service. 

So, I made the key to accept a string. Some strings are ootb just like table cell renderers. But you also have an option to register and pass custom key.

projects/common/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,43 @@
...
+  telemetryProviderConfig: TConfig;
+  enablePageTracking: boolean;
+  enableEventTracking: boolean;
+  enableErrorTracking: boolean;
+}
+
+export interface TelemetryProviderConfig {
```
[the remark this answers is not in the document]
anandtiwary: That's something all provider would need for them to work. How would we configure and initialize the ootb ones otherwise?

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,70 @@
...
+import { FullStoryTelemetry } from './providers/fullstory/full-story-provider';
+import { Injectable } from '@angular/core';
+import { Dictionary } from '../utilities/types/types';
+import { UserTelemetry, UserTelemetryConfig, UserTelemetryType, UserTraits } from './telemetry';
+import { NavigationEnd, Router } from '@angular/router';
+import { filter, map } from 'rxjs/operators';
+
+@Injectable({ providedIn: 'root' })
+export class UserTelemetryInternalService {
+  private readonly registeredTelemetryProviderMap: Map<string, UserTelemetry> = new Map();
```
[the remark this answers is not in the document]
anandtiwary: Yeah, something i have to change. I was thinking on what to do if a user tries to pass two separate configs for the same telemetry platform? May be with a different org in? Should we send data to both or mark it as error case ?

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
aaron-steinfeld: interfaces always reduce coupling. Take initialized as an example - who's the consumer of that? If it's the telemetry orchestrator (what you called the internal one), we don't need it on the interface, we can just track it locally. If it's the specific telemetry provider, it doesn't need to expose it via interface. It's up to the impl to decide whether or not it wants to track it which it can do when its initialize call is made. Instead think of this interface as answering two questions that are key to the relationship between one provider and the orchestrator (and which must be public by their nature):
- what actions does the orchestrator need to tell that provider to take?
- what state does the orchestrator need to know about the provider that it couldn't track itself (which I believe those two readonly props fall under)

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
aaron-steinfeld: Table cell renderers are a slightly, but not completely, different story. We register them exactly as I'm describing - by providing a class. The string association is there because we want to support configuring the table in json where we only have strings - otherwise, I'd just define my columns with components - way easier! The unfortunate concusion there - we didn't really need to build that string->renderer mapping into table, we could have left it as the responsibility of the table widget only.

So as to how this would work - imagine I'm given something like
```ts
    UserTelemetryModule.withProviders([
      {
        telemetryProvider: FullStoryTelemetryProvider, // Instead of: UserTelemetryType.FullStory,
        telemetryProviderConfig: { orgId: environment.fullStoryOrgId },
        enablePageTracking: true,
        enableEventTracking: true,
        enableErrorTracking: true
      }
    ]);
```

In the orchestrator, I can now use the provider definition to ingest an instance of FullStoryTelemetryProvider by asking the injector for that token, then use it in the same way. There's no benefit to a string lookup here - it's just an extra level of indirection on either side. The caller now needs to provide both the specific instance config, and define the mapping of string to provider; the orchestrator needs to get that specific instance config and go in reverse, finding that mapping to get the provider.

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
aaron-steinfeld: Each provider defines its own config. They could all have org ID, but the orchestrator doesn't need to know that or require it. In the world I'm describing, the _only_ difference in an OOTB provider is that it's already written. It's not treated as special in any other way.

projects/common/src/telemetry/user-telemetry-internal.service.ts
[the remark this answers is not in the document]
aaron-steinfeld: No reason not to support it, IMO. Every newly registered telemetry config should be added to the array - it would be more work to try to dedupe.

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
anandtiwary: what about `config`?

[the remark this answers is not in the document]
anandtiwary: You wouldn't want it to change or meddle with it after the provider has been initialized. As a public property, this becomes difficult to enforce.

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
anandtiwary: Makes sense. Can you help me understand how would you associate `FullStoryTelemetryProvider` token with its corresponding class impl `FullStoryTelemetry`?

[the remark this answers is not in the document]
aaron-steinfeld: I think just confusion in terminology here - in angular a provider is two things: a token to get an instance (as in, what a dependent give the injector when requesting injection), and a factory to build one (what the injector needs to actually build an instance for that token). Most of the time, we use a class constructor as both rolled into one, and that's the context I'm using the term here (I was calling the class `FullStoryTelemetryProvider`, you were calling it `FullStoryTelemetry`).

In other words, if my class was named `FullStoryTelemetry`, I'd pass in `telemetryProvider: FullStoryTelemetry` as part of my provider definition and the orchestrator would call the injector with `injector.get(providerDef.telemetryProvider)` to get an instance of `FullStoryTelemetry`.

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
aaron-steinfeld: the provider doesn't need to expose its config, it only needs to read it in. So given the example I had below:
```ts
UserTelemetryModule.withProviders([
  {
    telemetryProvider: FullStoryTelemetryProvider, // Instead of: UserTelemetryType.FullStory,
    telemetryProviderConfig: { orgId: environment.fullStoryOrgId },
    enablePageTracking: true,
    enableEventTracking: true,
    enableErrorTracking: true
  }
]);

interface ConfiguredUserTelemetryProvider<TConfig> {
    telemetryProvider: UserTelemetryProvider<TConfig>;
    telemetryProviderConfig: TConfig;
    enablePageTracking: boolean;
    enableEventTracking: boolean;
    enableErrorTracking: boolean;
}
```

`FullStoryTelemetryProvider` might implement `UserTelemetryProvider` which perhaps looks like:
```ts
export interface UserTelemetryProvider<TConfig> {
  initialize(config: TConfig): void;
  identify(userTraits: UserTraits): void;
  trackEvent(name: string, eventData: Dictionary<unknown>): void;
  trackPage(url: string): void;
  shutdown(): void;
}
```

projects/common/src/telemetry/telemetry.ts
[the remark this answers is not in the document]
anandtiwary: Got it. But here is a circular dep. As per the current design, the provider class needs to take a config in its constructor. In the above interface, the provider class is part of the **config**. They need to be separated now. Also, I need to check how we can pass a constructor argument in the injector.get method.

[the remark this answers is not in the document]
anandtiwary: okay. saw your other comment. Let me spend some more time over it.

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,155 @@
...
+  public identify(userTraits: UserTraits): void {
+    this.telemetryActionSubject.next({ type: TelemetryActionType.Identify, data: userTraits });
+  }
+
+  public shutdown(): void {
+    this.telemetryActionSubject.next({ type: TelemetryActionType.Shutdown });
+  }
+
+  public trackEvent(name: string, data: Dictionary<unknown>): void {
+    this.telemetryActionSubject.next({ type: TelemetryActionType.TrackEvent, name: name, data: data });
```
anandtiwary, line 66: Used observables to make event processing run in a separate execution loop.

projects/common/src/telemetry/user-telemetry.service.ts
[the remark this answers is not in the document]
anandtiwary: will change the name after this round of review

projects/common/src/telemetry/user-telemetry.module.ts
```diff
@@ -0,0 +1,45 @@
...
+  declarations: [TrackDirective],
+  exports: [TrackDirective],
+  providers: [
+    {
+      provide: USER_TELEMETRY_PROVIDER_TOKENS,
+      useValue: [],
+      multi: true
+    },
+    {
+      provide: ErrorHandler,
```
anandtiwary, line 17: Because I am injecting ErrorHandler here and plan to import the same module at places where I plan to use `htTrack`, I think I should have `forRoot()` and `forChild()` methods.

projects/components/src/link/link.component.ts
```diff
@@ -18,6 +18,7 @@ import { EMPTY, Observable } from 'rxjs';
       [queryParamsHandling]="navData?.extras?.queryParamsHandling"
       [skipLocationChange]="navData?.extras?.skipLocationChange"
       [replaceUrl]="navData?.extras?.replaceUrl"
+      htTrack="{{ navData && navData.path ? navData.path : '' }}"
```
anandtiwary, line 21: an example, (which we also want to keep) :)

anandtiwary:
@aaron-steinfeld ready for another round. I will be making more changes in parallel, but lemme know what you think

projects/common/src/telemetry/providers/freshpaint/freshpaint-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Dictionary } from './../../../utilities/types/types';
+import { FreshPaint, loadFreshPaint } from './load-snippet';
+
+@Injectable({ providedIn: 'root' })
+export class FreshPaintTelemetry<InitConfig extends TelemetryProviderConfig>
+  implements UserTelemetryProvider<InitConfig> {
+  private freshPaint?: FreshPaint;
+
+  public initialize(config: InitConfig): void {
+    this.freshPaint = loadFreshPaint();
```
anandtiwary, line 12: btw, the script loads one global object. So, we can't support multiple configs for the same telemetry provider.

projects/common/src/telemetry/user-telemetry.module.ts
anandtiwary: Probably I should just move this to L31 and rename `withProviders` to `forRoot()`

projects/common/src/telemetry/error-handler/telemetry-global-error-handler.ts
```diff
@@ -0,0 +1,26 @@
+import { LocationStrategy, PathLocationStrategy } from '@angular/common';
+import { ErrorHandler, Injectable, Injector } from '@angular/core';
+import { UserTelemetryInternalService } from '../user-telemetry-internal.service';
+
+@Injectable()
+export class TelemetryGlobalErrorHandler implements ErrorHandler {
+  public constructor(private readonly injector: Injector) {}
+
+  public handleError(error: Error): Error {
+    const telemetryService = this.injector.get(UserTelemetryInternalService);
```
aaron-steinfeld, line 10: Why are we working with the injector ini this class rather than just declaring dependencies? Circular?

projects/common/src/telemetry/error-handler/telemetry-global-error-handler.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    telemetryService.trackErrorEvent(message, {
+      message: message,
+      url: url,
+      stack: error.stack,
+      name: error.name,
+      isError: true
+    });
+
+    throw error;
```
aaron-steinfeld, line 24: Is this responsible for re-throwing? It doesn't seem to indicate that in the docs

projects/common/src/telemetry/providers/freshpaint/freshpaint-provider.ts
aaron-steinfeld: That's fine - can only do it if they support it. Now that this PR is getting more mature, can we start considering separating it down - I'd think the telemetry orchestration, tracking directive, and each specific provider can all be separate PRs.

projects/common/src/telemetry/track/track.directive.ts
```diff
@@ -0,0 +1,17 @@
...
+})
+export class TrackDirective {
+  @Input('htTrack')
+  public name!: string;
+
+  public constructor(private readonly userTelemetryInternalService: UserTelemetryInternalService) {}
+
+  @HostListener('click', ['$event'])
+  public trackClick(event: MouseEvent): void {
+    this.userTelemetryInternalService.trackEvent(`Click: ${this.name}`, { target: event.target, type: event.type });
```
aaron-steinfeld, line 15: what's the name meant to be here?

projects/common/src/telemetry/track/track.directive.ts
```diff
@@ -0,0 +1,17 @@
...
+@Directive({
+  selector: '[htTrack]'
+})
+export class TrackDirective {
+  @Input('htTrack')
+  public name!: string;
+
+  public constructor(private readonly userTelemetryInternalService: UserTelemetryInternalService) {}
+
+  @HostListener('click', ['$event'])
```
aaron-steinfeld, line 13: So how can we make this support events other than click? This is something we should consider now because adding that in the future will break usages.

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,155 @@
+import { Injectable, Injector, OnDestroy } from '@angular/core';
+import { NavigationEnd, Router } from '@angular/router';
+import { Observable, ReplaySubject, Subject } from 'rxjs';
+import { filter, map, takeUntil } from 'rxjs/operators';
+import { Dictionary } from '../utilities/types/types';
+import { UserTelemetryProvider, UserTelemetryRegistrationConfig, UserTraits } from './telemetry';
+
+@Injectable({ providedIn: 'root' })
+export class UserTelemetryInternalService implements OnDestroy {
```
aaron-steinfeld, line 9: there is no destroy for a root injectable, implementing ondestroy doesn't do anything since this lives the whole life of the app.

projects/common/src/telemetry/user-telemetry-internal.service.ts
aaron-steinfeld: Is that something we actually need? We wouldn't do that when using them directly, as I assume they're not doing anything heavy.

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,155 @@
...
+        .filter(provider => provider.enableErrorTracking)
+        .forEach(provider => provider.telemetryProvider.trackError?.(`Error: ${action.name!}`, action.data!))
+    );
+  }
+
+  private setupAutomaticPageTracking(): void {
+    this.router.events
+      .pipe(
+        filter((event): event is NavigationEnd => event instanceof NavigationEnd),
+        map(route => this.trackPageEvent(`Visited: ${route.url}`, { url: route.url })),
```
aaron-steinfeld, line 128: this should be in the subscription rather than mapping for side effects.

projects/common/src/telemetry/user-telemetry.module.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+  ]
+})
+// tslint:disable-next-line: no-unnecessary-class
+export class UserTelemetryModule {
+  public constructor(
+    userTelemetryInternalService: UserTelemetryInternalService,
+    @Inject(USER_TELEMETRY_PROVIDER_TOKENS) providerConfigs: UserTelemetryRegistrationConfig<unknown>[][]
+  ) {
+    userTelemetryInternalService.register(...providerConfigs.flat());
```
aaron-steinfeld, line 28: The way this is currently set up, we'll have dupe registrations since `USER_TELEMETRY_PROVIDER_TOKENS` is multi and will keep growing with each call to `withProviders`. Either need register to handle that or a way to dedupe like tracking last index.

projects/common/src/telemetry/user-telemetry.module.ts
aaron-steinfeld: Or just split the directive into a separate module - probably more straight forward.

projects/common/src/telemetry/user-telemetry.service.ts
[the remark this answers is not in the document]
aaron-steinfeld: Is it even still used?

projects/components/src/link/link.component.ts
aaron-steinfeld: but not in this PR!

projects/common/src/telemetry/error-handler/telemetry-global-error-handler.ts
anandtiwary: Yes, exactly.

projects/common/src/telemetry/error-handler/telemetry-global-error-handler.ts
anandtiwary: If we want to swallow the error from the console then we need to throw it here. I think we should keep the default behavior

projects/common/src/telemetry/providers/freshpaint/freshpaint-provider.ts
anandtiwary: Sure. let's take it to 90ish %. Splitting it up would be pretty easy

projects/common/src/telemetry/track/track.directive.ts
anandtiwary: yeah, so the name could be anything we configure.  The product team can specify what name they want for a particular click event

We can also set it to "Logout" when we use this directive on the logout button. We can also add this directive on the button component and use the button label as `name`.

projects/common/src/telemetry/error-handler/telemetry-global-error-handler.ts
aaron-steinfeld: yes, agreed

projects/common/src/telemetry/track/track.directive.ts
aaron-steinfeld: I'd probably take it as a directive arg

projects/common/src/telemetry/track/track.directive.ts
anandtiwary: From a Product/Marketing perspective, hover and mouseover events are not very interesting. 'Click' and possibly 'Context menu click' would be worth tracking. If you are trying to optimize for UX teams, then we really need  'auto-tracking' based telemetry tools.

So, I was debating between keeping different directives for each event type vs one directive with multiple config options. IMO, multiple config options would be better from a usability perspective (for our current use case). 

We can support properties like `enableClickTracking, enableContextMenuTracking, etc` and for each type, we can have a host Listener binding that would emit only when the corresponding flag is enabled.

Of course, a host listener would still be attached for disabled events. So, it's not ideal as well. But it would be better than having multiple tracking directives on the same element especially during the initial cases.

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: okay. Then I can remove the `takeUntil` logic as well.

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: They make a network call and may add a small delay. I still want to protect against it.

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: yes, I missed to update this one

projects/common/src/telemetry/user-telemetry.module.ts
anandtiwary: What would we call the new module?

projects/common/src/telemetry/user-telemetry.module.ts
anandtiwary: Lemme check

projects/common/src/telemetry/track/track.directive.ts
aaron-steinfeld: click, right click, enter/return - there are other relevant events. Why not just `htTrack="click" [trackEventName]="label"`? Using a host listener isn't a great fit here - see `EventBlockerComponent` - it had a similar requirement.

projects/common/src/telemetry/track/track.directive.ts
anandtiwary: It is a directive argument
  `@Input('htTrack') public name!: string;`

projects/common/src/telemetry/track/track.directive.ts
anandtiwary: lemme check

projects/common/src/telemetry/user-telemetry-internal.service.ts
aaron-steinfeld: that network call would be out of band anyway though, right?

projects/common/src/telemetry/user-telemetry.module.ts
aaron-steinfeld: TelemetryTrackModule or something like that? Same as our other single component modules.

projects/common/src/telemetry/track/track.directive.ts
anandtiwary: Yeah, we can take an array of event types as an input (and assign a reasonable default).

anandtiwary:
okay, I think I have enough to work with. I will address these pending comments and will create smaller PRs for easier review.

projects/common/src/telemetry/track/track.directive.ts
aaron-steinfeld: ah whoops - went straight to the listener and missed that.

anandtiwary:
Moving to smaller PRs

PR #1122, opened 2021-09-14
feat: score bar component

arjunlalb:
## Description
Score bar component.

The component can have 2 type of displays - single or pill.
Single bar display supports absolute values as well as percentage values.
Pill bar display supports only absolute values.
Screenshot 2021-09-14 at 1 36 28 PM

### Testing
Manually tested. Relevant UTs added.

### Checklist:


projects/components/src/score-bar/score-bar.component.ts
```diff
@@ -0,0 +1,110 @@
...
+            class="pill"
+            [htTooltip]="pill.value"
+            [ngStyle]="{ backgroundColor: pill.color, width: this.pillWidth }"
+          ></div>
+        </div>
+      </ng-template>
+    </div>
+  `
+})
+export class ScoreBarComponent implements OnChanges, AfterViewInit {
```
anandtiwary, line 35: "Progress bar" would sound better. Any particular reason for the chosen name?

arjunlalb: @anandtiwary I initially had this as a progress bar, but later changed it since progress bars usually have a slightly different API that includes animation type for non determinate values, etc.

This felt more oriented towards a particular score. Its the same reason why I didn't want to use the material progress bar underneath the hood for the single bar display type. Just leaving space for when we want to actually create a wrapper for the material progress bar.

anandtiwary: Okay, then should this be part of ht components? May be we put this in the main repo if it wont be shared.

PR #1126, opened 2021-09-15
feat: adding user telemetry config and service

anandtiwary:
## Description
feat: adding user telemetry config and service

### Testing

### Checklist:

### Documentation


package.json
```diff
@@ -37,6 +37,8 @@
     "@angular/platform-browser-dynamic": "^12.2.1",
     "@angular/router": "^12.2.1",
     "@apollo/client": "^3.4.9",
+    "@fullstory/browser": "^1.4.9",
```
aaron-steinfeld, line 40: package changes not part of this PR

projects/common/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,33 @@
+import { InjectionToken, ProviderToken } from '@angular/core';
+import { Dictionary } from './../utilities/types/types';
+export interface UserTraits extends Dictionary<unknown> {
+  email?: string;
+  companyName?: string;
+  name?: string;
+  displayName?: string;
+}
+
+export interface UserTelemetryRegistrationConfig<InitConfig> {
```
aaron-steinfeld, line 10: type params should follow convention (e.g. T, TConfig, TInitConfig, etc.)

projects/common/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,33 @@
...
+  trackPage?(url: string, eventData: Dictionary<unknown>): void;
+  trackError?(error: string, eventData: Dictionary<unknown>): void;
+  shutdown?(): void;
+}
+
+export interface TelemetryProviderConfig {
+  orgId: string;
+}
+
+export const USER_TELEMETRY_PROVIDER_TOKENS = new InjectionToken<UserTelemetryRegistrationConfig<unknown>[][]>(
```
aaron-steinfeld, line 31: This is a local concern to the module file and can be declared there/not exported.

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,144 @@
+import { Injectable, Injector } from '@angular/core';
+import { NavigationEnd, Router } from '@angular/router';
+import { Observable, ReplaySubject, Subject } from 'rxjs';
+import { filter } from 'rxjs/operators';
+import { Dictionary } from '../utilities/types/types';
+import { UserTelemetryProvider, UserTelemetryRegistrationConfig, UserTraits } from './telemetry';
+
+@Injectable({ providedIn: 'root' })
+export class UserTelemetryInternalService {
```
aaron-steinfeld, line 9: I thought we had gotten away from the concept of an Internal and external version of the service? Is the purpose to discourage programmatic usage by anyone but the directives (if yes - why? why prevent programmatic usage if the directive is not sufficient - that should be a shortcut rather than the only way this can be used).

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,144 @@
...
+  private readonly trackPageAction$: Observable<TelemetryAction>;
+  private readonly trackErrorAction$: Observable<TelemetryAction>;
+
+  public constructor(private readonly injector: Injector, private readonly router: Router) {
+    this.telemetryAction$ = this.telemetryActionSubject.asObservable();
+
+    this.identifyAction$ = this.telemetryAction$.pipe(filter(action => action.type === TelemetryActionType.Identify));
+
+    this.trackEventAction$ = this.telemetryAction$.pipe(
+      filter(action => action.type === TelemetryActionType.TrackEvent)
```
aaron-steinfeld, line 24: I still maintain that using observables here just adds complexity for no benefit, but it's internal so not a big deal.

projects/common/src/telemetry/user-telemetry-internal.service.ts
```diff
@@ -0,0 +1,144 @@
...
+       * NoOp
+       */
+    }
+  }
+
+  public identify(userTraits: UserTraits): void {
+    this.telemetryActionSubject.next({ type: TelemetryActionType.Identify, data: userTraits });
+  }
+
+  public shutdown(): void {
```
aaron-steinfeld, line 54: do we have any requirements that shutdown be called? who's in charge of calling it?

projects/common/src/telemetry/user-telemetry.module.ts
```diff
@@ -0,0 +1,45 @@
+import { ErrorHandler, Inject, ModuleWithProviders, NgModule } from '@angular/core';
+import { TelemetryGlobalErrorHandler } from './error-handler/telemetry-global-error-handler';
+import { UserTelemetryRegistrationConfig, USER_TELEMETRY_PROVIDER_TOKENS } from './telemetry';
+import { TrackDirective } from './track/track.directive';
```
aaron-steinfeld, line 4: this got moved out

projects/common/src/telemetry/user-telemetry.module.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+  ]
+})
+// tslint:disable-next-line: no-unnecessary-class
+export class UserTelemetryModule {
+  public constructor(
+    userTelemetryInternalService: UserTelemetryInternalService,
+    @Inject(USER_TELEMETRY_PROVIDER_TOKENS) providerConfigs: UserTelemetryRegistrationConfig<unknown>[][]
+  ) {
+    userTelemetryInternalService.register(...providerConfigs.flat());
```
aaron-steinfeld, line 28: So what happened with the investigation here? I'm almost positive this will result in duplicates being registered.

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: So, I want devs to use directives wherever they want to capture events. There is no reason why we need to expose it via service as well. In some cases, this could lead to duplicate events.

With custom events, we are trying to track user actions. Can you describe a use case where we would want to do it programmatically (Apart from page view and error handling that we have already covered)?

projects/common/src/telemetry/telemetry.ts
anandtiwary: TInitConfig +1

package.json
anandtiwary: oh yeah I will remove

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: Is it complicating the code? I agree it probably makes a small perf improvement, but theoretically, the browser should spend as few cycles as possible in the original execution loop. Let's say we start tracking the `mouseover` event for topology (since we show some contents over there), then this additional processing may have a more perceived impact.

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: The main service in-app (which is also calling the initialize method)

anandtiwary: Since we would only want to capture data on prod env, a separate initialize method would be useful

projects/common/src/telemetry/user-telemetry.module.ts
anandtiwary: I changed it to forRoot and removed multi. So, I am expecting that this would only be imported at the root level.

projects/common/src/telemetry/telemetry.ts
anandtiwary: We will have to export it so that consumers can provide the configs.

anandtiwary:
WIll add tests when we are 90ish % done

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: my bad. I removed the `initialize`call in the latest code. I need to bring it back.

projects/common/src/telemetry/telemetry.ts
aaron-steinfeld: That's done inside the module constructor - the consumer just calls `Module.withProviders/forRoot` I thought?

projects/common/src/telemetry/user-telemetry-internal.service.ts
aaron-steinfeld: How does that know when the app is shutting down? I can't remember, do we get an onDestroy hook for the root component or something?

projects/common/src/telemetry/telemetry.ts
anandtiwary: I don't know where my head is today.

projects/common/src/telemetry/user-telemetry-internal.service.ts
aaron-steinfeld: I think so as a reader (and I'm more comfortable with observables than most!) We're basically just splitting one minuscule event into two minuscule events - the call to this function, which pushes into the observable and then returns control, and then the subscription to that observable which pushes into each telemetry provider. Inside the telemetry providers, they're almost certainly designed to be very performant - either dropping (in a separate event) an async call to report the event, or storing it in a buffer for a later report. Because the foreach is on subscriber side, all of those calls to each provider still happen as part of one event anyway - and since we're running off one event loop (if I remember my JS, which admittedly is super rusty), any delay there would still be user perceived, same as if we had not used the observable - it's not running in a different thread - but we're talking so far outside the realm of perceived perf, it really shouldn't be relevant. Happy to be proved wrong though!

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: Yeah, the application frame component can call it on destroy. Or we can call it when the corresponding feature flag is turned off after the the UI is initialized. 

I don't have a strong use case for it. Maybe I can go and ready why these telemetry platforms expose this method in the first place.

projects/common/src/telemetry/user-telemetry-internal.service.ts
aaron-steinfeld: For example on a handled error - like an error server call that we caught and put a toast for. We're currently only capturing uncaught errors, right? Alternatively, some page event like an auto-refresh. User going idle perhaps? I feel like there's almost infinite possibilities and just don't see a huge benefit of funneling all usages into a single directive. IMO, that's provided as the 90% case, but the programmatic api should always be exposed by the lib (think angular's router-link but direct router navigate api).

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: One thing which is concerning is that we are loading these telemetry platforms through Js during the later stage of app initialization (it is not loaded via index.html). During this stage, I don't the telemetry tool to completely miss the first-page load event because it is not initialized. With an observable-based approach and a Replay subject, I can ensure that those initial events are not missed.

anandtiwary: `it's not running in a different thread ` I can use the `observeOn(async)` that would help.

In general, I agree it is verbose. But lemme know what you think about the above concern.

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: Yeah, so those programmatic things are not interesting metrics for "users". We are trying to capture user actions and that are more relevant for Product/maketing and ux. 

The use cases you are describing are more relevant to UI (imo) and I have looked into this aspect in detail. My recommendation for UI and UX use cases were to use an "autotracking" based telemetry tool like fullstory, logRocket or Zipy. These tools would automatically capture much more info, like cpu/memory usage, network perf and the rest. I don't think we want to replicate all that for this framework, since they are not useful data for a tool like GA and mixpanel. 

For any explicit trackEvent call, it will have to be thought through since some platforms charge on a per event storage basis. 

So... let's wait for an actual use case to come. I am okay with exposing those methods. But let's do it when we have a use case.

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: P.S I may have to make some edits to the existing observable code to take care of the above concern.

aaron-steinfeld: > With an observable-based approach and a Replay subject, I can ensure that those initial events are not missed.

Yeah that's a fair point - although how do we know when the providers are ready to receive events?

> observeOn(async)

Not sure what that would help - rx uses async by default. My main point was that making things async doesn't just background things like it does in when you put it on a different thread. Anyway, it's moot if we want to support the delayed start case anyway.

anandtiwary: > Yeah, that's a fair point - although how do we know when the providers are ready to receive events?

Telemetry tools that come with an npm package, usually allow a callback in their init function, which gets called after the load and init is done. We could use that. 

This would need some changes to the current design. 

For tools with only js script (that we will host. For example. fresh paint), we can build a way to introduce a callback once the object is availabl.e

projects/common/src/telemetry/user-telemetry-internal.service.ts
aaron-steinfeld: Agree to disagree. Anyway, my main thing was about having an "internal" and "external" service. We could get around that with this pattern: https://angular.io/guide/dependency-injection-in-action#class-interface

projects/common/src/telemetry/user-telemetry-internal.service.ts
anandtiwary: So, probably we should have `initialize(config, callback: () => void)` ? And then in the orchestration service, we can use this before we start pushing the events to the tool

aaron-steinfeld: I wouldn't work with callbacks when we're working with Rx everywhere else (especially since that was part of the argument for why it's Rx!). Perhaps pass in an observer?

anandtiwary: Lemme check

anandtiwary: So, I tried this but the whole design was getting complicated with too many subscriptions. I am reverting the whole observable approach for now. For a delayed start, I will first verify how these telemetry tools handle it. After that, I may bring back the observable approach if required.

package-lock.json
```diff
@@ -970,9 +970,9 @@
       }
     },
     "node_modules/@apollo/client": {
-      "version": "3.4.9",
-      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.4.9.tgz",
-      "integrity": "sha512-1AlYjRJ/ktDApEUEP2DqHI38tqSyhSlsF/Q3fFb/aCbLHQfcSZ1dCv7ZlC9UXRyDwQYc0w23gYJ7wZde6W8P4A==",
+      "version": "3.4.11",
```
aaron-steinfeld, line 973: can revert this file, no package json changes mean no package lock changes should be pulled in.

projects/common/src/telemetry/user-telemetry-helper.service.ts
```diff
@@ -0,0 +1,79 @@
...
+  public constructor(private readonly injector: Injector, private readonly router: Router) {
+    this.setupAutomaticPageTracking();
+  }
+
+  public register(...configs: UserTelemetryRegistrationConfig<unknown>[]): void {
+    try {
+      const providers = configs.map(config => this.buildTelemetryProvider(config));
+      this.telemetryProviders = [...this.telemetryProviders, ...providers];
+    } catch (error) {
+      /**
```
aaron-steinfeld, line 20: don't break, but also we should be logging this.

aaron-steinfeld (commented):
could you follow up on the comment about avoiding the two services

package-lock.json
anandtiwary: how do I revert it? I basically had checked in the old version in a commit. So, had to regenerate the package.lock

projects/common/src/telemetry/user-telemetry-helper.service.ts
anandtiwary: Are you suggesting console error?

package-lock.json
aaron-steinfeld: git checkout unless I'm misunderstanding? `git checkout origin/main package-lock.json`

projects/common/src/telemetry/user-telemetry-helper.service.ts
aaron-steinfeld: or the logger service, but we don't use that much any more.

anandtiwary: yeah, that's why I didn't want to create a new dependency.

package-lock.json
anandtiwary: my git knowledge is not good. I realized it when I messed up my angular repo PR (twice). Such an embarrassment.

anandtiwary:
> could you follow up on the comment about avoiding the two services

So for this one, I will have to create an abstract class and export it. I am not sure if it looks any better than exporting the public service directly.

package-lock.json
anandtiwary: done

aaron-steinfeld:
> > could you follow up on the comment about avoiding the two services
> 
> So for this one, I will have to create an abstract class and export it. I am not sure if it looks any better than exporting the public service directly.

It saves from having to create two implementations though - it just allows users to access the main implementation by a different token (in other words, provides an interface)

anandtiwary:
My preference would be to roll with what I have latest. We can make improvements later

projects/common/src/telemetry/user-telemetry-helper.service.ts
aaron-steinfeld: if you don't want to depend on logger service use console.error, either way. Right now, there's no way to see a miconfiguration without either seeing it on the third party side or putting in a breakpoint.

anandtiwary:
This token approach would also allow any consumer to provide their own impl for this service. what are your thoughts on that?

aaron-steinfeld:
> This token approach would also allow any consumer to provide their own impl for this service. what are your thoughts on that?

You _could_ rebind it with another, but only in the same way you could override an existing injection token always - same as in this implementation. (i.e. I could provide `UserTelemetryService` with a different implementation here, too - exactly the same)

In normal use, it just works how dependency injection is meant to work, I inject a class and don't care how it got there, what the specific impl is, or who put it there.

anandtiwary:
With the token approach, I will have to provide this token inside `forRoot` method. That means I will have to expose the `register` method as well. I don't want that.

Let's stay with the two-service approach for now. We can change it later.

aaron-steinfeld:
> With the token approach, I will have to provide this token inside `forRoot` method. That means I will have to expose the `register` method as well. I don't want that.
> 
> Let's stay with the two-service approach for now. We can change it later.

We can change it later if you really want, that's fine - but providing the token inside forRoot doesn't expose register any more than it currently is. The diff basically looks like this as I understand it:


```diff
diff --git a/projects/common/src/public-api.ts b/projects/common/src/public-api.ts
index aae6ba44..c4477d64 100644
--- a/projects/common/src/public-api.ts
+++ b/projects/common/src/public-api.ts
@@ -112,3 +112,6 @@ export * from './time/time';
 
 // Validators
 export * from './utilities/validators/email-validator';
+
+export * from './telemetry/user-telemetry.module';
+export * from './telemetry/user-telemetry.service';
diff --git a/projects/common/src/telemetry/user-telemetry-helper.service.ts b/projects/common/src/telemetry/user-telemetry-helper.service.ts
index 6c0d6fd1..dd684e08 100644
--- a/projects/common/src/telemetry/user-telemetry-helper.service.ts
+++ b/projects/common/src/telemetry/user-telemetry-helper.service.ts
@@ -3,12 +3,14 @@ import { NavigationEnd, Router } from '@angular/router';
 import { filter } from 'rxjs/operators';
 import { Dictionary } from '../utilities/types/types';
 import { UserTelemetryProvider, UserTelemetryRegistrationConfig, UserTraits } from './telemetry';
+import { UserTelemetryService } from './user-telemetry.service';
 
 @Injectable({ providedIn: 'root' })
-export class UserTelemetryHelperService {
+export class UserTelemetryHelperService extends UserTelemetryService {
   private telemetryProviders: UserTelemetryInternalConfig[] = [];
 
   public constructor(private readonly injector: Injector, private readonly router: Router) {
+    super();
     this.setupAutomaticPageTracking();
   }
 
diff --git a/projects/common/src/telemetry/user-telemetry.module.ts b/projects/common/src/telemetry/user-telemetry.module.ts
index f35d2094..e0e8ca18 100644
--- a/projects/common/src/telemetry/user-telemetry.module.ts
+++ b/projects/common/src/telemetry/user-telemetry.module.ts
@@ -1,6 +1,7 @@
-import { Inject, ModuleWithProviders, NgModule, InjectionToken } from '@angular/core';
+import { Inject, InjectionToken, ModuleWithProviders, NgModule } from '@angular/core';
 import { UserTelemetryRegistrationConfig } from './telemetry';
 import { UserTelemetryHelperService } from './user-telemetry-helper.service';
+import { UserTelemetryService } from './user-telemetry.service';
 
 @NgModule()
 export class UserTelemetryModule {
@@ -20,6 +21,10 @@ export class UserTelemetryModule {
         {
           provide: USER_TELEMETRY_PROVIDER_TOKENS,
           useValue: providerConfigs
+        },
+        {
+          provide: UserTelemetryService,
+          useExisting: UserTelemetryHelperService
         }
       ]
     };
diff --git a/projects/common/src/telemetry/user-telemetry.service.ts b/projects/common/src/telemetry/user-telemetry.service.ts
index d5adf7aa..4ae7cb1a 100644
--- a/projects/common/src/telemetry/user-telemetry.service.ts
+++ b/projects/common/src/telemetry/user-telemetry.service.ts
@@ -1,17 +1,7 @@
-import { Injectable } from '@angular/core';
 import { UserTraits } from './telemetry';
-import { UserTelemetryHelperService } from './user-telemetry-helper.service';
 
-@Injectable({ providedIn: 'root' })
-export class UserTelemetryService {
-  public constructor(private readonly userTelemetryHelperService: UserTelemetryHelperService) {}
+export abstract class UserTelemetryService {
+  public abstract identify(userTraits: UserTraits): void;
 
-  public initialize(userTraits: UserTraits): void {
-    this.userTelemetryHelperService.initialize();
-    this.userTelemetryHelperService.identify(userTraits);
-  }
-
-  public shutdown(): void {
-    this.userTelemetryHelperService.shutdown();
-  }
+  public abstract shutdown(): void;
 }

```

- I didn't do any renames, but don't love the name helper. It's independent of the discussion and not exported, so can always be changed later (but I'd probably call it InternalUserTelemetryService or something)
- the public api changes in my diff are needed either way to use this change
- there's a slight api change to make the external api align with the internal one. These should match, and if the old external api was how we want it to be used we should change the internal api to match.


So I'll go ahead and approve since I know there are more PRs following where we can address these things.

aaron-steinfeld (approved):

anandtiwary:
hmm.... so `useExisting` would inject it from the root. That makes sense. I was using useClass and for that I had to expose register as an abstract method too.

anandtiwary:
I will take care of this in the follow up.

PR #1127, opened 2021-09-15
feat: adding error handler and track directive

anandtiwary:
## Description
feat: adding error handler and track directive

### Testing

### Checklist:

### Documentation


projects/common/src/telemetry/track/track.directive.ts
```diff
@@ -0,0 +1,60 @@
...
+      this.trackedEventLabel = this.label ?? (this.host.nativeElement as HTMLElement)?.tagName;
+    }
+  }
+
+  public ngOnDestroy(): void {
+    this.clearListeners();
+  }
+
+  private setupListeners(): void {
+    this.clearListeners();
```
aaron-steinfeld, line 40: I can't remember, does a subscription allow new subscriptions to be added after unsubscribe is called?

projects/common/src/telemetry/track/track.directive.ts
```diff
@@ -0,0 +1,60 @@
...
+    private readonly userTelemetryImplService: UserTelemetryImplService
+  ) {}
+
+  public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (changes.userEvents) {
+      this.setupListeners();
+    }
+
+    if (changes.label) {
+      this.trackedEventLabel = this.label ?? (this.host.nativeElement as HTMLElement)?.tagName;
```
aaron-steinfeld, line 31: tag name doesnt seem very useful, maybe we should require label? Could make that the main arg and the event name the addtl arg since that's easier to default.

projects/common/src/telemetry/telemetry.ts
```diff
@@ -28,3 +28,8 @@ export interface UserTraits extends Dictionary<unknown> {
   name?: string;
   displayName?: string;
 }
+
+export const enum TrackUserEventsType {
+  Click = 'click',
```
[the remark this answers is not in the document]
anandtiwary: It makes sense to keep it flexible (string), but I want to expand this list after I test their behavior, how many events they generate and any perf impact. From Product/Marketing standpoint, the 4 event types we discussed previously are important. Events like MouseOver, MouseOut are not useful.

projects/common/src/telemetry/error-handler/telemetry-global-error-handler.ts
```diff
@@ -0,0 +1,26 @@
+import { LocationStrategy, PathLocationStrategy } from '@angular/common';
+import { ErrorHandler, Injectable, Injector } from '@angular/core';
+import { UserTelemetryImplService } from '../user-telemetry-impl.service';
+
+@Injectable()
```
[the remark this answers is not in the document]
anandtiwary: No, I need to provide it in the module. I missed it in this PR

projects/common/src/telemetry/track/track.directive.ts
anandtiwary: Yeah, it would immediately unsubscribe from it. Fixing now.

projects/common/src/telemetry/track/track.directive.ts
anandtiwary: So, the autotrack based telemetry tools extract certain fields from the element. I am yet to figure out the relevant ones. After I do that, I will add those fields here. 

Imo htTrack= "events" is more readable.  You can always miss providing a directive input and angular won't complain. So let's keep the default logic. I will use a better field than just <tagname> with a later PR.

projects/common/src/telemetry/track/track.directive.ts
```diff
@@ -48,12 +54,12 @@ export class TrackDirective implements OnChanges, OnDestroy {
...
   private clearListeners(): void {
-    this.subscription.unsubscribe();
+    this.activeSubscriptions.unsubscribe();
   }
 
-  private trackUserEvent(userEvent: TrackUserEventsType, eventObj: MouseEvent): void {
+  private trackUserEvent(userEvent: string, eventObj: MouseEvent): void {
     this.userTelemetryImplService.trackEvent(`${userEvent}: ${this.trackedEventLabel}`, {
-      target: eventObj.target,
+      ...(eventObj.target as HTMLElement),
```
aaron-steinfeld, line 62: why are you spreading the whole element onto the event?

anandtiwary: I wasn't sure what all fields would be interesting. Anyway, I have added few properties only now.

projects/common/src/telemetry/track/track.directive.ts
```diff
@@ -58,8 +58,12 @@ export class TrackDirective implements OnInit, OnChanges, OnDestroy {
   }
 
   private trackUserEvent(userEvent: string, eventObj: MouseEvent): void {
+    const targetElement = eventObj.target as HTMLElement;
     this.userTelemetryImplService.trackEvent(`${userEvent}: ${this.trackedEventLabel}`, {
-      ...(eventObj.target as HTMLElement),
+      tagName: targetElement.tagName,
+      innerHTML: targetElement.innerHTML,
```
aaron-steinfeld, line 64: This still feels like over-reporting (and potentially can report PII data). Why do we need the innerHTML of a clicked (for example) element? Can we start minimal and add stuff as use cases come up?

anandtiwary: I can remove innerHtml. But btw, you need to put controls in your telemetry tool to protect against storing pii. It shouldn't be completely dependent on the code

anandtiwary: done

aaron-steinfeld (approved):

PR #1129, opened 2021-09-15
feat: using telemetry tracker in few atomic components

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation


projects/common/src/telemetry/user-telemetry-impl.service.ts
```diff
@@ -9,7 +9,7 @@ import { UserTelemetryService } from './user-telemetry.service';
 export class UserTelemetryImplService extends UserTelemetryService {
   private telemetryProviders: UserTelemetryInternalConfig[] = [];
 
-  public constructor(private readonly injector: Injector, private readonly router: Router) {
+  public constructor(private readonly injector: Injector, @Optional() private readonly router?: Router) {
```
anandtiwary, line 12: So, if I don't make Router optional then I get test errors wherever the atomic component (that uses htTrack) is used. Alternatively, I will have to provide Router Mock service everywhere.

aaron-steinfeld: But shouldn't we be mocking the component in those tests? Why are they use the real track component?

projects/components/src/breadcrumbs/breadcrumbs.component.test.ts
```diff
@@ -13,7 +15,8 @@ describe('BreadcrumbsComponent', () => {
   const createHost = createHostFactory({
     declareComponent: false,
     component: BreadcrumbsComponent,
-    imports: [BreadcrumbsModule, HttpClientTestingModule, IconLibraryTestingModule],
+    imports: [BreadcrumbsModule, HttpClientTestingModule, IconLibraryTestingModule, RouterTestingModule],
+    declarations: [MockDirective(TrackDirective)],
```
aaron-steinfeld, line 19: So here's an example. The mocked track component should be sufficient, we shouldn't need the router testing module too (and in general, this test needs to be converted be shallow)

projects/components/src/button/button.component.test.ts
```diff
@@ -12,6 +14,7 @@ describe('Button Component', () => {
   const createHost = createHostFactory({
     component: ButtonComponent,
     imports: [ButtonModule, RouterTestingModule],
+    declarations: [MockDirective(TrackDirective)],
```
aaron-steinfeld, line 17: Likewise, should be shallow. So deep tests are forced to mock it, but shallow tests (which all of them need to eventually be migrated to) won't require any changes.

projects/components/src/button/button.component.ts
```diff
@@ -10,7 +10,7 @@ import { ButtonRole, ButtonSize, ButtonStyle } from './button';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <ht-event-blocker event="click" class="button-container" [enabled]="this.disabled">
-      <button class="button" [ngClass]="this.getStyleClasses()">
+      <button class="button" [ngClass]="this.getStyleClasses()" [htTrack] [htTrackLabel]="this.label">
```
aaron-steinfeld, line 13: do we capture something like the route path or url here? Seems like it's worth not just tracking "Submit" for example, but where that submit is.

projects/components/src/breadcrumbs/breadcrumbs.component.test.ts
anandtiwary: So, even the MockDirective still tries to inject Router since it is in the constructor.

projects/components/src/button/button.component.test.ts
anandtiwary: true. But that would be a bigger change :(

projects/components/src/button/button.component.ts
anandtiwary: Yes. URL change is tracked via router events and it can be enabled via `enablePageTracking` option

aaron-steinfeld: but those are separate events. If I'm looking at a click event for a button labeled "submit" how do I know what page it's on? Or do the tools provide that by tracking the event timeline?

projects/components/src/button/button.component.test.ts
aaron-steinfeld: yeah, not asking to change it here, but if it's just a question of adding a mock for the small (hopefully, but maybe not?) number of remaining deep tests, seems better. If that's a big scope increase, can do that later and keep it optional.

projects/components/src/breadcrumbs/breadcrumbs.component.test.ts
aaron-steinfeld: really? that seems off...

projects/components/src/button/button.component.ts
anandtiwary: These are the things we can capture and add to the "eventData" object from inside the telemetryImpl service. But AFAIU, the telemetry tools will timestamp these anyway, so we can connect the dots there. Let's iterate over this.

projects/components/src/button/button.component.test.ts
anandtiwary: okay. This rework had come up in the past too. Let's take this as a separate effort. I will keep it optional for now.

projects/components/src/breadcrumbs/breadcrumbs.component.test.ts
anandtiwary: Yep, they try to inject the services. I have seen it at few other places.

aaron-steinfeld (approved):

PR #1147, opened 2021-09-23
feat: make page header configurable to persist active tab

jake-bassett:
## Description
This allows consumers of the page header to provide a unique persitanceId that will turn on saving the active tab to browser storage.

### DIFF
--- a/projects/components/src/header/page/page-header.component.ts
+++ b/projects/components/src/header/page/page-header.component.ts
@@ -1,7 +1,7 @@
-import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
-import { Breadcrumb } from '@hypertrace/common';
-import { Observable } from 'rxjs';
-import { map } from 'rxjs/operators';
+import { ChangeDetectionStrategy, Component, Input, OnInit } from '@angular/core';
+import { Breadcrumb, isNonEmptyString, NavigationService, PreferenceService } from '@hypertrace/common';
+import { Observable, of } from 'rxjs';
+import { first, map } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


projects/components/src/header/page/page-header.component.ts
```diff
@@ -62,5 +65,42 @@ export class PageHeaderComponent {
...
 
-  public constructor(protected readonly breadcrumbsService: BreadcrumbsService) {}
+  public constructor(
+    protected readonly navigationService: NavigationService,
+    protected readonly preferenceService: PreferenceService,
+    protected readonly breadcrumbsService: BreadcrumbsService
+  ) {}
+
+  public ngOnInit(): void {
+    this.getPreferences().subscribe(preferences => {
```
anandtiwary, line 75: Do we need to unsubscribe this?

jake-bassett: Fixed

projects/components/src/tabs/navigable/navigable-tab-group.component.ts
```diff
@@ -57,7 +68,8 @@ export class NavigableTabGroupComponent implements AfterContentInit {
   public ngAfterContentInit(): void {
     this.activeTab$ = merge(this.navigationService.navigation$, this.tabs.changes).pipe(
       startWith(undefined),
-      map(() => this.findActiveTab())
+      map(() => this.findActiveTab()),
+      tap(activeTab => this.tabChange.emit(activeTab?.path))
```
anandtiwary, line 72: I wonder if this would emit duplicate events. May be we can put `distinctUntilChanged` operator before the tap.

anandtiwary (approved):

PR #1149, opened 2021-09-27
feat: drilldown selection feature added  to cartesian chart

jyothishjose6190:
brush added to cartesian chart
on a feature**: brush added to cartesian chart

### DIFF
--- a/projects/components/src/input/input.component.ts
+++ b/projects/components/src/input/input.component.ts
@@ -86,6 +86,7 @@ export class InputComponent<T extends string | number> implements ControlValueAc
   public registerOnTouched(onTouch: (value: T | undefined) => void): void {
     this.propagateControlValueChangeOnTouch = onTouch;
   }
+
   public setDisabledState(isDisabled?: boolean): void {
     this.disabled = isDisabled ?? false;
   }
--- a/projects/components/src/select/select.component.ts
+++ b/projects/components/src/select/select.component.ts
@@ -324,6 +324,10 @@ export class SelectComponent<V> implements ControlValueAccessor, AfterContentIni
     this.propagateControlValueChangeOnTouch = onTouch;
   }
 
+  public setDisabledState(isDisabled?: boolean): void {
+    this.disabled = isDisabled ?? false;
+  }
+
   private propagateValueChangeToFormControl(value: V | undefined): void {
     this.propagateControlValueChange?.(value);
     this.propagateControlValueChangeOnTouch?.(value);


projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -64,6 +65,9 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
   @Output()
   public readonly selectedIntervalChange: EventEmitter<IntervalValue> = new EventEmitter();
 
+  @Output()
+  public readonly selecteRangeChange: EventEmitter<any> = new EventEmitter();
```
anandtiwary, line 69: nit: SelectionChange

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -65,14 +68,35 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
   public constructor(
     protected readonly hostElement: Element,
     protected readonly injector: Injector,
     protected readonly renderingStrategy: RenderingStrategy,
     protected readonly svgUtilService: SvgUtilService,
     protected readonly d3Utils: D3UtilService,
     protected readonly domRenderer: Renderer2
-  ) {}
+  ) {
+    this.brushBehaviour = brushX<unknown>().on('end', () => this.onBrushSelection(_d3CurrentEvent));
```
anandtiwary, line 81: Please only set this up if the interaction handler is provided by parent component.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -65,14 +68,35 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
+
+    const { height } = this.hostElement.getBoundingClientRect();
+
+    const [startX, endX] = event.selection as [number, number];
+
+    let fromdata: any = this.allSeriesData.flatMap(viz => viz.dataForLocation({ x: startX, y: height }));
+    let todata: any = this.allSeriesData.flatMap(viz => viz.dataForLocation({ x: endX, y: height }));
+
+    this.eventListeners.forEach(listener => {
+      listener.onEvent({ start: fromdata[0].dataPoint.timestamp, end: todata[0].dataPoint.timestamp } as any);
```
anandtiwary, line 97: This depends on the api being discussed above.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -281,6 +305,20 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     this.moveDataOnTopOfAxes();
     this.drawMouseEventContainer();
     this.setupEventListeners();
+    this.attachBrush();
```
anandtiwary, line 308: Should be config driver as mentioned above.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -54,6 +55,10 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
     this.updateDataObservable();
   }
 
+  public onRangeChange(data: any): void {
```
anandtiwary, line 58: `onSelectionChange()` and type the data

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-navigation-handler.model.ts
```diff
@@ -0,0 +1,42 @@
...
+  type: 'cartesian-explorer-navigation-handler'
+})
+export class CartesianExplorerNavigationHandlerModel implements InteractionHandler {
+  @ModelInject(ExplorerService)
+  private readonly explorerService!: ExplorerService;
+
+  @ModelInject(NavigationService)
+  private readonly navigationService!: NavigationService;
+
+  public execute(dateRange: any): Observable<void> {
```
anandtiwary, line 20: this can be typed

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-navigation-handler.model.ts
```diff
@@ -0,0 +1,42 @@
...
+    const startDate = new Date(dateRange.start).getTime() - 6000;
```
anandtiwary, line 21: why the 6000 buffer?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
anandtiwary: also could be part of setupEventListerners

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-navigation-handler.model.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ModelInject } from '@hypertrace/hyperdash-angular';
+import { Observable, of } from 'rxjs';
+import { NavigationService } from '../../../../../../../../common/src/navigation/navigation.service';
+import { FilterOperator } from '../../../../../../../../components/src/filtering/filter/filter-operators';
+import { ExplorerService } from '../../../../../../pages/explorer/explorer-service';
+import { ScopeQueryParam } from '../../../../../../pages/explorer/explorer.component';
+import { InteractionHandler } from '../../../../interaction/interaction-handler';
+
+@Model({
+  type: 'cartesian-explorer-navigation-handler'
```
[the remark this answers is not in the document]
anandtiwary: or `cartesian-explorer-selection-handler`

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.ts
```diff
@@ -0,0 +1,42 @@
+import { ChangeDetectionStrategy, Component, EventEmitter, Input, Output } from '@angular/core';
+import { ButtonStyle } from '@hypertrace/components';
+
+@Component({
+  selector: 'ht-context-menu',
```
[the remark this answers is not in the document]
jyothishjose6190: done

projects/observability/src/shared/components/cartesian/cartesian-chart.module.ts
```diff
@@ -18,7 +19,8 @@ import { CartesianSummaryComponent } from './d3/legend/cartesian-summary.compone
     SelectModule,
     LayoutChangeModule,
     IntervalSelectModule,
-    ChartTooltipModule
+    ChartTooltipModule,
+    CartesianExplorerContextMenuModule
```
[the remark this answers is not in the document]
jyothishjose6190: done

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  private navigateToExplorer(start: Date, end: Date): void {
+    const params = this.timeRangeService.toQueryParams(start, end);
+
+    this.navigationService.navigate({
+      navType: NavigationParamsType.InApp,
+      path: ['/explorer'],
+      queryParams: params,
+      queryParamsHandling: 'merge',
+      replaceCurrentHistory: true
```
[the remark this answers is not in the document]
jyothishjose6190: uses custom time range  to filter data between the selected time ranges

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,45 @@
...
+  @ModelInject(NavigationService)
+  private readonly navigationService!: NavigationService;
+
+  // tslint:disable-next-line
+  public execute(selectionData: any): Observable<void> {
+    const startPoint = selectionData[0];
+    const endPoint = selectionData[1];
+
+    // When there is only one point in selected area, start timestamp  and end timestamp is same. This causes timerange exception
+    const startTime = new Date(startPoint.dataPoint.timestamp).getTime() - 6000;
```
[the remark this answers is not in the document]
jyothishjose6190: done

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  @ModelInject(NavigationService)
+  private readonly navigationService!: NavigationService;
+
+  // tslint:disable-next-line
+  public execute(selectionData: any): Observable<void> {
+    const startPoint = selectionData[0];
+    const endPoint = selectionData[1];
+
+    // When there is only one point in selected area, start timestamp  and end timestamp is same. This causes timerange exception
```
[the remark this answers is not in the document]
jyothishjose6190: done

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,45 @@
...
+export class CartesianExplorerSelectionHandlerModel implements InteractionHandler {
+  @ModelInject(TimeRangeService)
+  private readonly timeRangeService!: TimeRangeService;
+
+  @ModelInject(NavigationService)
+  private readonly navigationService!: NavigationService;
+
+  // tslint:disable-next-line
+  public execute(selectionData: any): Observable<void> {
+    const startPoint = selectionData[0];
```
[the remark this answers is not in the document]
jyothishjose6190: chnaged to timerange

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,45 @@
...
+})
+export class CartesianExplorerSelectionHandlerModel implements InteractionHandler {
+  @ModelInject(TimeRangeService)
+  private readonly timeRangeService!: TimeRangeService;
+
+  @ModelInject(NavigationService)
+  private readonly navigationService!: NavigationService;
+
+  // tslint:disable-next-line
+  public execute(selectionData: any): Observable<void> {
```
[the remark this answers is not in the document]
jyothishjose6190: chnaged

projects/common/src/time/time-range.service.ts
```diff
@@ -110,4 +110,13 @@ export class TimeRangeService {
   public static toFixedTimeRange(startTime: Date, endTime: Date): FixedTimeRange {
     return new FixedTimeRange(startTime, endTime);
   }
+
+  public toQueryParams(startTime: Date, endTime: Date): QueryParamObject {
```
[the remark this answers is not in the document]
jyothishjose6190: emitting on the subject needed to change the custom time range at the top

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -64,17 +75,37 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
...
+      icon: IconType.ArrowUpRight
+    }
+  ];
+
   public constructor(
     private readonly chartBuilderService: ChartBuilderService,
     private readonly chartTooltipBuilderService: ChartTooltipBuilderService,
-    private readonly renderer: Renderer2
+    private readonly renderer: Renderer2,
+    private readonly popoverService: PopoverService
```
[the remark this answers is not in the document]
jyothishjose6190: moved to cartesian widget renderer. now this component emits only data

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -1,6 +1,8 @@
 import { Injector, Renderer2 } from '@angular/core';
 import { TimeRange } from '@hypertrace/common';
-import { ContainerElement, mouse, select } from 'd3-selection';
+import { brush, BrushBehavior, D3BrushEvent } from 'd3-brush';
+// tslint:disable-next-line: no-restricted-globals weird tslint error. Rename event so we can type it and not mistake it for other events
```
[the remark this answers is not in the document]
jyothishjose6190: removed

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
jyothishjose6190: done

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -93,7 +97,10 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
         this.chartTooltipBuilderService.constructTooltip<TData, Series<TData>>(data =>
           this.convertToDefaultTooltipRenderData(data)
         )
-      );
+      )
+      .withEventListener(ChartEvent.Select, data => {
+        this.selecteRangeChange.emit(data);
```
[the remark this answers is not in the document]
jyothishjose6190: done

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
jyothishjose6190: done

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
jyothishjose6190: done

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
jyothishjose6190: done

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -114,6 +156,25 @@ export class CartesianWidgetRendererComponent<TSeriesInterval> extends Interacti
 
     return match || 'AUTO';
   }
+
+  private showContextMenu(): void {
+    this.popover = this.popoverService.drawPopover({
+      componentOrTemplate: this.contextMenuTemplate,
```
anandtiwary, line 162: You can pass the component here too. So, in the interaction handler, just pass a new component that has above  content

anandtiwary:
@jyothishjose6190 when can we incorporate the remaining review comments?

jyothishjose6190:
> @jyothishjose6190 when can we incorporate the remaining review comments?

I will incorporate them today.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -93,7 +98,10 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
         this.chartTooltipBuilderService.constructTooltip<TData, Series<TData>>(data =>
           this.convertToDefaultTooltipRenderData(data)
         )
-      );
+      )
+      .withEventListener(ChartEvent.Select, selectedData => {
+        this.selectionChange.emit(selectedData as CartesianSelectedData<TData>);
```
anandtiwary, line 103: Is this cast required? Can we type it so that `selectedData` come as         `CartesianSelectedData<TData>` ?

anandtiwary: may be we can do `if(data instanceOf CartesianSelectedData<TData>)` ?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -74,6 +76,38 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
+      return;
+    }
+
+    this.eventListeners.forEach(listener => {
+      if (listener.event === ChartEvent.Select) {
+        const { height } = this.mouseEventContainer!.getBoundingClientRect();
+
+        const [startPoint, endPoint] = event.selection as [number, number];
+
+        const startDate = this.allSeriesData[0].getXAxisValue(startPoint);
```
anandtiwary, line 90: What does `this.allSeriesData[0]` represent?  first series?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -256,11 +290,15 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
     eventContainer.on('mousemove', () => this.onMouseMove()).on('mouseleave', () => this.onMouseLeave());
 
-    this.eventListeners.forEach(listener =>
-      eventContainer.on(this.getNativeEventName(listener.event), () =>
-        listener.onEvent(this.getMouseDataForCurrentEvent())
-      )
-    );
+    this.eventListeners.forEach(listener => {
+      if (listener.event === ChartEvent.Select) {
+        this.attachBrush();
```
anandtiwary, line 295: We should ensure this code doesn't attach brush more than once

projects/observability/src/shared/components/cartesian/d3/scale/band/cartesian-band-scale.ts
```diff
@@ -40,4 +40,8 @@ export class CartesianBandScale<TData> extends CartesianScale<TData, string> {
   protected getEmptyScale(): ScaleBand<string> {
     return scaleBand().paddingInner(0.2).paddingOuter(0.1).align(1);
   }
+
+  public invert(x: number): number {
```
anandtiwary, line 44: this may need some changes.

anandtiwary: but not sure. Will check

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,63 @@
...
+  private readonly popoverService!: PopoverService;
+
+  public popover?: PopoverRef;
+
+  @ModelProperty({
+    key: 'show-context-menu',
+    displayName: 'Show Context Menu',
+    type: BOOLEAN_PROPERTY.type
+  })
+  public isContextMenuVisible: boolean = true;
```
anandtiwary, line 34: nit: Since this is set to true by default, we don't need to set it to true again in model jsons

anandtiwary: probably call it `showContextMenu`. `isContextMenuVisible` makes it sound like whether the context menu is currently visible or not

anandtiwary:
Looks mostly fine. Would like @aaron-steinfeld @itssharmasandeep to look into this as well.

jaywalker21:
@itssharmasandeep - Can you have a look at it whenever feasible?
@anandtiwary - Just wanted to confirm that the PR is good to be merged with `-0.46%` diff in coverage.

itssharmasandeep:
> @itssharmasandeep - Can you have a look at the earliest?

@jyothishjose6190 is making some changes, once he's done, I will take a look.

jaywalker21:
@jyothishjose6190 Where do we stand on this?
Can we close this if all is good?

anandtiwary:
LGTM

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
anandtiwary: @jyothishjose6190 looks like you missed this

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
anandtiwary: @jyothishjose6190

projects/observability/src/shared/components/cartesian/d3/scale/band/cartesian-band-scale.ts
anandtiwary: Band scale would return a string right? `CartesianBandScale<TData> extends CartesianScale<TData, string>`

so the domain is TData and range is a string. That means, the parameter for invert would be 'string', and return type would be TData.

anandtiwary (commented):
apart from few minor outstanding issues, this is good to go. @aaron-steinfeld @itssharmasandeep your thoughts?

aaron-steinfeld:
> apart from few minor outstanding issues, this is good to go. @aaron-steinfeld @itssharmasandeep your thoughts?

My thoughts haven't changed since the beginning - I don't think this is the right UX for the use case. A brush selection on the cartesian should change the time range, like it does in most products. If we want to support drilling into explorer, that's independent of a brush selection.

I do recognize that this has been open for months at this point, but I've probably brought this up a half dozen times in various contexts at this point.

projects/common/src/time/time-range.service.ts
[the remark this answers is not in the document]
aaron-steinfeld: This needs to be fixed, converting two values to query params should not have the side effect of changing the time range.

anandtiwary:
@aaron-steinfeld I agree with the concern here, but this ux had been discussed in the past with various stakeholders and we decided to go ahead with this approach for **hypertrace**. We will surely iterate over this

aaron-steinfeld:
> @aaron-steinfeld I agree with the concern here, but this ux had been discussed in the past with various stakeholders and we decided to go ahead with this approach for **hypertrace**. We will surely iterate over this

Discussed this with @anandtiwary and came to a good compromise. For context, my concern is for Hypertrace in that I don't think in its current form we solve either ticket (if there's any others that I missed, please clarify) :
- https://github.com/hypertrace/hypertrace/issues/220 - where the ask is to apply the global time range _in context_ like grafana, as I was describing above. In the current PR, we lose context because the only way to set the global time range on brush selection also navigtes.
- https://github.com/hypertrace/hypertrace/issues/208 - where the ask is to correlate traces corresponding to a particular chart. Here, the problem is we're not bringing the context with us when we navigate to explorer, we're just setting the global time range and going to explorer with a blank slate.

As a compromise that both maintains the existing work and helps better align to the use cases with minimal extra effort, my proposal is to add a "Set Time Range" option to the context menu as the first option before "Explore", whose affect is to just set the time range without navigating. With that https://github.com/hypertrace/hypertrace/issues/220 is solved I believe. As to https://github.com/hypertrace/hypertrace/issues/208 , we can keep iterating on that in later PRs to bring more context with us when clicking Explore to accomplish the described use case.

jyothishjose6190:
> > @aaron-steinfeld I agree with the concern here, but this ux had been discussed in the past with various stakeholders and we decided to go ahead with this approach for **hypertrace**. We will surely iterate over this
> 
> Discussed this with @anandtiwary and came to a good compromise. For context, my concern is for Hypertrace in that I don't think in its current form we solve either ticket (if there's any others that I missed, please clarify) :
> 
> * [[Feature request] Ability to drill down to a time range from aggregations hypertrace#220](https://github.com/hypertrace/hypertrace/issues/220) - where the ask is to apply the global time range _in context_ like grafana, as I was describing above. In the current PR, we lose context because the only way to set the global time range on brush selection also navigtes.
> * [Correlate Service and API dashboard metrics with related Traces hypertrace#208](https://github.com/hypertrace/hypertrace/issues/208) - where the ask is to correlate traces corresponding to a particular chart. Here, the problem is we're not bringing the context with us when we navigate to explorer, we're just setting the global time range and going to explorer with a blank slate.
> 
> As a compromise that both maintains the existing work and helps better align to the use cases with minimal extra effort, my proposal is to add a "Set Time Range" option to the context menu as the first option before "Explore", whose affect is to just set the time range without navigating. With that [hypertrace/hypertrace#220](https://github.com/hypertrace/hypertrace/issues/220) is solved I believe. As to [hypertrace/hypertrace#208](https://github.com/hypertrace/hypertrace/issues/208) , we can keep iterating on that in later PRs to bring more context with us when clicking Explore to accomplish the described use case.

@anandtiwary @aaron-steinfeld made changes as discussed above. please review

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.ts
```diff
@@ -38,16 +43,21 @@ export class CartesianExplorerContextMenuComponent<TData> {
...
-      this.selectionData.timeRange.startTime,
-      this.selectionData.timeRange.endTime
-    );
+    if (_menu.name === 'Explore') {
+      this.cartesainExplorerNavigationService.navigateToExplorer(
+        this.selectionData.timeRange.startTime,
+        this.selectionData.timeRange.endTime
+      );
+    } else {
+      this.timeRangeService.toQueryParams(this.selectionData.timeRange.startTime, this.selectionData.timeRange.endTime);
```
aaron-steinfeld, line 59: This should be calling set time range. There's an existing thread asking to fix the new method `toQueryParams` to **not** modify the global time range, it should only be producing a param object.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.ts
```diff
@@ -38,16 +43,21 @@ export class CartesianExplorerContextMenuComponent<TData> {
...
   ) {
     this.selectionData = data;
   }
 
   public menuSelectHandler = (_menu: ContextMenu): void => {
-    this.cartesainExplorerNavigationService.navigateToExplorer(
-      this.selectionData.timeRange.startTime,
-      this.selectionData.timeRange.endTime
-    );
+    this.timeRangeService.setFixedRange(this.selectionData.timeRange.startTime, this.selectionData.timeRange.endTime);
```
aaron-steinfeld, line 53: This shouldn't need to be set here, the time range is passed to the navigateToExplorer method and set as query params on https://github.com/hypertrace/hypertrace-ui/pull/1149/files#diff-707e0c6ee1d2d1e2a7f5365ff8a899673d103e05e8e8c619b5f3bbc1de0382f9R19

jaywalker21: @aaron-steinfeld I think there are two use cases.
1. You want to select a fraction of time and apply the same on the dashboard page. (You choose the menu option `Apply Time Range`)
2. You want to apply the time range and navigate to explorer dashboard. (you use `Explore` menu option).

So, we need to set the time range for case `1` here. For `2`, you are correct that the `navigateToExplorer` method handles setting the time range via `queryParams`, as we are not calling `navigateToExplorer` for menu option `1`.

aaron-steinfeld: Right, I think this belies the problem in the design though - every context menu item calls the same handler. We want this called in case 1 above, but not in case 2 where the navigation params for time are set directly.

At a minimum, we can change this method to switch on the calling item, but I think a significantly better solution and no real extra work would be to add a handler to the `ContextMenu` interface and change line 20 to `            (click)="menu.onClick()"`.

jaywalker21: So, to add the respective handler for each item within the `menu` object definition itself, right?
Like the `ContextMenu` interface will be having one more property called `onClick` which will be the reference of the function to be executed whenever the respective item is clicked.

If that's the case, I think it makes sense and will make it generic enough.

jaywalker21: Ahh, thanks for the clarification, I might have put up a reference like `this.someMethod` as you said. Will keep this in mind.

jyothishjose6190: @aaron-steinfeld when we navigate to explorer without setFixedRange, the global time range filter is not changing. see in the below screenshot it is showing 15 mins, on the other hand, URL shows with correct parameters
Screenshot 2022-01-06 at 9 25 15 AM

aaron-steinfeld: Hmm, that may be an outstanding bug. Guessing the time range service doesn't keep track of the URL after it initializes, which it should be. Do you mind filing it as a separate issue? For now, we can just separate the handlers and add the extra explicit set in the explorer nav handler to get this moving, then come back to fix the url handling and remove.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
```diff
@@ -0,0 +1,131 @@
+import { ChangeDetectionStrategy, Component, Inject, Injector, Optional } from '@angular/core';
+import { IconType } from '@hypertrace/assets-library';
+import { TimeRangeService } from '@hypertrace/common';
+import { ButtonComponent, DividerComponent, POPOVER_DATA } from '@hypertrace/components';
+import { createHostFactory, mockProvider, Spectator } from '@ngneat/spectator/jest';
+import { MockComponent } from 'ng-mocks';
+import { CartesianSelectedData, CartesianSeriesVisualizationType } from '../../../../../../../public-api';
```
aaron-steinfeld, line 7: bad import

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
```diff
@@ -0,0 +1,131 @@
...
+import { CartesainExplorerNavigationService } from '../cartesian-explorer-navigation.service';
+import { CartesianExplorerContextMenuComponent, ContextMenu } from './cartesian-explorer-context-menu.component';
+
+@Component({
+  // tslint:disable-next-line:component-selector
+  selector: 'test-context-menu-content',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<div class="test-modal-content">Test Component Content</div> `
+})
+class TestComponent {
```
aaron-steinfeld, line 17: This shouldn't be required here, doesn't look to be used.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  public constructor(
+    public readonly injector: Injector,
+    @Optional() @Inject(POPOVER_DATA) public readonly data: unknown
+  ) {}
+}
+
+describe('Sheet Overlay component', () => {
```
aaron-steinfeld, line 24: Describe text is describing a different component

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
```diff
@@ -0,0 +1,131 @@
...
+          hide: false
+        },
+        location: { x: 138, y: 82.58120000000001 }
+      }
+    ],
+    location: { x: 452, y: 763 }
+  };
+  let spectator: Spectator<CartesianExplorerContextMenuComponent<unknown>>;
+
+  const createHost = createHostFactory({
```
aaron-steinfeld, line 70: nit: you can use a host factory, but it's overkill for just testing a component without inputs. Component factory would work with less extra config.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    ]
+  });
+
+  const createConfiguredHost = ({}) =>
+    createHost(undefined, {
+      providers: [
+        {
+          provide: POPOVER_DATA,
+          deps: [Injector],
+          useFactory: (injector: Injector) => ({
```
aaron-steinfeld, line 94: not quite sure what this factory is meant to do, it looks like it produces an injector that holds POPOVER_DATA. Guessing this was copied from the popover container test, not necessary here. just use `mockProvider(POPOVER_DATA, {/*your mock */})`

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      onClick: () => {
+        spectator
+          .inject(CartesainExplorerNavigationService)
+          .navigateToExplorer(selectedData.timeRange.startTime, selectedData.timeRange.endTime);
+      }
+    };
+
+    menu.onClick();
+
+    expect(spectator.inject(CartesainExplorerNavigationService).navigateToExplorer).toHaveBeenCalled();
```
aaron-steinfeld, line 129: this test doesn't actually verify any behavior, because it creates a function then calls it - everything being verified is test code. It should be instantiating the component, then clicking in the dom and verifying the function has been called.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.test.ts
```diff
@@ -207,4 +254,20 @@ describe('Cartesian widget renderer component', () => {
...
+      series: [series],
+      maxSeriesDataPoints: 20
+    });
+    const spectator = buildComponent({
+      providers: [...mockDashboardWidgetProviders(mockModel)]
+    });
+
+    spectator.component.onSelectionChange(selectedData);
+
+    expect(spectator.component).toBeTruthy();
```
aaron-steinfeld, line 271: we're not verifying what the test says its verifying here.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      this.selectionData.timeRange.endTime
+    );
+  };
+
+  public readonly setTimeRangeHandler = () => {
+    this.timeRangeService.setFixedRange(this.selectionData.timeRange.startTime, this.selectionData.timeRange.endTime);
+  };
+}
+
+export interface ContextMenu {
```
aaron-steinfeld, line 61: this is a local type, doesn't need to be exported

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,63 @@
...
+  PopoverBackdrop,
+  PopoverFixedPositionLocation,
+  PopoverPositionType,
+  PopoverRef,
+  PopoverService
+} from '@hypertrace/components';
+import { BOOLEAN_PROPERTY, Model, ModelProperty } from '@hypertrace/hyperdash';
+import { ModelInject } from '@hypertrace/hyperdash-angular';
+import { Observable, of } from 'rxjs';
+import { CartesianSelectedData } from '../../../../../../public-api';
```
aaron-steinfeld, line 11: bad import. Please go through all changes and verify that no imports are either from any `public-api` file, importing the project they reside in `@hypertrace/observability`, in this case, or importing relative paths across projects.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.test.ts
```diff
@@ -0,0 +1,84 @@
+import { TimeRangeService } from '@hypertrace/common';
+import { PopoverService } from '@hypertrace/components';
+import { createModelFactory } from '@hypertrace/dashboards/testing';
+import { mockProvider } from '@ngneat/spectator/jest';
+import { CartesianSelectedData, CartesianSeriesVisualizationType } from '../../../../../../public-api';
```
aaron-steinfeld, line 5: bad import. looks to be widespread.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.ts
jyothishjose6190: using in test file, ie why exported

aaron-steinfeld: Commented on the test file too, it shouldn't be using it - it's defining what it's trying to test, rather than just interacting with the component.

If for some reason we did need to use it in another file and thus export it, the naming becomes more important because it shows up in autocomplete. In that case, we'd want to change the name to clarify that it's only used in here (e.g. `CartesianHandlerContextMenuIte`) - but again, we shouldn't need to.

aaron-steinfeld (commented):
Unresolved several comments that had been resolved but not addressed as well.

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
jyothishjose6190: Used host factory to inject POPOVER_DATA to the test
Screenshot 2022-01-17 at 8 25 02 AM

jyothishjose6190:
@aaron-steinfeld updated PR with your comments. Please check

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -1,6 +1,9 @@
 import { Injector, Renderer2 } from '@angular/core';
-import { TimeRange } from '@hypertrace/common';
-import { ContainerElement, mouse, select } from 'd3-selection';
+
+import { TimeRange, TimeRangeService } from '@hypertrace/common';
+import { BrushBehavior, brushX, D3BrushEvent } from 'd3-brush';
+// tslint:disable
```
aaron-steinfeld, line 5:
```suggestion
// tslint:disable-next-line: no-restricted-globals
```

Current version turns off tslint for the whole file

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.test.ts
```diff
@@ -14,12 +16,57 @@ import { ModelApi } from '@hypertrace/hyperdash';
 import { runFakeRxjs } from '@hypertrace/test-utils';
 import { createComponentFactory, mockProvider } from '@ngneat/spectator/jest';
 import { of } from 'rxjs';
+import { CartesianSelectedData } from '../../../../../public-api';
```
aaron-steinfeld, line 19: bad import

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
```diff
@@ -1,12 +1,15 @@
 import { ChangeDetectionStrategy, ChangeDetectorRef, Component, Inject } from '@angular/core';
 import { IntervalDurationService, TimeDuration } from '@hypertrace/common';
+
 import { InteractiveDataWidgetRenderer } from '@hypertrace/dashboards';
 import { Renderer } from '@hypertrace/hyperdash';
 import { RendererApi, RENDERER_API } from '@hypertrace/hyperdash-angular';
 import { NEVER, Observable } from 'rxjs';
 import { switchMap, tap } from 'rxjs/operators';
-import { Band, Series } from '../../../../components/cartesian/chart';
+import { CartesianSelectedData } from '../../../../../public-api';
```
aaron-steinfeld, line 9: bad import

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-context-menu/cartesian-explorer-context-menu.component.test.ts
```diff
@@ -0,0 +1,129 @@
+import { StaticProvider } from '@angular/core';
+import { IconType } from '@hypertrace/assets-library';
+import { TimeRangeService } from '@hypertrace/common';
+import { ButtonComponent, DividerComponent, POPOVER_DATA } from '@hypertrace/components';
+import { CartesianSelectedData, CartesianSeriesVisualizationType } from '@hypertrace/observability';
```
aaron-steinfeld, line 5: bad self-import (relative path)

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.test.ts
```diff
@@ -0,0 +1,84 @@
+import { TimeRangeService } from '@hypertrace/common';
+import { PopoverService } from '@hypertrace/components';
+import { createModelFactory } from '@hypertrace/dashboards/testing';
+import { CartesianSelectedData, CartesianSeriesVisualizationType } from '@hypertrace/observability';
```
aaron-steinfeld, line 4: bad self-import (relative path)

projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/interactions/cartesian-explorer-selection-handler.model.ts
```diff
@@ -0,0 +1,63 @@
+import {
+  PopoverBackdrop,
+  PopoverFixedPositionLocation,
+  PopoverPositionType,
+  PopoverRef,
+  PopoverService
+} from '@hypertrace/components';
+import { BOOLEAN_PROPERTY, Model, ModelProperty } from '@hypertrace/hyperdash';
+import { ModelInject } from '@hypertrace/hyperdash-angular';
+import { CartesianSelectedData } from '@hypertrace/observability';
```
aaron-steinfeld, line 10: self import

aaron-steinfeld (commented):
lgtm, didn't have a chance to review the bursh and cartesian code though, so relying on @anandtiwary 's review of that - @anandtiwary did you want to take one more pass or good to go?

anandtiwary (approved):

anandtiwary:
LGTM

anandtiwary:
@jyothishjose6190 Could you please merge the latest main in this PR branch?

anandtiwary:
@aaron-steinfeld something went wrong here. 

This commit has all the changes. 
https://github.com/hypertrace/hypertrace-ui/commit/4be37398ba629f2faeb7f12fed44f806bae3f1c2

aaron-steinfeld:
Looks like @jyothishjose6190 didn't merge main but instead force pushed main into his remote branch, overwriting all of the commits. Force pushing is very dangerous for this reason - I strongly discourage anyone from doing it unless you have a really strong understanding of git _and_ are working solo such that rewriting history will not affect anyone else. Hopefully he still has the original commit on his local.

jyothishjose6190:
@anandtiwary @aaron-steinfeld Sorry about this. I will keep this in mind

aaron-steinfeld:
It's your branch (looks like it still exists in the razorpay repo, so it's recoverable), you're welcome to do what you want with it - it's just riskier for the author to use force pushing in terms of data loss because it goes around version control and overwrites history each time.

jyothishjose6190:
new PR created with all the changes https://github.com/hypertrace/hypertrace-ui/pull/1394

PR #1160, opened 2021-09-28
feat: content functionality for select option

itssharmasandeep:


projects/components/src/select/directive/select-option-renderer.directive.test.ts
```diff
@@ -0,0 +1,21 @@
+import { createDirectiveFactory, SpectatorDirective } from '@ngneat/spectator/jest';
+import { SelectOptionRendererDirective } from './select-option-renderer.directive';
+
+describe('Let async directive', () => {
```
aaron-steinfeld, line 4: copy paste

projects/components/src/select/select.component.ts
```diff
@@ -89,37 +97,59 @@ import { SelectSize } from './select-size';
...
-            <ng-container
-              *ngTemplateOutlet="itemsTemplate; context: { items: items, showSelectionStatus: true }"
-            ></ng-container>
-          </div>
-
-          <ng-template #itemsTemplate let-items="items" let-showSelectionStatus="showSelectionStatus">
+            <!-- Select Option Items --->
             <div
-              *ngFor="let item of items"
+              *ngFor="let item of this.items"
```
aaron-steinfeld, line 135: probably could have left the items template and just pulled the item template out to avoid the repeated ngFors, but seems fine functionally.

projects/components/src/select/directive/select-option-renderer.directive.test.ts
itssharmasandeep: Changed

projects/components/src/select/select.component.ts
itssharmasandeep: Updated

aaron-steinfeld (approved):

PR #1163, opened 2021-09-29
fix: running telemetry methods on initialized list only

anandtiwary:
## Description
fix: running telemetry methods on initialized list only

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/common/src/telemetry/user-telemetry-impl.service.ts
+++ b/projects/common/src/telemetry/user-telemetry-impl.service.ts
@@ -7,7 +7,8 @@ import { UserTelemetryService } from './user-telemetry.service';
 
 @Injectable({ providedIn: 'root' })
 export class UserTelemetryImplService extends UserTelemetryService {
-  private telemetryProviders: UserTelemetryInternalConfig[] = [];
+  private allTelemetryProviders: UserTelemetryInternalConfig[] = [];
+  private initializedTelemetryProviders: UserTelemetryInternalConfig[] = [];
 
   public constructor(private readonly injector: Injector, @Optional() private readonly router?: Router) {
     super();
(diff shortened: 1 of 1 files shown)


projects/common/src/telemetry/user-telemetry-impl.service.ts
```diff
@@ -29,31 +30,32 @@ export class UserTelemetryImplService extends UserTelemetryService {
   }
 
   public initialize(): void {
-    this.telemetryProviders.forEach(provider => provider.telemetryProvider.initialize(provider.initConfig));
+    this.allTelemetryProviders.forEach(provider => provider.telemetryProvider.initialize(provider.initConfig));
+    this.initializedTelemetryProviders = [...this.allTelemetryProviders];
```
aaron-steinfeld, line 34: what are we trying to solve here? were we calling before initializing? That's the only behavior that I see that would change here. Wouldn't that imply a misconfig though?

anandtiwary: Currently, we don't 'initialize' telemetry when FF is disabled or when user is an internal user. But the track directive still pushes click events on the service. Our E2E tests fails when it sees any console error. In this case, the third party lib throws an error since we tried to send an 'event' without initializing the library.

anandtiwary: The other alternative is to put a init check in every method that will run every time the track event methods are called. We can avoid that by keeping a separate initialized list

anandtiwary: Still, a perf hit. But an ideal solution would be to load the entire module only if the feature flag is enabled.

aaron-steinfeld (approved):
Discussed offline. Will take this approach for now, in the future we should investigate splitting behavior so that we can differentiate between all of:
- enabled but not ready (error any calls)
- disabled (drop calls)
- enabled (forward calls)

PR #1227, opened 2021-11-01
fix: incorrect focus on editing filter chips

aaron-steinfeld:
## Description
Fixed a couple issues causing bugs in filter bar chip interactions

### Testing
Verified through unit tests and manual

### DIFF
--- a/projects/components/src/filtering/filter-bar/filter-bar.component.ts
+++ b/projects/components/src/filtering/filter-bar/filter-bar.component.ts
@@ -2,7 +2,6 @@ import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
-  ElementRef,
   EventEmitter,
   Input,
   OnChanges,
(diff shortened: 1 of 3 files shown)


projects/components/src/filtering/filter-bar/filter-bar.component.ts
```diff
@@ -168,7 +171,7 @@ export class FilterBarComponent implements OnChanges, OnInit, OnDestroy {
   }
 
   private resetFocus(): void {
-    this.filterInput?.nativeElement.focus();
+    this.filterInput?.focus();
```
aaron-steinfeld, line 174: This was the main change. We don't want to focus on the filter chip, we want to focus on the input inside it. Exposed focus method down the component stack.

projects/components/src/filtering/filter-bar/filter-bar.component.ts
```diff
@@ -123,7 +126,7 @@ export class FilterBarComponent implements OnChanges, OnInit, OnDestroy {
 
   private subscribeToUrlFilterChanges(): void {
     this.subscription = this.attributeSubject$
-      .pipe(mergeMap(attributes => this.filterUrlService.getUrlFiltersChanges$(attributes)))
+      .pipe(switchMap(attributes => this.filterUrlService.getUrlFiltersChanges$(attributes)))
```
aaron-steinfeld, line 129: When attributes come after load, a mergemap led to multiple subscriptions

projects/components/src/filtering/filter-bar/filter-bar.component.ts
```diff
@@ -83,14 +84,16 @@ export class FilterBarComponent implements OnChanges, OnInit, OnDestroy {
...
+  public readonly filterInput?: FilterChipComponent;
 
   public isFocused: boolean = false;
 
   private readonly attributeSubject$: BehaviorSubject<FilterAttribute[]> = new BehaviorSubject<FilterAttribute[]>([]);
   private readonly internalFiltersSubject$: BehaviorSubject<Filter[]> = new BehaviorSubject<Filter[]>([]);
-  public readonly internalFilters$: Observable<Filter[]> = this.internalFiltersSubject$.asObservable();
+  public readonly internalFilters$: Observable<Filter[]> = this.internalFiltersSubject$
+    .asObservable()
+    .pipe(distinctUntilChanged(isEqual));
```
aaron-steinfeld, line 96: Optimization, but sometimes multiple things (such as a url change and a child output) cause a filter change to be picked up, and we want to dedupe.

anandtiwary (approved):

PR #1228, opened 2021-11-01
Fix multi select view update

jake-bassett:
## Description
Fixes a bug that prevents trigger label from updating.

### DIFF
--- a/projects/components/src/multi-select/multi-select.component.ts
+++ b/projects/components/src/multi-select/multi-select.component.ts
@@ -10,7 +10,7 @@ import {
   QueryList
 } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
-import { queryListAndChanges$ } from '@hypertrace/common';
+import { queryListAndChanges$, SubscriptionLifecycle } from '@hypertrace/common';
 import { BehaviorSubject, combineLatest, EMPTY, Observable, Subject } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { ButtonRole, ButtonStyle } from '../button/button';
@@ -24,6 +24,7 @@ import { MultiSelectJustify } from './multi-select-justify';
   selector: 'ht-multi-select',
   styleUrls: ['./multi-select.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
+  providers: [SubscriptionLifecycle],
   template: `
     <div
       class="multi-select"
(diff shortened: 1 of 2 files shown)


projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -241,14 +244,16 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
       return;
     }
 
-    const selectedItems: SelectOptionComponent<V>[] | undefined = this.allOptionsList?.filter(item =>
-      this.isSelectedItem(item)
-    );
+    this.subscriptionLifecycle.add(
```
aaron-steinfeld, line 247: It doesn't make sense to add a new subscription every time we set the trigger label. What's the bug?

jake-bassett: The bug is that the `allOptionsList` is not updating.

aaron-steinfeld: Ever (this would surprise me)? Or that when options update, this logic isn't rerun (this I would expect).

jake-bassett: It's not updating from the call to ngOnChanges().

aaron-steinfeld: Right, it wouldn't - it's not coming in as an input but as content children exposed via a QueryList. That will emit when they change (and that's what allOptions is projected from). The problem is this page is a mix of async and sync behavior - the set trigger behavior is updating regular variables, the search and options are working in observables.

My suggestion would be to represent the selections as observables too. If we keep the hybrid, we're stuck managing subscriptions which is easy to mess up.

jake-bassett: I really don't think it makes sense to refactor this component right now.

anandtiwary: @jake-bassett What property do you want it to change on? Why should options list change when an input changes?

anandtiwary: I could help if you could explain the use case

anandtiwary: AllOptionsList will update as soon as you add/remove any select option component. Are you trying to change the properties of select option component and expecting that it would reflect in multi select?

jake-bassett: Talked a bit offline. Removed the subscription. Please take a look when you get a chance.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -236,22 +237,32 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
...
+    this.triggerValues$ = this.allOptions$?.pipe(
+      map(options => {
+        const selectedItems: SelectOptionComponent<V>[] = options.filter(item => this.isSelectedItem(item));
 
-    this.selectedItemsCount = selectedItems?.length ?? 0;
-
-    // Trigger label is placeholder in case there is element selected on multiselect
-    this.triggerLabel = this.selectedItemsCount === 0 ? this.placeholder : (selectedItems || [])[0]?.label;
+        return {
+          label: selectedItems.length === 0 ? this.placeholder : selectedItems[0]?.label,
```
anandtiwary, line 253: This is multi-select, IMO, this trigger should try to show all selected labels.

anandtiwary: Something like `<label1>,<label-2> and 5 more`

anandtiwary: I see that it is already happening :D

anandtiwary (dismissed):

anandtiwary (approved):

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -24,6 +24,7 @@ import { MultiSelectJustify } from './multi-select-justify';
   selector: 'ht-multi-select',
   styleUrls: ['./multi-select.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
+  providers: [SubscriptionLifecycle],
```
aaron-steinfeld, line 27: next time you're in here, this can be cleaned out

PR #1230, opened 2021-11-03
feat(span-detail): cookies as first class fields

skjindal93:
## Description
Exposing cookies as first class field for span detail view. Cookies would be displayed along with headers

### Testing
Added unit tests. Verified UI locally

### Checklist:


projects/observability/src/shared/components/span-detail/call/headers/span-detail-call-headers.component.ts
```diff
@@ -10,37 +10,58 @@ import { EMPTY, Observable, of } from 'rxjs';
...
+  public label?: string;
 
   public ngOnChanges(): void {
-    this.buildHeaderRecords();
+    this.label = this.getLabel();
+    this.buildRecords();
+  }
+
+  private getLabel(): string {
+    switch (this.mode) {
```
aaron-steinfeld, line 39: nit: if this enum is only used for the title, it'd be better to just take the title as an input (and make this named something more generic).

aaron-steinfeld (approved):

PR #1231, opened 2021-11-03
feat: table prefs now use sessionStorage instead of localStorage

jake-bassett:
## Description
This changes table widget to use sessionStorage instead of localStorage to persist view preferences.

### DIFF
--- a/projects/common/src/preference/preference.service.ts
+++ b/projects/common/src/preference/preference.service.ts
@@ -1,21 +1,33 @@
 import { Injectable } from '@angular/core';
 import { Observable, of, throwError } from 'rxjs';
 import { map, switchMap } from 'rxjs/operators';
+import { AbstractStorage } from '../utilities/browser/storage/abstract-storage';
 import { LocalStorage } from '../utilities/browser/storage/local-storage';
+import { SessionStorage } from '../utilities/browser/storage/session-storage';
 import { BooleanCoercer } from '../utilities/coercers/boolean-coercer';
 import { NumberCoercer } from '../utilities/coercers/number-coercer';
+
+export const enum StorageType {
+  Local = 'local',
+  Session = 'session'
+}
 
 @Injectable({
   providedIn: 'root'
 })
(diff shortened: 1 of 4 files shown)


anandtiwary (approved):

projects/common/src/preference/preference.service.ts
```diff
@@ -1,30 +1,46 @@
...
    * will continue to emit as the preference is updated, reverting to default value if the
    * preference becomes unset. If default value is not provided, the observable will
    * throw in the case the preference is unset.
    */
-  public get<T extends PreferenceValue>(key: PreferenceKey, defaultValue?: T): Observable<T> {
-    return this.preferenceStorage.watch(this.asStorageKey(key)).pipe(
+  public get<T extends PreferenceValue>(
+    key: PreferenceKey,
+    defaultValue?: T,
+    type: StorageType = PreferenceService.DEFAULT_STORAGE_TYPE
```
aaron-steinfeld, line 41: nit: I think I'd argue that get should not take a type argument. We should read this from wherever it is written, first looking at session storage, then local storage.

aaron-steinfeld (approved):

projects/common/src/preference/preference.service.ts
jake-bassett: Not sure I agree. That seems like it could cause unexpected behavior. Prefer to hold off on that idea.

aaron-steinfeld: We can hold off for now and continue the discussion async, but it allows the writer and reader to decouple. The attribute is just as valid in either location, so the reader shouldn't need to care.

aaron-steinfeld (approved):

projects/common/src/preference/preference.service.ts
jake-bassett: Yeah, that's a fair argument. My problem with this implementation is exactly that, so your suggestion does seem like a reasonable trade-off. We'll revisit.

PR #1234, opened 2021-11-04
feat: support for showing labels in toggle component

arjunlalb:
## Description
support for showing labels in toggle component

Screenshot 2021-11-04 at 9 53 11 PM

### Testing
Locally verified. UT updated.

### Checklist:

### DIFF
--- a/projects/components/src/toggle-group/toggle-item.component.ts
+++ b/projects/components/src/toggle-group/toggle-item.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectionStrategy, Component, ElementRef, Input } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
+import { Color } from '@hypertrace/common';
 import { IconSize } from '../icon/icon-size';
 
 @Component({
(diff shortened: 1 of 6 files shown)


projects/components/src/toggle-group/toggle-item.component.ts
```diff
@@ -11,6 +12,13 @@ import { IconSize } from '../icon/icon-size';
       <ht-icon *ngIf="this.icon; else labelBlock" class="icon" [icon]="this.icon" size="${IconSize.Medium}"></ht-icon>
       <ng-template #labelBlock>
         <ht-label class="label" [label]="this.label"></ht-label>
+        <ht-label-tag
```
anandtiwary, line 15: Should we do this via ng-content so that we can support more modifications in future.

arjunlalb: I tried exploring that. But its slightly more complicated in this case because the toggleGroup component does a bunch of things based on the view children today. If we support content here, we'll have to find a way to merge the content children into the list of items the toggle group handles. Didn't want to introduce such a change on a widely used component.

I can take another stab and see if there's a clean way to achieve what we want.

PR #1238, opened 2021-11-05
Allow configurable select or multiselect table controls

jake-bassett:
## Description
This change allows table dropdown controls to be configured for single or multi-select.


palbizu (approved):
LGTM!

PR #1253, opened 2021-11-10
fix: make divider 1px high

jake-bassett:
## Description
1px high divider

### DIFF
--- a/projects/components/src/divider/divider.component.scss
+++ b/projects/components/src/divider/divider.component.scss
@@ -2,6 +2,5 @@
 
 .divider {
   width: 100%;
-  border: 1px solid $gray-2;
-  border-radius: 1px;
+  border-bottom: 1px solid $gray-2;
 }


projects/components/src/divider/divider.component.scss
```diff
@@ -2,6 +2,5 @@
 
 .divider {
   width: 100%;
-  border: 1px solid $gray-2;
-  border-radius: 1px;
+  border-bottom: 1px solid $gray-2;
```
anandtiwary, line 5: are you sure?

anandtiwary (approved):

projects/components/src/divider/divider.component.scss
jake-bassett: I'm sure this makes it 1px. Not positive we don't have use cases for 2px, but the ones I know of are 1px. Not a huge deal either way.

PR #1254, opened 2021-11-11
refactor: breadcrumb to support additional specifications

anandtiwary:
## Description
refactor: breadcrumb to support additional specifications

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.test.ts
+++ b/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.test.ts
@@ -84,19 +84,40 @@ describe('Api detail breadcrumb resolver', () => {
       runFakeRxjs(({ expectObservable }) => {
         expectObservable(breadcrumb$).toBe('(abc|)', {
           a: {
+            [entityIdKey]: 'test-id',
+            [entityTypeKey]: ObservabilityEntityType.Api,
             label: 'test service',
             icon: ObservabilityIconType.Service,
-            url: ['services', 'service', 'test-service-id']
+            url: ['services', 'service', 'test-service-id'],
+            name: 'test api',
+            parentId: 'test-service-id',
+            parentName: 'test service',
+            serviceName: 'test service',
+            serviceId: 'test-service-id'
           },
           b: {
+            [entityIdKey]: 'test-id',
+            [entityTypeKey]: ObservabilityEntityType.Api,
             label: 'Endpoints',
             icon: ObservabilityIconType.Api,
-            url: ['services', 'service', 'test-service-id', 'endpoints']
+            url: ['services', 'service', 'test-service-id', 'endpoints'],
+            name: 'test api',
+            parentId: 'test-service-id',
+            parentName: 'test service',
+            serviceName: 'test service',
+            serviceId: 'test-service-id'
           },
           c: {
+            [entityIdKey]: 'test-id',
+            [entityTypeKey]: ObservabilityEntityType.Api,
             label: 'test api',
             icon: ObservabilityIconType.Api,
-            url: ['api', 'test-id']
+            url: ['api', 'test-id'],
+            name: 'test api',
+            parentId: 'test-service-id',
+            parentName: 'test service',
+            serviceName: 'test service',
+            serviceId: 'test-service-id'
           }
         });
       });
@@ -109,7 +130,7 @@ describe('Api detail breadcrumb resolver', () => {
         entityType: ObservabilityEntityType.Api,
         id: 'test-id'
       }),
-      { cacheability: GraphQlRequestCacheability.NotCacheable }
+      { cacheability: GraphQlRequestCacheability.Cacheable }
     );
   }));
 
@@ -123,9 +144,14 @@ describe('Api detail breadcrumb resolver', () => {
       runFakeRxjs(({ expectObservable }) => {
         expectObservable(breadcrumb$).toBe('(y|)', {
           y: {
+            [entityIdKey]: 'test-id',
+            [entityTypeKey]: ObservabilityEntityType.Api,
             label: 'test api',
             icon: ObservabilityIconType.Api,
-            url: ['api', 'test-id']
+            url: ['api', 'test-id'],
+            name: 'test api',
+            serviceName: 'test service',
+            serviceId: 'test-service-id'
           }
         });
       });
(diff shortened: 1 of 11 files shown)


projects/observability/src/pages/apis/api-detail/api-detail.service.ts
```diff
@@ -18,8 +19,9 @@ export class ApiDetailService extends EntityDetailService<ApiEntity> {
   }
 }
 
-export interface ApiEntity extends Entity {
+export interface ApiEntity extends Entity, Breadcrumb {
```
aaron-steinfeld, line 22: having ApiEntity extend breadcrumb doesn't seem intuitive. Does this service actually return all the breadcrumb info? It doesn't look like it.

projects/observability/src/shared/services/entity-breadcrumb/entity-breadcrumb.resolver.ts
```diff
@@ -0,0 +1,59 @@
...
+      switchMap(timeRange =>
+        this.graphQlQueryService.query<EntityGraphQlQueryHandlerService, T>(
+          {
+            requestType: ENTITY_GQL_REQUEST,
+            entityType: entityType,
+            id: id,
+            properties: [...this.getAttributeSpecification(), ...this.getAdditionalSpecifications()],
+            timeRange: new GraphQlTimeRange(timeRange.startTime, timeRange.endTime)
+          },
+          { cacheability: GraphQlRequestCacheability.NotCacheable }
```
aaron-steinfeld, line 42: Why are these not cacheable?

projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
```diff
@@ -1,101 +1,78 @@
...
     };
   }
 
-  protected getParentBreadcrumbs(api: ApiBreadcrumbDetails, parentEntityMetadata?: EntityMetadata): Breadcrumb[] {
+  protected getParentBreadcrumbs(api: T & Breadcrumb, parentEntityMetadata?: EntityMetadata): Breadcrumb[] {
     return parentEntityMetadata !== undefined
       ? [
           {
-            label: api.parentName,
+            label: api.parentName as string,
```
aaron-steinfeld, line 61: types are wrong here, shouldn't need to cast

projects/observability/src/shared/services/entity-breadcrumb/entity-breadcrumb.resolver.ts
anandtiwary: because we can do crud operations on entities -> labels update and probably rate limits

projects/observability/src/pages/apis/api-detail/api-detail.service.ts
anandtiwary: Yeah, we shouldn't really be using APIEntity in breadcrumb resolver but I am trying to limit the refactor in the PR.

projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
anandtiwary: Ts is not able to detect the types in the pipe -> map operator

aaron-steinfeld: What's the type on line 42? Are we sure we're not just dropping it on 59? That map should produce a `T & Partial<Pick<ApiBreadcrumbDetails, 'parentName' | 'parentId'>>` since we're spreading them together (that partial could use a type alias!).

aaron-steinfeld: (nit btw)

projects/observability/src/shared/services/entity-breadcrumb/entity-breadcrumb.resolver.ts
aaron-steinfeld: I think this is the main thing remaining. Labels would be the only thing that could affect it, but never caching breadcrumbs is overkill. Could we pull this out into a protected method as a starting point, we can go from there and make decisions per breadcrumb resolver.

projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
anandtiwary: Yeah, I am not able to make it work. IMO, `getParentPartial` is making the parentName and parentId optional and then we are required fields in `ApiBreadcrumbDetails`. 

We wouldn't have to cast if we can ensure that following code returns an `ApiBreadcrumbDetails` object
```
 map(apiEntity => ({
          ...apiEntity,
          ...this.getParentPartial(apiEntity, parentEntityMetadata)
        })),
```

anandtiwary: I can do that by assigning '' instead of undefined but i don't know if it would create any run time issue. So, I am ignoring it for this change.

aaron-steinfeld: we could just make those two fields optional on `ApiBreadcrumbDetails` I think? They _are_ optional, they're omitted if we can't resolve them, hence the check on line 60 before we access them.

aaron-steinfeld (approved):

anandtiwary:
Seeing some issues with test execution. Pushing another patch

anandtiwary:
Screen Shot 2021-11-11 at 12 30 42 PM


@aaron-steinfeld  See this. Not sure why it didn't fail.

anandtiwary:
back to normal time

projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.test.ts
```diff
@@ -108,7 +130,7 @@ describe('Api detail breadcrumb resolver', () => {
         entityType: ObservabilityEntityType.Api,
         id: 'test-id'
       }),
-      { cacheability: GraphQlRequestCacheability.NotCacheable }
```
anandtiwary, line 111: Will have to keep an eye on this change. Hopefully it doesn't break any existing case

aaron-steinfeld: you could change the default from cacheable to undefined too - use the default options.

aaron-steinfeld (approved):

aaron-steinfeld (approved):

PR #969, opened 2021-07-05
feat: move a common method into navigation service

arjunlalb:
## Description
Minor refactor

### Testing
UT added

### Checklist:
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #1258, opened 2021-11-13
fix: common project should not depend on components project

arjunlalb:
## Description
This PR -> https://github.com/hypertrace/hypertrace-ui/pull/969 introduced a change where the common project depends on components project. That should not happen. Refactoring code to avoid that unintended dependency.

### Testing
Verified with build. 

### Checklist:

### DIFF
--- a/projects/common/src/navigation/navigation.config.ts
+++ b/projects/common/src/navigation/navigation.config.ts
@@ -1,23 +1,4 @@
 import { Observable } from 'rxjs';
-import { Color } from '../color/color';
-import { FeatureState} from '../feature/state/feature.state';
-
-export type NavItemConfig = NavItemLinkConfig | NavItemHeaderConfig | NavItemDividerConfig;
-
-export interface NavItemLinkConfig {
-  type: NavItemType.Link;
-  icon: string;
-  iconSize?: string;
-  label: string;
-  matchPaths: string[]; // For now, default path is index 0
-  features?: string[];
-  replaceCurrentHistory?: boolean;
-  isBeta?: boolean;
-  trailingIcon?: string;
-  trailingIconTooltip?: string;
-  trailingIconColor?: Color;
-  featureState$?: Observable<FeatureState>;
-}
 
 export type FooterItemConfig = FooterItemLinkConfig;
 
--- a/projects/common/src/navigation/navigation.config.ts
+++ b/projects/common/src/navigation/navigation.config.ts
@@ -1,13 +1,13 @@
-import { Color, FeatureState } from '@hypertrace/common';
 import { Observable } from 'rxjs';
-import { IconSize } from '../../../components/src/icon/icon-size';
+import { Color } from '../color/color';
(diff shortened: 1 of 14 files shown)


projects/common/src/navigation/navigation.service.test.ts
```diff
@@ -3,8 +3,7 @@ import { Title } from '@angular/platform-browser';
 import { Router, UrlSegment } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
 import { IconType } from '@hypertrace/assets-library';
-import { APP_TITLE } from '@hypertrace/common';
-import { NavItemType } from '@hypertrace/components';
+import { APP_TITLE, NavItemType } from '@hypertrace/common';
```
aaron-steinfeld, line 6: circular ref, we're in the common project here

projects/common/src/navigation/navigation.service.ts
```diff
@@ -14,13 +14,13 @@ import {
...
   UrlTree
 } from '@angular/router';
-import { NavItemConfig, NavItemType } from '@hypertrace/components';
 import { uniq } from 'lodash-es';
 import { from, Observable, of } from 'rxjs';
 import { distinctUntilChanged, filter, map, share, skip, startWith, switchMap, take, tap } from 'rxjs/operators';
 import { isEqualIgnoreFunctions, throwIfNil } from '../utilities/lang/lang-utils';
 import { Dictionary } from '../utilities/types/types';
 import { APP_TITLE, HtRoute } from './ht-route';
+import { NavItemConfig, NavItemType } from './navigation.config';
```
aaron-steinfeld, line 23: IMO the better solution would be to move the method out of nav service into the nav component - that's the only caller and would allow leaving icon size on the config as is, too - but up to you.

aaron-steinfeld (dismissed):

projects/common/src/navigation/navigation.service.ts
anandtiwary: I was actually trying to fix this too as part of nx poc work. I think we should remove the `decorateNavItem` method from here and move it to a new service -> `navigation-list.service.ts`. This method is only used when we are using the navigation-list.component.  

We can move it to the component too like Aaron is suggesting but that may require some additional work.

projects/common/src/navigation/navigation.config.ts
```diff
@@ -1,13 +1,13 @@
...
+import { Color } from '../color/color';
+import { FeatureState} from '../feature/state/feature.state';
 
 export type NavItemConfig = NavItemLinkConfig | NavItemHeaderConfig | NavItemDividerConfig;
 
 export interface NavItemLinkConfig {
   type: NavItemType.Link;
   icon: string;
-  iconSize?: IconSize;
+  iconSize?: string;
```
anandtiwary, line 10: I don't think we should remove this type safety. We only support few size options. Moving the method to a new service would solve this.

PR #1270, opened 2021-11-18
feat: toggle able legend for cartesian chart

itssharmasandeep:
## Description

- Click on the legend entry to toggle the series on/off
- Reset to back to default

Screenshot 2021-11-18 at 8 03 09 PM

### Testing
Manual testing done!!


projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -50,6 +70,20 @@ export class CartesianLegend {
...
+    select(container)
+      .append('span')
+      .classed(CartesianLegend.RESET_CSS_CLASS, true)
+      .text('Reset')
+      .on('click', () => this.resetToDefault());
+
+    this.toggleClearAll(this.isDefault);
+  }
+
+  private toggleClearAll(isHidden: boolean): void {
```
arjunlalb, line 83: nit: update as per the new `reset` name

arjunlalb (commented):
LGTM. Will advise waiting for @aaron-steinfeld to take a look.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -271,8 +279,10 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     }
   }
 
-  private updateData(): void {
-    this.drawLegend();
+  private updateData(withLegend: boolean = true): void {
```
aaron-steinfeld, line 282: boolean flags to change behavior are a code smell. Why were we able to always draw legend up until now, and now need to configure it?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -283,6 +293,16 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     this.setupEventListeners();
   }
 
+  private redrawVisualization(): void {
+    const chartViz = select(this.chartContainerElement!).selectAll(
```
aaron-steinfeld, line 297: In d3, usually you don't remove nodes in order to redraw them - you update the bound data and let d3 optimize the dom manipulation.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -348,6 +368,14 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
           this.chartContainerElement,
           this.legendPosition
         );
+        if (this.activeSeriesSubscription) {
+          this.activeSeriesSubscription.unsubscribe();
+        }
+        this.activeSeriesSubscription = this.legend.activeSeries$.subscribe(activeSeries => {
+          this.series.length = 0;
```
aaron-steinfeld, line 375: Isn't series the value passed into the chart from outside? We should never be overwriting that, aren't we losing data? I'm guessing that's why we jump through hoops above to not redraw the legend, since it still has access to the old data. Instead, we should just hold the visible series in a separate variable that we update and bind that as the d3 data.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -23,7 +35,14 @@ export class CartesianLegend {
     private readonly injector: Injector,
     private readonly intervalData?: CartesianIntervalData,
     private readonly summaries: Summary[] = []
-  ) {}
+  ) {
+    this.activeSeries = [...this.series];
```
aaron-steinfeld, line 39: and here, we're duping the series because we mutate it above, which would otherwise effect here :) lots of side effects that different places need to be in sync with - better to not touch `this.series` as described above.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -50,6 +70,20 @@ export class CartesianLegend {
...
+    select(container)
+      .append('span')
+      .classed(CartesianLegend.RESET_CSS_CLASS, true)
+      .text('Reset')
+      .on('click', () => this.resetToDefault());
+
+    this.toggleReset(this.isDefault);
+  }
+
+  private toggleReset(isHidden: boolean): void {
```
aaron-steinfeld, line 83: nit: toggle implies we're flipping something without an arg, but since we'r'e providing the input here, it's just a set function.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -133,4 +183,33 @@ export class CartesianLegend {
       })
     );
   }
+
+  private resetToDefault(): void {
+    this.activeSeries = [...this.initialSeries];
+    this.isDefault = true;
```
aaron-steinfeld, line 189: this seems duplicative of a length comparison of active and initial series, can we calculate it on the fly so we don't duplicate information?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -133,4 +183,33 @@ export class CartesianLegend {
...
+        this.activeSeries.push(seriesEntry);
+      }
+    }
+    this.updateLegendTextClasses();
+    this.toggleReset(this.isDefault);
+    this.activeSeriesSubject.next();
+  }
+
+  private isThisLegendEntryActive(seriesEntry: Series<{}>): boolean {
+    return this.activeSeries.findIndex(series => series === seriesEntry) >= 0;
```
aaron-steinfeld, line 213: nit: `contains` (or something like that - there's a simple util in lodash)

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -80,6 +84,10 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     this.tooltip && this.tooltip.destroy();
     this.legend && this.legend.destroy();
 
+    if (this.activeSeriesSubscription) {
```
anandtiwary, line 87: why not just `this.activeSeriesSubscription?.unsubscribe();`

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -23,7 +35,14 @@ export class CartesianLegend {
     private readonly injector: Injector,
     private readonly intervalData?: CartesianIntervalData,
     private readonly summaries: Summary[] = []
-  ) {}
+  ) {
+    this.activeSeries = [...this.series];
+    this.initialSeries = [...this.series];
+    this.activeSeries$ = this.activeSeriesSubject.asObservable().pipe(
```
anandtiwary, line 41: we shouldn't need observables for this. This could be simplified.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Actually, this is not. This represents whether we're on the initial/default/reset stage (Where color of the text would be $gray-9).
In a scenario where suppose two legend entries are there, X and Y. If I select A,  then X becomes active ($blue-4) and Y becomes inactive ($gray-5), then if I click on the Y again, then Y also becomes active.

We can implement this in two ways-
1. Whenever the all the entires become active then put them to default state with $gray-9
2. Or we can have them the active state ($blue-4) with the indication of that they have made some changes, anyways they have one click reset button to go to default.

I like the second one more and implemented the same.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
itssharmasandeep: Fixed

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: used `includes`

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: changed it

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Any thoughts?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
itssharmasandeep: So the current behaviour was to draw everything when series changes. In this current case I do not want to draw the legend again as we need every state etc. for the legend.

aaron-steinfeld: Commented on this below - we don't want to hold state in the legend in that way.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
aaron-steinfeld: Yeah that makes sense. I go back and forth on it, it seems weird to have all series selected and it be possible to be in two different states, but it's even weirder to have the selection action change as a user selects - that's what you were describing right? Where if I select A, then select B, then select B again it should deselect B. Selecting a series means "exclusively this" from default, but it means "toggle this" after that.

aaron-steinfeld: Given that, maybe a slight wording change would help - something to convey we're in a selection mode, even though we may have the default series.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
itssharmasandeep: I have used a new variable now!!

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
itssharmasandeep: I looked into the code and tried to modify the code to use d3 update pattern but it seems that, we're using some underlying classes like band, axis, legend which are not using the same. To update this behaviour we need to update at many places, not sure about that much of effort!

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: > Where if I select A, then select B, then select B again it should deselect B

Correct

itssharmasandeep: > something to convey we're in a selection mode, even though we may have the default series.
We have that color change as well as that reset button, I think that can be good enough indication.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.scss
```diff
@@ -118,13 +119,41 @@
...
+          }
+
+          &.inactive {
+            color: $gray-5;
+          }
         }
       }
     }
 
+    .reset {
```
anandtiwary, line 141: Is this for legend-text only? If so, can we place it inside `.legend`  or  `.legend-text?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -338,16 +360,23 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
-          this.legendPosition
-        );
+        this.legend = new CartesianLegend(
+          this.activeSeries,
+          this.injector,
+          this.intervalData,
+          this.seriesSummaries
+        ).draw(this.chartContainerElement, this.legendPosition);
+        this.activeSeriesSubscription?.unsubscribe();
+        this.activeSeriesSubscription = this.legend.activeSeries$.subscribe(activeSeries => {
```
anandtiwary, line 376: @itssharmasandeep Is there a way to avoid the subscription? Can we not send a normal array?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
anandtiwary: Why don't we introduce a 'visible' property in the series object (default = true). With every change, we can set visible to false. So, Active series would be a filter on series with visible = true. Would this avoid the subs and observable?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
aaron-steinfeld: I meant to the reader of code. Default here refers to a selection mode, rather than which series are selected - the latter would be my assumption as a reader.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
aaron-steinfeld: Observables make sense to me here. We don't want to mutate the visibility of the series object that's passed in, and we'd still have to notify the chart when this changed in order to rerender.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
aaron-steinfeld: This is a continuously changing state driven by user interactions, it's a textbook case for observables IMO.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
aaron-steinfeld: Those classes shouldn't be either, but also should be independent of what we're doing with legend. The main thing I'd go for here is to remove the special casing for legend with the boolean - that's the bigger code smell, the update vs remove/redraw pattern is easy to fix later as long as it is contained within drawLegend.

That boolean, from my recollection was needed because we don't want to rebuild the legend now that it's holding state. It would be better to address that and not rebuild it but moving this check inside drawLegend so it's self contained works too if we need a quicker way.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.scss
itssharmasandeep: it is inside `.legend`

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
itssharmasandeep: What I have done here is, I broke the updateData function and apart from legend part, I put in another function called `drawVisualizations()`, So I am calling that function from `redrawVisualization`, by which we're not doing this boolean part. Now the part where not removing the chart Visualization and using D3 update flow, we can skip as mentioned above as of now.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: changed this!!

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -82,15 +116,41 @@ export class CartesianLegend {
...
   }
 
+  private updateLegendClassesAndStyle(): void {
+    const legendElementSelection = select(this.legendElement!);
+
+    // Legend entry symbol
+    legendElementSelection
+      .selectAll('.legend-symbol circle')
+      .style('fill', series =>
+        !this.isThisLegendEntryActive(series as Series<{}>) ? Color.Gray3 : (series as Series<{}>).color
```
aaron-steinfeld, line 137: All these casts should be avoidable, the type on line 131 is incorrect. See line 116 for an example.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -12,9 +14,19 @@ import { CartesianSummaryComponent, SUMMARIES_DATA } from './cartesian-summary.c
 
 export class CartesianLegend {
   private static readonly CSS_CLASS: string = 'legend';
+  private static readonly RESET_CSS_CLASS: string = 'reset';
+  private static readonly DEFAULT_CSS_CLASS: string = 'default';
+  private static readonly ACTIVE_CSS_CLASS: string = 'active';
+  private static readonly INACTIVE_CSS_CLASS: string = 'inactive';
   public static readonly CSS_SELECTOR: string = `.${CartesianLegend.CSS_CLASS}`;
 
+  public readonly activeSeries$: Observable<Series<{}>[]>;
```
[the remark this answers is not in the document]
aaron-steinfeld: I see that some of this was pre-existing. Since this is old code, I'm guessing this was actually from before unknown was introduced to the language, but if it's difficult to change we can defer; if it's a simple replacement, let's do it.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -133,4 +193,33 @@ export class CartesianLegend {
...
+    this.isSelectionModeOn = false;
+    this.updateLegendClassesAndStyle();
+    this.setResetVisibility(!this.isSelectionModeOn);
+    this.activeSeriesSubject.next();
+  }
+
+  private updateActiveSeries(seriesEntry: Series<{}>): void {
+    if (!this.isSelectionModeOn) {
+      this.activeSeries = [];
+      this.activeSeries.push(seriesEntry);
```
aaron-steinfeld, line 208: nit - why reset then mutate?
```suggestion
      this.activeSeries = [seriesEntry];
```

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -133,4 +193,33 @@ export class CartesianLegend {
...
+      this.isSelectionModeOn = true;
+    } else {
+      if (this.isThisLegendEntryActive(seriesEntry)) {
```
aaron-steinfeld, line 211: nit: extra nesting, use else if

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -23,7 +35,14 @@ export class CartesianLegend {
     private readonly injector: Injector,
     private readonly intervalData?: CartesianIntervalData,
     private readonly summaries: Summary[] = []
-  ) {}
+  ) {
+    this.activeSeries = [...this.series];
+    this.initialSeries = [...this.series];
+    this.activeSeries$ = this.activeSeriesSubject.asObservable().pipe(
+      startWith(undefined),
+      switchMap(() => of(this.activeSeries))
```
aaron-steinfeld, line 43: a. this need not use switchMap and wrap the series, it can just use `mapTo`
b. Why not make the subject hold the series rather than having it as a notification and then mapping it? Every place we `next` it we have the new value, so it's just an extra layer of code
c. We should make sure these are distinct - either with an operator like distinctUntilChanged (which would need to use isEqual due to the changing references like `distinctUntilChanged(isEqual)`), or by only invoking it on change. The only spot that I can see where we're potentially doubling up is when changing modes. If there's a single series (commented on preventing this elsewhere), selecting it won't change the activeSeries selection. Similarly, if all series are selected and we click reset, only the mode changes, not the activeSeries.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -82,15 +116,41 @@ export class CartesianLegend {
     const legendEntry = select<EnterElement, Series<{}>>(element).append('div').classed('legend-entry', true);
 
     this.appendLegendSymbol(legendEntry);
-
     legendEntry
       .append('span')
       .classed('legend-text', true)
-      .text(series => series.name);
+      .text(series => series.name)
+      .on('click', series => this.updateActiveSeries(series));
```
aaron-steinfeld, line 123: If there's only one series, can we avoid adding this? We don't need selection mode for a single series graph (which is many of them).

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -112,15 +109,15 @@ export class CartesianLegend {
...
+  private drawLegendEntry(element: EnterElement): Selection<HTMLDivElement, Series<TData>, null, undefined> {
+    const legendEntry = select<EnterElement, Series<TData>>(element).append('div').classed('legend-entry', true);
 
     this.appendLegendSymbol(legendEntry);
     legendEntry
       .append('span')
       .classed('legend-text', true)
       .text(series => series.name)
-      .on('click', series => this.updateActiveSeries(series));
+      .on('click', series => (this.series.length > 1 ? this.updateActiveSeries(series) : undefined));
```
aaron-steinfeld, line 120: do we need to change the css in this case too to remove the pointer?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -12,36 +12,33 @@ import {
...
     private readonly intervalData?: CartesianIntervalData,
     private readonly summaries: Summary[] = []
   ) {
     this.activeSeries = [...this.series];
     this.initialSeries = [...this.series];
-    this.activeSeries$ = this.activeSeriesSubject.asObservable().pipe(
-      startWith(undefined),
-      switchMap(() => of(this.activeSeries))
-    );
+    this.activeSeries$ = this.activeSeriesSubject.asObservable().pipe(distinctUntilChanged(), startWith(this.series));
```
aaron-steinfeld, line 41: This distinctUntilChanged doesn't solve our problem - see my previous comment about `isEqual` https://github.com/hypertrace/hypertrace-ui/pull/1270/files/d9be3992407b4436a1c84bc7b826c49942eeb888#r760277561

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Yes, Updated this!!

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: This is not working.
The problem I found it, distinctUntilChanged is not behaving the way we want. I tried to do this, to check where is the problem.

```
this.activeSeries$ = this.activeSeriesSubject.asObservable().pipe(
      tap(console.log),
      distinctUntilChanged((prev, curr) => {
        console.log(prev, curr);
        return prev.length === curr.length;
      })
    );
```
The problem I found was, `prev` is always the first one.

I think, we can take this later as you mentioned, there are only two case where we would need this, only one series (this has already been taken care of where we disable the click behaviour), second when everything is selected and and then we do a reset. I think this is fine, It is also very common to do, so we can skip

aaron-steinfeld (approved):

PR #1281, opened 2021-11-23
Form changes

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation


anandtiwary:
This PR touches up on a lot of different (but small) changes. Not too big though. Let me know if this doesn't work

projects/components/src/modal/modal.ts
```diff
@@ -36,15 +36,18 @@ export const getModalDimensions = (modalSize: ModalSize): ModalDimension => {
     case ModalSize.Large:
       return getModalDimensionObject(640, 720);
     case ModalSize.LargeTall:
-      return getModalDimensionObject(640, 800);
+      return getModalDimensionObject(640, '90vh');
```
arjunlalb, line 39: please ensure you test the change across various modals in the system today.

projects/graphql-client/src/model/graphql-argument.ts
```diff
@@ -23,3 +25,51 @@ export type GraphQlArgumentValue =
...
   | GraphQlArgumentArray
   | GraphQlEnumArgument<string>;
+
+/**
+ * @param value
+ * Convert the value to GraphQL argument value
+ */
+export const getValueAsGraphQlArgument = (
+  value: unknown,
+  considerAllCapsStringAsEnum: boolean = false,
```
arjunlalb, line 35: I am not sure if we should do this. This makes it hard for us to combine enum and non enum cases that might use upper case input.
I'd prefer to have a boolean indicating whether the value is enum or not.

projects/components/src/modal/modal.ts
anandtiwary: I did check a few modals, but it shouldn't be impactful anyway. Just making a better use of space when it is available.

projects/graphql-client/src/model/graphql-argument.ts
anandtiwary: Yeah, this would need more iterations in the future. The challenge is coming due to the nested nature of the object. I would like us to consider this as the first solution towards a better/generic solution. we can't have a boolean for each value.

arjunlalb: Yes, we can't have boolean for each value, but the only value that needs special treatment is enum. A string remains a string or becomes an enum. I feel its better to make the input param represent the type than use `considerAllCapsStringAsEnum`. That way we can use it for enums that aren't uppercased and strings that need to be uppercased without confusion.

arjunlalb (approved):
Rest of the changes LGTM.

projects/graphql-client/src/model/graphql-argument.ts
anandtiwary: Yeah, that's what I tried initially, but I am not sure how to differentiate between a string and an enum. `typeof` wouldn't work.

anandtiwary: happy to address it (if there is a solution) in a follow up pr

PR #1288, opened 2021-11-26
feat: grouped cartesian legend

itssharmasandeep:
## Description
This will provide the functionality for having groups in legend for cartesian chart!!
`Non-grouped`
Screenshot 2021-12-17 at 2 00 09 PM

`Without useGroupName`
Screenshot 2021-12-17 at 1 59 48 PM

`With useGroupName`
Screenshot 2021-12-17 at 1 59 14 PM

### Testing
Local testing done

### Checklist:


projects/observability/src/shared/components/cartesian/cartesian-chart.component.test.ts
```diff
@@ -185,13 +185,23 @@ describe('Cartesian Chart component', () => {
           name: 'test series 1',
           color: 'blue',
           type: CartesianSeriesVisualizationType.Column,
+          assignedGroup: {
+            id: 1,
+            groupName: 'first',
+            specName: 'sum(foo)'
```
anandtiwary, line 191: why does it need spec name? That should be unrelated to this api. Group Name should be sufficient, isn't it?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -36,6 +40,13 @@ export class CartesianLegend {
     private readonly intervalData?: CartesianIntervalData,
     private readonly summaries: Summary[] = []
   ) {
+    this.numberOfgroups =
+      this.series.length === 0 || isNil(this.series[0].assignedGroup)
+        ? 0
+        : this.series[this.series.length - 1].assignedGroup!.id;
```
anandtiwary, line 46: `assignedGroup` is optional right? We shouldn't override it here.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -85,13 +96,43 @@ export class CartesianLegend {
...
+        .enter()
+        .each((_, index, elements) => this.drawLegendEntry(elements[index]));
+    } else {
+      containerSelection
+        .selectAll('.legend-entries')
+        .data(this.groupedSeries)
+        .enter()
+        .append('div')
+        .attr('class', seriesGroup => `legend-entries group-${seriesGroup[0].assignedGroup!.id}`)
+        .each((seriesGroup, id, elems) => {
```
anandtiwary, line 115: nit: id or index?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -85,13 +96,43 @@ export class CartesianLegend {
...
+          const legendEntriesSelection = select(elems[id]);
+
+          legendEntriesSelection
+            .selectAll('.legend-entries-title')
+            .data([seriesGroup])
+            .enter()
+            .append('div')
+            .classed('legend-entries-title', true)
+            .text(group => `${group[0].assignedGroup!.specName}:`)
```
anandtiwary, line 124: assignedGroup could be missing.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -85,13 +96,43 @@ export class CartesianLegend {
...
+      containerSelection
+        .append('div')
+        .classed('legend-entries', true)
+        .selectAll('.legend-entry')
+        .data(this.series.filter(series => !series.hide))
+        .enter()
+        .each((_, index, elements) => this.drawLegendEntry(elements[index]));
+    } else {
+      containerSelection
+        .selectAll('.legend-entries')
```
anandtiwary, line 110: We should try to use d3 update pattern meaning we should add merge and exit calls as well. But we can do it later

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -202,16 +259,57 @@ export class CartesianLegend {
...
+  private getGroupedSeries(): Series<{}>[][] {
+    if (!this.isGrouped) {
+      return [];
+    }
+
+    const nonHiddenSeries = this.series.filter(series => !series.hide);
+    const groupedSeries: Series<{}>[][] = [];
+    const currentGroup: Series<{}>[] = [];
+
+    nonHiddenSeries.forEach(seriesEntry => {
```
anandtiwary, line 271: Could you please comment or just explain in this PR how exactly this solution is working? A step wise breakup would be great to understand

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.test.ts
```diff
@@ -1,3 +1,4 @@
+//
```
anandtiwary, line 1: unintended?

projects/observability/src/shared/components/cartesian/cartesian-chart.component.test.ts
aaron-steinfeld: +1 - specs should not be in the chart code, that's a data concept.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Actually, for this assigned group, if one of the series has this, every series will have this, This is the way group by working currently!! So It should be fine here

projects/observability/src/shared/components/cartesian/cartesian-chart.component.test.ts
itssharmasandeep: Fixed, used name instead! some code changes that as well, please check

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: The way we're using this component/class, it doesn't matter actually so whenever we change the series, cartesian chart will create a new instance of the legend, That should be fixed first.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
anandtiwary: We are introducing a tight coupling. If you keep it optional here then your app won't crash if the upstream stops sending `assignedGroup` obj (due to a bug).

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
anandtiwary: that's true.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
anandtiwary: Also, why are you assigning `id` to number of groups?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -85,11 +96,42 @@ export class CartesianLegend {
...
+        .selectAll('.legend-entries')
+        .data(this.groupedSeries)
+        .enter()
+        .append('div')
+        .attr('class', seriesGroup => `legend-entries group-${seriesGroup[0].assignedGroup!.id}`)
+        .each((seriesGroup, index, elements) => this.drawLegendEntriesTitleAndValues(seriesGroup, elements[index]));
+    }
+  }
+
+  private drawLegendEntriesTitleAndValues(seriesGroup: Series<{}>[], element: HTMLDivElement): void {
```
anandtiwary, line 119: may be do `Series<TData>` or `Series<unknown>`

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -85,11 +96,42 @@ export class CartesianLegend {
...
+    legendEntriesSelection
+      .selectAll('.legend-entries-title')
+      .data([seriesGroup])
+      .enter()
+      .append('div')
+      .classed('legend-entries-title', true)
+      .text(group => `${group[0].name}:`)
+      .on('click', () => this.updateActiveSeries(seriesGroup));
+
+    legendEntriesSelection
```
anandtiwary, line 130: This append is not bound by any data, so it will always be applied. If the legends dom are not cleared then this would add duplicate nodes.

Also, why are appending `div` inside an svg?

anandtiwary: hmm, looks like all of the legend is in div and spans. Please confirm.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -202,16 +260,57 @@ export class CartesianLegend {
...
-      this.activeSeries = [];
-      this.activeSeries.push(seriesEntry);
-      this.isDefault = false;
+  private getGroupedSeries(): Series<{}>[][] {
+    if (!this.isGrouped) {
+      return [];
+    }
+
+    const nonHiddenSeries = this.series.filter(series => !series.hide);
+    const groupedSeries: Series<{}>[][] = [];
```
anandtiwary, line 269: What is the structure of a groupedSeries? Why is it two dimensional?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -23,8 +24,11 @@ export class CartesianLegend {
   public readonly activeSeries$: Observable<Series<{}>[]>;
   private readonly activeSeriesSubject: Subject<void> = new Subject();
   private readonly initialSeries: Series<{}>[];
+  private readonly groupedSeries: Series<{}>[][];
+  private readonly numberOfgroups: number;
```
aaron-steinfeld, line 28:
```suggestion
  private readonly numberOfGroups: number;
```

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
aaron-steinfeld: This reads to me like an assumption that the ids are numbers - we really shouldn't care, or even have ids for that matter. We also don't want to have magic indexing - easy source of bugs. We can just map through each series optional group name, collect them into a set. We have everything we need with just a group name on the series to organize, count distinct values etc.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
aaron-steinfeld: Guessing a `Map<String, Series<TData>[]>` is a better data structure here. Looks like we're trying to get a list of series split by their groups.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -202,16 +260,57 @@ export class CartesianLegend {
...
+    const currentGroup: Series<{}>[] = [];
+
+    nonHiddenSeries.forEach(seriesEntry => {
+      if (
+        currentGroup.length === 0 ||
+        currentGroup[currentGroup.length - 1].assignedGroup!.id === seriesEntry.assignedGroup!.id
+      ) {
+        currentGroup.push(seriesEntry);
+      } else {
+        groupedSeries.push([...currentGroup]);
```
aaron-steinfeld, line 279: Are we assuming an order? Looks like it, easy to avoid.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -202,16 +260,57 @@ export class CartesianLegend {
...
+    const nonHiddenSeries = this.series.filter(series => !series.hide);
+    const groupedSeries: Series<{}>[][] = [];
+    const currentGroup: Series<{}>[] = [];
+
+    nonHiddenSeries.forEach(seriesEntry => {
+      if (
+        currentGroup.length === 0 ||
+        currentGroup[currentGroup.length - 1].assignedGroup!.id === seriesEntry.assignedGroup!.id
+      ) {
+        currentGroup.push(seriesEntry);
```
aaron-steinfeld, line 277: Functional is your friend here - you should be able to use utils like `groupBy`  to quickly and more importantly readably, reshape an array into a map.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -202,16 +260,57 @@ export class CartesianLegend {
...
+        currentGroup.length = 0;
+        currentGroup.push(seriesEntry);
+      }
+    });
+    groupedSeries.push([...currentGroup]);
+
+    return groupedSeries;
+  }
+
+  private updateActiveSeries(series: Series<{}> | Series<{}>[]): void {
```
aaron-steinfeld, line 289: If we can possibly keep the data consistent (e.g. always use arrays, even if ungrouped), that would help keep the code cleaner and avoid things like instanceof.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
```diff
@@ -222,4 +321,8 @@ export class CartesianLegend {
   private isThisLegendEntryActive(seriesEntry: Series<{}>): boolean {
     return this.activeSeries.includes(seriesEntry);
   }
+
+  private isThisLegendSeriesGroupActive(seriesGroup: Series<{}>[]): boolean {
```
aaron-steinfeld, line 325: Why do we care if a whole group is active? Are we styling the title? If so, why? If it's not interactive, we shouldn't change its color and instead leave it as a plain label.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
anandtiwary: I would recommend you keep it optional and just assign a default value if it's missing. It could be ''

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: We do care about this, We can active whole group in once by clicking on the title!! That is why we need this as we have style changes as well like border change as well!!

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: I thought of doing the same earlier, but I ran into one problem.
Since we have whole group selection thing as well, that creates an issue. Suppose for non-grouped case what I would do is, I will pass [seriesA] as an argument (If I go with array approach), now what I would do is, foe every entry present in the array, I would check whether it is active or not , which works well with non-grouped case.

Now since we have whole group selection this time, and if I pass [A, B, C, D, E] with already D and E active then the above mentioned approach would make A, B, C => active and D, E inactive. SO I needed a way to distinguish between whole group selection or individual one. I can use another argument to do that, which can remove that instance of atleast, but it might extra handling of the code inside.

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: It has `svg` but not as parent. `Visualizations` are `svgs` then `legend` is `div`. inside that we have `divs` and `spans`, and `svgs` (used for `dot` part).

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Removed `id` part from here!!

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Done

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Done

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: done!!

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Not needed now

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
itssharmasandeep: Used` Dictionary<Series<TData>[]>`

itssharmasandeep: We need to assume the order, otherwise I am not sure, how are we going to do it?

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
aaron-steinfeld: This foreach still doesn't really make sense to me - I would expect us to just group series by `series.groupName ?? series.name`. Once all grouped, render the groups of > 1 with the new style, the groups of 1 as the existing style.

Here, If I had three series and they belong to groups A, B and A respectively, it would break because we're unnecessarily depending on order. Further, having the same wrapper shape for every series allows us to remove all the if/else logic around such as what we have on 281 (I remember a thread started there, but not sure if I missed responding to it - will try to dig that up).

projects/observability/src/shared/components/cartesian/d3/legend/cartesian-legend.ts
aaron-steinfeld: OK, so given the additional requirement of toggling a whole series at once, how do we expect it to behave? In the scenario you described, would we want A,B,C to be set to active and D,E not to change? Or D,E to set to inactive and A,B,C not to change? That resolution logic of a grouped series could be it own thing, but my suggestion would be that this takes in our top level unit - (which, in my other recommendation would be a `SeriesGroup`, even if that `SeriesGroup` contains one).

In other words, we do want to be able to make the determination "what got clicked", but that's not at odds with keeping the data shape consistent. The right abstraction (which doesn't sound like it should be just an array of series), would allow that too and address the consistency problem.

anandtiwary:
@itssharmasandeep Is it ready for another round of review?

itssharmasandeep:
@anandtiwary this is updated now. I had a discussion with @aaron-steinfeld  and I have updated accordingly now.

aaron-steinfeld (commented):
I think fixing the useGroupName case to work the same as a regular series is the only thing that we should fix before merging, the rest can be deferred.

itssharmasandeep:
@aaron-steinfeld I have updated this PR, and addressed required comments please review again

aaron-steinfeld (approved):

PR #1295, opened 2021-11-30
feat: persisted expand collapse for explorer panels

itssharmasandeep:
## Description
It will provide preference persistence for panels/sections, present in the explorer on expand and collapse.

### Testing
Local testing done.

### Checklist:

### DIFF
--- a/projects/observability/src/pages/explorer/explorer.component.ts
+++ b/projects/observability/src/pages/explorer/explorer.component.ts
@@ -304,8 +304,6 @@ interface InitialExplorerState {
   series: ExploreSeries[];
   interval?: IntervalValue;
   groupBy?: GraphQlGroupBy;
-  visualizationExpanded?: boolean;
-  resultsExpanded?: boolean;
 }
 
 interface ExplorerContextScope {
--- a/projects/observability/src/pages/explorer/explorer.component.ts
+++ b/projects/observability/src/pages/explorer/explorer.component.ts
@@ -3,6 +3,7 @@ import { ActivatedRoute, ParamMap } from '@angular/router';
 import {
   assertUnreachable,
   NavigationService,
+  PreferenceService,
   QueryParamObject,
   TimeDuration,
   TimeDurationService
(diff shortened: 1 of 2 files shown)


projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -146,7 +154,16 @@ export class ExplorerComponent {
     this.vizDashboard$ = this.explorerDashboardBuilder.visualizationDashboard$;
     this.initialState$ = activatedRoute.queryParamMap.pipe(
       take(1),
-      map(paramMap => this.mapToInitialState(paramMap))
+      map(paramMap => this.mapToInitialState(paramMap)),
+      tap(initialState => {
```
aaron-steinfeld, line 158: we can do this without tap, please just use distinct observables.

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -146,7 +154,16 @@ export class ExplorerComponent {
...
+        // Updating initial visualization and results expanded state
+        this.visualizationExpanded = initialState.visualizationExpanded ?? true;
+        this.resultsExpanded = initialState.resultsExpanded ?? true;
+        this.navigationService.addQueryParametersToUrl({
```
aaron-steinfeld, line 162: I think there was a misunderstanding of the ask here - the expand/collapse shouldn't be in the URL, it should be persisted as a preference (see usages of preference service), similar to how table sorts and columns work.

itssharmasandeep: Can someone edit these preferences?

aaron-steinfeld: Currently, they're only indirectly exposed. A user's actions update their preferences. Eventually, some may be put in front of a UI (not things like expand collapse but formats, colors etc.)

The simplest example of interaction with preferences is probably the expand/collapse behavior of the left nav in `hypertrace-ui/src/app/shared/navigation/navigation.component.ts`

projects/observability/src/pages/explorer/explorer.component.ts
itssharmasandeep: Done!!

projects/observability/src/pages/explorer/explorer.component.ts
itssharmasandeep: Done!!

aaron-steinfeld (approved):

PR #1296, opened 2021-11-30
feat: table filters now always show their placeholder even when a sel…

jake-bassett:
Currently, when a filter is selected we only show the value and remove the placeholder text. This change will keep the placeholder text as a prefix for the selected values.

Screen Shot 2021-11-30 at 3 27 28 PM
.

### DIFF
--- a/projects/components/src/multi-select/multi-select.component.ts
+++ b/projects/components/src/multi-select/multi-select.component.ts
@@ -146,6 +146,9 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
   public placeholder?: string;
 
   @Input()
+  public prefix?: string;
+
+  @Input()
   public disabled: boolean = false;
 
   @Input()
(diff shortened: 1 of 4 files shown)


projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -275,13 +278,21 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
...
       })
     );
   }
 
+  private getLabel(selectedItems: SelectOptionComponent<V>[]): string {
+    if (selectedItems.length === 0) {
+      return this.placeholder === undefined ? '' : this.placeholder;
+    }
+
+    return this.prefix === undefined ? selectedItems[0]?.label : `${this.prefix}${selectedItems[0]?.label}`.trim();
```
anandtiwary, line 293: oooh.. don't you think this is confusing? Currently, we should placeholder, something like "Select labels" and once we select few labels, we would show "label 1 + 2" others. Now we are showing that upfront. IMO, we lose the distinction and it will be hard to distinguish between your new placeholder and existing selection option trigger

jake-bassett: Hmm. Fair point. There are probably a few uses that we don't want to follow this pattern. Let me make it configurable.

jake-bassett: Ah, so I just looked back at the code and I had the foresight to already make this configurable. If the user doesn't see the `prefix` input, the behavior should be the same as before this PR.

anandtiwary (approved):

PR #1299, opened 2021-12-02
Semgrep workflow

Sunn-y-Arora:
## Description
adding semgrep scan as workflow for any merge to rzp_main branch


Sunn-y-Arora:
was supposed to add this in our fork, my bad.

PR #1313, opened 2021-12-09
Fix date picker timezone issue

adisreyaj:
## Description
Updating the time sometimes changes the date due to timezone issues. The date string was constructed using `toISOString()` method which uses the UTC timezone as seen:

```ts
const date = new Date();  // Thu Dec 09 2021 23:25:27 GMT+0530 (India Standard Time)
date.toISOString(); // 2021-12-09T17:55:27.051Z
```
So if the timeframe is in the range of the user's timezone, the date can change
timezone

This PR fixes the issue by replacing the `toISOString()` method for getting the date value.

### Checklist:

### DIFF
--- a/projects/components/src/datetime-picker/datetime-picker.component.ts
+++ b/projects/components/src/datetime-picker/datetime-picker.component.ts
@@ -53,11 +53,14 @@ export class DatetimePickerComponent implements OnChanges {
   }
 
   public getInputDate(): string {
-    if (!this.date) return '';
+    if (!this.date) {
+      return '';
+    }
     const d = this.date;
     const YEAR = d.getFullYear();
     const MONTH = d.getMonth() + 1 > 9 ? d.getMonth() + 1 : `0${d.getMonth() + 1}`;
     const DATE = d.getDate() > 9 ? d.getDate() : `0${d.getDate()}`;
+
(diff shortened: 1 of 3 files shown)


projects/components/src/datetime-picker/datetime-picker.component.ts
```diff
@@ -53,7 +53,12 @@ export class DatetimePickerComponent implements OnChanges {
   }
 
   public getInputDate(): string {
-    return this.date?.toISOString().slice(0, 10) ?? '';
+    if (!this.date) return '';
```
anandtiwary, line 56: @adisreyaj thanks for picking it up. Could we use `Time()` with `isUTC` set to true? Just trying to keep conversion at one place

adisreyaj: Will work if all the provided times in the dropdown uses `isUTC: true` and we use `setUTCHours()` instead of `setHours()` to patch the date when time changes.

anandtiwary: Ok. Can we do that then?

adisreyaj: Yup. On it!

anandtiwary (approved):

PR #1315, opened 2021-12-10
fix: setting max-height for multi select options container

arjunlalb:
## Description
Setting max-height for multi select options container.

### Testing
N/A.

### Checklist:


projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -89,6 +89,7 @@
   min-width: 120px;
   display: flex;
   height: 100%;
+  max-height: 360px;
```
anandtiwary, line 92: I think this should be driven by a component input. 360px might be too less in some cases

anandtiwary: also, we should make it scroll

arjunlalb: It already scrolls.

360px is good for about 10 items. Do we want the consumer to be able to control the height via an input? Currently, it can be done with css override.

jake-bassett (approved):
lgtm

palbizu (approved):
lgtm but I would wait for @anandtiwary review

projects/components/src/multi-select/multi-select.component.scss
anandtiwary: no. I meant we should be able to do it via code.

anandtiwary: umm. don't want to block the pr though

anandtiwary (approved):

PR #1316, opened 2021-12-13
feat: marker syncup among charts

jyothishjose6190:
marker syncup among charts

### DIFF
--- a/projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
+++ b/projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
@@ -19,6 +19,7 @@ import { DefaultChartTooltipRenderData } from '../utils/chart-tooltip/default/de
 import { MouseLocationData } from '../utils/mouse-tracking/mouse-tracking';
 import { Axis, AxisLocation, AxisType, Band, CartesianChart, RenderingStrategy, Series } from './chart';
 import { ChartBuilderService } from './chart-builder.service';
+import { ChartEvent } from './chart-interactivty';
 import { defaultXDataAccessor, defaultYDataAccessor } from './d3/scale/default-data-accessors';
 
 @Component({
@@ -61,8 +62,19 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
   @Input()
   public selectedInterval?: IntervalValue;
 
+  @Input()
+  public sync?: boolean;
+
+  @Input()
+  public groupId?: string;
+
   @Output()
   public readonly selectedIntervalChange: EventEmitter<IntervalValue> = new EventEmitter();
+
+  @Output()
+  public readonly mousePositionChange: EventEmitter<
+    MouseLocationData<TData, Series<TData> | Band<TData>>[]
+  > = new EventEmitter();
 
   @ViewChild('chartContainer', { static: true })
   public readonly container!: ElementRef;
@@ -87,13 +99,16 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
     }
 
     this.chart = this.chartBuilderService
-      .build<TData>(this.strategy, this.container.nativeElement, this.renderer)
+      .build<TData>(this.strategy, this.container.nativeElement, this.renderer, this.sync, this.groupId)
       .withSeries(...this.series)
       .withTooltip(
         this.chartTooltipBuilderService.constructTooltip<TData, Series<TData>>(data =>
           this.convertToDefaultTooltipRenderData(data)
         )
-      );
+      )
+      .withEventListener(ChartEvent.Hover, data => {
+        this.mousePositionChange.emit(data);
+      });
 
     if (this.bands) {
       this.chart.withBands(...this.bands);
--- a/projects/observability/src/shared/components/cartesian/chart-builder.service.ts
+++ b/projects/observability/src/shared/components/cartesian/chart-builder.service.ts
@@ -12,7 +12,22 @@ export class ChartBuilderService {
     protected readonly d3Utils: D3UtilService
   ) {}
 
-  public build<TData>(strategy: RenderingStrategy, element: Element, renderer: Renderer2): CartesianChart<TData> {
-    return new DefaultCartesianChart(element, this.injector, strategy, this.svgUtilService, this.d3Utils, renderer);
+  public build<TData>(
+    strategy: RenderingStrategy,
+    element: Element,
+    renderer: Renderer2,
+    sync?: boolean,
+    groupId?: string
+  ): CartesianChart<TData> {
+    return new DefaultCartesianChart(
+      element,
+      this.injector,
+      strategy,
+      this.svgUtilService,
+      this.d3Utils,
+      renderer,
+      sync,
+      groupId
+    );
   }
 }
--- a/projects/observability/src/shared/components/cartesian/chart-interactivty.ts
+++ b/projects/observability/src/shared/components/cartesian/chart-interactivty.ts
@@ -4,7 +4,8 @@ import { AxisType, Band, Series } from './chart';
 export const enum ChartEvent {
   Click,
   DoubleClick,
-  RightClick
+  RightClick,
+  Hover
 }
 
 export type ChartEventListener<TData> = (data: MouseLocationData<TData, Series<TData> | Band<TData>>[]) => void;
--- a/projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
+++ b/projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
@@ -76,7 +76,9 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     protected readonly renderingStrategy: RenderingStrategy,
     protected readonly svgUtilService: SvgUtilService,
     protected readonly d3Utils: D3UtilService,
-    protected readonly domRenderer: Renderer2
+    protected readonly domRenderer: Renderer2,
+    protected readonly sync?: boolean,
+    protected readonly groupId?: string
   ) {}
 
   public destroy(): this {
@@ -515,6 +517,15 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
 
   private onMouseMove(): void {
     const locationData = this.getMouseDataForCurrentEvent();
+
+    if (this.sync && this.groupId) {
+      this.eventListeners.forEach(listener => {
+        if (listener.event === ChartEvent.Hover) {
+          listener.onEvent(locationData);
+        }
+      });
+    }
+
     if (this.tooltip) {
       this.tooltip.showWithData(this.mouseEventContainer!, locationData);
     }
--- a/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
+++ b/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget-renderer.component.ts
@@ -27,6 +27,8 @@ import { CartesianDataFetcher, CartesianResult, CartesianWidgetModel } from './c
         [selectedInterval]="this.selectedInterval"
         [intervalOptions]="this.intervalOptions"
         [legend]="this.model.legendPosition"
+        [sync]="this.model.mouseSync"
+        [groupId]="this.model.syncGroupId"
         (selectedIntervalChange)="this.onIntervalChange($event)"
       >
       </ht-cartesian-chart>
--- a/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget.model.ts
+++ b/projects/observability/src/shared/dashboard/widgets/charts/cartesian-widget/cartesian-widget.model.ts
@@ -146,6 +146,18 @@ export class CartesianWidgetModel<TInterval> {
   })
   public maxSeriesDataPoints?: number;
 
+  @ModelProperty({
+    key: 'mouse-sync',
+    type: BOOLEAN_PROPERTY.type
+  })
+  public mouseSync?: boolean = false;
+
+  @ModelProperty({
+    key: 'sync-group-id',
+    type: STRING_PROPERTY.type
+  })
+  public syncGroupId?: string;
+
   @ModelInject(MODEL_API)
   private readonly api!: ModelApi;
 
--- a/src/app/home/home.dashboard.ts
+++ b/src/app/home/home.dashboard.ts
@@ -418,6 +418,9 @@ export const homeDashboard: DashboardDefaultConfiguration = {
               'show-grid-lines': true,
               'min-upper-limit': 25
             },
+            'mouse-sync': true,
+            'sync-group-id': 'HOME_DASHBOARD',
+
             series: [
               {
                 type: 'series',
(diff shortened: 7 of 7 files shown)


projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -61,9 +62,20 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
   @Input()
   public selectedInterval?: IntervalValue;
 
+  @Input()
+  public sync?: boolean;
```
aaron-steinfeld, line 66: API wise, I think this change makes sense with one suggestion - removing the boolean flag. It doesn't make sense to mark this true and not put a groupId, and likewise a groupId doesn't make sense if this isn't true, so the two properties are redundant - we can infer the sync behavior by the presence of a group ID.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -515,6 +517,15 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
 
   private onMouseMove(): void {
     const locationData = this.getMouseDataForCurrentEvent();
+
+    if (this.sync && this.groupId) {
```
aaron-steinfeld, line 521: We're tying two things together in the implementation in an unclear way - not only do these flags need to be set, we're assuming that setting them means we should fire on hover, and if these aren't set, the hover listeners get ignored. My suspicion is that at this level, we'd want a distinct listener rather than any of these three things. The group id can be built into it, and its presence indicates it should fire.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -87,13 +99,16 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
...
+      .build<TData>(this.strategy, this.container.nativeElement, this.renderer, this.sync, this.groupId)
       .withSeries(...this.series)
       .withTooltip(
         this.chartTooltipBuilderService.constructTooltip<TData, Series<TData>>(data =>
           this.convertToDefaultTooltipRenderData(data)
         )
-      );
+      )
+      .withEventListener(ChartEvent.Hover, data => {
+        this.mousePositionChange.emit(data);
```
aaron-steinfeld, line 110: Who listens to this? We'd still need some service or similar to receive the event and notify the other charts that are supposed to match

jyothishjose6190: working on the service part

projects/observability/src/shared/components/cartesian/chart-builder.service.ts
```diff
@@ -12,7 +13,22 @@ export class ChartBuilderService {
...
   ) {}
 
-  public build<TData>(strategy: RenderingStrategy, element: Element, renderer: Renderer2): CartesianChart<TData> {
-    return new DefaultCartesianChart(element, this.injector, strategy, this.svgUtilService, this.d3Utils, renderer);
+  public build<TData>(
+    strategy: RenderingStrategy,
+    element: Element,
+    renderer: Renderer2,
+    groupId?: string,
+    chartSyncService?: ChartSyncService<TData>
```
aaron-steinfeld, line 21: This need not be passed in, the whole injector is already in the chart so it can look it up if needed without the extra arg.

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Band, Series } from '../../../public-api';
+import { MouseLocationData } from '../utils/mouse-tracking/mouse-tracking';
+
+@Injectable({ providedIn: 'root' })
+export class ChartSyncService<TData> implements OnDestroy {
+  public readonly locationChangeSubject: Subject<
+    MouseLocationData<TData, Series<TData> | Band<TData>>[]
+  > = new Subject();
+
+  public mouseLocationChange(data: MouseLocationData<TData, Series<TData> | Band<TData>>[]): void {
```
aaron-steinfeld, line 12: Where does the group id come in? neither the publisher nor subscriber is using the group id

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable, OnDestroy } from '@angular/core';
+import { Subject } from 'rxjs';
+import { Band, Series } from '../../../public-api';
+import { MouseLocationData } from '../utils/mouse-tracking/mouse-tracking';
+
+@Injectable({ providedIn: 'root' })
+export class ChartSyncService<TData> implements OnDestroy {
+  public readonly locationChangeSubject: Subject<
```
aaron-steinfeld, line 8: Only expose observables, not subjects (for encapsulation)

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,19 @@
...
+    MouseLocationData<TData, Series<TData> | Band<TData>>[]
+  > = new Subject();
+
+  public mouseLocationChange(data: MouseLocationData<TData, Series<TData> | Band<TData>>[]): void {
+    this.locationChangeSubject.next(data);
+  }
+
+  public ngOnDestroy(): void {
```
aaron-steinfeld, line 16: a root injectable doesn't have a destroy lifecycle hook - it lasts the full lifecycle of the application.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -515,6 +533,16 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
 
   private onMouseMove(): void {
     const locationData = this.getMouseDataForCurrentEvent();
+
+    // tslint:disable-next-line
+    if (this.groupId) {
```
aaron-steinfeld, line 538: group id should only be used when deciding how to interact with the sync service, not at the event level itself (a hover is a hover, whether it belongs to a group or not).

projects/observability/src/shared/components/cartesian/chart-interactivty.ts
```diff
@@ -8,9 +8,16 @@ export const enum ChartEvent {
...
+  data: MouseLocationData<TData, Series<TData> | Band<TData>>[] | CartesianHoverData<TData>
+) => void;
 
 export interface ChartTooltipTrackingOptions {
   followSingleAxis?: AxisType;
   radius?: number;
 }
+
+export interface CartesianHoverData<TData> {
+  groupId?: string;
```
aaron-steinfeld, line 21: This isn't relevant to the hovering, just the syncing - make this a separate arg to the call on chart sync service (that way we can guarantee it's present)

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -1,19 +1,22 @@
...
-    this.locationChangeSubject.next(data);
-  }
+  public locationChange$: Observable<
+    MouseLocationData<TData, Series<TData> | Band<TData>>[] | CartesianHoverData<TData>
+  > = this.locationChangeSubject.asObservable();
 
-  public ngOnDestroy(): void {
-    this.locationChangeSubject.complete();
+  public mouseLocationChange(
+    data: MouseLocationData<TData, Series<TData> | Band<TData>>[] | CartesianHoverData<TData>
```
aaron-steinfeld, line 18: see other comment. In addition, accept the the groupId rather than exposing the observable directly, that way we can give a filtered response stream and leave the implementation out of the API.

e.g. (type on observable isn't quite right, but hopefully conveys the idea)

```ts
public getLocationChangesForGroup<TData>(groupId: string): Observable<MouseLocationData<TData>>)
```

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -81,8 +86,12 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     protected readonly groupId?: string,
     private readonly chartSyncService?: ChartSyncService<TData>
   ) {
-    this.chartSyncService?.locationChangeSubject.subscribe(data => {
-      this.showTooltip(data);
+    this.chartSyncService?.locationChange$.subscribe(data => {
```
aaron-steinfeld, line 89: Should only be subscribed to if we have an ID.

projects/observability/src/shared/components/utils/chart-tooltip/chart-tooltip-popover.ts
```diff
@@ -11,7 +11,7 @@ export class ChartTooltipPopover<TData, TContext> implements ChartTooltipRef<TDa
   ) {}
 
   public showWithData(element: Element, data: MouseLocationData<TData, TContext>[]): void {
```
aaron-steinfeld, line 13: Since we're not showing the tooltip in the sync'd charts (discussed on slack), no changes should be needed in this file or any of the other popover files.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -76,8 +82,26 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
-  ) {}
+    protected readonly domRenderer: Renderer2,
+    protected readonly groupId?: string,
+    private readonly chartSyncService?: ChartSyncService<TData>
+  ) {
+    this.chartSyncService?.locationChange$.subscribe(data => {
+      const hoverData = data as CartesianHoverData<TData>;
+
+      if (this.groupId === hoverData.groupId) {
+        this.showTooltip(hoverData.selectedData);
```
aaron-steinfeld, line 93: sync crosshairs only.

jyothishjose6190:
https://github.com/hypertrace/hypertrace-ui/pull/1316#discussion_r796073079
 @aaron-steinfeld groupid should be sent from the primary chart along with the current location. right?. Here we are sending from the nonprimary charts. How can we compare wether these are same or not?

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
aaron-steinfeld: > @aaron-steinfeld groupid should be sent from the primary chart along with the current location. right?. Here we are sending from the nonprimary charts. How can we compare whether these are the same or not?

@jyothishjose6190 there is no "primary chart" unless you mean the one being hovered. The one being hovered would broadcast, across the chart sync service that its groupId is being hovered at some location. Here, we're listening - which we only need to do if we have a group ID, because we only want to listen for events relevant to our group ID.

projects/components/src/popover/popover.ts
```diff
@@ -34,6 +34,7 @@ export interface PopoverRelativePosition {
 export interface PopoverFixedPosition {
   type: PopoverPositionType.Fixed;
   location: PopoverFixedPositionLocation;
+  boundingElement?: Element;
```
aaron-steinfeld, line 37: no longer used

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -94,17 +99,26 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
...
       )
+      .withEventListener(ChartEvent.Hover, data => {
+        this.chartSyncService.mouseLocationChange(data, this.groupId);
+      })
       .withEventListener(ChartEvent.Select, selectedData => {
         this.selectionChange.emit(selectedData);
       });
 
+    if (this.groupId !== undefined) {
+      this.chartSyncService?.getLocationChangesForGroup(this.groupId).subscribe(data => {
```
aaron-steinfeld, line 117: use subscription lifecycle to unsubscribe on destroy

aaron-steinfeld: Actually, do we need this here? The comment about unsubscribing applies no matter where we do it, but the chart already has its group ID so if we move this into the chart code instead of the component, we don't have to expose the showCrosshair method.

projects/observability/src/shared/components/cartesian/chart.ts
```diff
@@ -16,6 +16,8 @@ export interface CartesianChart<TInterval> {
   withIntervalData(intervalData: CartesianIntervalData): this;
   withTooltip(tooltip: ChartTooltipRef<TInterval>): this;
   withTimeRange(timeRange: TimeRange): this;
+  // tslint:disable-next-line
+  showCrosshair(locationData: any): void;
```
aaron-steinfeld, line 20: See comment about avoiding exposing this; also no any.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -79,9 +79,17 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     protected readonly renderingStrategy: RenderingStrategy,
     protected readonly svgUtilService: SvgUtilService,
     protected readonly d3Utils: D3UtilService,
-    protected readonly domRenderer: Renderer2
+    protected readonly domRenderer: Renderer2,
+    protected readonly groupId?: string
   ) {}
 
+  // tslint:disable-next-line
```
aaron-steinfeld, line 86: Do not disable lint rules out of convenience - if it needs to be disabled, there needs to be a reason - and no-any never has a good reason.

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Band, CartesianSelectedData, Series } from '../../../public-api';
+import { MouseLocationData } from '../utils/mouse-tracking/mouse-tracking';
+
+@Injectable({ providedIn: 'root' })
+export class ChartSyncService<TData> {
+  private readonly locationChangeSubject: Subject<
+    MouseLocationData<TData, Series<TData> | Band<TData>>[] | CartesianSelectedData<TData>
+  > = new Subject();
+
+  private groupId?: string;
```
aaron-steinfeld, line 13: No need to store state in two places even if correct, it's making assumptions about ordering that aren't necessary. It's cleaner to hold this inside the subject e.g. `{groupId: string, locationData: ... }`, then filter against the pipeline value rather than reaching into service state as line 28 is currently doing.

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,33 @@
...
+  ): void {
+    this.locationChangeSubject.next(data);
+    this.groupId = groupId;
+  }
+
+  public getLocationChangesForGroup(
+    groupId: string
+  ): Observable<MouseLocationData<TData, Series<TData> | Band<TData>>[] | CartesianSelectedData<TData>> {
+    const selectedData = this.locationChangeSubject.pipe(
+      switchMap(async results => results),
```
aaron-steinfeld, line 27: What's the purpose of this line? results isn't actually async so this is just an identity map.

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,33 @@
...
+      filter(_result => this.groupId === groupId)
+    );
+
+    return selectedData;
```
aaron-steinfeld, line 31: nit: unnecessary variable

jyothishjose6190:
https://github.com/hypertrace/hypertrace-ui/pull/1454

PR #1333, opened 2021-12-21
feat: disabled state for input component reactive forms

itssharmasandeep:
## Description
By this, we can set disabled state directly from form control for input component (which is recommended in reactive forms)

### Testing
Manual testing done!!!


palbizu (approved):
It looks fine but I will check it with @anandtiwary

anandtiwary (approved):

PR #1343, opened 2021-12-28
Adding inputs to time picker

palbizu:
## Description
Adding new inputs in TimePicker:
- Disabled : boolean
- Display Mode : DisplayMode (with-border/with-background)
- IconSize: defines chevron size (Small/Extra-small)

### Testing
Visual Testing

image

image

image

### Checklist:

### Documentation


projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -63,7 +63,11 @@ import { MultiSelectJustify } from './multi-select-justify';
                 <span *ngIf="triggerValues.selectedItemsCount > 1" class="trigger-more-items"
                   >+{{ triggerValues.selectedItemsCount - 1 }}</span
                 >
-                <ht-icon class="trigger-icon" icon="${IconType.ChevronDown}" size="${IconSize.Small}"></ht-icon>
+                <ht-icon
+                  class="trigger-icon"
+                  icon="${IconType.ChevronDown}"
+                  [size]="this.size === '${IconSize.ExtraSmall}' ? '${IconSize.ExtraSmall}' : '${IconSize.Small}'"
```
arjunlalb, line 69: why not use `this.size` directly?

palbizu: because I would be modifying the regular case that already uses Small size as a regular

anandtiwary: `[size]="this.size"` should also work right?

anandtiwary: hmm.. you are comparing and assigning`SelectSize` with IconSize. that seems incorrect.

anandtiwary: This component also has iconSize btw.

projects/components/src/time-picker/time-picker.component.ts
```diff
@@ -9,8 +9,8 @@ import { PredefinedTimeService } from '../time-range/predefined-time.service';
   styleUrls: ['./time-picker.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="time-picker">
-      <ht-popover class="time-selector" [closeOnClick]="true">
+    <div class="time-picker" [ngClass]="[this.disabled ? 'disabled' : '']">
```
anandtiwary, line 12: `[ngClass]="{'disabled': this.disabled}"`

projects/components/src/form-field/form-field.component.ts
```diff
@@ -65,6 +65,9 @@ export class FormFieldComponent {
   @Input()
   public showBorder: boolean = false;
 
+  @Input()
+  public showOnlyErrorBorder: boolean = false;
```
anandtiwary, line 69: why this change?

palbizu: To show only red borders

palbizu: To show only red borders

anandtiwary: We need consistent behavior across forms. Why do we want to show borders but not error message?

palbizu: Yes, I agree but exist some cases where it would be good to show the red borders without messages, for example when I want to show two nested fields with the same message.

projects/components/src/multi-select/multi-select.component.ts
palbizu: thanks, done

projects/components/src/time-picker/time-picker.component.ts
palbizu: done

projects/components/src/form-field/form-field.component.ts
anandtiwary: In that case, we should be combining two fields inside the same form-field component

palbizu: So should I use an object (with a pair of variables inside) instead of showing the input separatly?

palbizu: showOnlyErrorBorder  and showTransparentBackground were Removed

jake-bassett (approved):

PR #1352, opened 2022-01-05
feat: datetime picker as form field

itssharmasandeep:
## Description
1. Datetime picker as form field
2. Text area as form field
3. Min width for time picker (style looks weird when value is null)
4. no error border when showBorder is false

### Testing
Manual testing done!!

### Checklist:

### DIFF
--- a/projects/components/src/form-field/form-field.component.ts
+++ b/projects/components/src/form-field/form-field.component.ts
@@ -23,7 +23,7 @@ import { IconSize } from '../icon/icon-size';
         class="content"
         [ngClass]="{
           'show-border': this.showBorder,
-          'error-border': this.showBorder && this.showFormError && this.errorLabel
+          'error-border': this.showFormError && this.errorLabel
         }"
       >
         <ng-content></ng-content>
--- a/projects/components/src/form-field/form-field.component.ts
+++ b/projects/components/src/form-field/form-field.component.ts
@@ -23,7 +23,7 @@ import { IconSize } from '../icon/icon-size';
         class="content"
         [ngClass]="{
           'show-border': this.showBorder,
-          'error-border': this.showFormError && this.errorLabel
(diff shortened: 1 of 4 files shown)


projects/components/src/form-field/form-field.component.ts
```diff
@@ -23,7 +23,7 @@ import { IconSize } from '../icon/icon-size';
         class="content"
         [ngClass]="{
           'show-border': this.showBorder,
-          'error-border': this.showFormError && this.errorLabel
+          'error-border': this.showBorder && this.showFormError && this.errorLabel
```
anandtiwary, line 26: We don't need both.

projects/components/src/time-picker/time-picker.component.scss
```diff
@@ -1,23 +1,14 @@
...
   display: flex;
 
   .time-selector {
     @include body-1-medium($gray-7);
     height: 32px;
+    border: 1px solid $gray-1;
+    background: $gray-1;
     border-radius: 4px;
-
-    &.with-background {
```
anandtiwary, line 12: Why are we removing them?

itssharmasandeep: Done by mistake, reverted

projects/components/src/form-field/form-field.component.ts
itssharmasandeep: So I want to show form error label but not error-border (because it does not look good for datetime picker), since `this.showFormError` is also used for error label I cannot use that

anandtiwary: This could have some side effects

anandtiwary: Why does it look bad? If double borders are showing up then we need to remove border from date-picker through an input option

anandtiwary (approved):

PR #1357, opened 2022-01-07
feat: providing functionality to use else template in if feature directive

itssharmasandeep:
## Description
After this, will be able to provide else template for `*htIfFeature`, similar to `*ngIf`
Example
```
<ng-container *htIfFeature="state as featureState; else elseTemplate">
  <div class="test-div">{{featureState}}</div>
</ng-container>
<ng-template #elseTemplate>
  <div class="else-div">Else template</div>
</ng-template>
```

### Testing
Manual testing done.

### Checklist:

### DIFF
--- a/projects/components/src/feature-check/if-feature.directive.ts
+++ b/projects/components/src/feature-check/if-feature.directive.ts
@@ -9,7 +9,11 @@ export class IfFeatureDirective implements OnChanges {
   @Input('htIfFeature')
   public featureState?: FeatureState;
 
-  private embeddedViewRef?: EmbeddedViewRef<FeatureFlagsContext>;
+  // tslint:disable-next-line:no-input-rename
+  @Input('htIfFeatureElse')
+  public else?: TemplateRef<unknown>;
+
+  private embeddedViewRef?: EmbeddedViewRef<FeatureFlagsContext | unknown>;
   private readonly context: FeatureFlagsContext = {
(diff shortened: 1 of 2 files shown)


projects/components/src/feature-check/if-feature.directive.ts
```diff
@@ -9,7 +9,11 @@ export class IfFeatureDirective implements OnChanges {
   @Input('htIfFeature')
   public featureState?: FeatureState;
 
-  private embeddedViewRef?: EmbeddedViewRef<FeatureFlagsContext>;
+  // tslint:disable-next-line:no-input-rename
```
anandtiwary, line 12: why is this needed?

anandtiwary (commented):
LGTM. but let's wait for @aaron-steinfeld to take a look as well

projects/components/src/feature-check/if-feature.directive.ts
itssharmasandeep: It gives lint error
```
In the class "IfFeatureDirective", the directive input property "elseContent" should not be renamed.
However, you should use an alias when the directive name is also an input property, and the directive name
doesn't describe the property. In this last case, you can disable this rule with `tslint:disable-next-line:no-input-rename`.
```

aaron-steinfeld (approved):

PR #1358, opened 2022-01-08
Add Nx build tool to Hypertrace UI

anandtiwary:
## Description
Please include a summary of the change, motivation, and context.
Migrating Hypertrace to NX

- Using NX 12
- Staying on Tslint for now. Separate PR will be created for migration to ESLint
- Renamed `projects` to `libs` to go with nx standard
- Each lib is now a light-weight buildable library using ng-packagr-lite
- Ng serve will source compile and run all lib files instead of building them as a separate library. This is OOTB with NX
- Migrated jest to the newer version. Fixed all test errors.
- Added relevant code for the `affected` command. I will propose and create a separate PR for using this in PR build mode
- Since tests are now split at the lib boundary, I had to remove the existing test batching (which requires all tests to be run via the test app command). Will revisit and bring back this intelligent test batching with a future pr.
- NX cloud is enabled (since `ht-ui` is an OS project)
- Please run `nx affected:*` command to only run the command in projects with the dependent code changes. This will save time! 
- Easy-to-use npm scripts are also added.
- Caching is disabled in CI as of now. 

Future PRs:
- migrate to eslint
- migrate to the latest version of NX. This will also migrate to the latest versions of dependent libs like angular, jest, eslint etc

### Testing

### Checklist:

### Documentation


arjunlalb (commented):
Not sure if this is intended or if this is github's interpretation, 

[c/back-button/back-button.component.scss → apps/.gitkeep](https://github.com/hypertrace/hypertrace-ui/pull/1358/files#diff-6c8af18d3430e2017bd40adfe7385a3556b26865c9d0d3b9e7aaebb6fa2d1a73)

libs/components/src/file-upload/file-display/file-display.ts
```diff
@@ -0,0 +1,6 @@
+export const enum FileUploadState {
```
arjunlalb, line 1: why are some of these changes in here?

arjunlalb (commented):
LGTM overall.

libs/components/src/file-upload/file-display/file-display.ts
anandtiwary: I will have to check

anandtiwary: reverted

libs/components/src/table/table.component.test.ts
```diff
@@ -457,57 +457,13 @@ describe('Table component', () => {
...
-          selections: [statefulRows[0]]
-        }
-      }
-    );
-
-    expect(spectator.component.shouldHighlightRowAsSelection(statefulRows[0])).toBeTruthy();
-    expect(spectator.component.shouldHighlightRowAsSelection(statefulRows[1])).toBeFalsy();
-  });
-
-  test('row should be highlighted (even) when in multi selection mode', () => {
```
anandtiwary, line 489: these tests somehow pass in the existing code, but they have issues.

itssharmasandeep (approved):

itssharmasandeep (approved):

Christian862:
is this integrated or packaged based Nx ?

nx.json
```diff
@@ -0,0 +1,83 @@
...
+      "unitTestRunner": "jest"
+    },
+    "@nrwl/angular:component": {
+      "style": "scss"
+    }
+  },
+  "defaultProject": "hypertrace-ui",
+  "projects": {
+    "assets-library": {
+      "tags": []
```
Christian862, line 62: Are we going to use tags do you think ? might be useful to automatically enforce rules between libs

Christian862:
Why we're all of the config files left in hypertrace-ui/projects/... ?

nx.json
anandtiwary: This was generated by default from nx workspace generator. I left it as it is.

anandtiwary:
> Why we're all of the config files left in hypertrace-ui/projects/... ?

couldn't find any. All of the project configs are deleted in the PR.

Christian862:
> couldn't find any. All of the project configs are deleted in the PR

Oh, just a local WS thing then...

arjunlalb (approved):
Did some playing around in my local. LGTM.

nx.json
LayZeeDK: Tags are not needed to detect circular dependencies.

You wouldn't gain much from adding tags since the library projects are currently not very fine-grained. They need further splitting to achieve more gains from using Nx. This doesn't have to be done in this PR.

PR #1364, opened 2022-01-10
fix: Fixing Time picker comparison time in dropdown

palbizu:
## Description
Fixing Time comparison

### Testing
Tested locally

### Checklist:

### Documentation

### DIFF
--- a/projects/common/src/time/time.test.ts
+++ b/projects/common/src/time/time.test.ts
@@ -14,4 +14,9 @@ describe('Time', () => {
   test('can get ISO String from time', () => {
     expect(mockedTime.toISOString()).toBe('08:30:00.000Z');
   });
+
+  test('can compare two times', () => {
+    expect(mockedTime.equals(new Time(8, 30))).toBeTruthy();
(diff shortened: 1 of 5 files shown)


projects/common/src/time/time.test.ts
```diff
@@ -14,4 +14,9 @@ describe('Time', () => {
   test('can get ISO String from time', () => {
     expect(mockedTime.toISOString()).toBe('08:30:00.000Z');
   });
+
+  test('can compare two times', () => {
+    expect(mockedTime.equals(new Time(8, 30))).toBeTruthy();
```
aaron-steinfeld, line 19: Hmm, trying to think if there's any way to test the UTC vs local comparison, given that the tests already run in UTC. I think there are some complicated ways, but it's not worth it... The complexity is due to limitations in our Time class, which would likely be easier to fix rather than figure out how to test.

(sorry, just rambling thoughts here, nothing to do)

aaron-steinfeld (approved):

Issue #1099, opened 2021-08-31
Support for grouping explorer results by custom tags


subintp:
## Use Case
 Visualize traffic patterns like calls count, P99 latency, error across IP, user_id, api_key, etc via the explorer tab. 
 
 This will help us debug issues like 
   1. Latency/Errors spike specific to a user or customer
   2. Request spikes from particular IPs or API keys
  
   
 
## Proposal
![Screenshot 2021-09-01 at 12 51 00 AM](https://user-images.githubusercontent.com/1742357/131563823-dfbf4f02-c3f6-4ef9-a128-e8e06e6cd0ee.png)

The above use cases can be solved in a generic fashion by grouping explorer query results based on custom tags. Currently, there is no way we can solve this issue via prometheus+grafana due to the high cardinality of these tags. Addition of feature will be makes explorer tab more powerful

kotharironak:
Currently, we are storing the span tags in a MAP attribute which is eventually stored in pinot as MAP column (two multi-valued arrays). On the MAP attribute, we do have support for `lookup` -> `Contains_Key` and `key value comparison -> `Contains_key_value`, and they are translated to pinot's mapValue queries in the query service layer. Similarly, we will need the support in translating `group by` query expression involving `mapValue`. 

@sarthak77 Can you look into the query service side of changes as part of this ticket?

kotharironak:
To handle the container object like map for a corresponding attribute (e.g spanTags - https://github.com/hypertrace/query-service/blob/main/query-service/src/main/resources/configs/common/application.conf#L75), I was thinking if we can extend query service expression definition (https://github.com/hypertrace/query-service/blob/main/query-service-api/src/main/proto/request.proto#L11) as bellow to handle transporting the information.

```
message Expression {
  oneof value {
    ColumnIdentifier columnIdentifier = 1;
    LiteralConstant literal = 2;
    Function function = 3;
    OrderByExpression orderBy = 4;
    ObjectIdentifier = 5;
  }
}

message ObjectIdentifier {
  string columnName = 1; // 1 & 2 can be replaced with ColumnIdentifier
  string alias = 2;
  string path_key = 3; // or path_expression
}
```

With the above, we will be able to handle containe_key_value as below.
.eg existing contain_key_value expression
```
childFilter {
        lhs {
          columnIdentifier {
            columnName: "API_TRACE.tags"
          }
        }
        operator: CONTAINS_KEYVALUE
        rhs {
          literal {
            value {
[...]

PR #1368, opened 2022-01-12
Explorer subpath support

aaron-steinfeld:
## Description
Part 2 of #1099 UI support, adding filtering support for attribute expressions.

### Testing
E2E, updated units tests.


aaron-steinfeld:
@jake-bassett this would be a good one for you to take a look at, all (or at least most of) the filter changes are in here

projects/observability/src/shared/graphql/model/schema/filter/graphql-filter.ts
```diff
@@ -23,6 +23,5 @@ export const enum GraphQlOperatorType {
   Like = 'LIKE',
   In = 'IN',
   NotIn = 'NOT_IN',
-  ContainsKey = 'CONTAINS_KEY',
-  ContainsKeyValue = 'CONTAINS_KEY_VALUE'
+  ContainsKey = 'CONTAINS_KEY'
```
jake-bassett, line 26: Aw, was hoping we could get rid of this operator too. Maybe in the future.

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.component.ts
```diff
@@ -120,8 +120,7 @@ export class FilterChipComponent implements OnInit, OnChanges {
   private mapToComboBoxOption(filter: IncompleteFilter): ComboBoxOption<IncompleteFilter> {
     return {
       text: filter.userString,
-      value: filter,
-      tooltip: `${filter.userString} (${filter.field})`
+      value: filter
```
jake-bassett, line 123: How are tooltips now handled? Or at least, how do we let users know the underlying field backing the display name?

projects/observability/src/shared/graphql/model/schema/filter/graphql-filter.ts
aaron-steinfeld: This is its own thing, it's distinct functionality unique to maps. It no longer takes an array though, so it's just like "Tags CONTAINS_KEY span.kind"

projects/components/src/filtering/filter-bar/filter-chip/filter-chip.component.ts
aaron-steinfeld: I actually removed them, this was one of the things I wanted to ask. In the existing UI, they're:
- The underlying field isn't actually used for matching a user's string that I could tell
- hard to find, since you have to hover on the popup
- very strange to read, since this key is thrown after the user string

Example from existing ui: 
image
image

jake-bassett: Fair points. Knowing the underlying keys was an ask early on, so we've always had the tooltip available. Not sure if users are even asking for that info. If they are, I'm surprised we haven't gotten feedback about the discoverability of it as you've noted.

I'm fine removing them as long as its a quick fix to add back if an issue arrises.

aaron-steinfeld: Yeah I remember the ask, but not sure why it was there given that we don't expose keys to users anywhere else. Easy enough to restore if the requirement comes back - hopefully can better understand the ask at that point.

arjunlalb (commented):
LGTM. Will wait for @jake-bassett to give a sign off on this since he has the most context around the filter bars.

jake-bassett (approved):

PR #1373, opened 2022-01-14
feat: Skeleton loader for LoadAsync directive

Christian862:
## Description
To finish the work of the LoadAsync directive, the SkeletonComponent is used to provide loading animations. It has the option of a few different shapes. Specify how many times to repeat the skeleton component by providing the `repeatCount` to the `LoadAsyncConfig` input - Note, as of this v1, repeatCount should only be used with `LoaderType.RectangleText`, `LoaderType.TableRow` and `LoaderType.ListItem`.  

### Testing
Updated LoaderComponent test to cover each possible new LoaderType and testing to ensure old LoaderType is used by default.  Added testing for SkeletonComponent, testing the expected default behaviour, testing the repeat input behaviour and testing the correct display for the respective `SkeletonType`. For the last test in SkeletonComponent, 'Should match the skeleton type to the corresponding element', i'm not sure if this is the best approach, thoughts ? 

### Checklist:

### Documentation


projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  @Input() public shapeStyle: SkeletonType = SkeletonType.Rectangle;
+
+  @Input() public repeat: number = 1;
+
+  @Input() public size: string = '';
+
+  public containerClass(): ContainerClass {
+    return {
+      skeleton: true,
+      'skeleton-rectangle': this.shapeStyle === 'rectangle',
```
arjunlalb, line 56: use the enum on the RHS.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  styleUrls: ['./skeleton.component.scss']
+})
+export class SkeletonComponent {
+  @Input() public shapeStyle: SkeletonType = SkeletonType.Rectangle;
+
+  @Input() public repeat: number = 1;
+
+  @Input() public size: string = '';
+
+  public containerClass(): ContainerClass {
```
arjunlalb, line 53: rename to `getContainerClass`

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,65 @@
...
+    return {
```
arjunlalb, line 54: This can be a property, updated when `shapeStyle` or `repeat` values are changed.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,65 @@
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+
+interface ContainerClass {
+  skeleton: boolean;
+  'skeleton-rectangle': boolean;
```
arjunlalb, line 5: use camelCasing for the property keys. This will also help you avoid string usages in L56-62.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+
+@mixin initial-location {
+  top: initial;
+  left: initial;
+  transform: initial;
+}
+
+.skeleton {
+  position: relative;
+  background-color: #dee2e6;
```
arjunlalb, line 22: use a color value from `_color-palette.scss`. If the required color is not defined, either use a suitable closest approximation or add an appropriate value there and consume it here. No explicit hex usages.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+.skeleton.skeleton-square {
+  width: 3rem;
+  height: 3rem;
+}
+
+.skeleton.skeleton-table-row {
+  height: 1.3rem;
+  width: 90%;
+
+  @include initial-location;
```
arjunlalb, line 61: inclusions should be in the beginning of a css rule definition so that overrides are clear.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+}
+
+.skeleton.skeleton-repeating {
+  margin-top: 1rem;
+}
+
+.skeleton.skeleton-list-item {
+  background-color: rgb(255, 255, 255);
+  @include initial-location;
```
arjunlalb, line 70: same comment here. inclusions in the beginning.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  .item-column {
+    display: flex;
+    flex-direction: column;
+    padding-left: 1rem;
+    width: 100%;
+  }
+
+  .item-column > div {
```
arjunlalb, line 91: do not use element styling. apply a class to the children and style the class.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+    background-color: #dee2e6;
```
arjunlalb, line 92: no explicit color hex codes.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+}
+
+.skeleton.skeleton-list-item {
+  background-color: rgb(255, 255, 255);
+  @include initial-location;
+  display: flex;
+  justify-content: flex-start;
+
+  .item-circle {
+    background-color: #dee2e6;
```
arjunlalb, line 75: no explicit color hex codes.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+
+  @include initial-location;
+}
+
+.skeleton.skeleton-repeating {
+  margin-top: 1rem;
+}
+
+.skeleton.skeleton-list-item {
+  background-color: rgb(255, 255, 255);
```
arjunlalb, line 69: use a color constant.

projects/components/src/skeleton/skeleton.component.scss
```diff
@@ -0,0 +1,114 @@
...
+  transform: translate(-50%, -50%);
+
+  border-radius: 6px;
+}
+
+.skeleton::after {
+  @include loading-animation;
+}
+
+.skeleton.skeleton-rectangle {
```
arjunlalb, line 36: instead of enumerating all child classes this way, you can do the following
```
.skeleton {
  &.skeleton-rectangle {}
  &.skeleton-rectangle-text {}
  ...
}
```

projects/components/src/load-async/load-async.module.ts
```diff
@@ -2,13 +2,14 @@ import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 import { IconModule } from '../icon/icon.module';
 import { MessageDisplayModule } from '../message-display/message-display.module';
+import { SkeletonModule } from '../skeleton/skeleton.module';
```
aaron-steinfeld, line 5: sounds intense ☠️

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -33,4 +47,29 @@ export class LoaderComponent implements OnChanges {
         return ImagesAssetPath.LoaderSpinner;
     }
   }
+
+  public loaderToSkeletonTypeMap(curLoaderType: LoaderType): SkeletonType {
```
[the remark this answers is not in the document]
aaron-steinfeld: that suggestion works for me, but map is really the verb here so `mapLoaderToSkeletonType` would also be fine.

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -33,4 +47,29 @@ export class LoaderComponent implements OnChanges {
...
+      case LoaderType.Circle:
+        return SkeletonType.Circle;
+      case LoaderType.Square:
+        return SkeletonType.Square;
+      case LoaderType.TableRow:
+        return SkeletonType.TableRow;
+      case LoaderType.ListItem:
+        return SkeletonType.ListItem;
+      default:
+        return SkeletonType.Rectangle;
```
aaron-steinfeld, line 64: We don't expect to hit this, so assertUnreachable would be a good fit here.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  Square = 'square',
+  Circle = 'circle',
+  TableRow = 'table-row',
+  ListItem = 'list-item'
+}
+
+@Component({
+  selector: 'ht-skeleton',
+  template: `
+    <ng-container *ngFor="let k of [].constructor(this.repeat)">
```
aaron-steinfeld, line 26: Can we move this array construction into TS for readability (ideally just do something like `this.repeat && new Array(this.repeat)` Also, be careful with undefined values here.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      <ng-container [ngSwitch]="this.shapeStyle">
+        <div *ngSwitchCase="'${SkeletonType.RectangleText}'" [ngClass]="containerClass()"></div>
```
aaron-steinfeld, line 28: Careful with function calls, especially ones that return new objects like this. Instead, just calculate the class dictionary in ngOnChanges.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      </ng-container>
+    </ng-container>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  styleUrls: ['./skeleton.component.scss']
+})
+export class SkeletonComponent {
+  @Input() public shapeStyle: SkeletonType = SkeletonType.Rectangle;
+
+  @Input() public repeat: number = 1;
```
aaron-steinfeld, line 49: In loader we're assigning an optional field to this input, make sure to handle undefined and fix types.

projects/components/src/load-async/loader/loader.component.ts
Christian862: wouldn't we hit it if the input loader type is either ExpandableRow, Page, or Spinner ? I attempted adding it but it wouldn't let me pass in `curLoaderType` : 

` Argument of type 'import("/Users/christian/traceable-ui/hypertrace-ui/projects/components/src/load-async/load-async.service").LoaderType' is not assignable to parameter of type 'never'`

aaron-steinfeld: I missed that not every case was listed here, but that's actually the point of assertUnreachable - it will give that compiler error you saw if a case isn't handled. So what I'd suggest is changing
```ts
      default:
        return SkeletonType.Rectangle;
```
to 
```ts
      case LoaderType.ExpandableRow:
      case LoaderType.Page:
      case LoaderType.Spinner:
        return SkeletonType.Rectangle;  
      default:
        return assertUnreachable(curLoaderType);
```

Although the behavior is currently identical, when a new loader type is added we're forcing the author to come look at this switch statement and make a decision on how to map it, rather than them potentially incorrectly being mapped to the default case.

projects/components/src/skeleton/skeleton.component.ts
Christian862: These property names are used as class names, does that still apply ? camel case class names ?

projects/components/src/skeleton/skeleton.component.ts
Christian862: Could you elaborate on 'fix types' ?

aaron-steinfeld: we're assigning something of type `number | undefined`, so we should declare that here - that'll force the compiler to keep us honest.

```suggestion
  @Input() 
  public repeat: number | undefined = 1;
```

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,25 +1,47 @@
...
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="ht-loader">
-      <img [ngClass]="[this.currentLoaderType]" [src]="this.getImagePathFromType(this.currentLoaderType)" />
+    <div class="ht-loader" [ngClass]="{ 'flex-centered': this.isOldLoaderType }">
+      <ng-container *ngIf="!this.isOldLoaderType; else oldLoader">
+        <ht-skeleton [shapeStyle]="this.skeletonType" [repeat]="this.repeatLoaderCount"></ht-skeleton>
+      </ng-container>
+
+      <ng-template #oldLoader>
```
arjunlalb, line 17: template names should be suffixed with `template`

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,25 +1,47 @@
...
+        <img [ngClass]="[this.currentLoaderType]" [src]="this.getImagePathFromType(this.currentLoaderType)" />
```
arjunlalb, line 18: since the image path is dependent on `currentLoaderType`, this can also be a property computed in `ngOnChanges`. Alternatively, use `htMemoize` here

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,25 +1,47 @@
...
+
   public ngOnChanges(): void {
     this.currentLoaderType = this.loaderType ?? LoaderType.Spinner;
+
+    this.isOldLoaderType =
+      this.currentLoaderType === LoaderType.ExpandableRow ||
+      this.currentLoaderType === LoaderType.Spinner ||
+      this.currentLoaderType === LoaderType.Page;
+
+    this.skeletonType = this.getSkeletonTypeForLoader(this.currentLoaderType);
```
arjunlalb, line 44: we need these to be recomputed only if `loaderType` input has changed. Add a condition to check for that around these statements.

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,25 +1,47 @@
...
+      </ng-template>
     </div>
   `
 })
 export class LoaderComponent implements OnChanges {
   @Input()
   public loaderType?: LoaderType;
 
+  @Input()
+  public repeatLoaderCount?: number;
```
arjunlalb, line 28: nit: name can be shortened to `repeatCount`

projects/components/src/load-async/loader/loader.component.test.ts
```diff
@@ -43,8 +49,132 @@ describe('Loader component', () => {
...
+    expect(spectator.query(SkeletonComponent)).toHaveAttribute('shapeStyle', SkeletonType.Square);
+    expect(spectator.query(SkeletonComponent)).toHaveAttribute('repeat', '1');
+  });
+
+  test('Should use corresponding skeleton component for loader type table row', () => {
+    spectator = createHost(
+      `<ht-loader [loaderType]="'${LoaderType.TableRow}'" [repeatCount]="repeatCount"></ht-loader>`,
+      {
+        hostProps: {
+          repeatCount: '4'
```
anandtiwary, line 138: this should be a number.

projects/components/src/load-async/loader/loader.component.test.ts
```diff
@@ -43,8 +49,132 @@ describe('Loader component', () => {
...
+      {
+        hostProps: {
+          repeatCount: '1'
+        }
+      }
+    );
+
+    expect(spectator.query('.ht-loader')).toExist();
+    expect(spectator.query('.ht-loader')).not.toHaveClass('flex-centered');
+    expect(spectator.query(SkeletonComponent)).toExist();
```
anandtiwary, line 176: nit: can store it as a local variable to avoid multiple dom query.

projects/components/src/load-async/loader/loader.component.ts
anandtiwary:
```
case LoaderType.ExpandableRow: 
case LoaderType.Page: 
case LoaderType.Spinner: 
return SkeletonType.Rectangle;
```

These cases wouldn't be used anyway since we only show skeleton component when we see the new "loader type".

projects/components/src/load-async/load-async.service.ts
```diff
@@ -62,7 +62,14 @@ export type AsyncState = LoadingAsyncState | SuccessAsyncState | NoDataOrErrorAs
 export const enum LoaderType {
   Spinner = 'spinner',
   ExpandableRow = 'expandable-row',
-  Page = 'page'
+  Page = 'page',
+  Rectangle = 'rectangle',
+  RectangleText = 'rectangle-text',
```
anandtiwary, line 67: Can we rename it to 'Text'?

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.containerClass = this.getContainerClass();
+  }
+
+  public ngOnChanges(): void {
+    this.numberOfIterations = Array(this.repeat).fill(1);
+
+    this.containerClass = this.getContainerClass();
+  }
+
+  public getContainerClass(): ContainerClass {
```
anandtiwary, line 50:
```
return [ 'skeleton', this.shapeStyle];

```

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    <ng-container *ngFor="let k of this.numberOfIterations">
+      <ng-container [ngSwitch]="this.shapeStyle">
+        <div *ngSwitchCase="'${SkeletonType.RectangleText}'" [ngClass]="this.containerClass"></div>
+        <div *ngSwitchCase="'${SkeletonType.Circle}'" [ngClass]="this.containerClass"></div>
+        <div *ngSwitchCase="'${SkeletonType.Square}'" [ngClass]="this.containerClass"></div>
+        <div *ngSwitchCase="'${SkeletonType.TableRow}'" [ngClass]="this.containerClass"></div>
+        <div *ngSwitchCase="'${SkeletonType.Donut}'" [ngClass]="this.containerClass">
+          <div class="donut-inner"></div>
+        </div>
+        <div *ngSwitchCase="'${SkeletonType.ListItem}'" [ngClass]="this.containerClass">
```
anandtiwary, line 15: Could you please change the switch case to contain:
1. TableRow
2. ListItem
3. Default template for all other

projects/components/src/skeleton/skeleton.component.ts
anandtiwary: replace ContainerClass with string[]. ContainerClass will no longer be needed.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,85 @@
...
+  'skeleton-rectangle-text': boolean;
+  'skeleton-table-row': boolean;
+  'skeleton-list-item': boolean;
+  'skeleton-repeating': boolean;
+  'skeleton-donut': boolean;
+}
+
+export const enum SkeletonType {
+  Rectangle = 'rectangle',
+  RectangleText = 'rectangle-text',
```
anandtiwary, line 79: rename it to Text

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,85 @@
...
+      'skeleton-rectangle-text': this.shapeStyle === SkeletonType.RectangleText,
+      'skeleton-table-row': this.shapeStyle === SkeletonType.TableRow,
+      'skeleton-list-item': this.shapeStyle === SkeletonType.ListItem,
+      'skeleton-donut': this.shapeStyle === SkeletonType.Donut,
+      'skeleton-repeating': this.repeat !== undefined && this.repeat > 1
+    };
+  }
+}
+
+interface ContainerClass {
```
anandtiwary, line 65: We can use SkeletonType as class names.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,85 @@
...
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  styleUrls: ['./skeleton.component.scss']
+})
+export class SkeletonComponent implements OnChanges {
+  @Input()
+  public shapeStyle: SkeletonType = SkeletonType.Rectangle;
+
+  @Input()
+  public repeat: number | undefined = 1;
```
anandtiwary, line 34: If we can repeat this with css (may be try background-repeat: repeat-y), then
1. Remove RepeatCount
2. Introduce a local variable "repeatInfinite" which will only be set to true for Table and List. Otherwise, false.

anandtiwary: If not, then I will recommend removing `repeat` as an input and we can set it to reasonable default like 4 or 5 and only set it when it is  a table or list.

projects/components/src/load-async/loader/loader.component.ts
```diff
@@ -1,25 +1,53 @@
...
 
-  public ngOnChanges(): void {
-    this.currentLoaderType = this.loaderType ?? LoaderType.Spinner;
+  public imagePath: ImagesAssetPath = ImagesAssetPath.LoaderSpinner;
+
+  public isOldLoaderType: boolean = true;
+
+  public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (changes.loaderType) {
+      this.currentLoaderType = this.loaderType ?? LoaderType.Spinner;
```
anandtiwary, line 40:
```
if (this.currentLoaderType === LoaderType.ExpandableRow ||
        this.currentLoaderType === LoaderType.Spinner ||
        this.currentLoaderType === LoaderType.Page ) {
isOldLoaderType = true;
imagePath = getImagePathFromType
} else {
this.skeletonType = this.getSkeletonTypeForLoader(this.currentLoaderType)
}
```

projects/components/src/load-async/loader/loader.component.ts
Christian862: yeah, as Anand said, we only end up using the return type from this switch if the loader type is not one of the three older types, so including them here is irrelevant, except for purpose of using the assertUnreachable. And with recent changes coming the switch statement won't even be hit if loader type is one of the three old types. So i'm going to revert it back to the original version using the default case.

aaron-steinfeld: That's trading a compiler check for a logic check - never a good trade. If you want the best of both worlds, the types can actually be improved to represent the behavior you want - make the loader type a union of two enums, one representing the old values, one the new. This function would only accept the new ones (or you may not even need to map types, since the same skeleton type enum can be used by both)

anandtiwary: Yeah, we could do that. But In a follow-up PR, we plan to migrate existing usages to new skeleton loaders, and then we will remove all the three **older** loader types and associated code. If you prefer, we can bring these cases back for correctness.  
```
case LoaderType.ExpandableRow: 
case LoaderType.Page: 
case LoaderType.Spinner: 
return SkeletonType.Rectangle;
```

aaron-steinfeld: If they're going to go away in a followup PR, the style isn't a big deal since that's just for maintainability. In that case either way is fine with me as long as the logic is right.

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,73 @@
...
+        <div *ngSwitchDefault [ngClass]="this.containerClass"></div>
+      </ng-container>
+    </ng-container>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  styleUrls: ['./skeleton.component.scss']
+})
+export class SkeletonComponent implements OnChanges {
+  @Input()
+  public shapeStyle: SkeletonType = SkeletonType.Rectangle;
```
anandtiwary, line 27: nit: can we call it `skeletonType`?

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,73 @@
...
+        return Array(5).fill(1);
+      case SkeletonType.ListItem:
+        return Array(4).fill(1);
+      default:
+        return Array(1).fill(1);
+    }
+  }
+
+  public getContainerClass(): string[] {
+    const classes = ['skeleton', this.shapeStyle];
```
anandtiwary, line 55: Can we declare `skeleton` and `repeating` as static string properties in the component?

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -0,0 +1,77 @@
...
+})
+export class SkeletonComponent implements OnChanges {
+  @Input()
+  public skeletonType: SkeletonType = SkeletonType.Rectangle;
+
+  public iterationsArray: number[] = Array(1).fill(1);
+
+  public containerClass: string[];
+
+  public static skeletonClass: string = 'skeleton';
```
anandtiwary, line 33:
```
private static readonly SKELETON_CLASS_NAME = 'skeleton';
private static readonly REPEATING_CLASS_NAME = 'repeating';
```

projects/components/src/skeleton/skeleton.component.ts
```diff
@@ -30,9 +30,9 @@ export class SkeletonComponent implements OnChanges {
 
   public containerClass: string[];
 
-  public static skeletonClass: string = 'skeleton';
+  private static readonly SKELETON_CLASS_NAME: string = 'skeleton';
```
anandtiwary, line 33: nit: we usually keep static properties at the top, right after class definition.

anandtiwary (dismissed):

anandtiwary (approved):

PR #1375, opened 2022-01-17
style(topology): adjusted application flow padding

23nobody:
Reduced padding from top. Padding below interaction control is adjusted.
Screenshot 2022-01-17 at 4 17 16 PM


anandtiwary:
@23nobody We also use this flow map on the service detail page. Could you please verify it there as well?

PR #1384, opened 2022-01-19
Cartesian column max width

anandtiwary:
## Description
fix: reducing max width of Cartesian column

### Testing

### Checklist:

### Documentation


aaron-steinfeld (commented):
Context? this is a major styling change on the bar chart and there's no explanation here. This would be especially egregious on a time series where a skinny bar wouldn't align with the time bucket it represents.

image

aaron-steinfeld:
> The chart in the screenshot above shouldn't be affected by this change. The alignment is also calculated based on the column width, so it wouldn't go wrong either.

That screenshot is just the UI with this change - certainlyaffected. The alignment of the left side (origin) of the bar is correct, aligning with the start of the interval but now the bar only extends for the shorter width.

> A higher column width when more space is available is a UX preference. The recent UX preference is to show a skinner chart even when more space is present.

Agreed at the end of the day it's a preference. If we're changing this everywhere though (which we are) did we take all the various usages into consideration when changing this - time series with bars, grouped time series, plain bar charts without time? Larger charts and smaller? Or is this based on one particular mock/screen? If we only want it one place, we should just make it configurable.

anandtiwary:
I have checked it at a few other places and it seems fine to me. And I remember when I originally wrote this code, I played around with this variable a lot and ensured changing it would not affect any other chart logic as such. At that time, I looked into Highcharts and kept their behavior (fat columns if more space is available). 


> That screenshot is just the UI with this change - certainlyaffected. The alignment of the left side (origin) of the bar is correct, aligning with the start of the interval but now the bar only extends for the shorter width.

So, this is not a problem, right? We do an origin x adjustment too which tries to align the column center at the interval (unless there are grouped series)

```
  private getOriginXAdjustment(columnWidth: number): number {
    const totalGroupedSeries = this.xScale.getGroupedColumnSeriesLength(this.series);
    const groupedColumnAdjustment = (this.xScale.getBandwidth() - columnWidth * totalGroupedSeries) / 2;

    return totalGroupedSeries === 1
      ? groupedColumnAdjustment
      : groupedColumnAdjustment + this.xScale.getGroupedColumnSeriesPosition(this.series) * columnWidth;
  }
```

aaron-steinfeld:
> So, this is not a problem, right? We do an origin x adjustment too which tries to align the column center at the interval (unless there are grouped series)

I guess a matter of opinion. Let me try and grab screens of each for comparison. To me, that's just an awful lot of whitespace in the narrower versions:

narrow/time series:
image
narrow/grouped time series:
image
narrow/grouped:
image

current/time series:
image

current/grouped time series:
image

current/grouped:
image

anandtiwary:
Yeah, there is a lot of white-space. How do we make a decision here? UX team has to own it and this is what I got from them.

anandtiwary:
Closing this for now.

PR #1385, opened 2022-01-21
feat: Adding state disable to dropdown

palbizu:
## Description
Adding disabling state from form.

### Testing
Tested locally, visually

image

### Checklist:

### Documentation


projects/components/src/select/select.component.ts
```diff
@@ -324,6 +324,10 @@ export class SelectComponent<V> implements ControlValueAccessor, AfterContentIni
     this.propagateControlValueChangeOnTouch = onTouch;
   }
 
+  public setDisabledState(isDisabled?: boolean): void {
```
jake-bassett, line 327: Since the property is public is there a need for a method to access it?

palbizu: I have no clear the reason. Probably @anandtiwary can clarify it. I followed the same structure already implemented in the input component to the Control Value Accessor. I will try to find the reason.

itssharmasandeep: Actually this is related to `ControlValueAccessor` interface, this function will be used by reactive form field to set the disabled state using form group itself. We will not pass any [disabled] input in the template while using the select component in the forms.

palbizu: So @itssharmasandeep , does it fine to keep it public?

palbizu:
Closing this PR because It's not necessary anymore but we're going to adding it in the future.

projects/components/src/select/select.component.ts
itssharmasandeep: Yeah it is fine.

itssharmasandeep (approved):

aaron-steinfeld (approved):

PR #1400, opened 2022-01-29
fix: changed menu-dropdown content style to body-1-regular

Greedy-Geek:


projects/components/src/menu-dropdown/menu-dropdown.component.scss
```diff
@@ -1,4 +1,5 @@
 @import 'interaction';
```
itssharmasandeep, line 1: Please remove this. `mixins` covers this

projects/components/src/menu-dropdown/menu-dropdown.component.scss
```diff
@@ -35,5 +36,5 @@
 }
 
 .dropdown-content {
-  @include popover-menu-content;
+  @include body-1-regular;
```
itssharmasandeep, line 39: is this the change we really want?
These are two completely different styles.
cc: @arjunlalb

arjunlalb: We need both here. We need the popover styling + a styling for the text.

@Greedy-Geek do not remove the existing container styling. Also, please test the changes before sending out a PR.

Greedy-Geek: Oh okay, I checked the changes before sending for the PR; maybe I missed some obvious things.

projects/components/src/menu-dropdown/menu-dropdown.component.scss
Greedy-Geek: Okay.

PR #1404, opened 2022-01-31
fix: checkbox on write refresh

palbizu:
## Description
Checkbox text wasn't refreshed when the data was changed from its FormControl and it needed a new event update.

### Testing
Visual test.

### Checklist:

### Documentation


projects/components/src/combo-box/combo-box.component.ts
```diff
@@ -445,6 +445,7 @@ export class ComboBoxComponent<TValue = string> implements AfterViewInit, OnChan
 
   public writeValue(text?: string): void {
     this.text = text;
+    this.changeDetectorRef.markForCheck();
```
anandtiwary, line 448: Do you need to call `setText` here? Everything except following

```
      this.textChange.emit(this.text);
      this.propagateValueChangeToFormControl(this.text);
 ```

palbizu: Applying setText without  markForCheck doesn't update the text

anandtiwary: Some function calls from `setText` still seems relevant whenever text is changed via form control. The emit and propogate method mentioned above would not be required. I will approve the PR, but you can play with it.

anandtiwary (approved):

PR #1409, opened 2022-02-03
feat(validators): add domain validator

adisreyaj:
## Description
Adds domain validation helper methods.

### Checklist:
- [x] I have added tests that prove my fix is effective or that my feature work


projects/common/src/utilities/validators/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './domain-validator';
```
itssharmasandeep, line 1: This change looks fine to me. I do not think we use `index.ts` file anywhere as of now.
IMO, I think we can use this pattern.

cc: @aaron-steinfeld @anandtiwary @arjunlalb

aaron-steinfeld: I don't have a problem with it

projects/common/src/utilities/validators/domain-validator.ts
```diff
@@ -0,0 +1,7 @@
+export const DOMAIN_REGEX = new RegExp(/^[a-z0-9\-]{3,}\.[a-z]{2,}$/);
```
aaron-steinfeld, line 1: Validating domains is _hard_ - do we have a requirement to do so? There's almost always domains that can get around it. This reads like it requires all lower case, ascii characters, but all kinds of fun stuff can show up in domain names. Country/second level domains (e.g. `co.uk`) I believe would fail this. Non english characters, upper cases, subdomains (e.g. `hyper.hypertrace.org`) would all fail here, too.

In the past, I've generally worked under the philosophy of undervalidating rather then over.

adisreyaj: I should probably name it differently. The function is intended to validate the `domain` part of an email.
Will make the change to accommodate domains like `.co.uk`

aaron-steinfeld: all of the cases I described are what you're saying. `aaron@hyper.hypertrace.org` is a valid email (well not that specifically, but that structure). Email addresses aren't case sensitive (at least I don't think they normally are), so `AARON@HYPERTRACE.ORG` is valid too, and likewise, the non-english characters can show up there as far as I'm aware.

aaron-steinfeld: my suggestion from slack - `^.+\.[a-zA-Z]{2,}$` - basically adds upper case support and gives up on trying to validate anything before the TLD, just makes sure it's at least one char.

adisreyaj: Updated. Thanks!

projects/common/src/utilities/validators/domain-validator.ts
```diff
@@ -0,0 +1,7 @@
+export const DOMAIN_REGEX = new RegExp(/^.+\.[a-zA-Z]{2,}$/);
+
+export const isDomainValid = (domain: string) => {
+  return DOMAIN_REGEX.test(domain);
+};
+
+export const areDomainsValid = (domains: string[]) => domains.filter(domain => !isDomainValid(domain)).length === 0;
```
arjunlalb, line 7: nit: can use `domains.every`

arjunlalb (approved):

projects/common/src/utilities/validators/domain-validator.test.ts
```diff
@@ -0,0 +1,11 @@
+import { areDomainsValid } from '@hypertrace/common';
+
+const VALID_DOMAINS = ['google.com', 'fig.io', 'traceable.ai', 'f4ncy-domain.cc'];
+const INVALID_DOMAINS = ['https://google.com', 'google.com/', '123123@.scom', 'g00#.dd', 's.ss', 'www.112323.com'];
```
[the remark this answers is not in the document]
adisreyaj: Yup. Updating test to instead assert individual domain's validities.

aaron-steinfeld (approved):

PR #1410, opened 2022-02-04
feat(navigable-tab): add change detection trigger

adisreyaj:
## Description
When tabs are dynamically added or removed change detection will now be automatically triggered.


projects/components/src/tabs/navigable/navigable-tab-group.component.ts
```diff
@@ -49,6 +49,10 @@ import { NavigableTabComponent } from './navigable-tab.component';
       </nav>
       <div class="divider"></div>
     </div>
+
+    <!-- When tabs input changes, change detection will be triggered automatically -->
```
aaron-steinfeld, line 53: The way we handle this elsewhere is just wrapping querylists with `queryListAndChanges` util. This should be used everywhere we use contentChildren

adisreyaj: Fixed.

projects/components/src/tabs/navigable/navigable-tab-group.component.ts
```diff
@@ -49,23 +56,28 @@ import { NavigableTabComponent } from './navigable-tab.component';
...
+    <ng-container *ngIf="tabs.changes | async"></ng-container>
```
anandtiwary, line 61: can you merge these observables on L80?

adisreyaj: Yup. This line will be removed.

arjunlalb (approved):

PR #1411, opened 2022-02-04
Async loader input config refactor - DRAFT DO NOT MERGE

Christian862:
## Description
Changed directive consumer interface. kept current config input for backwards compatibility. Changes made:
 - Provided standalone `loaderType` input.
 - Provided standalone `noData` input.
 - Made `noData` a property of the widgets base model. 

If the consumer provides a `config` input and either `noData` or `loaderType`, the standalone inputs will overwrite the respective properties of config.

### Testing

### Checklist:

### Documentation


projects/dashboards/src/widgets/base.model.ts
```diff
@@ -16,6 +17,13 @@ export abstract class BaseModel {
   })
   public id?: string;
 
+  @ModelProperty({
+    key: 'noDataLoadingConfig',
+    required: false,
+    type: PLAIN_OBJECT_PROPERTY.type
+  })
+  public noDataLoadingConfig?: NoDataOrErrorStateConfig;
```
itssharmasandeep, line 25: Why only base `no data` part? We could also use `loadAsyncConfig` .
cc: @anandtiwary

Christian862: After talking about it we decided only `noData` should be configurable to the consumer of widgets. Error state is the same across the app for `loadAsyncConfig` so it should be controlled within the directive itself.  The widgets should be responsible for `loaderType`, so it should be non-configurable to widget consumers. Please correct if needed Anand

PR #1412, opened 2022-02-04
feat: add minor multiselect improvements

aaron-steinfeld:
## Description
- Adds improved disable support (form integration, remove pointer on disable)
- Adds tooltip for overflow label so they can be viewed without opening multiselect (which isn't possible when disabled)

Action shot:
image


palbizu (approved):
LGTM!

PR #1413, opened 2022-02-05
fix(ui): centered toggle switch vertically

23nobody:
Screenshot 2022-02-05 at 2 40 38 PM


projects/components/src/toggle-switch/toggle-switch.component.scss
```diff
@@ -2,6 +2,7 @@
 @import 'font';
 
 :host {
+  display: flex;
```
aaron-steinfeld, line 5: No host styling please. Here, it's being used as a scope due to the also-problematic abuse of ng-deep, but we're not applying host styles.

The problem with host styling is that's breaking encapsulation - the parent is responsible for applying styles at the host level, and applying them inside the component can introduce non-obvious conflicts.

Please try to accomplish the same thing with regular css classes inside the component.

23nobody: Encapsulated it inside a div

aaron-steinfeld (approved):

projects/components/src/toggle-switch/toggle-switch.component.scss
anandtiwary: don't we need to add `align-items:center` too ?

23nobody: I was hoping it would be centered by default as there's only one child mat-slide-toggle

arjunlalb (approved):

PR #1422, opened 2022-02-09
feat: filter changes

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation


projects/components/src/filtering/filter-button/filter-button.component.ts
```diff
@@ -45,7 +45,7 @@ export class FilterButtonComponent implements OnChanges {
   public attribute?: FilterAttribute;
 
   @Input()
-  public value?: unknown;
+  public value?: FilterValue;
```
aaron-steinfeld, line 48: Seems completely dependent on the filter implementation - what's the point of typing it?

projects/components/src/filtering/filter/filter-operators.ts
```diff
@@ -9,6 +9,7 @@ export const enum FilterOperator {
   GreaterThanOrEqualTo = '>=',
   Like = '~',
   In = 'IN',
+  NotIn = 'NOT_IN',
```
aaron-steinfeld, line 12: Pretty sure this isn't supported, if it's being added please put in a separate PR

projects/components/src/filtering/filter/filter-operators.ts
```diff
@@ -44,6 +46,8 @@ export const toUrlFilterOperator = (operator: FilterOperator): UrlFilterOperator
       return UrlFilterOperator.In;
     case FilterOperator.ContainsKey:
       return UrlFilterOperator.ContainsKey;
+    case FilterOperator.NotIn:
+      return UrlFilterOperator.ContainsKey;
```
aaron-steinfeld, line 50: bad map

projects/components/src/filtering/filter/filter-operators.ts
```diff
@@ -78,6 +84,7 @@ export const incompatibleOperators = (operator: FilterOperator): FilterOperator[
   switch (operator) {
     case FilterOperator.In:
     case FilterOperator.Equals:
+    case FilterOperator.NotIn:
       return [
         FilterOperator.In,
```
aaron-steinfeld, line 89: this logic seems off and is not symmetric. `Not in` would return `in` is incompatible, but `in` would not return `not in` as incompatible.

projects/observability/src/shared/graphql/model/metadata/attribute-metadata.ts
```diff
@@ -7,6 +7,7 @@ export interface AttributeMetadata {
   units: string;
   type: AttributeMetadataType;
   scope: string;
+  category?: string;
```
aaron-steinfeld, line 10: not generally applicable nor part of GQL model, please remove

projects/observability/src/shared/services/filter-builder/graphql-filter-builder.service.ts
```diff
@@ -1,13 +1,36 @@
...
-import { FilterOperator, TableFilter } from '@hypertrace/components';
+import { FilterOperator } from '@hypertrace/components';
 import { GraphQlArgumentValue } from '@hypertrace/graphql-client';
 import { GraphQlFieldFilter } from '../../graphql/model/schema/filter/field/graphql-field-filter';
 import { GraphQlFilter, GraphQlOperatorType } from '../../graphql/model/schema/filter/graphql-filter';
 
 @Injectable({ providedIn: 'root' })
 export class GraphQlFilterBuilderService {
-  public buildGraphQlFilters(filters: TableFilter[]): GraphQlFilter[] {
+  public buildFiltersFromGraphQlFieldFilters(filters: GraphQlFieldFilter[]): FieldFilter[] {
```
aaron-steinfeld, line 12: Why do we want to convert back from GQL filters to field filters? If we're doing refactoring, we should change any components using GQL filters to use the UI construct instead of supporting going bidrectionally.

projects/components/src/filtering/filter-button/filter-button.component.ts
anandtiwary: This is required for Filter to GqlFilter transformation

projects/components/src/filtering/filter/filter-operators.ts
anandtiwary: I saw one operator in Gql. So thought it is supported. 

yes, this is just a draft PR. I will create smaller ones when it is ready

projects/components/src/filtering/filter/filter-operators.ts
anandtiwary: Yeah, I am not sure what this is code is doing. Still playing around with it.

projects/observability/src/shared/graphql/model/metadata/attribute-metadata.ts
anandtiwary: The category would be added to attributes eventually, right?

projects/observability/src/shared/services/filter-builder/graphql-filter-builder.service.ts
anandtiwary: Yeah, so that would be a big change. If we want to do it in iterations then temporarily this method will be used in the basic filter bar for now. Slowly, we will migrate the filters to the UI construct. 

We would also need to revisit the relationship between `IncompleteFilter`, `FieldFilter`, and `Filter`?

projects/components/src/filtering/filter/filter-operators.ts
anandtiwary: yeah. Do you know we need three operator types? GqlOperators, FilterOperator and URLFilterOperator. Imo We can combine FilterOperator and URLFilterOperator and only use one.

projects/components/src/filtering/filter-button/filter-button.component.ts
aaron-steinfeld: This is the kind of thing I was saying the other day we shouldn't try to make generic, because there is no one GQL filter representation.

projects/components/src/filtering/filter/filter-operators.ts
aaron-steinfeld: There's three different string representations of a user - the server's value (e.g. `EQUALS`), the value we display in the UI (`=`) and the value we use in a URL string where we can't use all specials chars (`_eq_`). We probably could try to consolidate such as by url encoding or similar, but seemed easy enough to maintain initially.

projects/components/src/filtering/filter/filter-operators.ts
aaron-steinfeld: I'm not 100%, but I believe this is used when adding filters via table controls, so if you do something like add an equals filter it replaces any current one.

projects/observability/src/shared/graphql/model/metadata/attribute-metadata.ts
aaron-steinfeld: I don't believe so - I thought the conclusion there was that concept was unique to a particular UI page and could be maintained there. It's not a universal concept across all attributes, nor is it necessarily universal that the same attribute belongs to the same category in all contexts - its use case specific.

projects/observability/src/shared/services/filter-builder/graphql-filter-builder.service.ts
anandtiwary: Things I am re-using:
1. Filter Objects and various filter builders 
2. Url string to Filter parsing logic
3. Basic filter bar value auto-completion

Things I am changing:
1. Moving UI side code to depend on `Filter` objects instead of GraphQlFilters. We will use a method that would have the conversion logic. This migration will have in phases and we can track it with stories.
2. Fix the dependency of Filter, FieldFilter, and IncompleteFilter
2. Using new design for basic view


@jake-bassett Migrating to UI construct (point 1) has a much bigger blast radius. We should do it in stages. Let's wait for other higher-priority tasks to get completed first.

jake-bassett: Okay, got it. Yeah, the items you are re-using are in desperate need of rework but I know that's not at all trivial. I'm good with doing this in phases. As much as I know that code is hard to work with, and making things a bit of a challenge, I don't think we can wait on these filter changes much longer.

I vote for not making this PR any larger that what it already is and revisiting replacing the rest of the filtering at a later date. Would appreciate if @aaron-steinfeld can chime in on any more quick wins we can do here in the meantime though.

anandtiwary: okay. approve this PR  :)

jake-bassett (approved):

PR #1425, opened 2022-02-11
style(table): remove border bottom for the last row

adisreyaj:
Remove the border that is added to the last row of the table.


itssharmasandeep (approved):
LGTM

PR #1426, opened 2022-02-14
feat(entity-renderer): ability to override icon size

adisreyaj:
Add an`Input()` to override the **Icon Size** in the entity renderer.


projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
```diff
@@ -55,6 +55,9 @@ export class EntityRendererComponent implements OnChanges {
   @Input()
   public iconColor?: string;
 
+  @Input()
+  public iconSize: IconSize = IconSize.Large;
```
arjunlalb, line 59: we do not use the icon anywhere today. Let's default to medium.

arjunlalb (approved):

projects/observability/src/shared/components/entity-renderer/entity-renderer.component.ts
arjunlalb: ignore the comment please

arjunlalb (approved):

PR #1431, opened 2022-02-15
fix: auto configure filterable attribute columns

jake-bassett:


projects/observability/src/shared/dashboard/widgets/table/table-widget-column.model.ts
```diff
@@ -104,33 +106,39 @@ export class TableWidgetColumnModel {
...
 
+  @ModelInject(TableWidgetColumnsService)
+  private readonly tableWidgetColumnsService!: TableWidgetColumnsService;
+
   public asTableColumnDef(scope?: string): Observable<SpecificationBackedTableColumnDef> {
     return scope !== undefined
       ? this.metadataService
-          .getAttributeKeyDisplayName(scope, this.value.name)
-          .pipe(map(displayName => this.toSpecificationBackedColumnDef(displayName)))
+        .getAttribute(scope, this.value.name)
```
jake-bassett, line 115: @anandtiwary @aaron-steinfeld for some reason this observable isn't completing when I change from `getAttributeKeyDisplayName` to just `getAttribute`.

I see that `getAttributeKeyDisplayName` has `defaultIfEmpty`, but that doesn't seem to be the difference causing the issue.

Any thoughts?

aaron-steinfeld: I really want us to work to unwind the coupling of attributes and table rather than doubling down on, that said to attempt to answer your question:

That really sounds like it's defaultIfEmpty, but it depends exactly what you mean by it doesn't complete - both should complete if the attribute isn't known, but only `getAttributeKeyDisplayName` will _also_ next a value (due to defaultIfEmpty) - `getAttribute` will simply complete without emitting, so if you're relying on an async pipe or a tap for side effect, it would behave the same as if it was just hanging.

PR #1435, opened 2022-02-15
fix: improve query error console messaging

jake-bassett:
## Description
This adds some additional messaging when the GQL response contains errors.


projects/graphql-client/src/graphql-request.service.ts
```diff
@@ -151,7 +151,15 @@ export class GraphQlRequestService {
         errorPolicy: 'all',
         fetchPolicy: options.cacheability
       })
-      .pipe(map(response => response.data));
+      .pipe(
```
aaron-steinfeld, line 154: should do the same in executeMutation below

aaron-steinfeld (approved):

PR #1450, opened 2022-02-22
feat: prefix icon for label tag component

itssharmasandeep:
## Description
Adding prefix icon for label tag component

### Testing
Local testing done.

### Checklist:

### DIFF
--- a/projects/components/src/label-tag/label-tag.component.scss
+++ b/projects/components/src/label-tag/label-tag.component.scss
@@ -12,12 +12,9 @@
   display: flex;
   align-items: center;
   justify-content: center;
+  gap: 4px;
 
   // Default color. Could be overridden by the component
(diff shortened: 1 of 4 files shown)


projects/components/src/label-tag/label-tag.component.scss
```diff
@@ -16,4 +16,8 @@
   // Default color. Could be overridden by the component
   background-color: $blue-1;
   border-radius: 2px;
+
+  .prefix-icon {
+    margin-right: 4px;
+  }
```
arjunlalb, line 22: perhaps use gap on the parent instead?

itssharmasandeep: Updated

arjunlalb (approved):

PR #1454, opened 2022-02-24
Marker visualisation on dashboards 1240 iterration1

jyothishjose6190:
marker syncup


projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,36 @@
...
+    groupId?: string
+  ): void {
+    this.locationChangeSubject.next({
+      groupId: groupId,
+      locationData: data as MouseLocationData<TData, Series<TData> | Band<TData>>[]
+    });
+  }
+
+  public getLocationChangesForGroup(
+    _groupId: string
```
itssharmasandeep, line 25: do we need this `_groupId` ? I think groupId is still present in the `MouseLocationData`, right?

jyothishjose6190:
@aaron-steinfeld @anandtiwary @itssharmasandeep please review

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.locationChangeSubject.next({
+      groupId: groupId,
+      locationData: data as MouseLocationData<TData, Series<TData> | Band<TData>>[]
+    });
+  }
+
+  public getLocationChangesForGroup(
+    groupId: string
+  ): Observable<MouseLocationData<TData, Series<TData> | Band<TData>>[]> {
+    return this.locationChangeSubject.pipe(map(data => this.filterLocationDataByGroup(data, groupId)));
```
aaron-steinfeld, line 27: use filter rather than map. We should be omitting events that aren't for the group, rather than making them empty arrays.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -79,9 +79,16 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     protected readonly renderingStrategy: RenderingStrategy,
     protected readonly svgUtilService: SvgUtilService,
     protected readonly d3Utils: D3UtilService,
-    protected readonly domRenderer: Renderer2
+    protected readonly domRenderer: Renderer2,
+    protected readonly groupId?: string
   ) {}
 
+  public showCrosshair(locationData: MouseLocationData<TData, Series<TData> | Band<TData>>[]): void {
+    const location = locationData[0].location;
```
aaron-steinfeld, line 87: I think in one of the other files, we were potentially making this array empty, which would cause a runtime error. Even if we fix that, we should be defensive here.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -97,20 +104,32 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
...
     if (this.rangeSelectionEnabled) {
       this.chart.withEventListener(ChartEvent.Select, selectedData => {
         this.selectionChange.emit(selectedData);
       });
     }
 
+    if (this.groupId !== undefined) {
+      this.subscriptionLifeCycle.unsubscribe();
+      this.subscriptionLifeCycle.add(
+        this.chartSyncService?.getLocationChangesForGroup(this.groupId).subscribe(data => {
```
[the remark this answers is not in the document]
jyothishjose6190: we are getting our own updates, to avoid this we need differentiation between sending and receiving components

projects/observability/src/shared/components/cartesian/chart-sync-service.ts
```diff
@@ -0,0 +1,29 @@
...
+    groupId?: string
+  ): void {
+    this.locationChangeSubject.next({
+      groupId: groupId,
+      locationData: data as MouseLocationData<TData, Series<TData> | Band<TData>>[]
+    });
+  }
+
+  public getLocationChangesForGroup(groupId: string): Observable<ChartHoverData<TData>> {
+    return this.locationChangeSubject.pipe(filter(data => data.groupId === groupId));
```
aaron-steinfeld, line 22: group ID doesn't need to come back on result, can pick the location data back off the object via map.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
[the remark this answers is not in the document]
aaron-steinfeld: a boolean could work, although that feels more complex to me - I'd probably send a chart identifier in addition to groupId on either side, which can accomplish the same thing - include it with group id and in the subject, then reverse the filter on subscribe - group id but not chart match. The identifier can probably just be as simple as a `this` ref.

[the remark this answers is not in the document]
jyothishjose6190: @aaron-steinfeld  I have updated the PR. Little confused over the part reverse the filter on subscribe.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -125,9 +126,12 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
 
     if (this.groupId !== undefined) {
       this.subscriptionLifeCycle.add(
-        this.chartSyncService?.getLocationChangesForGroup(this.groupId).subscribe(data => {
-          this.chart?.showCrosshair(data.locationData);
-        })
+        this.chartSyncService
+          ?.getLocationChangesForGroup(this)
```
aaron-steinfeld, line 130: both group id and this should go to the sync service - no reason to filter one thing internal to the service and the other external.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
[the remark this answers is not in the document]
aaron-steinfeld: It's just a two part identifier - a string for the group, and an object ref for the chart - both go in with the event, tag the object being pushed over the subject, then both apply as filters in the same way coming out. Right now, we've chosen to filter one of them inside the service, the other outside.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -79,9 +79,18 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
...
-    protected readonly domRenderer: Renderer2
+    protected readonly domRenderer: Renderer2,
+    protected readonly groupId?: string
   ) {}
 
+  public showCrosshair(locationData: MouseLocationData<TData, Series<TData> | Band<TData>>[]): void {
+    if (locationData.length > 0) {
+      const location = locationData[0].location;
+      const currentLocation = this.allSeriesData.flatMap(viz => viz.dataForLocation({ x: location.x, y: location.y }));
+      this.renderedAxes.forEach(axis => axis.onMouseMove(currentLocation));
```
aaron-steinfeld, line 90: I have no idea why this marked pending considering I've submitted reviews, but I'm guessing it never got seen, so reposting it here: 
image

jyothishjose6190: its not causing infinite loop, since data is emitted from onMouseMove method. 
Screenshot 2022-03-22 at 10 43 24 AM

aaron-steinfeld: Ah sorry - my mistake here was that we're calling `axis.onMouseMove`, I read it as `this.onMouseMove` which is what would have been the problem.

So this gets to two more points
1. We should expose explicit methods e.g. drawCrosshair and hideCrosshair that are called by the existing onMouseMove and onMouseLeave in axis. The problem doesn't exist in the current axis impl, but easily could - we don't want to rely on the handler not doing things other than drawing crosshairs, we're breaking encapsulation.
2. How is leave handled? When does the crosshair on the sync'd charts get hidden?

jyothishjose6190: 2: passed empty data while moving mouse out and then if empty then this.renderedAxes.forEach(axis => axis.onMouseLeave()); still marker is there

jyothishjose6190: 1 will work on this

aaron-steinfeld: > 2: passed empty data on mouseleave and then if empty then this.renderedAxes.forEach(axis => axis.onMouseLeave()); still marker is there

Not sure I understand the point - this line of code is how we hide it for the current chart when the mouse leaves it - for the two that have been sync'd, how do they know to hide their tooltips?

jyothishjose6190: 1) The marker was not removed on mouse leave for a single chart scenario. Reason class hidden was missing. when added a class named hidden marker is removed on mouse leave.
2) For removing the markers in other charts when the mouse leaves the first chart(current), the emitted  null as locationdata . On receiving null in other charts the above line of functionality called to hide the marker

aaron-steinfeld: > The marker was not removed on mouse leave for a single chart scenario. Reason - class hidden was missing. when added a class named hidden marker is removed on mouse leave.

Thanks, that's a bug. Maybe it used to have a css class previously, but there's a global style on the attribute `hidden` so let's use that instead please.

> For removing the markers in other charts when the mouse leaves the first chart(current), the emitted null as locationdata . On receiving null in other charts the above line of functionality called to hide the marker

I just pulled this and tried it out. I made the hidden changes:
```diff
diff --git a/projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis-crosshair.ts b/projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis-crosshair.ts
index 8bc91d39..6fbd047e 100644
--- a/projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis-crosshair.ts
+++ b/projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis-crosshair.ts
@@ -30,11 +30,11 @@ export class CartesianAxisCrosshair {
       position,
       locationData.map(data => data.location)
     );
-    crosshairSelection.classed('hidden', false);
+    crosshairSelection.attr('hidden', null);
   }
 
   public hide(axisElement: SVGGElement): void {
-    this.getCrosshairSelection(axisElement).classed('hidden', true);
+    this.getCrosshairSelection(axisElement).attr('hidden', true);
   }
 
   private initializeCrosshairIfMissing(axisElement: SVGGElement): void {
```

But I see the crosshair on the other two charts remain after I mouse off. Am I missing a step?

jyothishjose6190: @aaron-steinfeld updated the PR with the above changes and marker removal for all charts.

projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis-crosshair.ts
```diff
@@ -30,11 +30,12 @@ export class CartesianAxisCrosshair {
       position,
       locationData.map(data => data.location)
     );
-    crosshairSelection.classed('hidden', false);
+    // tslint:disable-next-line
```
aaron-steinfeld, line 33: only disable the specific rule you need

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -83,11 +83,13 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
     protected readonly groupId?: string
   ) {}
 
-  public showCrosshair(locationData: MouseLocationData<TData, Series<TData> | Band<TData>>[]): void {
-    if (locationData.length > 0) {
+  public showCrosshair(locationData: MouseLocationData<TData, Series<TData> | Band<TData>>[] | undefined): void {
```
aaron-steinfeld, line 86: calling showCrosshair(undefined) to hide a crosshair is counter intuitive. Instead, please expose hideCrosshair.

projects/observability/src/shared/components/cartesian/d3/chart/cartesian-chart.ts
```diff
@@ -599,6 +601,12 @@ export class DefaultCartesianChart<TData> implements CartesianChart<TData> {
       this.tooltip.hide();
     }
 
+    this.eventListeners.forEach(listener => {
+      if (listener.event === ChartEvent.Hover) {
+        listener.onEvent(undefined);
```
aaron-steinfeld, line 606: it's not intuitive that a mouse leave broadcasts a hover event with arg undefined - instead, create and broadcast a mouseleave event.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -113,6 +113,9 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
       )
       .withEventListener(ChartEvent.Hover, data => {
         this.chartSyncService.mouseLocationChange(data, this.groupId, this);
+      })
+      .withEventListener(ChartEvent.MouseLeave, data => {
+        this.chartSyncService.mouseLocationChange(data, this.groupId, this);
```
aaron-steinfeld, line 118: call mouse leave on sync service - it's not a location change. Then rather than switching on undefined (that's what we're trying to avoid) create a second leave subscriber. We've pushed it down one level, but we still want to avoid mixing undefined into the location event stream so we've got one more level to go.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -129,11 +129,13 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
...
-            this.chart?.showCrosshair(data);
-          } else {
-            this.chart?.hideCrosshair();
-          }
+          this.chart?.showCrosshair(data);
+        })
+      );
+
+      this.subscriptionLifeCycle.add(
+        this.chartSyncService?.getMouseLeave(this.groupId, this).subscribe(_data => {
```
aaron-steinfeld, line 137: nit
```suggestion
        this.chartSyncService?.getMouseLeave(this.groupId, this).subscribe(() => {
```

aaron-steinfeld (approved):

aaron-steinfeld (commented):
Please revert IDE and environment changes

jyothishjose6190:
@anandtiwary  @itssharmasandeep  please merge this pr

itssharmasandeep:
can you please update the branch with the latest code?

jyothishjose6190:
> can you please update the branch with the latest code?

@itssharmasandeep done

itssharmasandeep:
Please add some tests in follow-up PR.

jyothishjose6190:
> Please add some tests in follow-up PR.
sure

PR #1464, opened 2022-03-02
Feat/tooltip context

adisreyaj:
## Description
Add ability to pass context when Templates are used as tooltip.


projects/components/src/tooltip/tooltip-content-container.component.ts
```diff
@@ -8,18 +8,23 @@ import { POPOVER_DATA } from '../popover/popover';
   template: `
     <div class="tooltip-container">
       <ng-container *ngIf="this.isComplexContent; else simpleContent">
-        <ng-container *ngTemplateOutlet="this.content"> </ng-container>
+        <ng-container *ngTemplateOutlet="this.templateRef; context: this.data.context"></ng-container>
```
aaron-steinfeld, line 11: would we want this to be used as context directly, or made $implicit?

adisreyaj: Will add `$implicit`.

aaron-steinfeld: it's equivalent in terms of functionality, but I think implicit is easier to use if I'm understanding correctly. Take an obj like
```json
{
  a: "a-value",
  b: "b-value"
}
```

With $implicit, you'd write
```html
<ng-template let-data>
   A is {{ data.a }} and B is {{ data.b }}
</ng-template>
```

In the current form, wouldn't you have to let each value on the context individually and by name? 
```html
<ng-template let-a="a" let-b="b">
   A is {{ a }} and B is {{ b }}
</ng-template>
```

adisreyaj: Yes. It's easier with `$implicit`.

But what I meant was:
```html
<div [htTooltip]="tpl" [htTooltipContext]="{$implicit: data, index: i}">Something</div>
```
Since we have mentioned context as input, it can take an object which may or may not contain `$implicit` key. We are giving that flexibility to the parent.

PR #1469, opened 2022-03-03
feat: adding template tooltip

palbizu:
## Description
Adding Template Tooltip in Bar Gauge

### Testing
Visual Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
+++ b/projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
@@ -173,7 +173,7 @@ interface BarSegment extends Segment {
 }
 
 export interface SegmentContext {
-  $implicit: BarSegment
+  $implicit: BarSegment;
 }
 
 export const enum BarGaugeStyle {
--- a/projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
+++ b/projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
@@ -165,7 +165,7 @@ export interface Segment {
   label: string;
   value: number;
   color?: string;
-  tooltip?: TemplateRef<Segment>;
+  tooltip?: string | TemplateRef<Segment> | TemplateRef<{ $implicit: Segment }>;
 }
 
 interface BarSegment extends Segment {
--- a/projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
+++ b/projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
@@ -49,7 +49,7 @@ import {
               [ngClass]="{ 'hide-last-divider': this.nearMaxValue }"
               [style.background]="segment.color"
               [style.width.%]="segment.percentage"
-              [htTooltip]="this.segmentsTooltip ? this.segmentsTooltip : plainTooltip"
+              [htTooltip]="segment.tooltip ?? plainTooltip"
             >
               <div class="divider"></div>
               <ng-template #plainTooltip> {{ segment.label }} : {{ segment.value | htDisplayNumber }} </ng-template>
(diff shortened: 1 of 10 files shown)


projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -89,6 +91,9 @@ export class BarGaugeComponent implements OnChanges, AfterViewInit {
   @Input()
   public segments?: Segment[] = [];
 
+  @Input()
+  public segmentsTooltip?: TemplateRef<Segment>;
```
palbizu, line 95: This name can be updated

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -48,9 +49,10 @@ import {
               [ngClass]="{ 'hide-last-divider': this.nearMaxValue }"
               [style.background]="segment.color"
               [style.width.%]="segment.percentage"
-              htTooltip="{{ segment.label }} : {{ segment.value | htDisplayNumber }}"
+              [htTooltip]="this.segmentsTooltip ? this.segmentsTooltip : plainTooltip"
```
jake-bassett, line 52: Not a big deal so don't change it unless you are making other changes, but would `this.segmentsTooltip ?? plainTooltip` work here?

jake-bassett (dismissed):

aaron-steinfeld (changes requested):
Sorry, I forgot to press submit when I reviewed earlier

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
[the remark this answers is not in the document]
aaron-steinfeld: The context is needed otherwise you would only have a static template - same for every segment. If you want to define one per segment type, that would work without context.  Either way, should take the template as part of the segment to allow it to work either statically or dynamically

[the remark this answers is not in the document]
adisreyaj: > Depends on adding `[htTooltipContext]` as supported in #1464 CC @adisreyaj

Will be merged as soon as I add/update tests.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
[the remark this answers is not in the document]
palbizu: So would you add it to the Segment object as an attribute?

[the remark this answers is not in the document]
aaron-steinfeld: Yep, optional attribute

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -162,6 +165,7 @@ export interface Segment {
   label: string;
   value: number;
   color?: string;
+  tooltip?: string | TemplateRef<Segment> | TemplateRef<{ $implicit: Segment }>;
```
palbizu, line 168: @adisreyaj should I keep TemplateRef<Segment> or remove it?

adisreyaj:
```suggestion
  tooltip?: string | TemplateRef<{ $implicit: BarSegment }>;
```

palbizu: Disagree because there are cases where we use just the segment and not the barSegment

adisreyaj: We are looping the `barSegments` in the template:
```html
<div *ngFor="let segment of this.barSegments"></div>
```

So this means the `segment` here is of type `BarSegment` and so the context will also be `{ $implicit: BarSegment }`.

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
palbizu: removed

adisreyaj (approved):

PR #1477, opened 2022-03-10
Update sort icons and adding checkbox in header cell for multi select

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/table/header/table-header-cell-renderer.component.ts
+++ b/projects/components/src/table/header/table-header-cell-renderer.component.ts
@@ -200,7 +200,7 @@ export class TableHeaderCellRendererComponent implements OnInit, OnChanges {
       ? 'Some rows are selected'
       : this.allRowsSelected
       ? 'All rows in the current page are selected'
-      : 'Select all rows in the current page';
+      : 'None of the rows in the current page are selected';
   }
 
   private buildClasses(): string[] {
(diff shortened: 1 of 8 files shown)


projects/components/src/table/header/table-header-cell-renderer.component.ts
```diff
@@ -155,6 +190,19 @@ export class TableHeaderCellRendererComponent implements OnInit, OnChanges {
...
+  public onToggleAllSelectedChange(allSelected: boolean): void {
+    this.allRowsSelected = allSelected;
+    this.allRowsSelectionChange.emit(allSelected);
+  }
+
+  public getHeaderCheckboxTooltip(): string {
+    return this.indeterminateRowsSelected
+      ? 'Some rows are selected'
+      : this.allRowsSelected
+      ? 'All rows in the current page are selected'
```
jake-bassett, line 202: Two of these tooltips say what is currently the state. However, the final tooltip says what interacting with the control will do. Should we change the last one to 'None of the rows in the current page are selected' to align with the others? Or, should we change the others to state what action will happen if clicked?

projects/components/src/table/table.component.scss
```diff
@@ -71,11 +71,11 @@ $header-height: 32px;
 
   .header-cell-renderer {
     width: 100%;
-    padding: 10px 12px 10px 7px;
+    //  padding: 10px 12px 10px 7px;
```
jake-bassett, line 74: Commented code. Below as well.

projects/components/src/table/header/table-header-cell-renderer.component.ts
anandtiwary: Yeah, I added that because when the tooltip is not selected and the user hovers over it for the first time, they will get to know what the selection would do. I am fine with updating the text like you suggested.

jake-bassett (approved):

PR #1480, opened 2022-03-14
Feat | add error selectors

adisreyaj:
## Description
Add `role=alert` and `data-alert-type="failure"` for components showing error states:
1. Toast Error
2. Load Async Error 
3. Not Found

Selector:
```js
$('[role=alert][data-alert-type="failure"]')
```

### DIFF
--- a/projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
+++ b/projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
@@ -21,6 +21,9 @@ export const ASYNC_WRAPPER_PARAMETERS$ = new InjectionToken<Observable<LoadAsync
       </ng-container>
       <ng-container *ngSwitchDefault>
         <ht-message-display
+          aria-live="polite"
+          role="alert"
+          [attr.data-alert-type]="state.type === '${LoadAsyncStateType.GenericError}' ? 'failure' : 'no-data'"
           [icon]="this.icon"
           [title]="this.title"
           [description]="this.description"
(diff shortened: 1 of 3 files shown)


projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
```diff
@@ -21,6 +21,9 @@ export const ASYNC_WRAPPER_PARAMETERS$ = new InjectionToken<Observable<LoadAsync
       </ng-container>
       <ng-container *ngSwitchDefault>
         <ht-message-display
+          aria-live="polite"
+          role="alert"
+          [attr.data-alert-type]="state.type === '${LoadAsyncStateType.GenericError}' ? 'failure' : 'no-data'"
```
arjunlalb, line 26: can we use the `NotificationMode` enum here for the failure string?

projects/components/src/not-found/not-found.component.ts
```diff
@@ -9,7 +9,7 @@ import { ButtonRole, ButtonStyle } from '../button/button';
   styleUrls: ['./not-found.component.scss'],
   template: `
     <div class="not-found-container fill-container">
-      <div class="not-found-content">
+      <div class="not-found-content" role="alert" aria-live="assertive" [attr.data-alert-type]="'failure'">
```
arjunlalb, line 12: same here.

projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
adisreyaj: Updated.

projects/components/src/not-found/not-found.component.ts
adisreyaj: Updated.

anandtiwary (approved):

Issue #1250, opened 2021-11-10
Attribute based drill down should automatically come to Trace or Span tab on explorer as per attribute scope


JBAhire:
## Use Case
Current implementation of attribute based drill down always redirects to Trace tab of explorer with tag filter applied. But as some of the attributes don't exist in trace scope it shows no data. We should check the scope of attribute and then redirect accordingly.  

## Proposal
As there will always be data in span context, it make more sense to always go to Spans tab in that case.

jaywalker21:
I have started looking into this.
Following is the information I could gather around the changes we might need for the implementation to go through.
Currently the following attributes are sent as `Trace` related `attributes`.
```json
[
    {
        "name": "startTime",
        "displayName": "Start Time",
        "type": "LONG",
        "scope": "API_TRACE",
        "units": "",
        "onlySupportsAggregation": false,
        "onlySupportsGrouping": false,
        "supportedAggregations": [
            "SUM",
            "MIN",
            "MAX",
            "AVG"
        ],
        "groupable": false,
        "__typename": "AttributeMetadata"
    },
    {
        "name": "duration",
        "displayName": "Duration",
        "type": "LONG",
        "scope": "API_TRACE",
        "units": "ms",
        "onlySupportsAggregation": true,
        "onlySupportsGrouping": false,
        "supportedAggregations": [
            "SUM",
            "MIN",
            "MAX",
            "AVG",
            "AVGRATE",
            "PERCENTILE"
        ],
        "groupable": false,
        "__typename": "AttributeMetadata"
    },
    {
        "name": "apiTraceErrorSpanCount",
        "displayName": "Api Trace Error Span Count",
        "type": "LONG",
        "scope": "API_TRACE",
        "units": "",
        "onlySupportsAggregation": false,
        "onlySupportsGrouping": false,
        "supportedAggregations": [
            "SUM",
            "MIN",
            "MAX",
            "AVG"
        ],
        "groupable": false,
        "__typename": "AttributeMetadata"
    },
    {
        "name": "protocol",
        "displayName": "Protocol",
        "type": "STRING",
        "scope": "API_TRACE",
        "units": "",
        "onlySupportsAggregation": false,
        "onlySupportsGrouping": false,
        "supportedAggregations": [
            "DISTINCTCOUNT"
        ],
        "groupable": true,
        "__typename": "AttributeMetadata"
    },
    {
        "name": "statusCode",
        "displayName": "Status Code",
        "type": "STRING",
[...]

PR #1486, opened 2022-03-15
Explorer - Fix Retaining of filters on switching between Trace and Span view

jaywalker21:
## Description
***Fixes https://github.com/hypertrace/hypertrace-ui/issues/1250***
- Made the `filter-bar` component controlled in nature by providing a mechanism for two-way binding.
- Added logic in `explorer` component to update the filter values on toggling between `span` and `trace` views.

### Testing
- Carried out local testing and seems to be working well so far.


projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -194,11 +198,51 @@ export class ExplorerComponent {
...
+        type: applicableType,
+        onlySupportsAggregation: onlySupportsAggregation,
+        onlySupportsGrouping: onlySupportsGrouping
+      };
+    });
+  }
+
   public onContextUpdated(contextWrapper: ExplorerContextScope): void {
     this.attributes$ = this.metadataService.getFilterAttributes(contextWrapper.dashboardContext);
+    const listener = this.attributes$.subscribe(attributes => {
```
jaywalker21, line 218: Will try to move this block outside of this method in an independent structure.

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -327,6 +371,12 @@ interface ExplorerContextScope {
   scopeQueryParam: ScopeQueryParam;
 }
 
+type contextMap = {
```
aaron-steinfeld, line 374: PascalCase for type names (also, we generally use interfaces if constructing a brand new type that's not projected from another.

Naming here is also an issue. I think context is an overloaded term, and even knowing the goal of this PR it took me reading the object using this to understand what this was. Maybe something like
``` ts
type AttributeTranslationDictionary = { [key: string]: string; }

const spanToApiTraceAttributeTranslationDictionary = {...};
const apiTraceToSpanAttributeTranslationDictionary = {...};
```

projects/observability/src/pages/explorer/explorer.component.ts
aaron-steinfeld, line 218: We shouldn't use a subscription here. Make the filters an observable, and subscribe via async pipe in the template.

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -194,11 +198,51 @@ export class ExplorerComponent {
...
+  }
+
   public onContextUpdated(contextWrapper: ExplorerContextScope): void {
     this.attributes$ = this.metadataService.getFilterAttributes(contextWrapper.dashboardContext);
+    const listener = this.attributes$.subscribe(attributes => {
+      const lastTab = this.contextChangeSubject.getValue();
+      const newFilters = this.filters.map(eachFilter => {
+        // If the given filter has a different name for the selected tab, update the filter value
+        if (eachFilter.field in contextMapObject[lastTab]) {
+          const newFilter = this.filterChipService.autocompleteFilters(
```
aaron-steinfeld, line 223: We should be able to do this without the chip service, that's an impl detail from the filter bar. Can we use the filter builder service (I hate how complex we made this!) instead just read the details out of the old one and write them into the new one. May need to lookup the new attribute, too - but we shouldn't have to create it, it's already one of the ones in the array coming in.

projects/components/src/filtering/filter-bar/filter-bar.component.ts
```diff
@@ -175,10 +193,10 @@ export class FilterBarComponent implements OnChanges, OnInit, OnDestroy {
   }
 
   private updateFilter(oldFilter: Filter, newFilter: Filter): void {
-    this.onFiltersChanged(this.filterBarService.updateFilter(this.internalFiltersSubject$.value, oldFilter, newFilter));
+    this.onFiltersChanged(this.filterBarService.updateFilter(this.filters || [], oldFilter, newFilter));
```
aaron-steinfeld, line 196: When do we use the subject's value and when do we use this.filters? We seem to use different ways in different places which makes it very confusion to understand as a reader.

PR #1489, opened 2022-03-16
Fix | select change detection

adisreyaj:
When setting value using form control the change is not visible unless the user interacts with the select. Adding an explicitly CD check fixed the issue..

### DIFF
--- a/projects/components/src/select/select.component.ts
+++ b/projects/components/src/select/select.component.ts
@@ -314,7 +314,7 @@ export class SelectComponent<V> implements ControlValueAccessor, AfterContentIni
   private setSelection(value?: V): void {
     this.selected = value;
     this.selected$ = this.buildObservableOfSelected();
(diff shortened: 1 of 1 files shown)


projects/components/src/select/select.component.ts
```diff
@@ -314,6 +314,7 @@ export class SelectComponent<V> implements ControlValueAccessor, AfterContentIni
   private setSelection(value?: V): void {
     this.selected = value;
     this.selected$ = this.buildObservableOfSelected();
+    this.cdr.markForCheck();
```
anandtiwary, line 317: Since the issue is coming when using select with forms, please add this method inside the `writeValue` method. `setSelection` is being called for non form situations as well

adisreyaj:
Closing as this fix has landed in main as part of some other PR I guess.

PR #1491, opened 2022-03-17
Added RudderStack Telemetry Provider

jaywalker21:
- Added [RudderStack](rudderstack.com) Telemetry provider implementation with it's respective npm dependency.
- Imported `UserTelemetryModule` into `config.module.ts` with`[]` as default value for providers. (bug fix).
- Added one more field `orgTrackingURL` to `TelemetryProviderConfig` interface as Rudderstack requires a `orgId` and one more additional field.

### DIFF
--- a/projects/common/src/telemetry/providers/rudderstack/rudderstack-provider.ts
+++ b/projects/common/src/telemetry/providers/rudderstack/rudderstack-provider.ts
@@ -1,0 +1,29 @@
+import { Injectable } from '@angular/core';
+import { Dictionary } from '../../../utilities/types/types';
+
+import * as rudderanalytics from 'rudder-sdk-js';
+import { TelemetryProviderConfig, UserTelemetryProvider, UserTraits } from '../../telemetry';
+
+@Injectable({ providedIn: 'root' })
+export class RudderStackTelemetry<InitConfig extends TelemetryProviderConfig>
+  implements UserTelemetryProvider<InitConfig> {
+  public initialize(config: InitConfig): void {
+    rudderanalytics.load(config.orgId, config.orgTrackingURL!);
+  }
+
+  public identify(userTraits: UserTraits): void {
+    rudderanalytics.identify(undefined, userTraits as rudderanalytics.apiObject);
+  }
+
+  public trackEvent(name: string, eventData: Dictionary<unknown>): void {
+    rudderanalytics.track(name, eventData as rudderanalytics.apiObject);
+  }
+
+  public trackPage(name: string, eventData: Dictionary<unknown>): void {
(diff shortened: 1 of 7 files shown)


projects/common/src/telemetry/providers/rudderstack/rudderstack-provider.ts
```diff
@@ -0,0 +1,29 @@
+import { Injectable } from '@angular/core';
+import { Dictionary } from '../../../utilities/types/types';
+
+import * as rudderanalytics from 'rudder-sdk-js';
```
itssharmasandeep, line 4: nit: `rudderAnalytics`

itssharmasandeep:
LGTM, but will let @anandtiwary take a look.
@jaywalker21 can we add some tests as well?

projects/common/src/telemetry/telemetry.ts
```diff
@@ -20,6 +20,7 @@ export interface UserTelemetryProvider<TInitConfig = unknown> {
 
 export interface TelemetryProviderConfig {
   orgId: string;
+  orgTrackingURL?: string;
```
anandtiwary, line 23: As discussed in the sync, please add this in RudderTelemetryInitConfig which would extend from TelemetryProviderConfig

jaywalker21:
@anandtiwary @itssharmasandeep Made the changes as per comments, will add test cases in the future. (We currently don't have any test cases for all the analytics providers)

projects/common/src/telemetry/providers/rudderstack/rudderstack-provider.ts
```diff
@@ -0,0 +1,32 @@
+import { Injectable } from '@angular/core';
+import { Dictionary } from '../../../utilities/types/types';
+
+import { apiObject, identify, load, page, track } from 'rudder-sdk-js';
+import { TelemetryProviderConfig, UserTelemetryProvider, UserTraits } from '../../telemetry';
+
+export interface RudderStackConfig extends TelemetryProviderConfig {
+  writeKey: string;
```
anandtiwary, line 8: what is a `writeKey`? can we give it a better name?

jaywalker21: @anandtiwary Given we are creating an interface customised for RudderStack, it makes sense to use `WRITE_KEY` as the name here. Reference of RudderStack docs - https://www.rudderstack.com/docs/stream-sources/rudderstack-sdk-integration-guides/rudderstack-javascript-sdk/

anandtiwary (approved):

itssharmasandeep (approved):

PR #1492, opened 2022-03-18
Feat/select component clear option optional

adisreyaj:
Made `Clear Selected` option in the Select component configurable.


projects/components/src/select/select.component.ts
```diff
@@ -224,6 +224,9 @@ export class SelectComponent<V> implements ControlValueAccessor, AfterContentIni
   @Input()
   public showBorder: boolean = false;
 
+  @Input()
+  public showClearSelected: boolean = true;
```
aaron-steinfeld, line 228: Please default to false for compatibility

adisreyaj: Changing it to `false`.

cc: @anandtiwary

aaron-steinfeld (dismissed):

projects/components/src/select/select.component.ts
anandtiwary: Yes this makes sense.

PR #1494, opened 2022-03-18
Add dynamic draggable list

palbizu:
## Description
Allows the list to include new elements dynamically and adds some styles by default to provide the effect of dragging. 

### Testing
Tested locally

### Checklist:

### Documentation


projects/components/src/draggable-list/draggable-list.component.scss
```diff
@@ -9,3 +9,21 @@
     cursor: not-allowed;
   }
 }
+
+.draggable-list.cdk-drop-list-dragging .draggable-item:not(.cdk-drag-placeholder) {
+  transition: transform 250ms cubic-bezier(0, 0, 0.2, 1);
```
anandtiwary, line 14: could you please add a video for the animation?

projects/components/src/draggable-list/draggable-list.component.ts
```diff
@@ -39,14 +42,20 @@ export class DraggableListComponent<T> implements AfterContentInit {
...
-    this.draggableItems = this.draggableItemsRef.toArray();
+    this.draggableItems$ = queryListAndChanges$(this.draggableItemsRef).pipe(
+      map(draggableItems => draggableItems.toArray())
+    );
   }
 
   public dropList(event: CdkDragDrop<DraggableItemComponent<unknown>[]>): void {
-    moveItemInArray(this.draggableItems, event.previousIndex, event.currentIndex);
-    this.draggableListChange.emit(this.draggableItems.map(dragabbleItem => dragabbleItem.data!));
+    this.draggableItems$ = this.draggableItems$.pipe(
```
anandtiwary, line 54: I find this a little confusing. If you just want to perform an action on the current draggable items then you could just use it directly from `draggableItemsRef`. queryListAndChanges will just run the CD automatically on any addition/deletion of draggable items.

projects/components/src/draggable-list/draggable-list.component.scss
```diff
@@ -9,3 +9,21 @@
     cursor: not-allowed;
   }
 }
+
+.draggable-list.cdk-drop-list-dragging .draggable-item:not(.cdk-drag-placeholder) {
```
aaron-steinfeld, line 13: these cdk classes aren't ours - can we better control them - they shouldn't be stylable like this with encapsulation

projects/components/src/draggable-list/draggable-list.component.ts
aaron-steinfeld: yeah - what we're doing here isn't right, draggableItems$ should never be reassigned. This would create a sequence of new observables (since each dropList invocation builds on the draggableItems$ produced by the previous), building more and more tap functions along the chain each of which is emitting different things based on the enclosed event.

projects/components/src/draggable-list/draggable-list.component.scss
palbizu: Can I try to style it from outside but I should use ng-deep

aaron-steinfeld: I don't think these would work without ng deep, but ideally we're not using that either. Why do we need to change the out of the box animation? And the preview they give us explicit controls over if we want to customize it via `cdkDragPreviewClass`

palbizu: I'm adding this styles to animate the dragging effect. If I add it cdkDragPreviewClass I will need to add it in the parent class of <ng-container> so I'm not having the chance to add it from outside.

projects/components/src/draggable-list/draggable-list.component.ts
palbizu: What if I subscribe to the draggableItems$ and modify it internally.

aaron-steinfeld: I think I would start with Anand's suggestion - does the query list allow updates? If not, we probably want to use a subject and merge the query list and our own updates to create draggableItems

projects/components/src/draggable-list/draggable-list.component.scss
aaron-steinfeld: Got it - so I assumed some of this was default behavior, but it looks like the docs examlpe are doing this in the parent too (guessing that's the inspiration here)

projects/components/src/draggable-list/draggable-list.component.ts
anandtiwary: @palbizu Would this work? 
```
public dropList(event: CdkDragDrop<DraggableItemComponent<unknown>[]>): void {
   const draggableItems = this.draggableItemsRef.toArray();
   moveItemInArray(draggableItems, event.previousIndex, event.currentIndex);
   this.draggableListChange.emit(draggableItems.map(dragabbleItem => dragabbleItem.data!))
}
```

palbizu: done

projects/components/src/draggable-list/draggable-list.component.scss
palbizu: I will keep it for now.

aaron-steinfeld (approved):

PR #1499, opened 2022-03-22
Custom date range overflowing with page level time ranges

Christian862:
## Description
Before:
Screenshot 2022-03-22 at 10 03 40 AM

After:
Screen Shot 2022-03-22 at 12 33 19 PM

Added dropdown location to popover component for page level time range.

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/time-range/time-range.component.ts
+++ b/projects/components/src/time-range/time-range.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, EventEmitter, NgZone, Output } from '@angular/core';
+import { ChangeDetectionStrategy, Component, EventEmitter, Input, NgZone, Output } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
 import {
   FixedTimeRange,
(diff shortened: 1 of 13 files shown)


projects/components/src/time-range/time-range.component.ts
```diff
@@ -67,6 +72,12 @@ import { PopoverRef } from '../popover/popover-ref';
   `
 })
 export class TimeRangeComponent {
+  @Input()
+  public dropdownLocation: PopoverRelativePositionLocation[] = [
+    PopoverRelativePositionLocation.BelowLeftAligned,
+    PopoverRelativePositionLocation.BelowRightAligned
```
itssharmasandeep, line 78: why have we given two locations as default value??

Christian862: Initial thought was that this case wouldn't be hit with the time range and would be misleading to include it there, but that's more the case for the page time range since it's always on headers, my bad, adding the above options.

Christian862: fixed

aaron-steinfeld (approved):

PR #1500, opened 2022-03-23
fix: remove pagination from log-event-table

jake-bassett:
This change removes pagination from the log-event-table as it is unnecessary.

### DIFF
--- a/projects/observability/src/shared/components/log-events/log-events-table.component.test.ts
+++ b/projects/observability/src/shared/components/log-events/log-events-table.component.test.ts
@@ -51,7 +51,7 @@ describe('LogEventsTableComponent', () => {
         id: 'summary'
       })
     ]);
-    expect(spectator.query(TableComponent)!.pageable).toBe(true);
+    expect(spectator.query(TableComponent)!.pageable).toBe(false);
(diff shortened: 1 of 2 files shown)


projects/observability/src/shared/components/log-events/log-events-table.component.test.ts
```diff
@@ -51,7 +51,7 @@ describe('LogEventsTableComponent', () => {
         id: 'summary'
       })
     ]);
-    expect(spectator.query(TableComponent)!.pageable).toBe(true);
+    expect(spectator.query(TableComponent)!.pageable).toBe(false);
```
Greedy-Geek, line 54: can't we just remove this line altogether? Default state of `Pageable` is false, right?

anandtiwary (approved):

arjunlalb (approved):

projects/observability/src/shared/components/log-events/log-events-table.component.test.ts
Greedy-Geek: oh yes, and that way they won't miss testing it. Okay cool

Greedy-Geek (approved):

PR #1503, opened 2022-03-25
feat: support for specifying navigation params to open the page in a new tab where applicable

arjunlalb:
## Description
Support in sheet config to take `pageNavParams` that can be used by sheet overlay to display `open in new tab` button for a sheet.

### Testing
Added UTs

### Checklist:


projects/components/src/overlay/sheet/sheet-overlay.component.ts
```diff
@@ -61,6 +67,7 @@ export class SheetOverlayComponent {
   public readonly closeOnEscape: boolean;
   public readonly attachedTriggerTemplate?: TemplateRef<unknown>;
   public isViewCollapsed: boolean;
+  public navigationParams: ExternalNavigationParams | undefined;
```
anandtiwary, line 70: This should be `InAppNavigationParams`. Or you can keep just `NavigationParams`

arjunlalb: @anandtiwary open in new tab supports only external nav params. Hence the choice of types.

Alternative would be to use open in new tab conditionally if nav type is internal. But that can be added in a future iteration if we have a case for it. Currently all our hyperlinking cases from sheets are for opening in new tabs / external.

anandtiwary: But we might be opening an internal link right? So, can we update open-in-new-tab to accept  `NavigationParam` ?

arjunlalb: @anandtiwary Internal nav links are also converted to ExternalNavParams to OpenInNewTab currently.

If we want to extend support for InternalNavParams by default in OpenInNewTab, we can add support for that separately IMO. If you'd like that to be added, I can take that up after this.

anandtiwary: Okay. Sure. We can do it follow up

anandtiwary (approved):

PR #1509, opened 2022-03-29
New Design for Span Detail Component

adisreyaj:
Implement the new design for the `SpanDetailComponent`.
image

*List View Component* has been updated to take `display` as input:
```ts
export enum ListViewDisplay {
  Plain = 'PLAIN',
  Striped = 'STRIPED'
}
```


projects/observability/src/shared/components/log-events/log-events-table.component.ts
```diff
@@ -40,6 +41,7 @@ export const enum LogEventsTableViewType {
         <ht-list-view
           [records]="this.getLogEventAttributeRecords | htMemoize: row.attributes"
           [header]="this.header"
+          [display]="'${ListViewDisplay.Plain}'"
```
arjunlalb, line 44: display="${ListViewDisplay.Plain}" is more concise

projects/observability/src/shared/components/span-detail/call/headers/span-detail-call-headers.component.scss
```diff
@@ -1,9 +1,12 @@
 @import 'font';
 
+:host {
```
arjunlalb, line 3: can we avoid doing host styling?

adisreyaj: We'll have to add this is parent which I don't think is better than adding this host styling here.

projects/observability/src/shared/components/span-detail/span-detail.component.ts
```diff
@@ -79,12 +99,39 @@ export class SpanDetailComponent implements OnChanges {
...
+
   public showRequestTab?: boolean;
   public showResponseTab?: boolean;
   public showExitCallsTab?: boolean;
   public showLogEventsTab?: boolean;
   public totalLogEvents?: number;
 
+  public tabs: ToggleItem<SpanDetailTab>[] = [];
+
+  private readonly activeTabSubject: BehaviorSubject<ToggleItem<SpanDetailTab> | undefined> = new BehaviorSubject<
```
itssharmasandeep, line 125: do we really need this? can we not do it with a simple variable?

adisreyaj: Since we rely on observable for everything and this event can trigger side effects, thought a subject will make more sense here.

itssharmasandeep (dismissed):

arjunlalb (approved):

projects/observability/src/shared/components/span-detail/call/headers/span-detail-call-headers.component.scss
anandtiwary: @adisreyaj why not? why is `display: block`  needed here?

projects/observability/src/shared/components/span-detail/span-detail.component.ts
```diff
@@ -79,12 +99,39 @@ export class SpanDetailComponent implements OnChanges {
...
+    ToggleItem<SpanDetailTab> | undefined
+  >(undefined);
+  public readonly activeTab$: Observable<ToggleItem<SpanDetailTab> | undefined> = this.activeTabSubject.asObservable();
+
+  private readonly templateSubject: BehaviorSubject<TemplateRef<unknown> | undefined> = new BehaviorSubject<
+    TemplateRef<unknown> | undefined
+  >(undefined);
+  public readonly template$: Observable<TemplateRef<unknown> | undefined> = this.templateSubject.asObservable();
```
anandtiwary, line 133: This seems weird to me. Why not just do ngSwitch on SpanDetailTab in the template?

adisreyaj: Will refactor to use `ngSwitch` instead.

adisreyaj: Updated!

PR #1510, opened 2022-03-29
feat: support generating a color combination from a color palette

arjunlalb:
## Description
Adding support for generating a color combination from a string id.
This is to be used when a constant color has to be produced for an object based on its ID without having to define an explicit association.

Defining a string id to number hashing logic also to support the above requirement.

### DIFF
--- a/projects/common/src/color/color-palette.test.ts
+++ b/projects/common/src/color/color-palette.test.ts
@@ -31,4 +31,26 @@ describe('Color palette', () => {
     expect(() => new ColorPalette(['black'])).toThrow();
     expect(() => new ColorPalette(['white', 'black'])).not.toThrow();
   });
+
+  test('should return color combinations correctly', () => {
+    const palette = new ColorPalette(['#fffbeb', '#140300']);
+    expect(palette.getColorCombinations(2)).toEqual([
+      {
+        background: 'rgb(255, 251, 235)',
+        foreground: '#080909',
+      },
+      {
+        background: 'rgb(20, 3, 0)',
+        foreground: '#FFFFFF',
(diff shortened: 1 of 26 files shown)


aaron-steinfeld (approved):

aaron-steinfeld (commented):
Could be a worth a couple tests 😉

arjunlalb:
> Could be a worth a couple tests 😉

Sure. Adding them now. I was waiting for the approach to be reviewed before I added tests.

aaron-steinfeld (approved):

projects/common/src/color/color-palette.test.ts
```diff
@@ -31,4 +31,26 @@ describe('Color palette', () => {
...
+      {
+        background: 'rgb(20, 3, 0)',
+        foreground: '#FFFFFF'
+      }
+    ]);
+  });
+
+  test('should generate color for a string as expected from a limited set', () => {
+    const palette = new ColorPalette(['#fffbeb', '#140300', '#789ab7']);
+    expect(palette.getColorCombinationForId(2, 'test')).toEqual({
```
aaron-steinfeld, line 51: Looking at how this is used, we could probably make the count optional and the second param. so just `getColorCombinationForString('test')`

arjunlalb: makes sense. updated @aaron-steinfeld

aaron-steinfeld (dismissed):

aaron-steinfeld:
> @aaron-steinfeld accepted your suggestion. reapproval please.

damn, I failed prettier :(

PR #1511, opened 2022-03-29
feat: code viewer component

itssharmasandeep:
## Description
- `Code viewer` component
- `Download json` component  has been changed into `Download file`
- Follow up changes in other files

https://user-images.githubusercontent.com/78212931/160613447-97b17e26-8037-4d68-8343-2f9789634a59.mov

Screenshot 2022-03-30 at 9 25 54 PM

### Testing
Local Testing is completed.

### Checklist:


projects/components/src/code-viewer/code-viewer.component.ts
```diff
@@ -0,0 +1,110 @@
+import { ChangeDetectionStrategy, Component, Input, OnChanges, TemplateRef } from '@angular/core';
+import { Color, TypedSimpleChanges } from '@hypertrace/common';
+import { isEmpty, isNil } from 'lodash-es';
+import { DownloadFileMetadata } from '../download-file/download-file-metadata';
+
+@Component({
+  selector: 'ht-code-viewer',
```
aaron-steinfeld, line 7: How is this different than snippet viewer? Please consolidate.

aaron-steinfeld: It's possible all usages of it have been removed - if you would prefer to remove that in favor of this one, rather than updating it for the new use cases, that may be a fine option. We just don't want both hanging around.

anandtiwary: snippet viewer is very basic and we don't use it anymore since we started using the markdown-viewer. `snippet-viewer` is a subset of `ht-code-viewer` and eventually we could retire it.

projects/components/src/search-box/search-box.component.ts
```diff
@@ -46,6 +52,9 @@ export class SearchBoxComponent implements OnInit, OnChanges {
   @Input()
   public displayMode: SearchBoxDisplayMode = SearchBoxDisplayMode.Border;
 
+  @Input()
+  public backgroundColor: Color = Color.Transparent;
```
anandtiwary, line 56: can we set backgroundColor in the parent dom element inside code viewer?

projects/components/src/code-viewer/code-viewer.component.ts
```diff
@@ -0,0 +1,110 @@
...
+  public downloadCodeMetadata?: DownloadFileMetadata;
+
+  @Input()
+  public additionalHeaderContent?: TemplateRef<unknown>;
+
+  private codeLineElements?: HTMLElement[];
+
+  public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (changes.code) {
+      this.codeLineElements = undefined;
```
anandtiwary, line 72: hmm.. why don't we create all elements and line numbers, etc right here? Setting it to undefined and then doing it via html seems avoidable

projects/components/src/code-viewer/code-viewer.component.ts
```diff
@@ -0,0 +1,110 @@
...
+            backgroundColor="${Color.White}"
+            (valueChange)="this.onSearch($event)"
+          ></ht-search-box>
+          <ht-download-file *ngIf="this.downloadCodeMetadata" [metadata]="this.downloadCodeMetadata"></ht-download-file>
+        </div>
+      </div>
+      <div class="content">
+        <div class="line-numbers">
+          <div
+            *ngFor="let lineNumber of this.lineNumbers"
```
anandtiwary, line 29: Can we populate `lineNumbers` directly when `code` gets modified?

projects/components/src/code-viewer/code-viewer.component.ts
```diff
@@ -0,0 +1,110 @@
...
+    return new Array(this.code.length).fill(0).map((_, index) => index + 1);
+  }
+
+  public isHighlighted(lineNum: number): boolean {
+    return !isEmpty(this.highlightText) && this.code[lineNum].toLowerCase().includes(this.highlightText.toLowerCase());
+  }
+
+  public onSearch(searchText: string): void {
+    if (isNil(this.codeLineElements)) {
+      this.codeLineElements = Array.from(document.getElementsByClassName('code-line') ?? []) as HTMLElement[];
```
anandtiwary, line 86: please inject `Element` in constructor and run this query using dom api

projects/components/src/download-file/download-file.component.ts
```diff
@@ -45,31 +42,25 @@ export class DownloadJsonComponent {
...
+      this.downloadData(data);
+    });
   }
 
   private downloadData(data: string): void {
     this.renderer.setAttribute(
       this.dlJsonAnchorElement,
       'href',
-      `data:text/json;charset=utf-8,${encodeURIComponent(data)}`
+      `data:text/plain;charset=utf-8,${encodeURIComponent(data)}`
```
anandtiwary, line 61: why this change? How would this change the original json download?

projects/components/src/search-box/search-box.component.ts
anandtiwary: this property shouldn't be required

itssharmasandeep: Removed

projects/components/src/download-file/download-file.component.ts
itssharmasandeep: Since we have yaml as well to download, I moved to plain text, I does not affect original download

projects/components/src/code-viewer/code-viewer.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      const codeLine: string = this.code[index];
+      codeLineElem.innerHTML = this.getReplacedHtmlString(codeLine, searchText);
+    });
+  }
+
+  /**
+   * This gives the replaced HTML string after search.
+   * It uses regex to replace the searched text with a span if found
+   */
+  private getReplacedHtmlString(codeLine: string, searchText: string): string {
```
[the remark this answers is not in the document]
itssharmasandeep: I have used `htHighlight` pipe now. Let me know if we need to fix those issues, We can do that later or here as well.

projects/components/src/code-viewer/code-viewer.component.ts
itssharmasandeep: This is remove since used `htHighlight` pipe now

projects/components/src/code-viewer/code-viewer.component.ts
itssharmasandeep: Removed snippet viewer and moved those configs here

projects/components/src/viewer/code-viewer/code-viewer.component.scss
```diff
@@ -68,5 +70,18 @@
...
+    .copy-to-clipboard {
+      position: absolute;
+      color: $gray-9;
+      right: 0;
+      top: 0;
+    }
+  }
+
+  ::ng-deep {
+    mark {
```
anandtiwary, line 83: it won't work without ng-deep? 

also, we should put it inside `.code-lines` class

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -0,0 +1,127 @@
...
+            [ngClass]="{ highlight: this.isHighlighted(lineNumber - 1) }"
+          >
+            {{ lineNumber }}
+          </div>
+        </div>
+        <div class="code-lines">
+          <div
+            *ngFor="let codeLine of this.codeLines; let index = index"
+            class="code-line"
+            [ngClass]="{ highlight: this.isHighlighted(index) }"
```
anandtiwary, line 44: nit: can we rename it to `isLIneHighlighted()` and the class to `line-highlight`. It would help in conveying that L48 is for text highlight only

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -0,0 +1,127 @@
...
+          <ht-download-file
+            *ngIf="!!this.code && !!this.downloadFileName"
+            [metadata]="this.downloadCodeMetadata"
+          ></ht-download-file>
+        </div>
+      </div>
+      <div class="content">
+        <div class="line-numbers">
+          <div
+            *ngFor="let lineNumber of this.lineNumbers"
```
anandtiwary, line 33: even though `lineNumbers`, codeLines are getters, they would still be executed with every `do check` cycle.  if we have a huge spec then this could be a costly operation. We could just calculate it one time inside that ngOnChanges. Any reason for not doing that and choosing getter approach?

anandtiwary: same for `downloadCodeMetadata`

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: done

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: done

anandtiwary (approved):

PR #1514, opened 2022-03-29
refactor: re-structure of page time range to be query param centric

Christian862:
## Description
After discovering a page time range bug related to a race condition, we re-structured the way page time range works, to be query param centric. This means the time range is no set before/after in a reactionary way, but instead the time range is specified on each navigation as a query param, the TR service then takes the query param from the navigation event and sets it accordingly. 

This were the changes required:

- defaultTimeRange  to only be defined on root level routes ( to solve the lazy load issue of resolving default TR without the activated route)
- shouldSaveTimeRange  to be placed on every? child route that should share read/write prefs with siblings/parent for their time range (events/attackers). This to know when to save/when not to
- Use getRoutConfig  to resolve defaults in getDefaultTimeRangeForCurrentRoute  with the path (no lazy boundary issues with the above).
- Pass the time range down to the NavItemComponent  to be resolved and set as a query param
- Listen for nav end events in TR service and set the TR according to their time query param
- Modify query param building for time range to include relative time ranges.

DRAFT - testing not updated yet

### Testing

### Checklist:

### Documentation


projects/components/src/navigation/nav-item/nav-item.component.ts
```diff
@@ -57,12 +57,23 @@ export class NavItemComponent {
...
   public readonly navItemViewStyle?: NavViewStyle;
 
-  public buildNavigationParam = (item: NavItemLinkConfig): NavigationParams => ({
-    navType: NavigationParamsType.InApp,
-    path: item.matchPaths[0],
-    relativeTo: this.activatedRoute,
-    replaceCurrentHistory: item.replaceCurrentHistory
-  });
+  public buildNavigationParam = (item: NavItemLinkConfig): NavigationParams => {
+    const timeRange = this.config.timeRangeResolver ? this.config.timeRangeResolver() : undefined;
```
aaron-steinfeld, line 61: nit - optional chain
```suggestion
    const timeRange = this.config.timeRangeResolver?.();
```

projects/common/src/time/time-range.service.ts
```diff
@@ -62,16 +66,44 @@ export class TimeRangeService {
...
+    return this.navigationService.navigation$.pipe(
+      filter(
+        activatedRoute => !isNil(activatedRoute.snapshot.queryParamMap.get(TimeRangeService.TIME_RANGE_QUERY_PARAM))
+      ),
+      map(activeRoute => {
+        const timeRangeQueryParamString = activeRoute.snapshot.queryParamMap.get(
+          TimeRangeService.TIME_RANGE_QUERY_PARAM
+        );
+
+        return this.timeRangeFromUrlString(timeRangeQueryParamString!);
```
aaron-steinfeld, line 91: So when do we want to emit? 
- Current code here would change the time range every time we navigate. That would break flows like drilling down in the same TR
- We could do it only when the TR string changes (e.g. 1h -> 1d), but that means if we change top level items when the setting is the same, we wouldn't refresh (this is the behavior today, so seems acceptable)
- If we need to know when to refresh and when not to, we'll likely need another param

projects/common/src/time/time-range.service.ts
```diff
@@ -62,16 +66,44 @@ export class TimeRangeService {
...
-        take(1), // Wait for first navigation
-        switchMap(activatedRoute => activatedRoute.queryParamMap), // Get the params from it
-        take(1), // Only the first set of params
-        map(paramMap => paramMap.get(TimeRangeService.TIME_RANGE_QUERY_PARAM)), // Extract the time range value from it
-        filter((timeRangeString): timeRangeString is string => !isEmpty(timeRangeString)), // Only valid time ranges
-        map(timeRangeString => this.timeRangeFromUrlString(timeRangeString)),
-        catchError(() => EMPTY)
+        switchMap(featureState => {
+          if (featureState === FeatureState.Enabled) {
+            return this.pageTimeRangeInitialization();
```
aaron-steinfeld, line 102: nit - the method returns an observable of a time range, doesn't actually do the initialization and should be renamed (same with the global version).

src/app/routes/root-routing.module.ts
```diff
@@ -26,7 +26,8 @@ const ROUTE_CONFIG: HtRoute[] = [
                 icon: IconType.Dashboard,
                 label: 'Dashboard'
               },
-              defaultTimeRange: new RelativeTimeRange(new TimeDuration(1, TimeUnit.Hour))
+              defaultTimeRange: new RelativeTimeRange(new TimeDuration(1, TimeUnit.Hour)),
+              shouldSavePageTimeRange: true
```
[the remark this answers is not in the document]
Christian862: Yeah this crossed my mind. Correct, if the route defines it's own, we'd always want to save. However I'm gonna leave it  as is for readability - to anyone looking at these route configs without the full picture, I think it'd be easier to understand with this property here. One condition being satisfied based on one property rather than two, is a bit clearer

projects/common/src/time/time-range.service.ts
Christian862: No this only emits when we  navigate and have a time query param. Drill behaviour still works as expected.

projects/common/src/time/time-range.service.ts
```diff
@@ -66,7 +66,7 @@ export class TimeRangeService {
     this.setTimeRange(this.getCurrentTimeRange());
   }
 
-  private globalTimeRangeInitialization(): Observable<TimeRange> {
+  private globalTimeRangeInitConfigAndChanges(): Observable<TimeRange> {
```
aaron-steinfeld, line 69: nit: not changes - we're only take(1)'ing. Maybe `getInitialTimeRange()` and below `getPageTimeRangeChanges()`

projects/components/src/navigation/nav-item/nav-item.component.ts
```diff
@@ -67,9 +67,15 @@ export class NavItemComponent {
 
     if (this.config.pageLevelTimeRangeIsEnabled && this.config.timeRangeResolver) {
       const timeRange = this.config.timeRangeResolver();
+      const timeRangeQueryParam = this.timeRangeService.toQueryParams(timeRange);
+      const timeRangeRefreshQueryParam = this.timeRangeService.getRefreshTimeRangeQueryParam();
```
aaron-steinfeld, line 71: nit: could probably put this as a flag on the query param generation function rather than have two separate ones

aaron-steinfeld (dismissed):

anandtiwary (approved):

projects/common/src/time/time-range.service.ts
```diff
@@ -51,31 +53,67 @@ export class TimeRangeService {
   }
 
   public setRelativeRange(value: number, unit: TimeUnit): this {
-    return this.setTimeRange(TimeRangeService.toRelativeTimeRange(value, unit));
+    // return this.setTimeRange(TimeRangeService.toRelativeTimeRange(value, unit));
```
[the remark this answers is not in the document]
Christian862: removed thanks, noticed just have i pushed

projects/components/src/navigation/nav-item/nav-item.component.ts
```diff
@@ -9,7 +9,7 @@ import { NavItemLinkConfig, NavViewStyle } from '../navigation.config';
   styleUrls: ['./nav-item.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <ht-link *ngIf="this.config" [paramsOrUrl]="buildNavigationParam | htMemoize: this.config">
+    <ht-link *ngIf="this.config" [paramsOrUrl]="this.buildNavigationParam(this.config)">
```
aaron-steinfeld, line 12: we do want to bring memoization back here - no need to build this twice

projects/common/src/time/time-range.service.ts
```diff
@@ -93,6 +92,21 @@ export class TimeRangeService {
     );
   }
 
+  private initializeTimeRange(): void {
+    concat(this.getInitialTimeRange(), this.getPageTimeRangeChanges()).subscribe(timeRange => {
```
aaron-steinfeld, line 96: one option discussed offline - the two branches of the subscribe basically handle the respective emissions of these two observables. We could do it as two separate subscriptions and it would probably be easier to understand.

anandtiwary (approved):

anandtiwary (approved):

PR #1517, opened 2022-03-30
feat: add storage scoping

aaron-steinfeld:
## Description
Add the ability to configure storage to use a scope to contain its values, optionally falling back to the global scope. This can be used in several ways - storing by version, storing by user, storing by feature etc.

### Testing
Added UTs

### Checklist:

### DIFF
--- a/projects/common/src/utilities/browser/storage/abstract-storage.ts
+++ b/projects/common/src/utilities/browser/storage/abstract-storage.ts
@@ -1,17 +1,36 @@
 import { Observable, Subject } from 'rxjs';
 import { distinctUntilChanged, filter, map, startWith } from 'rxjs/operators';
+import { DictionaryStorageImpl } from './dictionary-storage-impl';
 
 // A small abstraction on browser's storage for mocking and cleaning up the API a bit
 export abstract class AbstractStorage {
   private readonly changeSubject: Subject<string> = new Subject();
-  public constructor(private readonly storage: Storage) {}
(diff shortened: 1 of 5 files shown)


projects/common/src/utilities/browser/storage/abstract-storage.ts
```diff
@@ -1,17 +1,36 @@
 import { Observable, Subject } from 'rxjs';
 import { distinctUntilChanged, filter, map, startWith } from 'rxjs/operators';
+import { DictionaryStorageImpl } from './dictionary-storage-impl';
 
 // A small abstraction on browser's storage for mocking and cleaning up the API a bit
 export abstract class AbstractStorage {
   private readonly changeSubject: Subject<string> = new Subject();
```
itssharmasandeep, line 7: is this redundant ?

aaron-steinfeld: Redundant with what? Not sure I follow - if you mean the type portion, we have a lint rule that requires all class members to have a declared type.

itssharmasandeep: I meant, this `changeSubject`. I saw we're using `changeSubject.next` but apart from that this is not being used, or am I missing something?

aaron-steinfeld: It's used in `watch` - it's not visible by default in this PR because it didn't change (neither did the subject, just the line below it 😉 )

PR #1521, opened 2022-03-31
feat: adding font size property to legend component and style fix

itssharmasandeep:
## Description
Adding font size property to legend component and style fix.

### Testing
Local testing done

### Checklist:

### DIFF
--- a/projects/observability/src/shared/components/utils/d3/d3-visualization.scss
+++ b/projects/observability/src/shared/components/utils/d3/d3-visualization.scss
@@ -17,7 +17,7 @@
     flex-direction: row-reverse;
     .chart-legend-container {
       margin-left: 12px;
-      margin-right: 12px;
+      margin-right: 36px;
     }
   }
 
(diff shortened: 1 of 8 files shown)


projects/observability/src/shared/components/utils/d3/d3-visualization.scss
```diff
@@ -17,7 +17,7 @@
     flex-direction: row-reverse;
     .chart-legend-container {
       margin-left: 12px;
-      margin-right: 12px;
+      margin-right: 36px;
```
anandtiwary, line 20: this may not be the case for all legend orientations. We should only apply larger right migin when we have left aligned legend

itssharmasandeep: Yeah this only the case for left aligned, `row-reverse` class will only be applied when it is left aligned

itssharmasandeep: in `d3-visualization.builder`

```ts
protected getOuterContainerLayoutClass(configuration: TInternalConfig): string {
    switch (configuration.legend) {
      case LegendPosition.Bottom:
        return 'column';
      case LegendPosition.Top:
      case LegendPosition.TopLeft:
      case LegendPosition.TopRight:
        return 'column-reverse';
      case LegendPosition.Right:
        return 'row';
      case LegendPosition.Left:
        return 'row-reverse';
      case LegendPosition.None:
        return '';
      default:
        return assertUnreachable(configuration.legend);
    }
  }
```

anandtiwary (approved):

PR #1522, opened 2022-03-31
feat: add operator argument to GraphQlIdFilter

jake-bassett:
## Description
Parameterize operator for GraphQLIdFilter.

### DIFF
--- a/projects/observability/src/shared/graphql/model/schema/filter/id/graphql-id-filter.ts
+++ b/projects/observability/src/shared/graphql/model/schema/filter/id/graphql-id-filter.ts
@@ -4,12 +4,16 @@ import { GraphQlFilter, GraphQlFilterType, GraphQlOperatorType } from '../graphq
 export class GraphQlIdFilter implements GraphQlFilter {
   public readonly key: string = 'id';
 
-  public constructor(public readonly id: string, public readonly idScope: string) {}
+  public constructor(
+    public readonly id: string,
+    public readonly idScope: string,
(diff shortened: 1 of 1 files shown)


projects/observability/src/shared/graphql/model/schema/filter/id/graphql-id-filter.ts
```diff
@@ -4,12 +4,16 @@ import { GraphQlFilter, GraphQlFilterType, GraphQlOperatorType } from '../graphq
...
+    public readonly id: string,
+    public readonly idScope: string,
+    public readonly operator: GraphQlOperatorType = GraphQlOperatorType.Equals
+  ) {}
 
   public asArgumentObjects(): [GraphQlIdFilterArgument] {
     return [
       {
-        operator: new GraphQlEnumArgument(GraphQlOperatorType.Equals),
+        operator: new GraphQlEnumArgument(this.operator),
```
anandtiwary, line 16: Any reason why we couldn't use a GraphQlFieldFilter if we want to use a different operator?

jake-bassett: Do you mean use a `GraphQlFieldFilter` for other operators? We can, at least in the case I need this for. But using the ID filter is a little more semantically correct and helps readers better understand intent.

Not sure if there is more under the hood that this also helps, maybe @aaron-steinfeld can comment.

PR #1523, opened 2022-04-01
feat: metric card component

itssharmasandeep:
## Description
Metric card component

### Testing
Local testing done.

### Checklist:


projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -0,0 +1,55 @@
...
+                [icon]="this.icon"
+                size="${IconSize.Small}"
+                [color]="this.indicatorColor"
+              ></ht-icon>
+            </ng-container>
+          </ng-container>
+        </div>
+        <div class="title-text">{{ this.titleText }}</div>
+      </div>
+      <div class="value">{{ value }}</div>
```
arjunlalb, line 30: we should apply the htDisplayNumber pipe if this is a number and htDisplayString pipe if its a string.

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    </div>
+  `
+})
+export class MetricCardComponent {
+  @Input()
+  public titleText?: string;
```
arjunlalb, line 36: can be simplified to `title`

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -0,0 +1,55 @@
...
+  public indicator: MetricCardIndicatorType = MetricCardIndicatorType.Dot;
+
+  @Input()
+  public icon?: string; // Only be used when indicator type is Icon
+
+  @Input()
+  public indicatorColor: Color = Color.Gray7;
+
+  @Input()
+  public backgroundColor: Color = Color.OffWhite;
```
arjunlalb, line 51: this should also default to white/transparent.

projects/components/src/metric-card/metric-card.ts
```diff
@@ -0,0 +1,4 @@
+export const enum MetricCardIndicatorType {
+  Dot = 'dot',
+  Icon = 'icon'
```
arjunlalb, line 3: `None` also should be supported if we just want a title and value.

arjunlalb (approved):
Comments can be addressed in a follow up.

projects/components/src/metric-card/metric-card.component.ts
itssharmasandeep: Actually, I do not want that, this basically collide with HTML `title` property and we can get weird tooltip

PR #1531, opened 2022-04-04
Query params resetting on nav item click

Christian862:
## Description
When navigating with nav items, the selected environment would reset due to it's query params being removed from the nav event. Before TR changes this wasn't an issue because there was no query params in the nav item link. 

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/navigation/nav-item/nav-item.component.ts
+++ b/projects/components/src/navigation/nav-item/nav-item.component.ts
@@ -45,8 +45,6 @@ import { NavItemLinkConfig, NavViewStyle } from '../navigation.config';
   `
 })
 export class NavItemComponent {
-  private static readonly QUERY_PARAM_HANDLING_STRATEGY = 'merge';
-
   @Input()
   public config!: NavItemLinkConfig;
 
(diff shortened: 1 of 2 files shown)


projects/components/src/navigation/nav-item/nav-item.component.ts
```diff
@@ -45,6 +45,8 @@ import { NavItemLinkConfig, NavViewStyle } from '../navigation.config';
   `
 })
 export class NavItemComponent {
+  private static readonly QUERY_PARAM_HANDLING_STRATEGY = 'merge';
```
aaron-steinfeld, line 48: Have we considered fixing nav service rather than the particular call? Do we know of use cases in nav service where we'd not want this behavior? seems like a bug to me.

aaron-steinfeld: Specifically I'm referring to changing 
```ts
        queryParams: params?.queryParams ?? this.buildQueryParam(),
        queryParamsHandling: params?.queryParamsHandling,
```
to something like
```ts
        queryParams: {...this.buildQueryParam(), ...(params?.queryParams ?? {})
        queryParamsHandling: params?.queryParamsHandling,
```
The idea being that assuming we're doing in app navigation, we never want to remove the global query params that are preserved by the buildQueryParam() call.

aaron-steinfeld: The more that I think about it, I think the current PR would have more issues. Take the scenario where you navigate away from a page that has its own query params - e.g. explorer or anything with tables or filters. Merging query params would send all of explorer's query params to the new page.

aaron-steinfeld (approved):

anandtiwary (approved):
seems fine to me.

PR #1537, opened 2022-04-06
feat: allow client side pagination in table

jake-bassett:
## Description

### Testing

### Checklist:

### Documentation


projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -328,11 +328,19 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
-     * index off the start of the result rows.
+     * With server-side pagination, the "rows" here are the results that are fetched. Since they are fetched with an
+     * offset in the request, we just index off the start of the result rows.
+     *
+     * However for when the backend APIs do not support pagination, the backend will return all results. In this case
+     * we can see if there are more results than we requested. If so, paginate within those results here.
+     *
+     * Historical Note & Warning:
+     * In the distant past for even server-side pagination we would prefetch additional rows, cache them, and
+     * paginate within those results. This caching will no longer work. In fact, it will break the below logic
```
aaron-steinfeld, line 339: This seems fairly easily fixable though, which would be an additional win. If we pass the request into this, we have the offset and thus know what the data's start is too, right?

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -328,11 +328,19 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
+     * due to the `length > pageSize` comparison we are doing here.
      */
-    const start = 0;
-    const end = pageConfig.pageSize;
+    const start = rows.length > pageConfig.pageSize ? pageConfig.pageSize * pageConfig.pageIndex : 0;
```
anandtiwary, line 342: I have seen another issue where the pagination row is hidden if the number of rows < page size. Could you please look into that as well?

aaron-steinfeld: Pretty sure that was by design

jake-bassett: Yeah, this was an early ask of us. Agreed we should remove it as it is confusing to users.

projects/components/src/table/data/table-cdk-data-source.ts
jake-bassett: I like this, but not sure it will work. The table is the one that generates the request - it applies the limits, sorting, filtering, etc based on the usage of the table component and paginator.

aaron-steinfeld: I agree, that alone won't work here - but in broad strokes, let me run through how it used to work before refactor.

The table generates the requests, sends it to the data source, who we presume to respect the request. Whether it does that client or server side is not a concern of the table. So if we sent a request with offset 0 and limit 50, we expect the first row back to be the first result at the request offset.

So if we get 100 results back, we should be able to slice that into 0-50 for display, easy enough. Now, what's missing is that when we get a new pagination event, rather than blindly firing a new request we used to check our current rows and the request that generated them. If we're just requesting the next page and the prior result had enough rows to satisfy that, we can just reslice and be on our way. If not, we can fire a query like we are now.

projects/observability/src/shared/dashboard/data/graphql/table/entity/entity-table-data-source.model.ts
```diff
@@ -63,7 +63,7 @@ export class EntityTableDataSourceModel extends TableDataSourceModel {
       requestType: ENTITIES_GQL_REQUEST,
       entityType: this.entityType,
       properties: request.columns.map(column => column.specification).concat(...this.additionalSpecifications),
-      limit: this.limit !== undefined ? this.limit : request.position.limit * 2, // Prefetch 2 pages
+      limit: this.limit !== undefined ? this.limit : request.position.limit,
```
jake-bassett, line 66: We could technically keep these prefetches if we want now.

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -274,19 +283,57 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
-            : rows
-        )
-      );
+    return this.hasCachedRowsForPageEvent(pageEvent)
+      ? this.fetchCachedData(pageEvent)
+      : this.fetchNewData(this.buildRequest(columnConfigs, pageEvent, filters, queryProperties));
+  }
+
+  private hasCachedRowsForPageEvent(pageEvent: PageEvent): boolean {
+    if (this.cachedRows.length < pageEvent.pageSize) {
```
anandtiwary, line 292: should it be `pageEvent.pageSize * pageEvent.pageIndex`?

jake-bassett: I think the check you are looking for is below.. This line is just a quick sanity check to fail early if we don't even have as many results cached as we have requested.

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -274,19 +283,57 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
+  private calcOffsetWithinCachedRows(pageEvent: PageEvent): number {
+    const cachedOffset = this.cachedPosition?.startIndex ?? 0;
+    const currentOffset = pageEvent.pageIndex * pageEvent.pageSize;
+
+    return currentOffset - cachedOffset;
+  }
+
+  private fetchNewData(request: TableDataRequest): Observable<StatefulTableRow[]> {
+    if (this.tableDataSourceProvider.data === undefined) {
+      return of([]);
```
arjunlalb, line 312: this will lead us to absorb errors, right? shouldn't this be an error if the datasource provider is undefined?

arjunlalb: I know this isn't new change. Just a question in general :)

arjunlalb (approved):

projects/components/src/table/data/table-cdk-data-source.ts
jake-bassett: Yeah, the check has been there a while. Pretty sure we never actually hit this.

anandtiwary (approved):

PR #1538, opened 2022-04-06
fix: avoid extra emit on deep link

aaron-steinfeld:
## Description
Previously, when loading the app with a time range parameter, the first time we navigate would _always_ trigger a time range change call, whether it was caused by one or not. 

This was because we concat'd two different observables - one to initialize the time range from URL, and another to consume changes from the URL. The change consuming observable always fired the first time since it could not yet tell if it was a change (distinctUntilChanged). That led to the behavior where if the first initializing observable did anything, the first change consuming action would be a duplicate (unless that change happened to be the time range).  If the first initializing observable did nothing (because there was no TR), things worked as expected.

To fix this, we separated out the two subscriptions (so they can happen concurrently), and always discard the first one from the change consumer _after_ it loads its distinctUntilChanged cache. The init code is left to determine what, if anything to do with the first nav. 

### Testing
Local e2e testing of various scenarios

### DIFF
--- a/projects/common/src/time/time-range.service.test.ts
+++ b/projects/common/src/time/time-range.service.test.ts
@@ -78,10 +78,7 @@ describe('Time range(TR) service', () => {
                 }
               } as unknown) as ActivatedRoute
             }),
-            getQueryParameter: jest
-              .fn()
-              .mockReturnValueOnce(firstArrivingTimeRange.toUrlString())
-              .mockReturnValue(secondArrivingTimeRange.toUrlString()),
+            getQueryParameter: jest.fn().mockReturnValue(secondArrivingTimeRange.toUrlString()),
             getCurrentActivatedRoute: () =>
               (({ snapshot: { queryParams: { time: 'test-value' } } } as unknown) as ActivatedRoute)
           })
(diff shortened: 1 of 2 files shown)


projects/common/src/time/time-range.service.test.ts
```diff
@@ -78,10 +78,7 @@ describe('Time range(TR) service', () => {
                 }
               } as unknown) as ActivatedRoute
             }),
-            getQueryParameter: jest
-              .fn()
-              .mockReturnValueOnce(firstArrivingTimeRange.toUrlString())
-              .mockReturnValue(secondArrivingTimeRange.toUrlString()),
+            getQueryParameter: jest.fn().mockReturnValue(secondArrivingTimeRange.toUrlString()),
```
aaron-steinfeld, line 81: this function is only used to determine changes which have their own subscriber now, so the first nav doesn't trigger it.

projects/common/src/time/time-range.service.test.ts
```diff
@@ -92,8 +89,8 @@ describe('Time range(TR) service', () => {
 
       expect(() => spectator.service.getCurrentTimeRange()).toThrow();
 
-      expectObservable(recordedTimeRanges).toBe('-x----y', {
-        x: new FixedTimeRange(new Date(1573255100253), new Date(1573255111159)),
+      expectObservable(recordedTimeRanges).toBe('-x---y', {
```
aaron-steinfeld, line 92: not sure why this didn't line up before with the emit pattern, now it does.

anandtiwary (approved):

PR #1543, opened 2022-04-08
fix: trimming search text

itssharmasandeep:
## Description
Trimming search text before emitting

### Testing
Local testing is done.

### Checklist:

### DIFF
--- a/projects/components/src/search-box/search-box.component.ts
+++ b/projects/components/src/search-box/search-box.component.ts
@@ -90,7 +90,7 @@ export class SearchBoxComponent implements OnInit, OnChanges {
     this.subscriptionLifecycle.add(
       this.debouncedValueSubject
         .pipe(debounceTime(this.debounceTime ?? 0))
-        .subscribe(value => this.valueChange.emit(value))
+        .subscribe(value => this.valueChange.emit(value.trim()))
     );
   }
 }


projects/components/src/search-box/search-box.component.ts
```diff
@@ -90,7 +90,7 @@ export class SearchBoxComponent implements OnInit, OnChanges {
     this.subscriptionLifecycle.add(
       this.debouncedValueSubject
         .pipe(debounceTime(this.debounceTime ?? 0))
-        .subscribe(value => this.valueChange.emit(value))
+        .subscribe(value => this.valueChange.emit(value.trim()))
```
arjunlalb, line 93: this trim should be handled in `onValueChange`. Not this method's responsibility.

projects/components/src/search-box/search-box.component.ts
```diff
@@ -69,11 +69,11 @@ export class SearchBoxComponent implements OnInit, OnChanges {
   }
 
   public onSubmit(): void {
-    this.submit.emit(this.value);
+    this.submit.emit(this.value.trim());
   }
 
   public onValueChange(): void {
-    this.debouncedValueSubject.next(this.value);
+    this.debouncedValueSubject.next(this.value.trim());
```
arjunlalb, line 76: we should be doing `this.value = this.value.trim()` here and then emit so that the change reflects in the input box as well.

itssharmasandeep: We do not want that.

itssharmasandeep: If we do that it will not even let us put a space, right?

arjunlalb: it shouldn't prevent you from entering spaces unless we are triggering CD cycle right after.

arjunlalb (approved):

PR #1544, opened 2022-04-08
feat: code viewer search and highlight with scroll into view

itssharmasandeep:
## Description
Code viewer search and highlight with scroll into view

### Testing
Local testing is done.

### Checklist:


projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -126,4 +154,26 @@ export class CodeViewerComponent implements OnChanges {
...
+    const codeViewerContentElement = this.element.nativeElement.querySelector('#code-viewer-content') as Node;
+    this.domMutationObserver.observe(codeViewerContentElement, {
+      subtree: true,
+      childList: true,
+      attributes: true,
+      attributeFilter: ['class']
+    });
+  }
+
+  private onDomMutation(mutations: MutationRecord[]): void {
```
anandtiwary, line 168: If a text search creates 10 mark elements, would we call this method 10 times?

anandtiwary, line 168: If a text search creates 10 mark elements, would we call this method 10 times?

anandtiwary: Can we do this instead? 
```
  public readonly debouncedOnDomMutation: (mutations: MutationRecord[]) => void = debounce(
    this.onDomMutation,
    500,
     {leading: true}
  );
  
  ```

Comment (anandtiwary) [reply projects/components/src/viewer/code-viewer/code-viewer.component.ts:168]: Can we do this instead? 
```
  public readonly debouncedOnDomMutation: (mutations: MutationRecord[]) => void = debounce(
    this.onDomMutation,
    500,
     {leading: true}
  );
  
  ```

itssharmasandeep: We can do it, but for search debounce has already been added in the `search-box` component, do we really need this?
To this we would be making the search box debounce as `0` and then I guess we can surely use this

anandtiwary: Cool

anandtiwary: This debounce would be for avoiding multiple multiple mutation action. Since, you mentioned this is not happening in the other comment then i guess we don't need it.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -103,6 +117,16 @@ export class CodeViewerComponent implements OnChanges {
   public lineNumbers: number[] = [];
   public searchText: string = '';
 
+  private readonly domMutationObserver: MutationObserver = new MutationObserver(mutations =>
```
aaron-steinfeld, line 120: uh oh someone's getting fancy!

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -126,4 +154,26 @@ export class CodeViewerComponent implements OnChanges {
   public onSearch(searchText: string): void {
     this.searchText = searchText;
   }
+
+  private observeDomMutations(): void {
+    const codeViewerContentElement = this.element.nativeElement.querySelector('#code-viewer-content') as Node;
```
aaron-steinfeld, line 159: Select via child queries rather than using native API

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -126,4 +154,26 @@ export class CodeViewerComponent implements OnChanges {
...
+    });
+  }
+
+  private onDomMutation(mutations: MutationRecord[]): void {
+    const mutationType: MutationRecordType = mutations[0].type;
+
+    if (mutationType === 'childList') {
+      const markElement: HTMLElement = this.element.nativeElement.querySelector('mark');
+      markElement?.scrollIntoView();
+    } else if (mutationType === 'attributes') {
```
aaron-steinfeld, line 174: Is this all to scroll into view on highlight? Doesn't' CDK give us some similar functionality without us building it ourselves? I guess a second question is - do we need to do this on the fly? This seems like it was an excuse to use dom mutation observer 😉

anandtiwary: heh.. Most of these dom elements are added dynamically after a search or highlight happens. In those cases, we don't get the reference to a new element unless they are actually added to the dom. Mutation observer is providing that callback through which we can deterministically call `scrollIntoView` on the first match

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: I didn't get this, can you provide an example?
Since I have merged this PR, I can take this in follow-up as well.

aaron-steinfeld: For example from combo box:
```ts
  @ViewChild('input', { read: ElementRef })
  public readonly input!: ElementRef;
```

projects/components/src/viewer/code-viewer/code-viewer.component.ts
aaron-steinfeld: ah search makes sense - with highlight, I was assuming that we can just onChange to know when to rescroll

PR #1563, opened 2022-04-21
fix: order items in multi select alphabetically and bubble selected items to top

arjunlalb:
## Description
order items in multi select alphabetically and bubble selected items to top

### Testing
Updated UTs.


aaron-steinfeld (commented):
If you want to make this a config option, that's fine, but I wouldn't change the default. Often, things have their own order and that's for the caller to decide (e.g. low, medium, high)

arjunlalb:
> If you want to make this a config option, that's fine, but I wouldn't change the default. Often, things have their own order and that's for the caller to decide (e.g. low, medium, high)

Good point. Parking this until I can get back on this then.

PR #1564, opened 2022-04-25
fix: escape regex reserve characters in highlight pipe

arjunlalb:
## Description
Highlight pipe uses regex matching to replace strings with corresponding highlight HTML.
This approach fails when the string to be replaced itself is a regex. The regex based matching fails since the method tries to parse it as a string. Hence, escaping reserved regex characters before doing the replacement step.

### Testing
UT added.

### Checklist:

### DIFF
--- a/projects/common/src/utilities/formatters/string/highlight.pipe.ts
+++ b/projects/common/src/utilities/formatters/string/highlight.pipe.ts
@@ -2,7 +2,6 @@ import { Pipe, PipeTransform } from '@angular/core';
 import { isArray } from 'lodash-es';
 import { assertUnreachable } from '../../lang/lang-utils';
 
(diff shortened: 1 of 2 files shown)


projects/common/src/utilities/formatters/string/highlight.pipe.ts
```diff
@@ -5,6 +5,10 @@ import { assertUnreachable } from '../../lang/lang-utils';
 // TODO: Currently htHighlight does not escape reserved regex characters
```
itssharmasandeep, line 5: remove this TODO??

arjunlalb: good catch. removed it. @itssharmasandeep

itssharmasandeep (approved):

PR #1570, opened 2022-04-26
feat: syntax highlighting for code viewer

itssharmasandeep:
## Description
This PR contains the changes related to:
- Syntax highlighting using `highlight.js`
- Changes in the approach for search.
- Adding no data message if code is empty.

`Search with syntax highlighting `

https://user-images.githubusercontent.com/78212931/165265547-59e39133-b99a-41c9-9b80-5aaf8f1fd55b.mov

`No data message`

Screenshot 2022-04-26 at 2 32 13 PM

### Testing
Local testing is done.

### Checklist:


projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +159,66 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+    const codeTexts = this.codeTexts.map(text => text.toLowerCase());
+
+    // Remove existing child background elements
+    this.codeLineBackgroundElements.forEach(codeLineBackgroundElement => {
+      const codeLineBackgroundElemNode: Node = codeLineBackgroundElement.nativeElement;
+
+      while (codeLineBackgroundElemNode.firstChild) {
+        codeLineBackgroundElemNode.removeChild(codeLineBackgroundElemNode.firstChild);
+      }
+    });
```
arjunlalb, line 184: instead of this, see if you can use the `replaceChildren()` method to just clear everything in one go.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +159,66 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+
+    const searchLen: number = searchText.length;
+    if (searchLen === 0) {
+      return;
+    }
+
+    const searchedPositions: Position[][] = Array.from(Array(codeTexts.length), () => new Array(0));
```
arjunlalb, line 191: isn't this the same as?
```suggestion
    const searchedPositions: Position[][] = Array.from(Array(codeTexts.length), () => ([]));
```

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +159,66 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+          i = i + searchLen - 1;
+        }
+      }
+    });
+
+    // Add background elements for searched positions
+    this.codeLineBackgroundElements.forEach((codeLineBackgroundElement, index) => {
+      searchedPositions[index].forEach(searchedPosition =>
+        (codeLineBackgroundElement.nativeElement as Node).appendChild(this.getBackgroundElement(searchedPosition))
+      );
```
arjunlalb, line 207: this is inefficient.

Instead, create a document fragment with all the new Elements inside it and add the fragment to the DOM in one go.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +159,66 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+    if (searchLen === 0) {
+      return;
+    }
+
+    const searchedPositions: Position[][] = Array.from(Array(codeTexts.length), () => new Array(0));
+
+    // Get all searched text positions
+    codeTexts.forEach((codeText, index) => {
+      for (let i = 0; i <= codeText.length - searchLen; i++) {
+        if (codeText.substr(i, searchLen) === searchText) {
```
arjunlalb, line 196: instead of this, you can use findIndex to progressively look for match in the remaining string. `findIndex` can take a second argument which instructs it on where to begin the search from. So, after each match you can just move the start index to that point and continue the search. 

You wouldn't have to do substring extraction and matching at every index with that approach.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: Used innerHTMl method

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: So basically we can only create a fragment for each line, that's what I have done.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: Yeah. I have used `indexOf` as I am using string here, but thanks though.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -6,12 +6,17 @@ import {
...
   OnChanges,
   OnDestroy,
-  TemplateRef
+  QueryList,
+  TemplateRef,
+  ViewChildren
 } from '@angular/core';
+import { IconType } from '@hypertrace/assets-library';
 import { Color, TypedSimpleChanges } from '@hypertrace/common';
+import * as HighlightJs from 'highlight.js';
```
anandtiwary, line 15: please avoid * import.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +159,69 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+  }
+
+  // Background elements for search
+  private getBackgroundElements(positions: Position[]): DocumentFragment {
+    const backgroundElemDocFragment: DocumentFragment = document.createDocumentFragment();
+
+    positions.forEach(position => {
+      const backgroundElem = document.createElement('div');
+      backgroundElem.className = 'bg-searched';
+      backgroundElem.style.height = '100%';
```
anandtiwary, line 215: you can possibly put height, background color and position in `bg-searched` class

tslint.json
```diff
@@ -23,7 +23,8 @@
         "@hypertrace/graphql-client",
         "@hypertrace/test-utils",
         "@hypertrace/observability",
-        "ng-mocks"
+        "ng-mocks",
+        "highlight.js"
```
anandtiwary, line 27: this wouldn't be required if you declare highlight.js as a peer dependency

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: Done

package.json
```diff
@@ -61,6 +61,7 @@
     "d3-zoom": "^1.8.3",
     "graphql": "^15.8.0",
     "graphql-tag": "^2.12.6",
+    "highlight.js": "^11.5.1",
```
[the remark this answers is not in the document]
itssharmasandeep: Why do we need this? I mean we don't add any peer dependencies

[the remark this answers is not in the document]
anandtiwary: We do. Peer dependency is a way to declare third-party dependencies. This needs to be declared in the components project.

tslint.json
anandtiwary: add peer in components -> package.json

itssharmasandeep: Done, thanks

package.json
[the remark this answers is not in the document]
itssharmasandeep: Done thanks

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
 import { Color, TypedSimpleChanges } from '@hypertrace/common';
-import HighlightJs from 'highlight.js';
+import hljs from 'highlight.js';
```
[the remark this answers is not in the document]
itssharmasandeep: No this function is not exported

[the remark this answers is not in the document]
anandtiwary: discussed offline. This is fine

anandtiwary:
@itssharmasandeep don't merge this pr unless you have its t-ui pr ready. Otherwise, the build will break

itssharmasandeep:
> @itssharmasandeep don't merge this pr unless you have its t-ui pr ready. Otherwise, the build will break

already ready!!

projects/components/src/viewer/code-viewer/code-viewer.component.ts
arjunlalb: we can create a fragment with all the children in it and then add it to dom.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
arjunlalb: that works too. `replaceChildren()` is a fairly new introduction https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +157,69 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+    const searchLen: number = searchText.length;
+    if (searchLen === 0) {
+      return;
+    }
+
+    const searchedPositions: Position[][] = Array.from(Array(codeTexts.length), () => []);
+
+    // Get all searched text positions
+    codeTexts.forEach((codeText, index) => {
+      for (let i = 0; i <= codeText.length - searchLen; i++) {
```
arjunlalb, line 189: nit: with this approach of finding matching indices, since you don't need to keep searching for every index, you can clean this up by doing a while loop on `searchedIndex > 1`.

That way, you can avoid the else - break.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +157,69 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
-  public onSearch(searchText: string): void {
-    this.searchText = searchText;
+  public onSearch(searchString: string): void {
+    // For case-insensitive search
+    const searchText = searchString.toLowerCase();
+    const codeTexts = this.codeTexts.map(text => text.toLowerCase());
+
+    // Remove existing child background elements
+    this.codeLineBackgroundElements.forEach(
+      codeLineBackgroundElement => (codeLineBackgroundElement.nativeElement.innerHTML = '')
```
arjunlalb, line 177: why are we clearing individual elements in the list and not the whole list itself?

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -145,14 +157,69 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+          i = searchedIndex + searchLen - 1;
+        } else {
+          break;
+        }
+      }
+    });
+
+    // Add background elements for searched positions
+    this.codeLineBackgroundElements.forEach((codeLineBackgroundElement, index) => {
+      codeLineBackgroundElement.nativeElement.appendChild(this.getBackgroundElements(searchedPositions[index]));
```
arjunlalb, line 202: since these elements were already cleared, we can just do an append in the parent, right?

projects/components/src/viewer/code-viewer/code-viewer.component.ts
itssharmasandeep: Added in the https://github.com/hypertrace/hypertrace-ui/pull/1574

PR #1578, opened 2022-04-29
feat: fixing color selection

palbizu:
## Description
Fixing the bug that did not select the color if it did not belong to the color palette previously.

### Testing
Local testing.

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/color-picker/color-picker.component.ts
+++ b/projects/components/src/color-picker/color-picker.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import { ChangeDetectionStrategy, Component, EventEmitter, Input, OnChanges, Output } from '@angular/core';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { IconType } from '@hypertrace/assets-library';
 import { Color } from '@hypertrace/common';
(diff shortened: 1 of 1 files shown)


projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -58,9 +58,12 @@ export class ColorPickerComponent implements ControlValueAccessor {
   private propagateControlValueChange?: (value: string | undefined) => void;
   private propagateControlValueChangeOnTouch?: (value: string | undefined) => void;
 
+  public ngOnInit(): void {
```
aaron-steinfeld, line 61: OnChanges please - selected can be changed

palbizu: Yes, but also can be changed after clicking on a new color and I don't. want this logic.

aaron-steinfeld: Yes, it can be changed by clicking on a color, which calls `selectColor` (we know in this case it's already part of the palette), via input (the onChanges case) or via the programmatic writeValue api. In the first one we'd emit (already correct) and in the last two we wouldn't, but we'd need to make sure it's part of the palette. All that I believe is correct here. The issue is that rather than looking for changes whenever component input changes via OnChanges, we're only looking once via OnInit.

palbizu: clear. done

projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -58,7 +58,7 @@ export class ColorPickerComponent implements ControlValueAccessor, OnInit {
   private propagateControlValueChange?: (value: string | undefined) => void;
   private propagateControlValueChangeOnTouch?: (value: string | undefined) => void;
 
-  public ngOnInit(): void {
+  public ngOnChanges(): void {
     this.initSelectedColor();
```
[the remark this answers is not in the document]
palbizu: done

aaron-steinfeld (approved):

PR #1602, opened 2022-05-13
Handle zero case for time range query param

Christian862:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/common/src/time/time-range.service.test.ts
+++ b/projects/common/src/time/time-range.service.test.ts
@@ -97,7 +97,6 @@ describe('Time range(TR) service', () => {
   });
 
   test('Emits default TR when set, then subsequent first and second TRs from query param changes', () => {
-    const invalidTimeRange = new FixedTimeRange(new Date(1573277100277), new Date(1573277100277));
     const defaultTimeRange = new FixedTimeRange(new Date(1573277100277), new Date(1573277100999));
     const firstArrivingTimeRange = new FixedTimeRange(new Date(1573255100253), new Date(1573255111159));
     const secondArrivingTimeRange = new FixedTimeRange(new Date(1573255111159), new Date(1573455111990));
@@ -120,8 +119,7 @@ describe('Time range(TR) service', () => {
             addQueryParametersToUrl: (newParams: QueryParamObject) => mockNavigation$.next(newParams.time as string),
             getQueryParameter: jest
               .fn()
-              .mockReturnValueOnce(invalidTimeRange.toUrlString())
-              .mockReturnValueOnce(defaultTimeRange.toUrlString())
+              .mockReturnValueOnce('')
               .mockReturnValueOnce(firstArrivingTimeRange.toUrlString())
               .mockReturnValue(secondArrivingTimeRange.toUrlString()),
             getCurrentActivatedRoute: () =>
@@ -130,8 +128,6 @@ describe('Time range(TR) service', () => {
           })
         ]
       });
-
-      cold('u').subscribe(() => spectator.service.setDefaultTimeRange(invalidTimeRange));
 
       cold('x').subscribe(() => spectator.service.setDefaultTimeRange(defaultTimeRange));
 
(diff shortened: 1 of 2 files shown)


projects/common/src/time/time-range.service.ts
```diff
@@ -65,13 +65,17 @@ export class TimeRangeService {
...
+  }
+
   private getInitialTimeRange(): Observable<TimeRange> {
     return this.navigationService.navigation$.pipe(
       take(1), // Wait for first navigation
       switchMap(activatedRoute => activatedRoute.queryParamMap), // Get the params from it
       take(1), // Only the first set of params
       map(paramMap => paramMap.get(TimeRangeService.TIME_RANGE_QUERY_PARAM)), // Extract the time range value from it
-      filter((timeRangeString): timeRangeString is string => !isEmpty(timeRangeString)), // Only valid time ranges
+      filter((timeRangeString): timeRangeString is string => this.isValidTimeRangeString(timeRangeString ?? '')), // Only valid time ranges
```
aaron-steinfeld, line 78: instead of trying to build dupe logic to parse/understand the string, after you map the time range you can filter it out if the time range duration (end - start ) is not a positive int.

Christian862: done

projects/common/src/time/time-range.service.ts
```diff
@@ -127,6 +132,7 @@ export class TimeRangeService {
       return fixedTimeRange;
     }
 
+    console.log('Throwing');
```
aaron-steinfeld, line 135: test artifact (otherwise, change lgtm)

Christian862: good catch, thank you

aaron-steinfeld (dismissed):

projects/common/src/time/time-range.service.test.ts
```diff
@@ -97,7 +97,7 @@ describe('Time range(TR) service', () => {
   });
 
   test('Emits default TR when set, then subsequent first and second TRs from query param changes', () => {
-    const defaultTimeRange = new FixedTimeRange(new Date(1573277100277), new Date(1573277100277));
+    const defaultTimeRange = new FixedTimeRange(new Date(1573277100277), new Date(1573277100999));
```
aaron-steinfeld, line 100: Would actually be good to have a test case showing it skips an invalid tr

aaron-steinfeld (approved):

projects/common/src/time/time-range.service.test.ts
```diff
@@ -119,16 +120,19 @@ describe('Time range(TR) service', () => {
             addQueryParametersToUrl: (newParams: QueryParamObject) => mockNavigation$.next(newParams.time as string),
             getQueryParameter: jest
               .fn()
-              .mockReturnValueOnce('1573277100277-1573277100277')
-              .mockReturnValueOnce('1573255100253-1573255111159')
-              .mockReturnValue('1573255111159-1573455111990'),
+              .mockReturnValueOnce(invalidTimeRange.toUrlString())
+              .mockReturnValueOnce(defaultTimeRange.toUrlString())
```
aaron-steinfeld, line 124: this one we wouldn't want to come from the query param, it's the default time range (not sure why it did before). Similarly line 134 shouldn't be set as the default. It would probably be more straight forward to split this into a different test, see that the default gets set when the url returns invalid.

projects/common/src/time/time-range.service.test.ts
```diff
@@ -119,9 +119,9 @@ describe('Time range(TR) service', () => {
             addQueryParametersToUrl: (newParams: QueryParamObject) => mockNavigation$.next(newParams.time as string),
             getQueryParameter: jest
               .fn()
-              .mockReturnValueOnce('1573277100277-1573277100277')
-              .mockReturnValueOnce('1573255100253-1573255111159')
-              .mockReturnValue('1573255111159-1573455111990'),
+              .mockReturnValueOnce('')
```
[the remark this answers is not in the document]
Christian862: i thought so but no, after checking. Fixing

projects/common/src/time/time-range.service.test.ts
```diff
@@ -147,6 +147,44 @@ describe('Time range(TR) service', () => {
...
+            addQueryParametersToUrl: (newParams: QueryParamObject) => mockNavigation$.next(newParams.time as string),
+            getQueryParameter: jest.fn().mockReturnValueOnce('').mockReturnValueOnce(defaultTimeRange.toUrlString()),
+            getCurrentActivatedRoute: () =>
+              (({ snapshot: { queryParams: { time: 'test-value' } } } as unknown) as ActivatedRoute),
+            replaceQueryParametersInUrl: jest.fn()
+          })
+        ]
+      });
+
+      cold('u').subscribe(() => mockNavigation$.next(invalidTimeRange.toUrlString()));
```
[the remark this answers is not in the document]
Christian862: For the timing of the observables i thought, but i guess since we're only expecting a single emission we shouldn't need to use it ? i'll try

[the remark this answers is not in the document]
aaron-steinfeld: in the earlier one it was used for delaying calls and seeing how late arriving things show up in the time range observable, but here we don't need that, since we're just checking the default

PR #1617, opened 2022-05-18
feat: file upload component

itssharmasandeep:
## Description
`Drag and drop with progress bar (delete button disabled when uploading in progress)`

https://user-images.githubusercontent.com/78212931/169384017-e00244cc-c5f4-4a78-84e2-2d4247c362c3.mov

`Click to upload with 2 different files and delete functionality`

https://user-images.githubusercontent.com/78212931/169385207-b8ac0ea6-519e-4b0b-b29b-da9b5f2adbaf.mov

`Files with no progress bar`

Screenshot 2022-05-20 at 12 28 15 AM

### Testing

### Checklist:


projects/components/src/file-display/file-display.component.ts
```diff
@@ -0,0 +1,55 @@
...
+        size="${IconSize.Large}"
+        [color]="!this.isDeleteDisabled ? '${Color.Gray4}' : '${Color.Gray2}'"
+        (click)="this.onDeleteClick()"
+      ></ht-icon>
+    </div>
+  `
+})
+export class FileDisplayComponent implements OnChanges {
+  @Input()
+  public file?: FileItem;
```
anandtiwary, line 35: Should we flatten the FileItem as two inputs on this component? That would allow us to trigger a CD whenever progress changes. Otherwise, we will have to create new fileItem objects.

projects/components/src/file-upload/drop-zone/drop-zone.directive.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@Directive({
+  selector: '[htDropZone]'
+})
+export class DropZoneDirective {
+  @Output()
+  public readonly dropped: EventEmitter<FileList> = new EventEmitter();
+
+  @Output()
+  public readonly dragOver: EventEmitter<boolean> = new EventEmitter();
```
anandtiwary, line 11: When would we use dragOver output (and it's true and false value)

projects/components/src/file-upload/drop-zone/drop-zone.directive.ts
```diff
@@ -0,0 +1,31 @@
...
+  @HostListener('dragover', ['$event'])
+  public onDragover(event: DragEvent): void {
+    event.preventDefault();
+    this.dragOver.emit(true);
+  }
+
+  @HostListener('dragleave', ['$event'])
+  public onDragleave(event: DragEvent): void {
+    event.preventDefault();
+    this.dragOver.emit(false);
```
anandtiwary, line 29: Probably, we should add both dragOver and dragLeave as a separate output. It is not clear that dragOver: false is same as dragLeave

projects/components/src/file-upload/file-upload-validators.ts
```diff
@@ -0,0 +1,133 @@
...
+import { UploadFileType } from './upload-file-type';
+
+/**
+ * @description
+ * Provides a set of validators specifically for file upload component that can be used by form controls.
+ */
+// tslint:disable:no-null-keyword
+// tslint:disable:no-unnecessary-class
+export class FileUploadValidators {
+  /**
```
anandtiwary, line 13: Lots of comments!!!

projects/components/src/file-upload/file-upload-validators.ts
```diff
@@ -0,0 +1,133 @@
...
+import { FileItem } from '../file-display/file-item';
+import { UploadFileType } from './upload-file-type';
+
+/**
+ * @description
+ * Provides a set of validators specifically for file upload component that can be used by form controls.
+ */
+// tslint:disable:no-null-keyword
+// tslint:disable:no-unnecessary-class
+export class FileUploadValidators {
```
anandtiwary, line 12: Do we need to make this a class? We can always export it as const methods no?

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,133 @@
...
+      </div>
+    </div>
+  `
+})
+export class FileUploadComponent implements ControlValueAccessor {
+  @Input()
+  public files: FileItem[] = []; // Can be used to update progress
+
+  @Input()
+  public subText: string = 'max size 2 GB';
```
anandtiwary, line 58: Let's keep the default to undefined or an empty string

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  @Input()
+  public disabled: boolean = false;
+
+  @Output()
+  private readonly fileUpload: EventEmitter<FileItem[]> = new EventEmitter();
```
anandtiwary, line 64: this should be public readonly

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  public isDragOver: boolean = false;
+
+  public constructor(private readonly cdr: ChangeDetectorRef) {}
+
+  public writeValue(value?: FileItem[]): void {
+    this.files.splice(0).push(...(value ?? []));
```
anandtiwary, line 71: We shouldn't be modifying input. I would recommend maintaining a separate internal property for this

anandtiwary: Actually, should `Files` be an input for this component? Users are adding Files through this component right?

projects/components/src/file-upload/upload-file-type.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @description
+ * Current set of upload file types
+ *
+ * @usageNotes
+ * This is used to provide supported file types as array into `FileUploadValidators.supportedFileTypes(<>)`
+ *
+ */
+export const enum UploadFileType {
```
anandtiwary, line 9: We should keep this with the validator. Imo, file upload component should not put this restriction in unless we bake in the validation into this component.

projects/components/src/public-api.ts
```diff
@@ -102,6 +102,18 @@ export * from './feature-check/feature-config-check.module';
 export * from './feature-check/feature.pipe';
 export * from './feature-check/if-feature.directive';
 
+// File Display
+export * from './file-display/file-item';
+export * from './file-display/file-display.component';
+export * from './file-display/file-display.module';
+
+// File Upload
+export * from './file-upload/drop-zone/drop-zone.directive';
```
anandtiwary, line 111: We don't need to export `file-display.component`, `file-display.module` and `drop-zone.directive` because they are internal details.

setup-jest.ts
```diff
@@ -1 +1,5 @@
 import 'jest-preset-angular/setup-jest';
+
+Object.defineProperty(window, 'DragEvent', {
```
anandtiwary, line 3: Is dragEvent not supported in JsDom?  Can we try putting this code in the corresponding test file? That way, it will stay isolated. I kind of did something similar with ResizeObservers

anandtiwary (commented):
Looks really good!

projects/components/src/file-upload/drop-zone/drop-zone.directive.ts
itssharmasandeep: This is being used to show the `blue dashed` border while dropping the files.

projects/components/src/file-upload/file-upload-validators.ts
itssharmasandeep: Yeah, basically I have used the same approach as angular to add new custom validators. These comments will be seen in VSCode hover for these functions.

projects/components/src/file-upload/file-upload-validators.ts
itssharmasandeep: I mean yeah, I was also confused here, just wanted to use the same approach which angular does for Validators

projects/components/src/file-display/file-display.component.ts
```diff
@@ -0,0 +1,72 @@
...
+      ></ht-icon>
+    </div>
+  `
+})
+export class FileDisplayComponent {
+  @Input()
+  public file?: File;
+
+  @Input()
+  public state: FileUploadState = FileUploadState.NotStarted;
```
anandtiwary, line 53: Can this be done via our form field?

projects/components/src/file-display/file-display.module.ts
```diff
@@ -0,0 +1,14 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+import { FormattingModule } from '@hypertrace/common';
+import { IconModule } from '../icon/icon.module';
+import { ProgressBarModule } from '../progress-bar/progress-bar.module';
+import { TooltipModule } from '../tooltip/tooltip.module';
+import { FileDisplayComponent } from './file-display.component';
+
+@NgModule({
+  imports: [CommonModule, IconModule, ProgressBarModule, FormattingModule, TooltipModule],
```
anandtiwary, line 10: ProgressBarModule doesn't seem to be used. Please remove the ones not needed.

projects/components/src/file-upload/drop-zone/drop-zone.directive.ts
anandtiwary: ping. Also, isn't `DragLeave` the same as dropped? Do we need both?

projects/components/src/file-upload/drop-zone/drop-zone.directive.ts
anandtiwary: see below

projects/components/src/file-upload/file-upload-validators.ts
anandtiwary: I think we should just make these exported validator methods. That way we won't have to disable the tslint rule

projects/components/src/file-upload/file-upload-validators.ts
```diff
@@ -0,0 +1,136 @@
...
+   *
+   * console.log(control.errors); // {maxFileSize: true}
+   * ```
+   *
+   * @returns A validator function that returns an error map with the
+   * `maxFileSize` property if the validation check fails, otherwise `null`.
+   *
+   *
+   */
+  public static maxFileSize = (maxSize: number) => (control: AbstractControl) =>
```
anandtiwary, line 71: We can append `Validator` in the name

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,142 @@
...
+  public showProgress: boolean = true; // To show bulk upload progress
+
+  @Input()
+  public progress: number = 0; // Bulk upload progress
+
+  @Input()
+  public uploadState: FileUploadState = FileUploadState.NotStarted;
+
+  @Output()
+  public readonly fileUpload: EventEmitter<File[]> = new EventEmitter();
```
anandtiwary, line 72: nit: may be `selectedFileChanges`?

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,142 @@
...
+    this.files.push(...this.getFilesFromFileList(files));
+    this.fileUpload.emit(this.files);
+    this.propagateValueChangeToFormControl(this.files);
+  }
+
+  /**
+   * Converts the FileList into File[]
+   */
+  private getFilesFromFileList(files?: FileList): File[] {
+    return !isNil(files) ? Array.from(files) : [];
```
anandtiwary, line 140: nit: `[...(files ?? [])]` I think this might also work

projects/components/src/file-display/file-display.component.ts
itssharmasandeep: This is not needed now, since it is not an input anymore

projects/components/src/file-upload/drop-zone/drop-zone.directive.ts
itssharmasandeep: So `mouseover` and `mouseleave` are combined as `hover`. So I can combine it as `dragHover` output.
No `DragLeave` is different than `dropped`, in drop we're dropping the files but in the drag leave we're moving away without dropping.

projects/components/src/file-upload/file-upload.component.ts
itssharmasandeep: This is not an input property anymore.

projects/components/src/file-upload/upload-file-type.ts
itssharmasandeep: done

projects/components/src/public-api.ts
itssharmasandeep: Done

projects/components/src/file-upload/file-upload-validators.ts
itssharmasandeep: Done

projects/components/src/file-display/file-display.module.ts
itssharmasandeep: Done

projects/components/src/file-upload/file-upload-validators.ts
itssharmasandeep: done

projects/components/src/file-upload/file-upload.component.ts
itssharmasandeep: No, this is not working!

projects/components/src/file-upload/file-upload.component.ts
itssharmasandeep: done

projects/components/src/file-display/file-display.component.ts
itssharmasandeep: I am not sure, but I would keep it here itself as this is being used to enable/disable delete buttons as well.

setup-jest.ts
itssharmasandeep: I tried but it didn't work. For `ResizeObserver` as well, we put in the `setup-jest.ts` file itself, right?

itssharmasandeep:
> Not urgent, but with a future pr please revert the name to dragOver and see if you can separate dragOver and dragLeave.

Sure :)

PR #1620, opened 2022-05-19
fix: extract zero filling function

Greedy-Geek:


projects/observability/src/shared/graphql/model/metric/metric-timeseries.ts
```diff
@@ -8,3 +11,33 @@ export interface MetricTimeseriesBandInterval {
   upperBound: number;
   lowerBound: number;
 }
+
+export const getZeroFilledMetricTimeseriesIntervals = (
```
arjunlalb, line 15: why do we need this as a const method?

The explore result class has a method `getTimeSeriesData`. that is already giving you access to the zero filled data.

PR #1623, opened 2022-05-19
feat: Adding disable property to ColorPicker

palbizu:
## Description
Disabling ColorPicker

image

### Testing
Locally tested

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/color-picker/color-picker.component.ts
+++ b/projects/components/src/color-picker/color-picker.component.ts
@@ -47,7 +47,7 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   public selected?: string;
 
   @Input()
-  public disabled: boolean = false;
+  public disabled?: boolean = false;
 
   @Output()
   private readonly selectedChange: EventEmitter<string> = new EventEmitter<string>();
--- a/projects/components/src/color-picker/color-picker.component.ts
+++ b/projects/components/src/color-picker/color-picker.component.ts
@@ -20,13 +20,18 @@ import { IconSize } from '../icon/icon-size';
       <div
         class="color"
         *ngFor="let color of this.paletteColors"
-        [ngClass]="{ selected: color === this.selected }"
+        [ngClass]="{ selected: color === this.selected, disabled: this.disabled }"
         [style.backgroundColor]="color"
         (click)="this.selectColor(color)"
       ></div>
-      <ht-popover>
+      <ht-popover [disabled]="this.disabled">
         <ht-popover-trigger>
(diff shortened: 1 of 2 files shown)


projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -94,6 +106,10 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
     this.propagateControlValueChangeOnTouch = onTouch;
   }
 
+  public setDisabledState(isDisabled?: boolean): void {
```
jake-bassett, line 109: This is fine, but could also do `public setDisabledState(isDisabled: boolean = false): void {`

And remove the `??` check below.

projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -41,6 +46,9 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   @Input()
   public selected?: string;
 
+  @Input()
+  public disabled: boolean = false;
```
jake-bassett, line 50: There is alway a risk of the user passing in a `null` or `undefined` here, so it's safest to type this as optional `?`.

jake-bassett (dismissed):

projects/components/src/color-picker/color-picker.component.ts
aaron-steinfeld: careful - if these are called by angular and it sends in nulls, the current code would assign false and the suggestion would assign null. May not be relevant, but something to keep in mind when working in code that interfaces with angular.

palbizu: I prefer to keep this change to keep the consistency with the rest of CVA components.

projects/components/src/color-picker/color-picker.component.ts
palbizu: done

projects/components/src/color-picker/color-picker.component.ts
jake-bassett: Is there a concern for angular or other code we don't own calling this method? If so, my suggestion above is safer.

aaron-steinfeld: this method is from the CVA interface, angular forms calls it. I'm not sure why we default it at all thoguh, that's not part of angular's contract:
https://angular.io/api/forms/ControlValueAccessor#setdisabledstate

palbizu: I took a look at the documentation and it's no concern. I just tried to keep the consistency with the rest of the components but already changed.

jake-bassett (approved):

PR #1628, opened 2022-05-23
feat: adding style and fixing data in Form Control

palbizu:
## Description
Adding new style and fixing data in NumberInputComponent

### Testing
Locally tested

### Checklist:

### Documentation


projects/components/src/number-input/number-input.component.scss
```diff
@@ -23,6 +23,14 @@
   border-bottom: 1px solid $gray-2;
 }
 
+.no-border {
```
anandtiwary, line 26: It should be no border by default (i think)

projects/components/src/number-input/number-input.component.ts
```diff
@@ -67,7 +67,7 @@ export class NumberInputComponent implements ControlValueAccessor {
     this.value = value;
 
     this.enforceMinMaxAndEmit();
-    this.propagateValueChangeToFormControl(value);
+    this.propagateValueChangeToFormControl(this.value);
```
anandtiwary, line 70: what difference does it make?

palbizu: local value could be update it in 'EnforceMinMaxAndEmit' for this reason is not fine to propagate the value without this specific change.

projects/components/src/number-input/number-input.component.scss
palbizu: Yep, I think the same but there was used previously in many places so for this reason, a created a new style

anandtiwary: i mean, if we don't pass any style then would it not hide the borders on its own?

projects/components/src/number-input/number-input.component.ts
anandtiwary: Hmm.. might want to rearrange the code here to make it more obvious. Also, please put a comment here explaining this requirement

anandtiwary (approved):
approved but please check the comments

projects/components/src/number-input/number-input.component.scss
palbizu: There are two previous scenarios 
1 - Border: sets all borders
2- Underline: sets just the border-bottom

It's necessary to add a new Appearance to have no border input because Border is set by default currently. We could remove the default border but could affect multiple places through the app.

projects/components/src/number-input/number-input.component.ts
palbizu: @anandtiwary I refactored the component to keep the logic clearer. Please take a look and let me know if you agree with this.

jake-bassett (approved):

PR #1643, opened 2022-05-30
fix: showing total in cartesian chart tooltips

Greedy-Geek:


aaron-steinfeld (commented):
Total is not relevant to all charts. Instead, make this a new series and don't modify the common component

Greedy-Geek:
> Total is not relevant to all charts. Instead, make this a new series and don't modify the common component

Got it, will create a new series

PR #1652, opened 2022-06-03
fix(sheet overlay): break-word for header

arjunlalb:
## Description
If the header text is too long, the sheet overlay component header becomes messed up. Fixing it by breaking the header text for really long text.

### Testing
Screenshot 2022-06-03 at 10 48 37 AM

### Checklist:


projects/components/src/overlay/sheet/sheet-overlay.component.scss
```diff
@@ -19,12 +19,12 @@ $box-shadow: -3px 0px 24px rgba(0, 0, 0, 0.12), -1px 0px 8px rgba(0, 0, 0, 0.08)
     display: flex;
     justify-content: space-between;
     align-items: center;
-    flex-wrap: wrap-reverse;
```
anandtiwary, line 22: why are we removing this?

PR #1682, opened 2022-06-13
fix: modified metric-card component to have multiple layouts

Greedy-Geek:


projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -42,14 +57,39 @@ export class MetricCardComponent {
...
   @Input()
-  public icon?: string; // Only be used when indicator type is Icon
+  public icon?: string; // Only be used when indicator type is Icon in CardWithIndicator layout
 
   @Input()
   public indicatorColor: Color = Color.Gray7;
 
   @Input()
-  public backgroundColor: Color = Color.OffWhite;
+  public backgroundColor?: Color;
```
arjunlalb, line 66: restore the defaults.

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -42,14 +57,39 @@ export class MetricCardComponent {
...
+
+  @Input()
+  public borderColor?: Color;
```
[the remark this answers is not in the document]
Greedy-Geek: restored

projects/components/src/metric-card/metric-card.component.ts
Greedy-Geek: restored

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -1,33 +1,51 @@
...
   template: `
-    <div class="metric-card" [style.backgroundColor]="this.backgroundColor" [style.borderColor]="this.borderColor">
-      <div class="indicator-and-title">
-        <div class="indicator">
-          <ng-container [ngSwitch]="this.indicator">
-            <ng-container *ngSwitchCase="'${MetricCardIndicatorType.Dot}'">
-              <div class="dot" [style.backgroundColor]="this.indicatorColor"></div>
+    <div
+      class="metric-card"
+      [ngClass]="{ selected: this.isSelected, clickable: this.clickable }"
```
arjunlalb, line 13: `selected` class doesn't seem to be used anywhere.

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -1,33 +1,51 @@
...
+      [ngStyle]="{
+        'background-color': this.isSelected ? '${Color.Blue1}' : this.backgroundColor,
+        'border-color': this.isSelected ? '${Color.Blue4}' : this.borderColor
+      }"
```
arjunlalb, line 17: define a method

projects/components/src/metric-card/metric-card.component.ts
Greedy-Geek: oh yes, missed to remove it after adding condition in ngStyle. Removed

projects/components/src/metric-card/metric-card.component.ts
Greedy-Geek: updated

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -52,4 +67,35 @@ export class MetricCardComponent {
...
+  public onSelectionChange(): void {
+    if (!this.clickable) {
+      return;
+    }
+    this.selectedValue.emit(this.selectionValue);
+  }
+
+  public getStylesForCard(): { 'background-color': string; 'border-color': string } {
+    return this.isSelected
+      ? { 'background-color': Color.Blue1, 'border-color': Color.Blue4 }
```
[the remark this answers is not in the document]
Greedy-Geek: we can't beacuse the HTML template will override the CSS and we won't be able to see the style for selection if done via CSS.

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -52,4 +67,35 @@ export class MetricCardComponent {
...
+  public layoutType: MetricCardLayout = MetricCardLayout.CardWithIndicator;
+
+  @Input()
+  public iconColor: Color = Color.Gray9;
+
+  @Input()
+  public isSelected: boolean = false;
+
+  @Input()
+  public clickable: boolean = false;
```
anandtiwary, line 81: we probably don't need clickable property. Parent component can decide to put `(click)` handler directly over this component host

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -1,33 +1,48 @@
...
-          <ng-container [ngSwitch]="this.indicator">
-            <ng-container *ngSwitchCase="'${MetricCardIndicatorType.Dot}'">
-              <div class="dot" [style.backgroundColor]="this.indicatorColor"></div>
+    <div
+      class="metric-card"
+      [ngClass]="{ clickable: this.clickable }"
+      [ngStyle]="this.getStylesForCard()"
+      (click)="this.onSelectionChange()"
+    >
+      <div class="card-with-indicator" *ngIf="this.layoutType === '${MetricCardLayout.CardWithIndicator}'">
```
aaron-steinfeld, line 17: If you find yourself ngIf'ing with two different templates and the two modes have different inputs, they shouldn't be the same component. Create a different component for the use case.

Greedy-Geek: we have different layouts for almost the same set of inputs, so won't it be better to keep these layouts in the same component instead of creating multiple small components with same inputs?

projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -52,4 +67,35 @@ export class MetricCardComponent {
...
+  public iconColor: Color = Color.Gray9;
+
+  @Input()
+  public isSelected: boolean = false;
+
+  @Input()
+  public clickable: boolean = false;
+
+  @Input()
+  public selectionValue?: string;
```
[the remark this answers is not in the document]
Greedy-Geek: oh yes, we don't need it here, removed

projects/components/src/metric-card/metric-card.component.ts
Greedy-Geek: yeah, removed

projects/components/src/metric-card/metric-card.component.ts
aaron-steinfeld: Even if they're identical, what do we save by having two different implementations inside the same component? The boilerplate of declaring a new component? It's the same reason that putting a flag argument on a method with otherwise identical inputs to affect its behavior is an antipattern. If it's a small styling difference and the majority of the template is reused that's one thing (e.g. primary vs secondary button), but if it's a completely different template that just makes each change that much harder - you have to track each input and output to two different places in code.

Anyway it's up to you guys, just my 2 cents.

aaron-steinfeld: Actually it's even more problematic, because on top of the layout type there's already an indicator type, and now the two don't play nice. The new layout type ignores the indicator type, and it uses inputs differently. If I'm using an icon indicator, it gets its colors from indicatorColor; if I'm using cardWithLargeIcon, it gets its color from iconColor.

So at a minimum, please make your change by extending the existing inputs so everything behaves consistently.

Greedy-Geek: Creating a new component for it

PR #1683, opened 2022-06-13
fix(ht-gauge-list): added optional input for total

23nobody:
## Description
Currently, `ht-gauge-list` sets percentage and bar values by default using calculated values. Added support to provide `total` to be used for these calculations.

### Testing
Added test for the two cases of percentages.

### Checklist:

### Documentation


projects/observability/src/shared/components/gauge-list/gauge-list.component.ts
```diff
@@ -82,7 +88,7 @@ export class GaugeListComponent<T extends GaugeItem = GaugeItem> implements OnCh
     this.itemOptions = this.items.map(option => ({
       label: option.label,
       color: colorLookupMap.get(option.colorKey ?? option.label),
-      width: `${((option.value / maxValue!) * 100).toFixed(2)}%`,
+      width: `${((option.value / this.total!) * 100).toFixed(2)}%`,
```
23nobody, line 91: Now that we have total being supplied from outside, that total needs to be used for percentage calculation in L93 below.
Thoughts?

projects/observability/src/shared/components/gauge-list/gauge-list.component.ts
```diff
@@ -71,22 +74,35 @@ export class GaugeListComponent<T extends GaugeItem = GaugeItem> implements OnCh
...
-    }));
+
+    if (this.total !== undefined && this.total !== 0) {
+      this.itemOptions = this.items.map(option => ({
+        label: option.label,
+        color: colorLookupMap.get(option.colorKey ?? option.label),
+        width: `${((option.value / this.total!) * 100).toFixed(2)}%`,
+        value: option.value,
+        percentage: (option.value / this.total!) * 100,
+        original: option
```
itssharmasandeep, line 86: This looks duplicate, same as L97 to L103

itssharmasandeep: can we create a function for those and pass the values as arguments

arjunlalb:
Could you please describe what the PR is intending to do?

projects/observability/src/shared/components/gauge-list/gauge-list.component.ts
23nobody: Updated. Thanks!

arjunlalb:
Why do we want to use total here?

When there are more than 2 or 3 items, the visualization of the bar doesn't add much value to the user if the fill portion is calculated as a percentage of the total.

Ex:
If there are 5 items to show, each with values 21, 22, 23, 20, 14, all the visual bars would look sort of similar since they are all plotted against the series total of 100.
Also, if the data set contains more than 5 items and we show only 5 in the guage list, the total will be even bigger and will capture no useful information.

We should re-evaluate this.

23nobody:
> Why do we want to use total here?
> 
> When there are more than 2 or 3 items, the visualization of the bar doesn't add much value to the user if the fill portion is calculated as a percentage of the total.
> 
> Ex: If there are 5 items to show, each with values 21, 22, 23, 20, 14, all the visual bars would look sort of similar since they are all plotted against the series total of 100. Also, if the data set contains more than 5 items and we show only 5 in the guage list, the total will be even bigger and will capture no useful information.
> 
> We should re-evaluate this.

Doesn't that give wrong impression of the distribution, that current 5 items that are in the list contribute in larger extent to the data. 
Showing some items to have more share than the actual is also saying that  the `rest` items comprise lesser shares in the distribution.

aaron-steinfeld:
> > Why do we want to use total here?
> > When there are more than 2 or 3 items, the visualization of the bar doesn't add much value to the user if the fill portion is calculated as a percentage of the total.
> > Ex: If there are 5 items to show, each with values 21, 22, 23, 20, 14, all the visual bars would look sort of similar since they are all plotted against the series total of 100. Also, if the data set contains more than 5 items and we show only 5 in the guage list, the total will be even bigger and will capture no useful information.
> > We should re-evaluate this.
> 
> Doesn't that give wrong impression of the distribution, that current 5 items that are in the list contribute in larger extent to the data. Showing some items to have more share than the actual is also saying that the `rest` items comprise lesser shares in the distribution.

So it sounds like there's two cases 
- Showing distribution of one total - how much does each element contribute to it (e.g. calls to API A as a percentage of all API calls) 
- Showing relative size of related values - how large is each element relative to the largest

We used to exclusively use this for the second case, because the former has much better representations (like a donut, or a single gauge with multiple series which I think we also built). 

If we're trying to extend it to the first case, I would agree we should reconsider the use case and if this is the most appropriate visualization.

projects/observability/src/shared/components/gauge-list/gauge-list.component.ts
```diff
@@ -71,18 +74,28 @@ export class GaugeListComponent<T extends GaugeItem = GaugeItem> implements OnCh
       return;
     }
 
-    let maxValue = maxBy(this.items, option => option.value)?.value;
-    if (maxValue === undefined || maxValue === 0) {
-      maxValue = 1;
+    if (this.total !== undefined && this.total !== 0) {
+      this.setItemOptions(this.items, this.total, this.total);
```
aaron-steinfeld, line 78: There's two different values here - totalCount looks like it got added later (not sure I agree with that from a visualization user case perspective, but a bit besides the point), and it represents that total that would be passed in here. FullCount is the visual indicator, representing the max value.

So I think there's a decent argument to be made to fix those percentages by passing the accurate total from the parent, but the bars should still be based on max value. Once we change that, we're really changing the whole purpose of the visualization.

23nobody:
Wasn't the correct choice for visualization of the case.

PR #1689, opened 2022-06-14
feat: metric card v2 component

Greedy-Geek:
## Description
Added a new metric card

### Checklist:
- My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works


projects/components/src/metric-card-v2/metric-card-v2.component.ts
```diff
@@ -0,0 +1,37 @@
...
+    <div class="metric-card" [ngClass]="{ selected: this.isSelected }">
+      <div class="value-with-icon">
+        <span class="value">{{ this.value }}</span>
+        <ht-icon class="icon" [icon]="this.icon" [size]="this.iconSize" [color]="this.iconColor"></ht-icon>
+      </div>
+      <ht-label class="text" [label]="this.text"></ht-label>
+    </div>
+  `
+})
+export class MetricCardV2Component {
```
arjunlalb, line 19: clickable or not should also be an input so that hover styling can be applied.

projects/components/src/metric-card-v2/metric-card-v2.component.ts
```diff
@@ -0,0 +1,40 @@
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { Color } from '@hypertrace/common';
+import { IconSize } from '../icon/icon-size';
+
+@Component({
+  selector: 'ht-metric-card-v2',
```
aaron-steinfeld, line 6: Does this actually replace the other metric card? Continuing the earlier conversation, 
- A separate component makes sense if we want to maintain two different ways to display a metric card AND they require different inputs - but in that case, they will both coexist so they shouldn't be versioned and we need to rethink this component's name.
- If we want to maintain two different ways to display a metric card, but we can align their inputs, then updating the existing component makes sense by extending the existing enum.
- If we only plan on maintaining one way to display a metric, we should be updating the existing component (without a new enum, since we want everyone consistent)

Greedy-Geek: We need to display almost the same inputs in quite different ways (major distinction that can be pointed out is the indicator and icon), i.e the separate templates will be required. 
I wasn't able to come up with a good name, so I named it metric card v2

PR #1692, opened 2022-06-14
feat: sheet title as template

itssharmasandeep:
## Description
Sheet title as a template

### Testing
Local testing is done,

### Checklist:

### DIFF
--- a/projects/components/src/overlay/sheet/sheet-overlay.component.ts
+++ b/projects/components/src/overlay/sheet/sheet-overlay.component.ts
@@ -18,7 +18,9 @@ import { SheetOverlayConfig, SheetSize } from './sheet';
       <div class="sheet-overlay" [style.padding.px]="this.hasDefaultPadding ? 24 : 0">
         <ng-container *ngIf="!this.isViewCollapsed">
           <div *ngIf="this.showHeader" class="header">
-            <h3 class="header-title">{{ sheetTitle }}</h3>
+            <ng-container *ngIf="!this.isSheetTitleAString; else defaultSheetTitleTemplate">
+              <ng-container *ngTemplateOutlet="this.sheetTitle"></ng-container>
+            </ng-container>
             <div class="action-buttons">
               <ht-open-in-new-tab
                 *ngIf="this.navigationParams"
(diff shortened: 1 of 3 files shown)


projects/components/src/overlay/sheet/sheet-overlay.component.ts
```diff
@@ -33,6 +35,9 @@ import { SheetOverlayConfig, SheetSize } from './sheet';
               >
               </ht-button>
             </div>
+            <ng-template #defaultSheetTitleTemplate>
+              <h3 class="default-header-title">{{ sheetTitle }}</h3>
```
arjunlalb, line 39:
```suggestion
              <h3 class="default-header-title">{{ this.sheetTitle }}</h3>
```

arjunlalb, line 39:
```suggestion
              <h3 class="default-header-title">{{ this.sheetTitle }}</h3>
```

projects/components/src/overlay/sheet/sheet-overlay.component.ts
```diff
@@ -81,6 +87,7 @@ export class SheetOverlayComponent {
     this.hasDefaultPadding = sheetConfig.hasDefaultPadding ?? true;
     this.showHeader = sheetConfig.showHeader === true;
     this.sheetTitle = sheetConfig.title === undefined ? '' : sheetConfig.title;
+    this.isSheetTitleAString = typeof this.sheetTitle === 'string';
```
arjunlalb, line 90: would be cleaner to check the inverse with `instanceof TemplateRef`

itssharmasandeep: Not sure, I did the same thing at many places, so continued with the same :)

arjunlalb (approved):

projects/components/src/overlay/sheet/sheet-overlay.component.ts
```diff
@@ -58,7 +63,7 @@ import { SheetOverlayConfig, SheetSize } from './sheet';
   `
 })
 export class SheetOverlayComponent {
-  public readonly sheetTitle: string;
+  public readonly sheetTitle: string | TemplateRef<unknown>;
```
anandtiwary, line 66: Title is supposed to be text only. If we allow it to be a template then technically we can pass any custom content here. I would propose that we keep title as it is and introduce a new property `customHeaderTemplate` that will be rendered between the title text and the header controls

projects/components/src/overlay/sheet/sheet-overlay.component.ts
anandtiwary: we can avoid this :)

PR #1695, opened 2022-06-16
feat: export drilldown filter type

arjunlalb:
## Description
Export a type.

### DIFF
--- a/projects/observability/src/pages/explorer/explorer-service.ts
+++ b/projects/observability/src/pages/explorer/explorer-service.ts
@@ -17,7 +17,7 @@ export class ExplorerService {
   ) {}
   public buildNavParamsWithFilters(
     scopeQueryParam: ScopeQueryParam,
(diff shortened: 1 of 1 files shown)


Christian862 (approved):

anandtiwary (changes requested):
Why do we need to export it? I would prefer if we can avoid it as it will just create more and more duplicate types.

projects/observability/src/pages/explorer/explorer-service.ts
```diff
@@ -49,4 +49,4 @@ export class ExplorerService {
   }
 }
 
-type DrilldownFilter = Omit<Filter, 'metadata' | 'userString' | 'urlString'>;
+export type ExplorerDrilldownFilter = Omit<Filter, 'metadata' | 'userString' | 'urlString'>;
```
anandtiwary, line 52: can we use `FieldFilter` instead?

PR #1705, opened 2022-06-21
feat/regex validator

Greedy-Geek:


projects/common/src/utilities/validators/regex-validator.test.ts
```diff
@@ -0,0 +1,14 @@
+import { isValidRegex } from './regex-validator';
+
+const VALID_REGEX = ['2+2', '(2+2) * 3*9'];
+const INVALID_REGEX = ['\\', '['];
```
arjunlalb, line 4: why are these outside the test description? please move them inside.

projects/common/src/utilities/validators/regex-validator.test.ts
```diff
@@ -0,0 +1,14 @@
...
+describe('Regex validator', () => {
+  test('can validate regex lists correctly', () => {
+    VALID_REGEX.forEach(regex => {
+      expect(isValidRegex(regex)).toBeTruthy();
+    });
+    INVALID_REGEX.forEach(regex => {
+      expect(isValidRegex(regex)).toBeFalsy();
+    });
```
arjunlalb, line 12: these are two tests. 

1 set for valid and another for invalid. Split them into two tests.

projects/common/src/utilities/validators/regex-validator.test.ts
```diff
@@ -0,0 +1,14 @@
+import { isValidRegex } from './regex-validator';
+
+const VALID_REGEX = ['2+2', '(2+2) * 3*9'];
+const INVALID_REGEX = ['\\', '['];
+describe('Regex validator', () => {
+  test('can validate regex lists correctly', () => {
+    VALID_REGEX.forEach(regex => {
+      expect(isValidRegex(regex)).toBeTruthy();
```
arjunlalb, line 8: no `toBeTruthy`. check for false explicity.

projects/common/src/utilities/validators/regex-validator.ts
```diff
@@ -0,0 +1,10 @@
+export const isValidRegex = (regex: string) => {
```
arjunlalb, line 1: add appropriate comments at the top please.

projects/common/src/utilities/validators/regex-validator.test.ts
Greedy-Geek: image
Followed this convention so that's why. Updating

projects/common/src/utilities/validators/regex-validator.test.ts
Greedy-Geek: updated

projects/common/src/utilities/validators/regex-validator.ts
Greedy-Geek: updated

arjunlalb (approved):

PR #1706, opened 2022-06-22
style(table): set min height for row

adisreyaj:
## Description

Add `min-height` to table alongside `height`.


adisreyaj:
Closing in favour of https://github.com/hypertrace/hypertrace-ui/pull/1711

PR #1719, opened 2022-06-27
FF resolver support for receiving dictionary map

Christian862:
## Description

### Testing

### Checklist:

### Documentation


projects/common/src/feature/state/feature-state.resolver.ts
```diff
@@ -24,3 +34,10 @@ export abstract class FeatureStateResolver {
     return FeatureState.Enabled;
   }
 }
+
+interface FeaturePageTimeRangeMap {
+  [path: string]: string;
+}
+
+export type GraphQlFeatureFlagValue = boolean | string | number | FeaturePageTimeRangeMap;
```
aaron-steinfeld, line 42:
```suggestion
export type FeatureFlagValue = boolean | string | number | Dictionary<string>;
```

projects/common/src/time/page-time-range-preference.service.ts
```diff
@@ -32,10 +31,10 @@ export class PageTimeRangePreferenceService {
   public getTimeRangePreferenceForPage(rootLevelPath: string): Observable<TimeRangeResolver> {
     return combineLatest([
       this.pageTimeRangeStringDictionary$,
-      this.featureStateResolver.getFeatureState(ApplicationFeature.PageTimeRange)
+      this.featureStateResolver.getFeatureFlagValue<boolean>(ApplicationFeature.PageTimeRange)
```
aaron-steinfeld, line 34: this should still be calling `getFeatureState`, it doesn't need the raw value

projects/components/src/header/application/application-header.component.ts
```diff
@@ -48,8 +47,8 @@ export class ApplicationHeaderComponent {
     private readonly featureStateResolver: FeatureStateResolver
   ) {
     this.pageLevelTimeRangeDisabled$ = this.featureStateResolver
-      .getFeatureState(ApplicationFeature.PageTimeRange)
-      .pipe(map(featureState => featureState === FeatureState.Disabled));
+      .getFeatureFlagValue<boolean>(ApplicationFeature.PageTimeRange)
```
aaron-steinfeld, line 50: likewise here and a couple other spots in later files

src/app/shared/feature-resolver/feature-resolver.service.ts
```diff
@@ -1,15 +1,17 @@
...
 export class FeatureResolverService extends FeatureStateResolver {
-  public getFeatureState(flag: string): Observable<FeatureState> {
-    switch (flag) {
+  public getFeatureFlagValue<T extends FeatureFlagValue = FeatureFlagValue>(feature: string): Observable<T> {
+    switch (feature) {
       case ApplicationFeature.PageTimeRange:
-        return of(FeatureState.Disabled);
+        return of(false as T);
+      case ApplicationFeature.FeatureDefaultTimeRange:
+        return of(({ 'path-1': '1h', 'path-2': '1d' } as unknown) as T); //dummy data
```
aaron-steinfeld, line 12: Should default to empty. These are really just overrides to allow config to override what's in code.

src/app/shared/feature-resolver/feature-resolver.service.ts
```diff
@@ -1,15 +1,17 @@
...
+  public getFeatureFlagValue<T extends FeatureFlagValue = FeatureFlagValue>(feature: string): Observable<T> {
+    switch (feature) {
       case ApplicationFeature.PageTimeRange:
-        return of(FeatureState.Disabled);
+        return of(false as T);
+      case ApplicationFeature.FeatureDefaultTimeRange:
+        return of(({ 'path-1': '1h', 'path-2': '1d' } as unknown) as T); // Dummy data
       default:
-        return of(FeatureState.Enabled);
+        return EMPTY;
```
aaron-steinfeld, line 14: We should stick with previous default behavior unless a reason to switch.

projects/common/src/feature/state/feature-state.resolver.ts
Christian862: fixed

projects/common/src/time/page-time-range-preference.service.ts
Christian862: fixed

projects/components/src/header/application/application-header.component.ts
Christian862: fixed

aaron-steinfeld (approved):

src/app/shared/feature-resolver/feature-resolver.service.ts
```diff
@@ -9,7 +9,7 @@ export class FeatureResolverService extends FeatureStateResolver {
       case ApplicationFeature.PageTimeRange:
         return of(false as T);
       case ApplicationFeature.FeatureDefaultTimeRange:
-        return of({} as T);
+        return of(({} as unknown) as T);
```
aaron-steinfeld, line 12: nit - don't cast to unknown. What was the issue? Either `FeatureFlagValue` is missing the type, or the fact that `{}` is empty means it's not matching the dictionary type in which case we'd rather cast it to something like `{} as Dictionary<string> & T`

Christian862: it was a tslint issue `no-object-literal-type-assertion`

Christian862: object literals can only be cast to `any` or `unknown`

aaron-steinfeld: You can either split it into declaration and usage or just ignore the lint rule. We don't want to change the casts to be less accurate to satisfy the linter, it's better to turn off the linter if we consider this an exception.

```
        const value = {};
        
        return of(value as T);
```
or
```
        // tslint:disable-next-line: no-object-literal-type-assertion
        return of({} as T);
```
would both be preferable.

aaron-steinfeld (approved):

PR #1722, opened 2022-06-28
fix: added large card config to metric card component

Greedy-Geek:


projects/components/src/metric-card/metric-card.component.ts
```diff
@@ -21,10 +25,22 @@ import { MetricCardIconPosition, MetricCardIndicatorType } from './metric-card';
             </ng-container>
           </ng-container>
         </div>
-        <div class="title-text">{{ this.titleText }}</div>
+        <div
+          [ngClass]="{
+            'large-title-text': this.cardSize === '${MetricCardSize.Large}',
+            'title-text': this.cardSize === '${MetricCardSize.Small}'
```
itssharmasandeep, line 31: instead of doing this, please do this.
```html
<div class="title-text" [ngClass]="this.cardSize">{{ this.titleText }}</div>
```
```css
.title-text {
  &.large {
      text-transform: capitalize;
    }
}
```

similarly below as well

Greedy-Geek: oh, got it, updated

itssharmasandeep (approved):

PR #1725, opened 2022-06-30
Tenant based custom page TR default support

Christian862:
## Description
Allows configuring the page defaults based on the values in the feature flag dictionary map `ApplicationFeature.TenantDefaultTimeRangeMap`.  The page TR defaults now follow the priority: User selected in app -> FF value in the dictionary `map<rootLevelPath, trStringValue>` -> route config define default.

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/common/src/constants/application-constants.ts
+++ b/projects/common/src/constants/application-constants.ts
@@ -1,4 +1,4 @@
 export const enum ApplicationFeature {
   PageTimeRange = 'ui.page-time-range',
-  TenantDefaultTimeRangeMap = 'ui.feature-default-time-range-map'
+  FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map'
 }
--- a/projects/common/src/constants/application-constants.ts
+++ b/projects/common/src/constants/application-constants.ts
@@ -1,4 +1,4 @@
 export const enum ApplicationFeature {
   PageTimeRange = 'ui.page-time-range',
-  FeatureDefaultTimeRange = 'ui.feature-default-time-range-map'
+  TenantDefaultTimeRangeMap = 'ui.feature-default-time-range-map'
 }
--- a/projects/common/src/time/page-time-range-preference.service.ts
+++ b/projects/common/src/time/page-time-range-preference.service.ts
@@ -1,9 +1,9 @@
 import { Injectable } from '@angular/core';
-import { isEmpty, isNil } from 'lodash-es';
+import { isNil } from 'lodash-es';
 import { combineLatest, Observable } from 'rxjs';
 import { map, shareReplay, take } from 'rxjs/operators';
 import { ApplicationFeature } from '../constants/application-constants';
-import { FeatureFlagValue, FeatureStateResolver } from '../feature/state/feature-state.resolver';
+import { FeatureStateResolver } from '../feature/state/feature-state.resolver';
 import { FeatureState } from '../feature/state/feature.state';
 import { NavigationService } from '../navigation/navigation.service';
 import { PreferenceService, StorageType } from '../preference/preference.service';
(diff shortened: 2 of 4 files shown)


projects/common/src/constants/application-constants.ts
```diff
@@ -1,4 +1,4 @@
 export const enum ApplicationFeature {
   PageTimeRange = 'ui.page-time-range',
-  FeatureDefaultTimeRange = 'ui.feature-default-time-range-map'
+  TenantDefaultTimeRangeMap = 'ui.feature-default-time-range-map'
```
aaron-steinfeld, line 3:
```suggestion
  DefaultTimeRangeMap = 'ui.feature-default-time-range-map'
```

Tenant is redundant in the context of the UI, since the whole UI is loaded in a single tenant context.

projects/common/src/time/page-time-range-preference.service.ts
```diff
@@ -32,11 +33,19 @@ export class PageTimeRangePreferenceService {
...
       this.pageTimeRangeStringDictionary$,
-      this.featureStateResolver.getFeatureState(ApplicationFeature.PageTimeRange)
+      this.featureStateResolver.getFeatureState(ApplicationFeature.PageTimeRange),
+      this.featureStateResolver.getFeatureFlagValue(ApplicationFeature.TenantDefaultTimeRangeMap)
     ]).pipe(
-      map(([pageTimeRangeStringDictionary, featureState]) => {
+      map(([pageTimeRangeStringDictionary, featureState, tenantDefaultTimeRangeMap]) => {
         if (featureState === FeatureState.Enabled) {
           if (isNil(pageTimeRangeStringDictionary[rootLevelPath])) {
+            if (this.tenantTimeRangeMapHasValue(tenantDefaultTimeRangeMap, rootLevelPath)) {
```
aaron-steinfeld, line 42: nit: flip around the conditions to avoid nesting:
if has pref override, return
if has map override, return
else return default for path

projects/common/src/time/page-time-range-preference.service.ts
```diff
@@ -49,6 +58,14 @@ export class PageTimeRangePreferenceService {
     );
   }
 
+  public tenantTimeRangeMapHasValue(tenantDefaultTimeRangeMap: FeatureFlagValue, rootLevelPath: string): boolean {
```
aaron-steinfeld, line 61: should already be typed coming in here. Change above to avoid casts: `this.featureStateResolver.getFeatureFlagValue<Dictionary<string>>(ApplicationFeature.TenantDefaultTimeRangeMap)` . Then can just do `rootLevelPath in tenantDefaultTimeRangeMap` since we don't really care about the empty string case (I'd probably inline it, but if you prefer you can leave the function).

projects/common/src/time/page-time-range-preference.service.ts
```diff
@@ -34,22 +34,18 @@ export class PageTimeRangePreferenceService {
     return combineLatest([
       this.pageTimeRangeStringDictionary$,
       this.featureStateResolver.getFeatureState(ApplicationFeature.PageTimeRange),
-      this.featureStateResolver.getFeatureFlagValue(ApplicationFeature.TenantDefaultTimeRangeMap)
+      this.featureStateResolver.getFeatureFlagValue<Dictionary<string>>(ApplicationFeature.FeatureDefaultTimeRangeMap)
     ]).pipe(
-      map(([pageTimeRangeStringDictionary, featureState, tenantDefaultTimeRangeMap]) => {
+      map(([pageTimeRangeStringDictionary, featureState, featureDefaultTimeRangeMap]) => {
         if (featureState === FeatureState.Enabled) {
```
aaron-steinfeld, line 40: can do the same thing here and remove another level of nesting. If disable, return global default.

aaron-steinfeld (approved):

projects/common/src/time/page-time-range-preference.service.ts
Christian862: Gonna leave this one, its easier to read imo. It quickly reads what is behind the feature flag, and what the behavior is when disabled. Oppose to reading each condition more carefully with all of them flat

PR #1728, opened 2022-07-01
Fix prefix match routing

aaron-steinfeld:
## Description
Fetching the route config previously used a "startsWith" match based on an incorrect assumption on how angular's prefix matching worked. It should actually still match full segments. Prefix instead refers to whether later segments invalidate the match. The logic here isn't perfect (corner case scenarios like a multi segment match like `x/y` against a route configs that merge segments like `path: 'x/y'`didn't work and will continue not to), but it's strictly better than what was there previously. We don't generally declare multiple paths in a single route with the exception of path params, which are correctly handled.

### Testing
Verified E2E, updated unit tests

### Checklist:

### DIFF
--- a/projects/common/src/navigation/navigation.service.test.ts
+++ b/projects/common/src/navigation/navigation.service.test.ts
@@ -94,6 +94,12 @@ describe('Navigation Service', () => {
     expect(spectator.service.getRouteConfig(['second/second-second'])).toEqual(secondSecondChildRouteConfig);
   });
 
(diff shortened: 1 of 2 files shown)


palbizu (approved):
Looks good to me but probably we should need more reviewers.

aaron-steinfeld:
> Looks good to me but probably we should need more reviewers.

sorry, had automerge on, but anyone feel free to continue to review and can push any further changes.

PR #1747, opened 2022-07-15
feat: Adding sort by filters to explorer view

palbizu:
## Description
With this change, it will be adding order by the filter to the explorer view. Including three selects one for metric, aggregation and sort by direction. They will be shown only if the corresponding FF is enabled and the interval select option is 'None'

### Testing
Locally tested

### Checklist:

### Documentation


projects/common/src/constants/application-constants.ts
```diff
@@ -1,4 +1,5 @@
 export const enum ApplicationFeature {
   PageTimeRange = 'ui.page-time-range',
-  FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map'
+  FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map',
+  TraceableViewOnly = 'ui.traceable-view-only'
```
aaron-steinfeld, line 4: Generic flag please e.g. `ExplorerOrderBySupport`

palbizu: Generic flag added

projects/observability/src/shared/services/metadata/handler/metadata-graphql-query-handler.service.ts
```diff
@@ -46,7 +46,8 @@ export class MetadataGraphQlQueryHandlerService
       onlySupportsAggregation: result.onlySupportsAggregation,
       onlySupportsGrouping: result.onlySupportsGrouping,
       allowedAggregations: result.supportedAggregations.flatMap(convertFromGraphQlMetricAggregationType),
-      groupable: result.groupable
+      groupable: result.groupable,
+      sortable: true // Assumption: All fields are sortable
```
aaron-steinfeld, line 50: Why do we need to add this attribute? The rest of these are being served by the backend and can be configured (since they depend on things like data type), but sortability we don't need any extra data to determine.

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -262,4 +284,12 @@ export interface ExploreSeries {
   visualizationOptions: ExploreSeriesVisualizationOptions;
 }
 
+export interface GraphQlOrderBy {
+  aggregation?: MetricAggregationType;
+  direction?: GraphQlSortDirection;
```
aaron-steinfeld, line 289: order by not valid without a direction and expression.

projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
```diff
@@ -137,6 +162,10 @@ export class ExploreQueryEditorComponent implements OnChanges, OnInit {
     this.visualizationBuilder.interval(interval === 'NONE' ? undefined : interval);
   }
 
+  public setOrderBy(orderBy: GraphQlOrderBy): void {
```
aaron-steinfeld, line 165: nit: unnecessary wrapper, it's identical to `updateOrderByExpression` above.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,239 @@
...
+import { GraphQlOrderBy } from '../explore-visualization-builder';
+@Component({
+  selector: 'ht-explore-query-order-by-editor',
+  styleUrls: ['./explore-query-order-by-editor.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="order-by-container" *ngIf="this.orderByExpression">
+      <span class="order-by-main-label"> Order by </span>
+      <div class="fields-container">
+        <div class="order-by-input-container" *ngIf="this.metricOptions$ | async as metricOptions">
```
aaron-steinfeld, line 20: please use our async constructs like letAsync and loadAsync, rather than ngIf for side effects. I don't typically review PRs any more, but if this has become common, please push back on it - there's a reason we introduced those other directives.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,239 @@
...
+            [disabled]="aggregationOptions && aggregationOptions.length === 1"
+          >
+            <ht-select-option
+              *ngFor="let aggregationOption of aggregationOptions"
+              [value]="aggregationOption.value"
+              [label]="aggregationOption.label"
+            ></ht-select-option>
+          </ht-select>
+        </div>
+        <div class="order-by-input-container" *ngIf="this.sortByOptions$ | async as sortByOptions">
```
aaron-steinfeld, line 51: these don't really need to come from an observable, there's always going to be exactly two - ascending and descending.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  public onMetricChange(newAttribute: AttributeMetadata): void {
+    this.buildSpecAndEmit(
+      combineLatest([of(newAttribute), this.selectedAggregation$, of(this.orderByExpression?.direction)])
+    );
+  }
+
+  public onAggregationChange(newAggregation: MetricAggregationType): void {
+    this.buildSpecAndEmit(
+      combineLatest([this.selectedMetric$, of(newAggregation), of(this.orderByExpression?.direction)])
```
aaron-steinfeld, line 127: Why are we building this strange observable and passing it to a function? I can't imagine how this will work without a subscribe later. Code smell here, should try to refactor to pass in sync values. Will read on.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,239 @@
...
+    orderByData$: Observable<
+      [AttributeMetadata | undefined, MetricAggregationType | undefined, GraphQlSortDirection | undefined]
+    >
+  ): void {
+    orderByData$
+      .pipe(
+        take(1),
+        switchMap(orderByData => this.buildOrderExpression(...orderByData))
+      )
+      .subscribe(value => this.orderByExpressionChange.emit(value));
```
aaron-steinfeld, line 145: Here it is. So would suggest combining some of the observables going into the template such that we can capture the state in one observable rather than having to manage each one independently and subscribe to all the rest when any one changes.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,239 @@
...
+          label: this.metadataService.getAttributeDisplayName(attribute)
+        }))
+      )
+    );
+  }
+
+  private getSortByOptions(): Observable<SelectOption<GraphQlSortDirection | undefined>[]> {
+    return of([
+      {
+        value: undefined,
```
aaron-steinfeld, line 202: Should not be an option. But this makes me wonder, what's the process like for removing a sort? maybe some kind of reset icon is needed

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,239 @@
...
+        label: 'Asc'
+      },
+      {
+        value: 'DESC',
+        label: 'Desc'
+      }
+    ]);
+  }
+
+  private buildOrderExpression(
```
aaron-steinfeld, line 216: should require the args it needs to actual build, keep the callers honest

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,239 @@
...
+    attribute?: AttributeMetadata,
+    aggregation?: MetricAggregationType,
+    sortBy?: GraphQlSortDirection
+  ): Observable<GraphQlOrderBy> {
+    if (attribute === undefined || aggregation === undefined) {
+      return EMPTY;
+    }
+
+    const aggregationToUse = attribute.allowedAggregations.includes(aggregation)
```
aaron-steinfeld, line 225: no defaulting please - only default in terms of selection (e.g. I pick some metric, and we default the selector to the first aggregation)

projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
```diff
@@ -32,27 +33,38 @@ import {
...
+            [limit]="currentVisualization.groupBy?.limit"
+            (limitChange)="this.updateGroupByLimit(currentVisualization.groupBy!, $event)"
+            [includeRest]="currentVisualization.groupBy?.includeRest"
+            (includeRestChange)="this.updateGroupByIncludeRest(currentVisualization.groupBy!, $event)"
+          >
+          </ht-explore-query-limit-editor>
+        </div>
+        <div class="filters-row" *htIfFeature="'${ApplicationFeature.ExplorerOrderBySupport}' | htFeature">
+          <ht-explore-query-order-by-editor
+            *ngIf="!currentVisualization.interval"
```
aaron-steinfeld, line 61: If the interval is added back, the order by would be hidden - but would the state be cleared? We need to make sure we're not generating a query with both things.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: there are three options indeed: None, asc, desc

palbizu: done

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: We're using the 'None' option in other filters like order by. So I just keeping the same logic here.

projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
palbizu: removed

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
aaron-steinfeld: just to clarify, filters don't have that behavior - they have a clear icon. I believe group by does though. If we want to follow that pattern, we should be putting that none option in the attribute selection rather than the direction selector.

projects/observability/src/shared/services/metadata/handler/metadata-graphql-query-handler.service.ts
palbizu: yes, you're right

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
palbizu: optional fields removed

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -248,7 +254,14 @@ export class ExplorerComponent {
           }
         : undefined,
       interval: this.decodeInterval(param.get(ExplorerQueryParam.Interval)),
-      series: param.getAll(ExplorerQueryParam.Series).flatMap(series => this.tryDecodeExploreSeries(series))
+      series: series,
+      orderBy: {
```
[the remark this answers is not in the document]
palbizu: done, removed

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -138,11 +147,23 @@ export class ExploreVisualizationBuilder implements OnDestroy {
...
         filters: state.filters && this.graphQlFilterBuilderService.buildGraphQlFieldFilters(state.filters),
         groupBy: state.groupBy,
+        orderBy: state.orderBy && this.mapOrderByToGraphQlSpecification(state.orderBy),
         limit: state.resultLimit
       }))
     );
   }
 
+  private mapOrderByToGraphQlSpecification(orderBy: GraphQlOrderBy): GraphQlSortBySpecification[] | undefined {
+    return orderBy.keyExpression !== undefined && orderBy.aggregation !== undefined && orderBy.direction !== undefined
```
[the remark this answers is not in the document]
palbizu: done. replaced by `orderBy !== undefined`

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: done

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: done

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: It has a subscription later. So I will keep i.

projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
palbizu: done

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: done

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -248,7 +255,14 @@ export class ExplorerComponent {
...
+        aggregation: series[0].specification.aggregation!,
```
anandtiwary, line 260: can we avoid the `!` ?

anandtiwary: probably 

```
const orderBy = series?.[0].specification.aggregation ? {<entire object>} : undefined.
``

File: projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.test.ts [hunk 16]
```diff
@@ -87,6 +89,10 @@ describe('Explore query editor', () => {
         getAutoDurationFromTimeDurations: () => new TimeDuration(15, TimeUnit.Second),
         getAutoDuration: () => new TimeDuration(15, TimeUnit.Second)
       }),
+      mockProvider(FeatureStateResolver, {
+        getCombinedFeatureState: () => of(FeatureState.Disabled),
+        getFeatureState: () => of(FeatureState.Disabled)
```
anandtiwary, line 94: we probably only need to override one of the two methods.

palbizu: removed combinedFeatureState

projects/observability/src/pages/explorer/explorer.component.ts
palbizu: done

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,243 @@
...
+      switchMap(() => this.getCurrentSelectedAggregation())
+    );
+
+    this.selectedMetric$ = combineLatest([this.contextSubject, this.incomingOrderByExpressionSubject]).pipe(
+      switchMap(() => this.getCurrentSelectedAttribute())
+    );
+
+    this.selectedSortBy$ = this.incomingOrderByExpressionSubject.pipe(switchMap(() => this.getCurrentSelectedSortBy()));
+
+    this.metricOptions$ = this.contextSubject.pipe(switchMap(context => this.getAttributeOptionsForContext(context)));
```
anandtiwary, line 112: Since i am seeing multiple subscriptions for `metricOptions`, it might be better to add a `share` operator here.

projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
```diff
@@ -107,6 +124,10 @@ export class ExploreQueryEditorComponent implements OnChanges, OnInit {
     if (changeObject.groupBy && this.groupBy?.keyExpressions.length) {
       this.updateGroupByExpression(this.groupBy, this.groupBy.keyExpressions[0]);
     }
+
+    if (changeObject.orderBy && this.orderBy) {
```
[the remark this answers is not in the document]
palbizu: removed

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,244 @@
...
+  public constructor(private readonly metadataService: MetadataService) {
+    this.selectedAggregation$ = combineLatest([this.contextSubject, this.incomingOrderByExpressionSubject]).pipe(
+      switchMap(() => this.getCurrentSelectedAggregation())
+    );
+
+    this.selectedMetric$ = combineLatest([this.contextSubject, this.incomingOrderByExpressionSubject]).pipe(
+      switchMap(() => this.getCurrentSelectedAttribute())
+    );
+
+    this.selectedSortBy$ = this.incomingOrderByExpressionSubject.pipe(switchMap(() => this.getCurrentSelectedSortBy()));
```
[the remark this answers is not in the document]
palbizu: replaced by orderBy

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: added.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    this.metricOptions$ = this.contextSubject.pipe(switchMap(context => this.getAttributeOptionsForContext(context)));
```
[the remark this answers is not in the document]
palbizu: renamed to attribute in all cases

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -138,11 +152,23 @@ export class ExploreVisualizationBuilder implements OnDestroy {
...
         filters: state.filters && this.graphQlFilterBuilderService.buildGraphQlFieldFilters(state.filters),
         groupBy: state.groupBy,
+        orderBy: state.orderBy && this.mapOrderByToGraphQlSpecification(state.orderBy),
         limit: state.resultLimit
       }))
     );
   }
 
+  private mapOrderByToGraphQlSpecification(orderBy: GraphQlOrderBy): GraphQlSortBySpecification[] | undefined {
+    return orderBy !== undefined
```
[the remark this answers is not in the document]
palbizu: fixed. Check removed

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -138,11 +152,23 @@ export class ExploreVisualizationBuilder implements OnDestroy {
...
+      ? [
+          {
+            direction: orderBy.direction,
+            key: this.exploreSpecBuilder.exploreSpecificationForKey(orderBy.keyExpression.key, orderBy.aggregation)
```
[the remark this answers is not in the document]
palbizu: replaced

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -262,4 +289,12 @@ export interface ExploreSeries {
   visualizationOptions: ExploreSeriesVisualizationOptions;
 }
 
+export interface GraphQlOrderBy {
+  aggregation: MetricAggregationType;
+  direction: GraphQlSortDirection;
+  keyExpression: {
+    key: string;
```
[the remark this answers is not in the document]
palbizu: done

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -337,5 +356,6 @@ const enum ExplorerQueryParam {
   Group = 'group',
   OtherGroup = 'other',
   GroupLimit = 'limit',
-  Series = 'series'
+  Series = 'series',
+  Sort = 'sort'
```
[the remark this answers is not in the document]
palbizu: Currently is not generating an issue related to this new query param. Data is getting fine from backend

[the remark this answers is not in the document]
aaron-steinfeld: Because this isn't being used - https://github.com/hypertrace/hypertrace-ui/pull/1747#discussion_r971062922 . Take a look at how the other query params are being written and read from the url.

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -88,9 +90,20 @@ export class ExploreVisualizationBuilder implements OnDestroy {
...
+    });
+  }
+
   public interval(interval?: TimeDuration | 'AUTO'): this {
     return this.updateState({
-      interval: interval
+      interval: interval,
+      ...(interval !== undefined
+        ? {
+            orderBy: undefined
```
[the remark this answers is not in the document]
palbizu: moved to mapStateToExploreQuery function

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -235,6 +238,20 @@ export class ExplorerComponent {
   }
 
   private mapToInitialState(param: ParamMap): InitialExplorerState {
+    const series: ExploreSeries[] = param
+      .getAll(ExplorerQueryParam.Series)
+      .flatMap((seriesString: string) => this.tryDecodeExploreSeries(seriesString));
+
+    const orderBy: GraphQlOrderBy | undefined = series?.[0].specification.aggregation
```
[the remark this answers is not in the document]
palbizu: done

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    if (changeObject.orderByExpression) {
+      this.incomingOrderByExpressionSubject.next(this.orderByExpression);
+    }
+  }
+
+  public onMetricChange(newAttribute: AttributeMetadata): void {
+    this.buildSpecAndEmit(
+      combineLatest([
```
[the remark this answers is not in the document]
palbizu: I would prefer to refactor it later if it's necessary but up to you.

projects/observability/src/pages/explorer/explorer.component.ts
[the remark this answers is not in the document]
palbizu: I've already added the query params read and write actions. Does it all or do I need to add something extra for the table?

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
```diff
@@ -95,24 +95,30 @@ export class ExploreQueryOrderByEditorComponent implements OnChanges {
...
 
-    this.selectedSortBy$ = this.incomingOrderByExpressionSubject.pipe(switchMap(() => this.getCurrentSelectedSortBy()));
+    this.selectedOrderBy$ = this.incomingOrderByExpressionSubject.pipe(
+      switchMap(() => this.getCurrentSelectedOrderBy())
+    );
 
-    this.metricOptions$ = this.contextSubject.pipe(switchMap(context => this.getAttributeOptionsForContext(context)));
+    this.attributeOptions$ = this.contextSubject.pipe(
+      switchMap(context => this.getAttributeOptionsForContext(context)),
+      shareReplay()
```
aaron-steinfeld, line 116: Would this ever clean up? The shareReplay lives on after it's been unsubscribed and pretty sure we can drop it here. Plain share would probably accomplish what you're looking for, but honestly It's overkill for a cheap compute, it's cached upstream. (applies in multiple places in this file).

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -235,6 +249,14 @@ export class ExplorerComponent {
   }
 
   private mapToInitialState(param: ParamMap): InitialExplorerState {
+    const series: ExploreSeries[] = param
```
aaron-steinfeld, line 252: This could be reverted right? it's no longer being used by the order by decode. 

That said, do we handle the default order by from series case too? I don't see that logic moved anywhere else, just removed (I very well may have missed it though).

Basically, I would expect to be able to do all of
- Change interval to none, and have a default order by added based on my current first series (no order by in the url to start, but added after)
- Every time I change the ordering details, the URL is updated appropriately.
- At any time, I can grab the url and reload the page (or in another tab) with the order by I had selected being restored.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: Using share operator produces an incorrect result. Can I remove directly this shareReplay()?

projects/observability/src/pages/explorer/explorer.component.ts
palbizu: I should remove the series and orderBy const and assign them to the underneath object. These 3 scenarios are currently satisfied.

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
aaron-steinfeld: That may be the case if there's late subscribers, would need to go through the logic. In that case I'd just remove entirely.

projects/observability/src/pages/explorer/explorer.component.ts
aaron-steinfeld: Can you point me to the code that does the first one?

palbizu: You are right. Currently, I'm taking undefined as None. When there is no URL param for order, the attribute will be taken as None and the order by will not be applied. I suppose that I will need a new state there, the NONE state, the undefined state and the regular.

aaron-steinfeld: We don't actually want a non state though - if the url param isn't defined, we should default to the value of the first series metric desc. there should be no way to get to a none state (other than setting an interval, in which case the order by won't be visible)

palbizu: I already added a None state (in attribute) to prevent the re-ordering. Do you say that I need to remove it? If the user clicks on None, the order is not applied.

aaron-steinfeld: I think the idea had been to remove None, correct. There's no value in having it - everything has some order, so let's pick a reasonable default that the user can change. There's no use case that I can think of where a user would say - "I want random order".

projects/observability/src/shared/components/explore-query-editor/order-by/explore-query-order-by-editor.component.ts
palbizu: done. removed

projects/observability/src/shared/components/explore-query-editor/explore-query-editor.component.ts
```diff
@@ -56,7 +56,7 @@ import {
           >
           </ht-explore-query-limit-editor>
         </div>
-        <div class="filters-row" *htIfFeature="'${ApplicationFeature.ExplorerOrderBySupport}' | htFeature">
```
aaron-steinfeld: this needs to be restored right?

palbizu: yes, ignore it for now. I made a mistake when I push it. I have pending one change yet.

palbizu: rolled back

projects/observability/src/pages/explorer/explorer.component.ts
palbizu: done all scenarios

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -117,18 +127,44 @@ export class ExploreVisualizationBuilder implements OnDestroy {
   }
 
   private buildRequest(state: ExploreRequestState): ExploreVisualizationRequest {
+    const orderBy = this.getOrderBy(state.series[0], state.interval, state.orderBy);
```
palbizu, line 130: @aaron-steinfeld I do not totally agree with this approach but I should apply the new orderBy to the attribute and also to the exploreQuery$ to remove the param or add it when it's necessary. It's working fine but any suggestion here?

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -325,22 +324,32 @@ export class ExplorerComponent {
...
-  private tryDecodeExploreOrderBy(orderByString?: string): ExploreOrderBy | undefined {
+  private tryDecodeExploreOrderBy(
+    interval: IntervalValue,
+    selectedSeries: ExploreSeries,
+    orderByString?: string
+  ): ExploreOrderBy | undefined {
     const matches = orderByString?.match(/(\w+)\((\w+)\):(\w+)/);
 
     if (matches?.length !== 4) {
+      if (interval === 'NONE') {
```
aaron-steinfeld, line 335: If the interval is not none, the whole thing is undefined in any case. We could put this as the first line of this function or better yet, outside it. Only try to decode order by if interval is not none e.g. `orderBy: interval === 'NONE' ? this.tryDecodeExploreOrderBy(series[0], param.get(ExplorerQueryParam.Order)) : undefined`

projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
aaron-steinfeld: Not sure I'm following why this is separate. The order by should already be part of the state. Let me read on.

aaron-steinfeld: My understanding after reading it is we're using this function to set defaults, but we shouldn't be because this only translates the UI to a data request and we want the defaults to show in the actual UI - so that should have already happened at this point. It looks like there's two behaviors: clearing order by if interval is set and setting the default to the value of the current first series.

But we previously did the clearing immediately when interval was set, so why the change? This had been brought up in https://github.com/hypertrace/hypertrace-ui/pull/1747#discussion_r971068145 and the concern is that we want to make sure whenever interval is cleared, a default order is set, but when an interval is assigned, the order is cleared. That concern still holds. In this PR, we assign the default in two places - when decoding the URL (only if interval is set to none, so this only is relevant in restoring particular URL patterns), and when sending the query. So we've added that logic on the request building side, but not on the UI state I believe? How does the state (and thus the UI) get updated with the default?

For example - what's the result if I load explorer with the default state (calls time series, interval AUTO), then change the interval to None? The default order by was undefined, so my expectation from reading the code (and I certainly could have missed it), would that the query would be right, the UI order by selector would be wrong.

aaron-steinfeld: Will try to pull this to test it out and see if I'm misunderstanding.

aaron-steinfeld: It does appear to work, will try to understand the code better.

aaron-steinfeld: There's some funkiness going on, but it may not be on the UI side. Easier to chat.

palbizu: fixed with workaround

aaron-steinfeld (approved):

PR #1758, opened 2022-07-21
feat: adding a file upload service and refactoring some code

anandtiwary:
## Description
feat: adding a file upload service and refactoring some code

### Testing

### Checklist:

### Documentation


projects/components/src/file-upload/file-upload.service.ts
```diff
@@ -0,0 +1,127 @@
...
+import { Injectable } from '@angular/core';
+import { uniqueId } from 'lodash-es';
+import { fromEvent, Observable, of, Subscription } from 'rxjs';
+import { catchError, delay, filter, finalize, map, shareReplay } from 'rxjs/operators';
+import { NotificationService } from '../notification/notification.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class FileUploadService {
```
anandtiwary, line 11: Tests will be added later. This is the first PR. Need to test this approach too

projects/components/src/file-upload/file-upload.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (this.beforeUnloadSubscription === undefined) {
+      this.addBeforeUnloadHandler();
+    }
+
+    const fileUploadId = uniqueId('file-upload');
+    const upload$ = this.uploadFilesAsFormData<TResponse>(url, formData).pipe(
+      delay(100),
+      finalize(() => {
+        this.onUploadComplete(fileUploadId);
+      }),
```
arjunlalb, line 30: do we expect failures here?

arjunlalb (approved):

projects/components/src/file-upload/file-upload.service.ts
anandtiwary: Yes, it could fail. We are capturing error in `uploadFilesAsFormData`

arjunlalb: finalize would be called on both success and failure. is that our intention?

if rest of the logic is in the `uploadFilesAsFormData` method, it'd make sense for this also to be moved there so that all the operators are in one place and readable. Or lift up the success/failure handling here. Both works.

palbizu (approved):

projects/components/src/file-upload/file-display/file-display.component.ts
```diff
@@ -50,23 +59,16 @@ export class FileDisplayComponent {
...
-  @Input()
-  public errorStateTooltipText: string = '';
+  public latestUploadEvent?: FileUploadEvent<unknown>;
 
   @Output()
   public readonly deleteClick: EventEmitter<void> = new EventEmitter();
 
-  // Only show file delete option if file upload has not started or it has resulted in failure
-  public get showDelete(): boolean {
-    return this.state === FileUploadState.NotStarted || this.state === FileUploadState.Failure;
```
itssharmasandeep, line 66: Now we're also showing the delete when the upload is in progress? I don't think we should do that.
Because if a big file upload is in progress and then what happens if a user removes it in between. After the success(Only if we have a way to delete it later) and before the start is fine showing that.

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -46,16 +45,8 @@ import { FileUploadState } from './file-display/file-display';
...
-        <ht-file-display
-          *ngFor="let file of this.files; let index = index"
-          [file]="file"
-          [state]="this.uploadState"
-          (deleteClick)="this.deleteFile(index)"
-        >
-        </ht-file-display>
-      </div>
+
+      <ng-content></ng-content>
```
itssharmasandeep, line 49: So we're moving the logic of the file display and progress to the parent?

projects/components/src/file-upload/file-upload.service.ts
```diff
@@ -0,0 +1,127 @@
...
+import { fromEvent, Observable, of, Subscription } from 'rxjs';
+import { catchError, delay, filter, finalize, map, shareReplay } from 'rxjs/operators';
+import { NotificationService } from '../notification/notification.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class FileUploadService {
+  private readonly activeSubscriptionsMap: Map<string, Subscription> = new Map<string, Subscription>();
+  private beforeUnloadSubscription?: Subscription;
```
itssharmasandeep, line 13: is this a typo? should be `beforeUploadSubscription`?

projects/components/src/file-upload/file-upload.service.ts
```diff
@@ -0,0 +1,127 @@
...
+              };
+            }
+
+            // Error Response
+            throw new Error(`File upload failed with status ${event.status}`);
+          }
+
+          return undefined;
+        }),
+        this.notificationService.withNotification(`File upload successful`, `File upload failed`),
```
itssharmasandeep, line 75: Where will be showing exact error message coming from backend for file upload

projects/components/src/file-upload/file-upload.component.ts
anandtiwary: yes. This makes the api more flexible and also allows the consumer to either show one progress bar for bulk upload or many for individual file upload.

projects/components/src/file-upload/file-upload.service.ts
anandtiwary: The observable still contains the original error. It can be shown inside file display component

projects/components/src/file-upload/file-display/file-display.component.ts
anandtiwary: added back

projects/components/src/file-upload/file-upload.service.ts
anandtiwary: This method just does a long lived subscription. I am not sure i understand the change

itssharmasandeep (approved):

Christian862 (approved):

PR #1774, opened 2022-08-05
style(page-header): vertically center timerange

adisreyaj:
## Description
Vertically center the time range. Because adding additional controls in the page header causes the controls and time range to be mis-aligned.

### DIFF
--- a/projects/components/src/header/page/page-header.component.scss
+++ b/projects/components/src/header/page/page-header.component.scss
@@ -2,9 +2,10 @@
 @import 'font';
 
 .page-header {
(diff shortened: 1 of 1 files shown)


projects/components/src/header/page/page-header.component.scss
```diff
@@ -45,6 +46,8 @@
   }
 
   .time-range {
+    align-items: center;
```
adisreyaj, line 49: Changes Here ⬆️

itssharmasandeep (approved):

PR #1775, opened 2022-08-05
feat: added ht-status-display component

chahat99:
Added ht-status-display component

### DIFF
--- a/projects/observability/src/shared/components/status-display/status-display-text.pipe.ts
+++ b/projects/observability/src/shared/components/status-display/status-display-text.pipe.ts
@@ -1,0 +1,24 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { displayString } from '@hypertrace/common';
+import { isNil } from 'lodash-es';
+
+@Pipe({
+  name: 'htStatusDisplayText'
+})
+export class StatusDisplayTextPipe implements PipeTransform {
+  /**
+   * Appends the `statusMessage` with a `-` if it's available.
+   * If not just show the `statusCode`
+   *
+   * @param statusCode - status code
+   * @param statusMessage - status message
+   */
+  public transform(statusCode: string | number, statusMessage?: string): string {
+    let statusDisplayText = displayString(statusCode);
+    if (!isNil(statusMessage) && statusMessage !== 'null') {
+      statusDisplayText = `${statusDisplayText} - ${statusMessage}`;
+    }
+
+    return statusDisplayText;
+  }
+}
--- a/projects/observability/src/shared/components/status-display/status-display.component.scss
+++ b/projects/observability/src/shared/components/status-display/status-display.component.scss
@@ -1,0 +1,12 @@
+@import 'mixins';
+
+.status {
+  display: flex;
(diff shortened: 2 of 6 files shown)


projects/observability/src/shared/components/status-display/status-display.component.scss
```diff
@@ -0,0 +1,12 @@
+@import 'mixins';
+
+.status {
+  display: flex;
+  align-items: center;
```
Greedy-Geek, line 5: nit: you can use display-flex(column) from mixins here

Greedy-Geek (approved):

projects/observability/src/shared/components/status-display/status-display.component.scss
itssharmasandeep: why? But it is not a column flow?

Greedy-Geek: oh sorry, my mistake, didn't see that

projects/observability/src/shared/components/status-display/status-display.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  @Input()
+  public statusCode?: number | string;
+
+  @Input()
+  public status?: string;
+
+  @Input()
+  public statusMessage?: string;
+
+  // TODO: Add display for different styles if required
```
itssharmasandeep, line 25: @arjunlalb @adisreyaj 
We have this component as gray background and since we want this to be used at the status table cell renderer (which has different styling according to the status), how do we wanna approach this?

1. Should we replace that styling with this new one
2. Or add that styling as a config here as well?

itssharmasandeep:
Please export this as well from `public-api.ts` file present in the components project

projects/observability/src/shared/components/status-display/status-display-text.pipe.ts
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * Appends the `statusMessage` with a `-` if it's available.
+   * If not just show the `statusCode`
+   *
+   * @param statusCode - status code
+   * @param statusMessage - status message
+   */
+  public transform(statusCode: string | number, statusMessage?: string): string {
+    let statusDisplayText = displayString(statusCode);
+    if (!isNil(statusMessage) && statusMessage !== 'null') {
```
arjunlalb, line 18: why are we looking for string `null` here?

projects/observability/src/shared/components/status-display/status-display.component.ts
arjunlalb: We should use a single style. I recommend using the existing style. After consulting with design team, we can switch it over to the new one or keep the old one.

itssharmasandeep: I agree on keeping the same styling throughout the product and yeah, the existing one seems better to me as well. But let's wait for @adisreyaj to comment on this since I think he has more context around that.

I think, We can leave this here as it is until we get final confirmation.

PR #1781, opened 2022-08-08
fix: table row height

jake-bassett:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/table/table.component.ts
+++ b/projects/components/src/table/table.component.ts
@@ -159,6 +159,7 @@ import { TableColumnConfigExtended, TableService } from './table.service';
           }"
           class="data-row"
           [style.min-height]="this.rowHeight"
+          [style.max-height]="this.rowHeight"
         ></cdk-row>
 
         <!-- Expandable Detail Rows -->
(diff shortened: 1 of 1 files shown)


projects/components/src/table/table.component.ts
```diff
@@ -158,7 +158,7 @@ import { TableColumnConfigExtended, TableService } from './table.service';
             selectable: this.supportsRowSelection()
           }"
           class="data-row"
-          [style.height]="this.rowHeight"
+          [style.min-height]="this.rowHeight"
```
aaron-steinfeld, line 161: I feel like I see a PR changing his back and forth between height and min height like twice a month. Please tread carefully and dig through PR history for context.

aaron-steinfeld: Here's one, looks like it was probably reverted more recently - https://github.com/hypertrace/hypertrace-ui/pull/1711/files

aaron-steinfeld: Here's another https://github.com/hypertrace/hypertrace-ui/pull/1724/files

itssharmasandeep: Yeah, this is an issue, but the `height` change, created more issues in the table, sometime it looks like all the tables are broken due to that.

arjunlalb, line 161: We should be ok with using min-height here as long as we verify all the existing tables and override any broken places with explicit height.

We reverted this change last time (I was the author) since we didn't consider individual height overrides then @jake-bassett

arjunlalb (approved):

PR #1782, opened 2022-08-12
fix: added function to get root path url in navigation service

Greedy-Geek:
## Description
Added get root path url in navigation service.

### Testing
Tested on the local

### Checklist:
-  My changes generate no new warnings

### DIFF
--- a/projects/common/src/navigation/navigation.service.ts
+++ b/projects/common/src/navigation/navigation.service.ts
@@ -273,7 +273,7 @@ export class NavigationService {
   }
 
   public getRootPathUrl(): string {
-    return window.location.host;
+    return new URL(this.getAbsoluteCurrentUrl()).origin;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


projects/common/src/navigation/navigation.service.ts
```diff
@@ -272,6 +272,10 @@ export class NavigationService {
     return this.platformLocation.href;
   }
 
+  public getRootPathUrl(): string {
+    return window.location.host;
```
arjunlalb, line 276: please refer to L271

arjunlalb: what you need will be in `origin` in `platformLocation`

Greedy-Geek: platformLocation does not provide host, through that we'll need to add hostname and port, is that fine?

Greedy-Geek: updated

arjunlalb (approved):

PR #1787, opened 2022-08-16
Add plain text filter operator to escape special chars

Christian862:
## Description
Search input would generally use the `FilterOperator.Like` , which is the regex filter operator. Because of this, you were unable to search with special characters, because they were not escaped. This PR introduces the `FilterOperator.Contains` in order to escape the special characters and treat the string as plain text.

In this PR, we look for the `FilterOperator.Contains` filters and do two things:
1. Transform the value of the filter to escape the special character's, allowing the GQL to treat it as plain text
2. Change the operator type from: `FilterOperator.Contains` to: `FilterOperator.Like`
  - This is because the backend doesn't support a plain text field yet, so we transform the regex string, to act like plain text, then pass it to the query as a `FilterOperator.Like`.  This way In the future when backend support is in for this, we just correct the mapping and remove the string transformation and its working.

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/filtering/filter/filter-operators.ts
+++ b/projects/components/src/filtering/filter/filter-operators.ts
@@ -9,6 +9,12 @@ export const enum FilterOperator {
   GreaterThanOrEqualTo = '>=',
   Like = '~',
   In = 'IN',
+  /**
+   * The difference between 'Contains' and 'Like' is that 'Like' will be treated as a regex filter, not escaping
+   * special characters. Whereas 'Contains' will escape special characters in order to treat it as plain text - currently
+   * this assumes no backend support and is done clientside, which is why FilterOperator.Contains gets mapped to
+   * GraphQlOperatorType.Like
+   */
   Contains = 'CONTAINS',
   ContainsKey = 'CONTAINS_KEY',
   ContainsKeyLike = 'CONTAINS_KEY_LIKE'
--- a/projects/components/src/filtering/filter/filter-operators.ts
+++ b/projects/components/src/filtering/filter/filter-operators.ts
@@ -9,6 +9,7 @@ export const enum FilterOperator {
   GreaterThanOrEqualTo = '>=',
   Like = '~',
   In = 'IN',
+  Contains = 'CONTAINS',
   ContainsKey = 'CONTAINS_KEY',
   ContainsKeyLike = 'CONTAINS_KEY_LIKE'
 }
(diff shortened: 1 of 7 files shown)


projects/observability/src/shared/services/filter-builder/graphql-filter-builder.service.ts
```diff
@@ -37,10 +38,16 @@ export class GraphQlFilterBuilderService {
         new GraphQlFieldFilter(
           { key: filter.field, subpath: filter.subpath },
           toGraphQlOperator(filter.operator),
-          filter.value as GraphQlArgumentValue
+          this.transformValueIfOperatorIsContains(filter)
         )
     );
   }
+
+  private transformValueIfOperatorIsContains(filter: Filter | FieldFilter | TableFilter): GraphQlArgumentValue {
```
aaron-steinfeld, line 46: nit: would rename to just something like "extractGraphQlFilterValue" or similar. Basically, the contains bit is internal to this function.

aaron-steinfeld (commented):
overall lgtm

projects/observability/src/shared/services/filter-builder/graphql-filter-builder.service.ts
Christian862: good point, fixed

aaron-steinfeld (approved):

arjunlalb (dismissed):

projects/components/src/filtering/filter/filter-operators.ts
```diff
@@ -9,6 +9,7 @@ export const enum FilterOperator {
   GreaterThanOrEqualTo = '>=',
   Like = '~',
   In = 'IN',
+  Contains = 'CONTAINS',
```
arjunlalb, line 12: recommendation : would be helpful to add couple lines describing the differences (or use cases) between LIKE and CONTAINS

Christian862: added

projects/components/src/filtering/filter/filter-operators.ts
```diff
@@ -9,6 +9,12 @@ export const enum FilterOperator {
   GreaterThanOrEqualTo = '>=',
   Like = '~',
   In = 'IN',
+  /**
+   * The difference between 'Contains' and 'Like' is that 'Like' will be treated as a regex filter, not escaping
```
aaron-steinfeld, line 13: nit: I would get list into impl details here. CONTAINS is a filter if the field contains the provided string LIKE is a regex match filter.

aaron-steinfeld (approved):

projects/components/src/filtering/filter/filter-operators.ts
aaron-steinfeld: Please follow up on this if you agree

PR #1789, opened 2022-08-18
feat: add support for not in operator

adisreyaj:
## Description
Add support for `NOT_IN` filter.


projects/components/src/filtering/filter/parser/filter-parser-lookup.service.ts
```diff
@@ -28,6 +29,8 @@ export class FilterParserLookupService {
         return new ComparisonFilterParser();
       case FilterOperator.In:
         return new InFilterParser();
+      case FilterOperator.NotIn:
```
aaron-steinfeld, line 32: This should be able to use the same parser as IN, no need for a new one (rename may be appropriate though)

adisreyaj: Updated!

aaron-steinfeld (dismissed):

projects/components/src/filtering/filter/parser/types/in-not-in-filter-parser.ts
```diff
@@ -5,13 +5,18 @@ import { FilterOperator } from '../../filter-operators';
...
 
-export class InFilterParser extends AbstractFilterParser<PossibleValuesTypes> {
+export class InNotInFilterParser extends AbstractFilterParser<PossibleValuesTypes> {
   public supportedAttributeTypes(): FilterAttributeType[] {
-    return [FilterAttributeType.String, FilterAttributeType.Number, FilterAttributeType.StringMap];
+    return [
+      FilterAttributeType.String,
+      FilterAttributeType.Number,
+      FilterAttributeType.StringMap,
+      FilterAttributeType.StringArray
```
Christian862, line 14: Just curious, why do we need to add `FilterAttributeType.StringArray` here ?

aaron-steinfeld: good catch - I almost commented on this myself since it's independent of the rest of the change :) What it's doing is adding support both for the existing IN operator and the new NOT_IN operator to support string array attributes. For example, MyEntity.labels IN [label1, label2]. So an array on both sides of the operator.

arjunlalb (approved):

PR #1798, opened 2022-08-26
fix: Group by Response Alias special characters

palbizu:
Replace text to support alias (ignoring special chars or spaces)

### Testing
Tested Locally

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
+++ b/projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
@@ -7,7 +7,7 @@ import {
   TimeRangeService
 } from '@hypertrace/common';
 import { Filter } from '@hypertrace/components';
-import { uniqBy } from 'lodash-es';
+import { camelCase, uniqBy } from 'lodash-es';
 import { BehaviorSubject, Observable, of, Subject } from 'rxjs';
 import { debounceTime, defaultIfEmpty, distinctUntilChanged, map, shareReplay, takeUntil } from 'rxjs/operators';
 import { AttributeMetadata } from '../../graphql/model/metadata/attribute-metadata';
@@ -84,7 +84,7 @@ export class ExploreVisualizationBuilder implements OnDestroy {
 
   public groupBy(groupBy?: GraphQlGroupBy): this {
     return this.updateState({
-      groupBy: groupBy
+      groupBy: groupBy && this.encodeGroupBy(groupBy)
     });
   }
 
@@ -233,6 +233,18 @@ export class ExploreVisualizationBuilder implements OnDestroy {
     }
 
     return of(interval);
+  }
+
+  private encodeGroupBy(groupBy: GraphQlGroupBy): GraphQlGroupBy {
+    return {
+      ...groupBy,
+      keyExpressions: [
+        {
+          ...groupBy.keyExpressions[0],
+          subpath: camelCase(groupBy.keyExpressions[0].subpath)
+        }
+      ]
+    };
   }
 }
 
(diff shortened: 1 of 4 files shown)


projects/observability/src/shared/components/explore-query-editor/explore-visualization-builder.ts
```diff
@@ -234,6 +234,21 @@ export class ExploreVisualizationBuilder implements OnDestroy {
...
+
+  private encodeGroupBy(groupBy: GraphQlGroupBy): GraphQlGroupBy {
+    return {
+      ...groupBy,
+      keyExpressions:
+        groupBy.keyExpressions.length > 0
+          ? [
+              {
+                ...groupBy.keyExpressions[0],
+                subpath: camelCase(groupBy.keyExpressions[0].subpath)
```
aaron-steinfeld, line 246: Is the only goal here to camel case this?
1. If this needs to be done, this isn't the place to do it. It should be where the group by expression is built, so we don't have two different objects with the same type one valid, one invalid.
2. Why does it need to be done in the first place? Subpaths are user-defined schemas, so if they gave us something with a particular casing, we should be respecting that.

palbizu: @aaron-steinfeld the issue here appears when we try to use the hyphened result as a variable. So I took this strategy as a workaround. The particular issue appears on the selection, is not possible to use it as a variable (the same for other types of special characters like `'/'` or `'.'`. I understand probably this would be not the best fix but I'm not able to find a better solution. 

image

aaron-steinfeld: Discussed offline, the issue is the alias can't have the same special chars a subpath supports, so the appropriate fix would be to update the spec builder which generates the alias given an expression to replace those special chars and leave the subpath untouched (it already does this with some special chars, but not `-`)

palbizu: done

aaron-steinfeld (dismissed):

projects/observability/src/shared/components/explore-query-editor/series/explore-query-series-group-editor.component.test.ts
```diff
@@ -127,6 +127,6 @@ describe('Explore Query Series Group Editor component', () => {
     spectator.queryAll(ExploreQuerySeriesEditorComponent)[0].seriesChange.emit(buildSeries('modified first'));
 
     expect(onChange).toHaveBeenCalledTimes(1);
-    expect(onChange).toHaveBeenCalledWith([matchSeriesWithName('modified first'), matchSeriesWithName('second')]);
+    expect(onChange).toHaveBeenCalledWith([matchSeriesWithName('modified_first'), matchSeriesWithName('second')]);
```
aaron-steinfeld, line 130: this one's actually interesting. I believe we use name and alias as the output of that regex, and maybe it would be better to leave name untouched and just do the replace on the alias? I guess can go with this since it matches existing behavior for now (and I don't know if code relies on the name in other ways), and revisit if needed. I'm guessing the effect is that something like this would show up in the legend with underscores rather than spaces.

aaron-steinfeld (approved):
If it were me, I'd probably explore just doing alias instead of both alias and name, but up to you/rest of UI.

projects/observability/src/shared/components/explore-query-editor/series/explore-query-series-group-editor.component.test.ts
palbizu: I would prefer to do it later if rise some issue

PR #1808, opened 2022-09-02
feat: add getShareableUrl method to navigation service

arjunlalb:
## Description
Expose a method to construct a shareableUrl that contains the current domain info.

### Testing
Added UT

### Checklist:


projects/common/src/navigation/navigation.service.ts
```diff
@@ -272,8 +272,8 @@ export class NavigationService {
     return this.platformLocation.href;
   }
 
-  public getHostUrl(): string {
-    return new URL(this.getAbsoluteCurrentUrl()).origin;
+  public getShareableUrl(path: string[], queryParams?: Dictionary<string>): string {
```
aaron-steinfeld, line 275: Could we make this interface match some of the other ones? They're a little inconsistent but either taking NavParams like our main nav method
```
navigate(paramsOrUrl: NavigationParams | string)
```
or at least taking the query params as a Params object like getUrlTree (not sure what the type of UrlSegment is - if it's an alias for string, even better - if it needs to be constructed, string[] is fine)
```
getUrlTree(urlSegments: UrlSegment[], queryParams?: Params)
```

arjunlalb: done.

projects/common/src/navigation/navigation.service.test.ts
```diff
@@ -50,7 +50,8 @@ describe('Navigation Service', () => {
     providers: [
       mockProvider(Location),
       mockProvider(Title, { setTitle: jest.fn().mockReturnValue(undefined) }),
-      { provide: APP_TITLE, useValue: 'defaultAppTitle' }
+      { provide: APP_TITLE, useValue: 'defaultAppTitle' },
+      mockProvider(PlatformLocation, { href: 'https://test.traceable.ai' })
```
aaron-steinfeld, line 54:
```suggestion
      mockProvider(PlatformLocation, { href: 'https://test.hypertrace.orgi' })
```

arjunlalb: done.

aaron-steinfeld (approved):

PR #1813, opened 2022-09-09
feat(dynamic-component): add method to attach component to given view…

adisreyaj:
## Description
Add a new method for dynamic component creation. This new method takes in a view container and attaches the component to that container.

### Checklist:

### DIFF
--- a/projects/common/src/utilities/angular/dynamic-component.service.ts
+++ b/projects/common/src/utilities/angular/dynamic-component.service.ts
@@ -9,12 +9,9 @@ import {
   Type,
   ViewContainerRef
 } from '@angular/core';
-import { isNil } from 'lodash-es';
 
 @Injectable({ providedIn: 'root' })
 export class DynamicComponentService {
-  public constructor(private readonly componentFactoryResolver: ComponentFactoryResolver) {}
-
   public insertComponent<C>(
     container: Element,
     componentClass: Type<C>,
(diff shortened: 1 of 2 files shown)


projects/common/src/utilities/angular/dynamic-component.service.ts
```diff
@@ -29,4 +34,31 @@ export class DynamicComponentService {
...
+      }
+
+      return componentRef;
+    }
+
+    return undefined;
+  }
+}
+
+export interface DynamicComponentOptions<T> {
```
aaron-steinfeld, line 60: We should make this consistent with the existing one and generally spell things out. 
component -> componentClass
vcr -> viewContainerRef
props -> don't use props and object assignment, that's just asking for bugs (this doesn't trigger things like change detection, which is why we don't use inputs for dynamically rendered components historically). Use injector and providers.

adisreyaj: Updated.

projects/common/src/utilities/angular/dynamic-component.service.ts
```diff
@@ -29,4 +33,32 @@ export class DynamicComponentService {
...
   }
+
+  /**
+   * Dynamically create a component and insert it into the view container. Data and services can be provided to the dynamic
+   * component via `providers`
+   *
+   * @returns ComponentRef or undefined if the component could not be created
+   */
+  public attachComponentToViewContainer<T>(opts: DynamicComponentOptions<T>): ComponentRef<T> | undefined {
+    if (!isNil(opts.viewContainerRef) && !isNil(opts.componentClass)) {
```
aaron-steinfeld, line 44: Why would we allow either to be nil? Can we enforce this by types instead so we can avoid returning undefined? Looks like componentClass already is, so that check is redundant, but not sure why viewContainerRef would be optional.

projects/common/src/utilities/angular/dynamic-component.service.ts
```diff
@@ -29,4 +33,32 @@ export class DynamicComponentService {
...
+      const componentFactory = this.componentFactoryResolver.resolveComponentFactory(opts.componentClass);
```
aaron-steinfeld, line 45: This should come from the provided injector. See line 24 above. Different child injectors have different components available to them, so using the service level one would only allow access to components available to the root injector.

projects/common/src/utilities/angular/dynamic-component.service.ts
adisreyaj: Just bringing down the null check down to the lowest level so the consumer doesn't have to check.

Do you think its a good idea to let the consumer to first check and then call this method?

projects/common/src/utilities/angular/dynamic-component.service.ts
aaron-steinfeld: Actually, now that I think about it - think a view container ref provides a number of these things already (injector and element) so it's really just a more specialized form of the above API. We should look to see if we can make the API support both cases, but if it's easier to have two methods, we should at least look to consolidate their implementation.

adisreyaj: Got it. Updating.

projects/common/src/utilities/angular/dynamic-component.service.ts
aaron-steinfeld: The consumer has to check one or the other - if we fix the type, the compiler will force the consumer to check the input, but trust the output. If we leave the type and output optional, then the compiler will allow the consumer to trust the input, but check the output. My preference is for the former, it leaves the consumer less coupled with the impl details of the method - it doesn't need to care about what's required and what's not to get it to work, the method API tells is that.

arjunlalb (approved):

aaron-steinfeld (approved):

PR #1821, opened 2022-09-14
fix(nav config): support for full path matching

arjunlalb:
## Description
Our current nav item config skips navigation path segments beyond the first one.
This PR does two things
1. Being able to specify multi-segment paths to construct URLs for nav items.
2. Correctly identifies full path match URLs and the nav item to be highlighted if present. If not present, falls back to existing logic.

### Testing
Tested locally on existing pages as well as with a page conforming to full path match requirement.

### Checklist:


itssharmasandeep (dismissed):

projects/components/src/navigation/navigation.config.ts
```diff
@@ -10,6 +10,7 @@ export interface NavItemLinkConfig {
   iconSize?: IconSize;
   label: string;
   matchPaths: string[]; // For now, default path is index 0
+  isFullPathMatch?: boolean;
```
aaron-steinfeld, line 13: rather than a boolean that change the meaning of matchPaths, we should allow matchPaths to accept multi-segment paths. Right now, you can either give multiple options or a multisegment path but not both.

This also doesn't match the meaning of full matches from the router (where `a/b/c` would not match `a/b`, but here it would)

arjunlalb: @aaron-steinfeld matchPaths accept multi-segment paths, but when the routes for nav items are being constructed (as well as in active item identification), we are skipping everything except the first item.

My goal of introducing the boolean was to fix this behaviour without changing any existing definitions that may have relied upon the bug / previous logic itself.

I intended to do a follow up where all nav items would be audited and updated according to full path match requirements.

aaron-steinfeld: I think there's a misunderstanding (could be on my side, haven't been in this code in a long time). From my recollection, matchPaths represents various alternative paths that can match. The first is used for navigate, but others are used to allow multiple different alternatives to match (which is why, active item identification also only uses first).

So for example, imagine we had an "APIs" nav item, and we wanted both the routes "apis" (representing the api list) and "api/{id}" to light up that same nav item, which, upon click would navigate to apis list. That would be configured with `matchPaths: ["apis", "api"]`

Does that reconcile with the behavior/code you're seeing?

If yes, my suggestion for supporting more than one segment in a match would be changing the array to like `(string | string[])[]` - or could use an object rather than string array inside, might be clearer like `(string | {segments: string[]})[]`. 

Old routes would behave as is, new ones that want to match more than one segment can use `matchPaths:[{segments: [ "parent/first-child"]}]`

arjunlalb: @aaron-steinfeld what you summarized is indeed the behaviour, the catch here is that the same matchPath logic (of picking just the first one) is also used for creating the navigation target for the nav entries as well. This is where the problem is. Highlighting the parent based on matchPath requirement is fine, but ignoring rest of the segments in an array while constructing navigation path isn't ideal.

Let me take another look at this problem to see if there's a better solution in the direction you proposed. Thank you! :)

aaron-steinfeld: > catch here is that the same matchPath logic (of picking just the first one) is also used for creating the navigation target for the nav entries as well.  

That's correct, that's expected. I believe there's still a disconnect here. Let me try again, but may be easier to just chat since I don't think I'm explaining it well.

In the current impl, there's no concept of multi-segment routes. Each entry in the array is the entirety of another route that's expected to light up the left nav entry. Because we can only ultimately navigate to a single route when clicking the nav entry, the first one is used.

So back to the api example (not sure if these routes or behavior are currently accurate, so treat it as a hypothetical). We have a single left nav entry "APIs" but 2 routes. one is `/apis` that maps to a list of APIs. The other is `/api/{id}` that maps to the API details. Both of these should light up the same "APIs" left nav entry, but if it's clicked, we just want to navigate to "/apis" so we can use `matchPaths: ["apis", "api"]`.

If we had instead hosted the api detail page at `/apis/api/{id}`, we wouldn't need the second entry. The first one would be sufficient since it matches both. If we had instead used "/apis/list" and "/apis/detail/{id}" for example, we'd want to support multiple segments in each array entry. That's where I was suggesting something like `matchPaths: [["apis", "list"], ["apis", "details"]]` or `matchPaths: ["apis/list", "apis/details"]` or some other equivalent where we allow multiple segments in each array entry.

arjunlalb: ah. got it. thanks for explaining it. In the earlier comment, when you meant multi-segment, I equated it with multi-segments of matchPaths.

PR #1824, opened 2022-09-17
fix(modal container): support for custom styles

arjunlalb:
## Description
Support applying custom styles in modal container

### Testing
Tested manually. 

### Checklist:

### DIFF
--- a/projects/components/src/modal/modal.ts
+++ b/projects/components/src/modal/modal.ts
@@ -12,6 +12,7 @@ export interface ModalConfig<TData = unknown> {
   styles?: ModalStyles;
 }
 
+// TODO: Revisit this
 export interface ModalStyles {
   'background-color'?: Color;
 }
--- a/projects/components/src/modal/modal.ts
+++ b/projects/components/src/modal/modal.ts
@@ -1,4 +1,5 @@
 import { InjectionToken, TemplateRef, Type } from '@angular/core';
+import { Dictionary } from '@hypertrace/common';
 import { Observable } from 'rxjs';
 
 export interface ModalConfig<TData = unknown> {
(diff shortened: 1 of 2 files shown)


projects/components/src/modal/modal.ts
```diff
@@ -8,6 +9,7 @@ export interface ModalConfig<TData = unknown> {
   title?: string;
   data?: TData;
   closeOnEscapeKey?: boolean;
+  styles?: Dictionary<unknown>;
```
anandtiwary, line 12: allowing arbitrary styles here could be bad. Can we just support only those customization that we need? Let's quickly discuss this on monday.

arjunlalb: @anandtiwary fixed it by defining an interface for style.

projects/components/src/modal/modal.ts
```diff
@@ -9,7 +9,11 @@ export interface ModalConfig<TData = unknown> {
   title?: string;
   data?: TData;
   closeOnEscapeKey?: boolean;
-  styles?: Dictionary<unknown>;
+  styles?: ModalStyles;
+}
+
+export interface ModalStyles {
```
anandtiwary, line 15: I still have reservations about this. Could you please add a todo here? We should come back and discuss this before adding more styles

anandtiwary (approved):

projects/components/src/modal/modal.ts
anandtiwary: approving for now.

arjunlalb: @anandtiwary added a todo. please reapprove :)

anandtiwary (approved):

PR #1825, opened 2022-09-18
fix(stepper): update stepper styles

arjunlalb:
## Description
Update stepper component styles to 
1. Position the action buttons absolutely.
2. Make content itself scrollable


projects/components/src/stepper/stepper.component.scss
```diff
@@ -25,6 +25,19 @@
         color: white;
       }
     }
+
+    .mat-stepper-horizontal {
```
anandtiwary, line 29: would this break styling for any existing stepper usage?

anandtiwary: cc @23nobody

anandtiwary (approved):
LGTM

projects/components/src/stepper/stepper.component.scss
arjunlalb: @anandtiwary I tested on existing stepper usage and it worked fine. Will wait for @adisreyaj and @23nobody to review.

23nobody (approved):

adisreyaj (approved):

PR #1832, opened 2022-09-21
feat: adding multiple ways to highlight a code line

itssharmasandeep:
## Description
Adding multiple ways to highlight a code line
- String (Already Present)
- String Array (New)
- RegExp (New)

### Testing
Local testing is done and added tests for newer support

### Checklist:

### DIFF
--- a/projects/components/src/viewer/code-viewer/code-viewer.component.ts
+++ b/projects/components/src/viewer/code-viewer/code-viewer.component.ts
@@ -157,9 +157,15 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy 
     }
 
     if (changes.code || changes.highlightText) {
+      /**
+       * `isEmpty(new RexExp('anything'))` returns true, so here we need to make another check for RegExp
+       */
+      const emptyValue =
+        this.highlightText instanceof RegExp ? isEmpty(this.highlightText.source) : isEmpty(this.highlightText);
+
       this.lineHighlights = new Array(this.codeLines.length)
         .fill(false)
-        .map((_, index) => this.isLineHighlighted(index));
+        .map((_, index) => (emptyValue ? false : this.isLineHighlighted(index)));
     }
   }
 
@@ -206,15 +212,6 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy 
   }
 
   private isLineHighlighted(lineNum: number): boolean {
-    /**
-     * `isEmpty(new RexExp('anything'))` returns true, so here we need to make another check for RegExp
-     */
-    const emptyValue =
-      this.highlightText instanceof RegExp ? isEmpty(this.highlightText.source) : isEmpty(this.highlightText);
-    if (emptyValue) {
-      return false;
-    }
-
     if (typeof this.highlightText === 'string') {
       return this.codeTexts[lineNum].toLowerCase().includes(this.highlightText.toLowerCase());
     }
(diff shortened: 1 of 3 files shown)


arjunlalb (dismissed):

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -204,6 +205,31 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
     });
   }
 
+  private isLineHighlighted(lineNum: number): boolean {
+    /**
+     * `isEmpty(new RexExp('anything'))` returns true, so here we need to make another check for RegExp
+     */
+    const emptyValue =
```
anandtiwary, line 212: can we keep out some checks outside this method? the highlightText is only changing once so we don't need to check its type for every line.

we could also avoid lowercase conversation multiple times for the same line.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -204,6 +205,31 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
+
+    if (Array.isArray(this.highlightText)) {
+      return this.highlightText.some(text => this.codeTexts[lineNum].toLowerCase().includes(text.toLowerCase()));
+    }
+
+    if (this.highlightText instanceof RegExp) {
+      return (this.codeTexts[lineNum].match(this.highlightText) ?? []).length > 0;
+    }
+
+    throw new Error('Invalid type for highlight text, only string, string[] and RegExp is allowed');
```
anandtiwary, line 230: Would it not break the ui? we should rather fail it gracefully

itssharmasandeep: I don't think so, and it there for the dev people, as they can see the error if they pass wrong values in here

anandtiwary: This is dependent on them doing the manual testing and verifying it. This could be easily miss and cause a regression imo. The problem is this would likely break the entire page/angular if this happens.  I would suggest that you remove the throw Error.

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -204,6 +211,22 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
     });
   }
 
+  private isLineHighlighted(lineNum: number): boolean {
+    if (typeof this.highlightText === 'string') {
```
anandtiwary, line 215: Could you also do this?
```
const lowercaseCodeText  = this.codeTexts[lineNum].toLowerCase();
```

itssharmasandeep: Updated the function signature and passed it from the parent of this function

projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -211,17 +211,17 @@ export class CodeViewerComponent implements AfterViewInit, OnChanges, OnDestroy
...
     }
 
     if (Array.isArray(this.highlightText)) {
-      return this.highlightText.some(text => this.codeTexts[lineNum].toLowerCase().includes(text.toLowerCase()));
+      return this.highlightText.some(text => codeText.includes(text.toLowerCase()));
     }
 
     if (this.highlightText instanceof RegExp) {
-      return (this.codeTexts[lineNum].match(this.highlightText) ?? []).length > 0;
+      return (codeText.match(this.highlightText) ?? []).length > 0;
```
anandtiwary, line 224: qq.. You were not using lowercase here before. Would this change cause any issues?

itssharmasandeep: Nope

itssharmasandeep: Since regex will come from parent, it shouldn't matter

anandtiwary (approved):

PR #1835, opened 2022-09-22
fix(stepper): next disabled fix

adisreyaj:
## Description
When using step control with the stepper, the `Next/Submit` buttons were not getting `disabled/enabled` until user actually hovers over the component (Change detection issue). This PR adds a fix to the issue, by switching to an observable.

### Checklist:

### DIFF
--- a/projects/components/src/stepper/stepper.component.ts
+++ b/projects/components/src/stepper/stepper.component.ts
@@ -11,9 +11,9 @@ import {
   ViewChild
 } from '@angular/core';
 import { MatStepper } from '@angular/material/stepper';
-import { queryListAndChanges$ } from '@hypertrace/common';
-import { Observable } from 'rxjs';
-import { map } from 'rxjs/operators';
+import { queryListAndChanges$, SubscriptionLifecycle } from '@hypertrace/common';
+import { BehaviorSubject, merge, Observable, of } from 'rxjs';
+import { debounceTime, map, startWith, switchMap } from 'rxjs/operators';
 import { ButtonRole, ButtonStyle } from '../button/button';
 import { IconSize } from '../icon/icon-size';
 import { StepperTabComponent } from './stepper-tab.component';
@@ -29,7 +29,13 @@ import { StepperTabComponent } from './stepper-tab.component';
         (selectionChange)="this.selectionChange.emit($event)"
       >
         <ng-container *ngFor="let step of steps">
-          <mat-step [completed]="step.completed" [stepControl]="step.stepControl" [optional]="step.optional">
+          <-- NOTE: completed and step control both should not be present when in linear flow. So when step control is
+          provided, we explicitly set completed to false so that the stepper can work with teh form validity status -->
+          <mat-step
+            [completed]="step.stepControl ? false : step.completed"
+            [stepControl]="step.stepControl"
+            [optional]="step.optional"
+          >
             <ng-template matStepLabel>
               <ht-label class="header-label" [label]="step.label"></ht-label>
             </ng-template>
@@ -57,23 +63,16 @@ import { StepperTabComponent } from './stepper-tab.component';
           ></ht-button>
           <ht-button
             class="next"
-            *ngIf="stepper.selectedIndex !== steps.length - 1"
             [label]="this.getActionButtonLabel | htMemoize: stepper.selectedIndex:steps"
-            (click)="stepper.next()"
-            [disabled]="this.isNextDisabled(stepper)"
-          ></ht-button>
-          <ht-button
-            class="submit"
-            *ngIf="stepper.selectedIndex === steps.length - 1"
-            [label]="this.getActionButtonLabel | htMemoize: stepper.selectedIndex:steps"
-            (click)="this.submitted.emit()"
-            [disabled]="this.isSubmitDisabled(stepper)"
+            (click)="this.nextOrSubmit(stepper)"
+            [disabled]="this.isNextDisabled$ | async"
           ></ht-button>
         </div>
       </div>
     </div>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
+  providers: [SubscriptionLifecycle],
   styleUrls: ['./stepper.component.scss']
 })
 export class StepperComponent implements AfterContentInit {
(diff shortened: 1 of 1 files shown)


projects/components/src/stepper/stepper.component.ts
```diff
@@ -138,14 +167,32 @@ export class StepperComponent implements AfterContentInit {
...
+    return step.stepControl?.valid ?? step.completed;
+  }
 
-    return step?.completed ?? false;
+  /**
+   * Check if all the steps are valid.
+   * If the step has a step control , we check the forms validity and if not, we check the step completed property.
+   */
+  private areAllStepsValid(): boolean {
+    return this.steps.toArray().every(step => step.stepControl?.valid ?? step.completed);
```
arjunlalb, line 195: you can use the above defined `isStepCompleted` method here.

projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +105,54 @@ export class StepperComponent implements AfterContentInit {
...
+    const sub = merge(
+      this.selectionChange.pipe(map(({ selectedStep }) => selectedStep)),
+      this.steps$.pipe(map(([firstStep]) => firstStep))
+    )
+      .pipe(
+        switchMap(step => {
+          if (step.stepControl) {
+            return step.stepControl.statusChanges.pipe(
+              startWith(step.stepControl.status),
+              map(status => status === 'VALID')
```
arjunlalb, line 140: define a meaningful const value for this string.

arjunlalb: also, is there a value we can look for instead of doing a string comparison?

arjunlalb (commented):
Left some minor comments. Approach looks alright to me.

projects/components/src/stepper/stepper.component.ts
```diff
@@ -83,9 +82,11 @@ export class StepperComponent implements AfterContentInit {
   private readonly steps!: QueryList<StepperTabComponent>;
 
   @ViewChild(MatStepper)
-  private readonly stepper?: MatStepper;
+  public readonly stepper?: MatStepper;
```
anandtiwary, line 85: Why make it public?

projects/components/src/stepper/stepper.component.ts
adisreyaj:
```
type FormControlStatus = 'VALID' | 'INVALID' | 'PENDING' | 'DISABLED';
```

This is what angular uses.

projects/components/src/stepper/stepper.component.ts
adisreyaj: Updated.

projects/components/src/stepper/stepper.component.ts
adisreyaj: Updated.

projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +105,54 @@ export class StepperComponent implements AfterContentInit {
...
+          if (step.stepControl) {
+            return step.stepControl.statusChanges.pipe(
+              startWith(step.stepControl.status),
+              map(status => status === 'VALID')
+            );
+          }
+
+          return of(step.completed);
+        }),
+        debounceTime(100) // <-- debounce to avoid multiple emits
```
arjunlalb, line 146: do we need this? if we are only looking for distinct values, we could use distinct or similar operators.

adisreyaj: Since we are using the `this.step` in the subscribe, in order to get the latest state, we need to fetch the state in the next tick. Hence added a debounce which will prevent multiple emits as well as tick, so we get the current state in subscribe.

adisreyaj:
Add more tests to support the use case with `completed` and `stepControl` inputs.

projects/components/src/stepper/stepper.component.ts
arjunlalb: is there a possibility the debounce will mask the last set of relevant changes within the 100ms interval? Debounce does not replay the new values later, it ignores anything within the 100ms period.

adisreyaj: image

It waits till that mentioned time and emits the last value. So I think we should be fine.

projects/components/src/stepper/stepper.component.ts
palbizu: should not be more useful to have an enum in this case?

projects/components/src/stepper/stepper.component.ts
palbizu: IMO looks good but why the 100ms, don't we have a default for this kind of case?

adisreyaj: I'm not sure. @arjunlalb ?

projects/components/src/stepper/stepper.component.ts
adisreyaj: Yeah totally. But I didn't want to define a separate enum for a framework-defined type as we have to keep parity with it in the long run.

projects/components/src/stepper/stepper.component.ts
arjunlalb: this is a case we haven't encountered before. so, there's no such established defaults. I'm also torn about this approach. I feel like there may be a better way to do this.

arjunlalb: @aaron-steinfeld any thoughts here?

aaron-steinfeld: discussed briefly on a chat - summary was basically to prefer using a function over a subject and subscription here to sidestep the issue. If function is called extra times, that's likely not going to be a problem perf wise even if not ideal (and likely able to be memoized away).

As long as there's no human-discernible perf impact, we typically should value readability over the number of clock cycles.

adisreyaj: @aaron-steinfeld Tried with methods instead. But the problem is that the methods are not called unless there is a change that is detected. The only right trigger here is the `statusChanges` for a form that changes when the form validity changes and we should be using that trigger to re-evaluate the `disabled` status.

aaron-steinfeld: updating with conversation from slack. Sounds like the issue is the caller is passing in a control object, mutating it and not updating the ref, and we're expecting UI updates.

> so I think there’s a few ways to approach it.
>
> Ideally we change callers to pass in a new object when there’s a change. That’s the whole point of onPush, so we’re kind of misusing it there. Not sure how difficult that is given the form integration.
>
> The other option that I think would still be significantly simpler would just add some explicit change detection - for example, in stepper tab you could maintain a subscription like
> `stepControl?.statusChanges.subscribe(() => this.cdr.markForCheck()`
> would just need to make sure to manage the subscription on the way in (in onChanges) and either subs lifecycle or onDestroy.

adisreyaj:
Closing in favor or https://github.com/hypertrace/hypertrace-ui/pull/1857

projects/components/src/stepper/stepper.component.ts
adisreyaj: https://github.com/hypertrace/hypertrace-ui/pull/1835#issuecomment-1270027689

PR #1836, opened 2022-09-22
File download service

itssharmasandeep:
## Description
Basic implementation of the generic (all in one) file download service. It currently has
- Download as text
- Download as CSV

It also contains the refactoring for the download file component

`Sample`

https://user-images.githubusercontent.com/78212931/191737350-3004d542-6c35-427c-88d7-4c984871a9b0.mov

### Testing
Local testing is done and text cases added for both the functions

### Checklist:


projects/components/src/download-file/download-file.component.ts
```diff
@@ -1,67 +1,41 @@
...
       <ht-button
-        *ngIf="!this.dataLoading"
+        *ngIf="!(this.dataLoading$ | async)"
         class="download-button"
         icon="${IconType.Download}"
         display="${ButtonStyle.Text}"
         size="${ButtonSize.Large}"
       ></ht-button>
-      <ht-icon *ngIf="this.dataLoading" icon="${IconType.Loading}" size="${IconSize.Large}"></ht-icon>
+      <ht-icon *ngIf="this.dataLoading$ | async" icon="${IconType.Loading}" size="${IconSize.Large}"></ht-icon>
```
anandtiwary, line 24: oh i see. We are hiding the button and showing the icon.

projects/components/src/download-file/download-file.component.ts
```diff
@@ -1,67 +1,41 @@
...
-  private downloadData(data: string): void {
-    this.renderer.setAttribute(
-      this.dlJsonAnchorElement,
-      'href',
-      `data:text/plain;charset=utf-8,${encodeURIComponent(data)}`
-    );
-    this.renderer.setAttribute(this.dlJsonAnchorElement, 'download', this.metadata!.fileName);
-    this.renderer.setAttribute(this.dlJsonAnchorElement, 'display', 'none');
-    this.dlJsonAnchorElement.click();
+    this.fileDownloadService.downloadAsText(this.metadata!);
```
anandtiwary, line 39: Shouldn't be using `!`. you could pass in the defined `metadata` from the template. That way it will always be defined

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    this.download({ ...config, dataSource: csvData$ }, csvData => {
+      const csvFile = new Blob([csvData], { type: 'text/csv' });
+
+      return window.URL.createObjectURL(csvFile);
+    });
+  }
+
+  public ngOnDestroy(): void {
+    this.downloadElement.remove();
```
anandtiwary, line 67: So, would it cancel the download if we moved out of the page?

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -0,0 +1,130 @@
+import { Injectable, OnDestroy } from '@angular/core';
+import { Dictionary } from '@hypertrace/common';
+import { isEmpty, startCase } from 'lodash-es';
+import { BehaviorSubject, combineLatest, Observable, of, Subscription } from 'rxjs';
+import { catchError, map, take } from 'rxjs/operators';
+import { NotificationService } from '../../notification/notification.service';
+
+@Injectable()
+export class FileDownloadService implements OnDestroy {
```
anandtiwary, line 9: will go through it post demo

projects/components/src/download-file/service/file-download.service.ts
itssharmasandeep: - If this happens before `downloadElement.click()`, then Yes as the flow is not completed. (Which I think is okay)
- After that(on browser download), I don't think so, I will check.

projects/components/src/download-file/download-file.component.ts
```diff
@@ -1,67 +1,41 @@
...
 @Component({
   selector: 'ht-download-file',
   changeDetection: ChangeDetectionStrategy.OnPush,
   styleUrls: ['./download-file.component.scss'],
+  providers: [FileDownloadService],
   template: `
     <div *ngIf="this.metadata" class="download-file" (click)="this.triggerDownload()">
       <ht-button
-        *ngIf="!this.dataLoading"
+        *ngIf="!(this.dataLoading$ | async)"
```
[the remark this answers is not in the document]
palbizu: +1

projects/components/src/download-file/download-file.component.ts
palbizu: I agree but my question here is doing the assignation in the constructor (just to learn)

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -0,0 +1,130 @@
...
+   * @param config Csv download config
+   */
+  public downloadAsCsv(config: CsvDownloadFileConfig): void {
+    // If given header is empty, then create header from the data keys
+    const header$ = isEmpty(config.header)
+      ? config.dataSource.pipe(
+          map(data => Object.keys(data[0] ?? [])),
+          map(keys => keys.map(startCase))
+        )
+      : of(config.header!);
```
palbizu, line 43: Does it this assertion necessary?

itssharmasandeep: Actually, I have already checked for this at L38, since `isEmpty(config.header)` doesn't give the correct type back, so yeah we need it here

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    // Value replacer for null and undefined values
+    const replacer = (_: string, value: string) => value ?? '-';
```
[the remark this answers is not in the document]
itssharmasandeep: So if you see at L50, where we're using this function. The callback has two attributes (key, value), since we do not need a key here, I used underscore since it doesn't give a `not used` error.
It is actually a pattern that we use. We could have used `_key` as well

projects/components/src/download-file/download-file.component.ts
itssharmasandeep: > my question here is doing the assignation in the constructor (just to learn)

`metadata` is an Input field so we can't assign anything in the constructor. It will be `undefined`

palbizu: So you can use OnChanges or in the worst case on OnInit

itssharmasandeep: Yeah, we can. As we're already doing ngIf in the template, Anand's method will also work. We can choose any of these

projects/components/src/download-file/download-file.component.ts
```diff
@@ -1,67 +1,41 @@
...
+  public constructor(private readonly fileDownloadService: FileDownloadService) {}
 
-  public constructor(
-    @Inject(DOCUMENT) private readonly document: Document,
-    private readonly renderer: Renderer2,
-    private readonly changeDetector: ChangeDetectorRef,
-    private readonly notificationService: NotificationService
-  ) {
-    this.dlJsonAnchorElement = this.document.createElement('a');
+  public get dataLoading$(): Observable<boolean> {
```
anandtiwary, line 34: Move it to constructor please

itssharmasandeep: Not needed, did refactoring after offline discussion

projects/components/src/download-file/download-file.component.ts
```diff
@@ -1,41 +1,43 @@
...
-  }
+  public constructor(
+    private readonly cdr: ChangeDetectorRef,
+    private readonly fileDownloadService: FileDownloadService
+  ) {}
 
   public triggerDownload(): void {
-    this.fileDownloadService.downloadAsText(this.metadata!);
+    this.dataLoading = true;
+    this.fileDownloadService.downloadAsText(this.metadata!).subscribe(() => {
```
anandtiwary, line 38: avoid `!`. If there is an if check on the template then please pass `metadata` as an argument

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -82,38 +76,47 @@ export class FileDownloadService implements OnDestroy {
...
+      map(data => {
+        try {
+          this.downloadElement.href = getHref(data);
+          this.downloadElement.download = config.fileName;
+          this.downloadElement.click();
+
+          const successEvent: FileDownloadSuccessEvent = { type: FileDownloadEventType.Success };
+
+          return successEvent;
+        } catch (error) {
```
anandtiwary, line 91: Is a try-catch needed here?

projects/components/src/download-file/download-file.component.test.ts
```diff
@@ -19,36 +21,32 @@ describe('Download File Component', () => {
...
+    // Before data loading
+    expect(spectator.query(ButtonComponent)).toExist();
+    expect(spectator.query(IconComponent)).not.toExist();
+
+    // While data loading
+    spectator.click('.download-file');
     expect(spectator.query(ButtonComponent)).not.toExist();
     expect(spectator.query(IconComponent)).toExist();
-  });
+    spectator.tick(2500);
```
anandtiwary, line 44: Can we keep a smaller tick? We can reduce the timer as well. This will delay the tests unneccessarily

projects/components/src/download-file/service/file-download.service.ts
anandtiwary: we are rethrowing the same error

projects/components/src/download-file/download-file.component.test.ts
aaron-steinfeld: Didn't review the PR, but just want to point out this isn't a sleep, it's fake async - won't change test runtime.

anandtiwary: Okay

projects/components/src/download-file/service/file-download.service.ts
itssharmasandeep: Yeah, actually it doesn't. Thanks

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -57,11 +56,6 @@ export class FileDownloadService implements OnDestroy {
     });
   }
 
-  public ngOnDestroy(): void {
```
anandtiwary: why are we removing this method?

anandtiwary (approved):

PR #1840, opened 2022-09-23
feat: popover zone component

itssharmasandeep:
## Description
This adds a new component called `PopoverZoneComponent`.
- It is a content holder as well as gives the functionality of showing a custom popover on mouse enter and closing the same on mouse leave.
- This has the functionality of having an actionable popover, where we can interact with the popover (click, hover, etc.).
- Popover gets closed automatically after the mouse leaves from popover content as well.

### Testing
Local testing is done.

### Checklist:

### Documentation
Example

```html
<div class="test-div">
  <ht-popover-zone class="popover-zone" [popoverContent]="popoverContent">
    <div class="hover-div">Hover me</div>
   </ht-popover-zone>
</div>

<ng-template #popoverContent>
  <div class="popover-content">
    <a href="https://hypertrace.org/">Go to hypertrace</a>
  </div>
</ng-template>
```

Output (After some styling)

https://user-images.githubusercontent.com/78212931/191997467-e045250f-97ae-43a8-9dea-9e7e66a33196.mov


aaron-steinfeld (commented):
This sounds almost identical to the existing popover component which has a trigger (aka zone) and content child.

Can you explain the differences in behavior and why a new component is required to achieve them?

anandtiwary:
+1

> Popover gets closed automatically after the mouse leaves from popover content as well.

Could we add this behavior as a config option to existing trigger component?

aaron-steinfeld:
> 1. The current popover works on click and the configurations added (in terms of input and output) are specifically for on-click which doesn't make any sense in the case of hover (which is what this new component is all about).
> 2. Yeah, it is the same in terms of the content and trigger(zone), but fundamental working is different as it works on hover.
> 3. I tried implementing the same using the existing component, but I found these issues
>    
>    * The properties have mismatches and some properties don't make any sense to each other.
>    * It would require a considerable amount of refactoring to make new API work (which may lead to possible breaking changes, which I am trying to avoid right now)


Got it, but I'm not sure if we're looking at the same thing. If the APIs do diverge significantly, then a separate component would make sense. Here though, none of the inputs and outputs I see are related to click.`closeOnClick` kind of sounds like it, but it relates to whether clicking an open popover should close it. Similarly closeOnNavigate, disabled, locationPrefs are all generally applicable, as are the two outputs emitting when a popover opens and closes. There's some impl code assuming a click, but that's easily modified. 

Two ways I could see to integrate hover behavior in where there is click today 
  - add an input on the trigger element (defaulting to click) that indicates the event it's listening for
  - add a second child type like `PopoverHoverTrigger` and look for one or the other (or better yet, use abstraction and move the listener code into the child trigger component).

You could probably mix and match some bits of this too, like adding an input on the child _and_ moving the listener code to the child.

Does that make sense? Or am I misunderstanding

itssharmasandeep:
Updated the PR with using existing component
cc: @aaron-steinfeld @anandtiwary

projects/components/src/popover/popover.component.ts
```diff
@@ -62,6 +69,30 @@ export class PopoverComponent implements OnDestroy {
...
   public onClick(): void {
+    if (this.popoverTriggerType === PopoverTriggerType.Click) {
+      this.handleClickPopoverTrigger();
+    }
+  }
+
+  @HostListener('mouseenter')
+  public onMouseenter(): void {
+    if (this.popoverTriggerType === PopoverTriggerType.Hover) {
+      this.handleHoverPopoverTrigger('mouseenter');
```
aaron-steinfeld, line 80: Please don't hardcode a method param to get completely separate behavior. Suggest just making two methods - handlePopoverOpen, handlePopoverClose. Ideally, should make the click method delegate as needed.

projects/components/src/popover/popover.component.ts
```diff
@@ -90,4 +121,29 @@ export class PopoverComponent implements OnDestroy {
...
   }
+
+  private handleHoverPopoverTrigger(event: 'mouseenter' | 'mouseleave'): void {
+    if (this.disabled) {
+      return;
+    }
+
+    switch (event) {
+      case 'mouseenter':
+        this.popoverHoverTriggerService.showPopover({
```
aaron-steinfeld, line 132: Clicking on a trigger and hovering a trigger shouldn't go down different code paths. At least the bulk should be shared, could have different listeners etc.

projects/components/src/popover/popover.component.ts
```diff
@@ -90,4 +121,29 @@ export class PopoverComponent implements OnDestroy {
...
+          }
+        });
+        break;
+
+      case 'mouseleave':
+        this.popoverHoverTriggerService.closePopover();
+        break;
+
+      default:
+        assertUnreachable(event);
```
aaron-steinfeld, line 146: typically want to have switch cases return and return this one too. It's moot given the comment about getting rid of the switch though.

projects/components/src/popover/popover.component.ts
aaron-steinfeld: In fact, it looks like this path is just doing its own thing and ignoring all the inputs and outputs.

anandtiwary:
Some of this code is already in the tooltip. Let's discuss more about this in our sync.

PR #1848, opened 2022-09-28
feat: add topology specification builder

itssharmasandeep:
## Description
Add topology specification builder

### Testing
Local testing is done

### Checklist:

### DIFF
--- a/projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
+++ b/projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
@@ -1,3 +1,4 @@
+import { Dictionary } from '@hypertrace/common';
 import { GraphQlSelection } from '@hypertrace/graphql-client';
 import { Specification } from '../../../../../model/schema/specifier/specification';
 import { GraphQlObservabilityArgumentBuilder } from '../../../../builders/argument/graphql-observability-argument-builder';
@@ -45,6 +46,12 @@ export interface NeighborSpecificationBuildConfig {
   edgeDirection: TopologyEdgeDirection;
   nodeSpecifications: Specification[];
   neighborType: string;
+}
+
+export interface TopologyNeighborSpecResult {
+  results: {
+    neighbor: Dictionary<unknown>;
+  };
 }
 
 export const enum TopologyEdgeDirection {
--- a/projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
+++ b/projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
@@ -48,9 +48,9 @@ export interface NeighborSpecificationBuildConfig {
   neighborType: string;
 }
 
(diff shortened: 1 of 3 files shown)


projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
```diff
@@ -0,0 +1,60 @@
...
+              }
+            ]
+          }
+        ]
+      }
+    ];
+  }
+}
+
+export interface NeighborSpecificationBuildConfig {
```
anandtiwary, line 45: Should we rename it to NodeSpecificationBuildConfig?

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
```diff
@@ -0,0 +1,60 @@
...
+  edgeDirection: TopologyEdgeDirection;
+  nodeSpecifications: Specification[];
+  neighborType: string;
+}
+
+export interface TopologyNeighborSpecResult {
+  results: {
+    neighbor: Dictionary<unknown>;
```
anandtiwary, line 53: Or we could type the result with 'T' (generics)

anandtiwary:
Also, please don't forget to write tests :)

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
itssharmasandeep: It is used for incoming edges and outgoing edges.
It will look like this.

```ts
incomingEdges_SERVICE: incomingEdges(neighborType: SERVICE) {
        results {
          neighbor {
            id
            name: attribute(expression: { key: "name" })
          }
        }
      }
```

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
itssharmasandeep: Not really, neighbor is like an entity query. where the value truly unknown

itssharmasandeep:
> Also, please don't forget to write tests :)

Added :)

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/topology-specification-builder.ts
anandtiwary: Could we try this?  
```
TopologyNeighborSpecResult<T = unknown> {
results: {
neighbor: Dictionary<T>
...
}
```

anandtiwary: This way, if you know the type, you could just pass it from the caller's side

itssharmasandeep: Done

anandtiwary (approved):

PR #1857, opened 2022-10-06
Fix/stepper validation

adisreyaj:
## Description
When using step control with the stepper, the Next/Submit buttons were not getting disabled/enabled until user actually hovers over the component (Change detection issue). This PR adds a fix to the issue, by switching to an observable.

Continued from https://github.com/hypertrace/hypertrace-ui/pull/1835
Based on the discussion. Implemented methods with explicit change detection checks.

### DIFF
--- a/projects/components/src/stepper/stepper.component.ts
+++ b/projects/components/src/stepper/stepper.component.ts
@@ -130,12 +130,13 @@ export class StepperComponent implements AfterContentInit {
       return true;
     }
 
-    const isValid = currentTab.stepControl ? currentTab.stepControl?.status === 'VALID' : currentTab?.completed;
-    const isLastStep = this.stepper ? this.stepper?.selectedIndex === this.stepper?.steps.length - 1 : true;
-    const isNextDisabled = this.isLinear && !isValid;
-    const isSubmitDisabled = !this.areAllStepsValid();
+    const isValid = currentTab.stepControl ? currentTab.stepControl.status === 'VALID' : currentTab.completed;
+    const isLastStep = this.stepper ? this.isLastStep(this.stepper) : true;
+    if (isLastStep) {
+      return !this.areAllStepsValid();
+    }
 
-    return isLastStep ? isSubmitDisabled : isNextDisabled;
+    return this.isLinear && !isValid;
   }
 
   /**
@@ -151,8 +152,12 @@ export class StepperComponent implements AfterContentInit {
   }
 
   public nextOrSubmit(stepper: MatStepper): void {
-    const isLastStep = stepper.selectedIndex === stepper.steps.length - 1;
-    isLastStep ? this.submitted.emit() : stepper.next();
+    const isLastStep = this.isLastStep(stepper);
+    if (isLastStep) {
+      this.submitted.emit();
+    } else {
+      stepper.next();
+    }
   }
 
   public getActionButtonLabel(selectedIndex: number, steps: StepperTabComponent[]): string {
@@ -176,6 +181,10 @@ export class StepperComponent implements AfterContentInit {
    */
   private areAllStepsValid(): boolean {
     return this.steps.toArray().every((_, index) => this.isStepCompleted(index));
+  }
+
+  private isLastStep(stepper: MatStepper): boolean {
+    return stepper.selectedIndex === stepper.steps.length - 1;
   }
 }
 
--- a/projects/components/src/stepper/stepper.component.ts
+++ b/projects/components/src/stepper/stepper.component.ts
@@ -2,6 +2,7 @@ import { StepperSelectionEvent } from '@angular/cdk/stepper';
 import {
   AfterContentInit,
   ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   ContentChildren,
   EventEmitter,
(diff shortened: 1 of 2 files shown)


projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +107,35 @@ export class StepperComponent implements AfterContentInit {
 
   public steps$?: Observable<StepperTabComponent[]>;
 
+  public constructor(private readonly cdr: ChangeDetectorRef, private readonly subs: SubscriptionLifecycle) {}
+
   public ngAfterContentInit(): void {
     this.steps$ = queryListAndChanges$(this.steps).pipe(map(list => list.toArray()));
+    this.subs.add(
+      merge(
+        this.selectionChange.pipe(map(({ selectedStep }) => selectedStep)),
```
aaron-steinfeld, line 116: This part looks fine (although more complicated than I was intending! I was just thinking each step component can watch itself so you don't need to do the change merging here).

projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +107,35 @@ export class StepperComponent implements AfterContentInit {
...
-  public isNextDisabled(stepper: MatStepper): boolean {
-    const selectedStep = stepper.steps.get(stepper.selectedIndex);
-    const formValid = selectedStep?.stepControl?.valid;
-    const stepCompleted = selectedStep?.completed;
-
-    return this.isLinear && !stepCompleted && !formValid;
-  }
+  public isNextDisabled(stepIndex: number): boolean {
+    const currentTab = this.steps?.get(stepIndex);
+    if (!currentTab) {
```
aaron-steinfeld, line 129: Expected? Or just a base case?

projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +107,35 @@ export class StepperComponent implements AfterContentInit {
...
+      return true;
+    }
 
-  // TODO: can be made a directive
-  public isSubmitDisabled(stepper: MatStepper): boolean {
-    const selectedStep = stepper.steps.get(stepper.selectedIndex);
-    const formValid = selectedStep?.stepControl?.valid;
-    const stepCompleted = selectedStep?.completed;
+    const isValid = currentTab.stepControl ? currentTab.stepControl?.status === 'VALID' : currentTab?.completed;
```
aaron-steinfeld, line 133: Looks like we already know currentTab is defined (and stepControl in the true branch)
```suggestion
    const isValid = currentTab.stepControl ? currentTab.stepControl.status === 'VALID' : currentTab.completed;
```

projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +107,35 @@ export class StepperComponent implements AfterContentInit {
...
+    const isLastStep = this.stepper ? this.stepper?.selectedIndex === this.stepper?.steps.length - 1 : true;
```
aaron-steinfeld, line 134: similar extra optional access here.

projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +107,35 @@ export class StepperComponent implements AfterContentInit {
...
+    const isNextDisabled = this.isLinear && !isValid;
+    const isSubmitDisabled = !this.areAllStepsValid();
 
-    return !stepCompleted && !formValid;
+    return isLastStep ? isSubmitDisabled : isNextDisabled;
```
aaron-steinfeld, line 138: nit: so we do this work calculating everything, then only use one or the other. Can move the conditional first.

projects/components/src/stepper/stepper.component.ts
```diff
@@ -138,14 +150,32 @@ export class StepperComponent implements AfterContentInit {
     }
   }
 
+  public nextOrSubmit(stepper: MatStepper): void {
+    const isLastStep = stepper.selectedIndex === stepper.steps.length - 1;
+    isLastStep ? this.submitted.emit() : stepper.next();
```
aaron-steinfeld, line 155: avoid ternary with side effect. Use a standard if here.

aaron-steinfeld (commented):
Seems like there's a bunch of issues being addressed here, mostly due to having the stepper support two modes.

projects/components/src/stepper/stepper.component.ts
```diff
@@ -104,26 +107,36 @@ export class StepperComponent implements AfterContentInit {
...
+    if (!currentTab) {
+      return true;
+    }
 
-  // TODO: can be made a directive
-  public isSubmitDisabled(stepper: MatStepper): boolean {
-    const selectedStep = stepper.steps.get(stepper.selectedIndex);
-    const formValid = selectedStep?.stepControl?.valid;
-    const stepCompleted = selectedStep?.completed;
+    const isValid = currentTab.stepControl ? currentTab.stepControl.status === 'VALID' : currentTab.completed;
```
aaron-steinfeld, line 133: nit: can move after, only used if not last step

projects/components/src/stepper/stepper.component.ts
```diff
@@ -138,14 +151,40 @@ export class StepperComponent implements AfterContentInit {
     }
   }
 
+  public nextOrSubmit(stepper: MatStepper): void {
+    const isLastStep = this.isLastStep(stepper);
+    if (isLastStep) {
```
aaron-steinfeld, line 156: nit
```suggestion
    if (this.isLastStep(stepper)) {
```

projects/components/src/stepper/stepper.component.ts
```diff
@@ -138,14 +151,40 @@ export class StepperComponent implements AfterContentInit {
...
+      this.submitted.emit();
+    } else {
+      stepper.next();
+    }
+  }
+
   public getActionButtonLabel(selectedIndex: number, steps: StepperTabComponent[]): string {
-    return steps[selectedIndex]?.actionButtonLabel ?? (selectedIndex === steps.length ? 'Submit' : 'Next');
+    const isLastStep = selectedIndex === steps.length - 1;
```
aaron-steinfeld, line 164: nit - could work with stepper instead and reuse isLastStep. Doesn't really even need args.

aaron-steinfeld (approved):

adisreyaj:
I'm gonna try and address the newer comments in a follow up possibly.

PR #1858, opened 2022-10-10
feat(multi-select): add config for select all

adisreyaj:
## Description
Make `Select All` configurable via the `showSelectAll` input. Enabled by default.

### DIFF
--- a/projects/components/src/multi-select/multi-select.component.test.ts
+++ b/projects/components/src/multi-select/multi-select.component.test.ts
@@ -629,4 +629,26 @@ describe('Multi Select Component', () => {
     expect(onChange).not.toHaveBeenCalled();
     flush();
   }));
+
+  test('should not show the select all button', fakeAsync(() => {
+    spectator = hostFactory(
+      `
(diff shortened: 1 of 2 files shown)


projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -629,4 +629,26 @@ describe('Multi Select Component', () => {
...
+          options: selectionOptions,
+          selected: [selectionOptions[1].value],
+          showSelectAll: false,
+          searchMode: MultiSelectSearchMode.CaseInsensitive,
+          triggerLabelDisplayMode: TriggerLabelDisplayMode.Selection
+        }
+      }
+    );
+
+    spectator.tick();
```
itssharmasandeep, line 651: I think we shouldn't need this and fakeAsync?

adisreyaj: Updated.

itssharmasandeep (approved):

Greedy-Geek:
Thanks for this one, had I not seen this one I would have raised another PR 😅

Greedy-Geek (approved):

PR #1868, opened 2022-10-13
fix: used cytoscape instead of d3-hierarchy

Greedy-Geek:
Trying to use Cytoscape library instead of d3-hierarchy to get positions of the nodes


Greedy-Geek:
Closed this PR because this was just for rough work (kind of experimenting), the fix is now in https://github.com/hypertrace/hypertrace-ui/pull/1875

PR #1870, opened 2022-10-17
feat(stepper): explicit disabling of stepper controls

adisreyaj:
Adds support for explicitly disabling stepper control buttons:
- Back
- Cancel
- Next
- Submit

**Use-case**:
One common use case, is to `disable` the **Submit** button when data is getting saved.

### DIFF
--- a/projects/components/src/stepper/stepper.component.ts
+++ b/projects/components/src/stepper/stepper.component.ts
@@ -52,6 +52,7 @@ import { StepperTabComponent } from './stepper-tab.component';
         <ht-button
           class="cancel"
           label="Cancel"
+          [disabled]="this.isCancelButtonDisabled()"
           (click)="this.cancelled.emit()"
           role="${ButtonRole.Tertiary}"
         ></ht-button>
@@ -61,6 +62,7 @@ import { StepperTabComponent } from './stepper-tab.component';
             *ngIf="stepper.selectedIndex !== 0"
             display="${ButtonStyle.Outlined}"
             label="Back"
+            [disabled]="this.isBackButtonDisabled()"
             (click)="stepper.previous()"
           ></ht-button>
           <ht-button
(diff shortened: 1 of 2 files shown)


projects/components/src/stepper/stepper.component.ts
```diff
@@ -186,10 +202,23 @@ export class StepperComponent implements AfterContentInit {
...
 export type StepperSelectionChange = StepperSelectionEvent;
+
+export enum StepperButton {
+  Cancel = 'cancel',
+  Back = 'back',
+  Next = 'next',
+  Submit = 'submit'
+}
+
+export type StepperButtonDisabledExplicitStatus = Partial<Record<StepperButton, boolean>>;
```
arjunlalb, line 224: `Explicit` is not required in the name.

23nobody:
Why do we want to be disabling `back` and `cancel`?

arjunlalb:
it’ll be better if we can bake this config onto the stepper tab component as individual controls so that change detection can be leveraged

adisreyaj:
> Why do we want to be disabling `back` and `cancel`?

Future proofing.

projects/components/src/stepper/stepper.component.ts
```diff
@@ -96,6 +98,9 @@ export class StepperComponent implements AfterContentInit {
   @Input()
   public orientation: StepperOrientation = StepperOrientation.Horizontal;
 
+  @Input()
+  public buttonDisabled: StepperButtonDisabledExplicitStatus = {};
```
Christian862, line 102: because this is not a boolean flag, it is like a map of boolean flags. I think it should be renamed, this name is misleading. Maybe `controlButtonsStatus`  ?

anandtiwary: Imo, we should flatten it. That way the api would be simpler and we don't need to always pass a new object if only the state for one button changes.

PR #1875, opened 2022-10-17
fix: used cytoscape to generate the graph nodes positions for application flow

Greedy-Geek:
Trying to use Cytoscape library instead of d3-hierarchy to get positions of the nodes

### DIFF
--- a/projects/observability/src/shared/components/topology/d3/layouts/custom-tree-layout.ts
+++ b/projects/observability/src/shared/components/topology/d3/layouts/custom-tree-layout.ts
@@ -1,33 +1,91 @@
-import { hierarchy, HierarchyNode } from 'd3-hierarchy';
-import { RenderableTopology, TopologyEdge, TopologyNode } from '../../topology';
-import { D3ProxyNode, TreeLayout } from './tree-layout';
+import cytoscape from 'cytoscape';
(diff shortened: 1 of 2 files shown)


aaron-steinfeld:
As discussed offline, before even testing cytoscape - how does it affect the bundle size? If we're adding multiple MBs of libs to switch layout providers, then this approach doesn't make a ton of sense and should be revisited.

projects/observability/src/shared/components/topology/d3/layouts/custom-tree-layout.ts
```diff
@@ -1,33 +1,91 @@
-import { hierarchy, HierarchyNode } from 'd3-hierarchy';
-import { RenderableTopology, TopologyEdge, TopologyNode } from '../../topology';
-import { D3ProxyNode, TreeLayout } from './tree-layout';
+import cytoscape from 'cytoscape';
+import { max } from 'lodash-es';
+import { RenderableTopology, TopologyCoordinates, TopologyEdge, TopologyNode } from '../../topology';
 
-export class CustomTreeLayout extends TreeLayout {
+export class CustomTreeLayout {
```
anandtiwary, line 5: Please create a separate layout file for cytoscape

PR #1876, opened 2022-10-18
fix: created new layout for graphs

Greedy-Geek:
A new algorithm to get positions of the graph nodes.


projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (node.incoming.length === 0) {
+        nodeToLevelMap.set(node, 0);
+        levelToNodesMap.has(0) ? levelToNodesMap.get(0)?.push(node) : levelToNodesMap.set(0, [node])
+      }
+    });
+  }
+
+  private fillNodeAndLevelMaps(levelToNodesMap: Map<number, RenderableTopologyNode[]>, nodeToLevelMap: Map<RenderableTopologyNode, number>) {
+
+    const nodes: RenderableTopologyNode[] = levelToNodesMap.get(0) ?? [];
```
Greedy-Geek, line 27: rootNodes

Greedy-Geek: done

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
+import { RenderableTopology, RenderableTopologyNode, TopologyCoordinates, TopologyEdge, TopologyNode } from "../../topology";
+
+export class GraphLayout {
+  public layout(topology: RenderableTopology<TopologyNode, TopologyEdge>): void {
+
+    const levelToNodesMap = new Map<number, RenderableTopologyNode[]>();
+    const nodeToLevelMap = new Map<RenderableTopologyNode, number>();
+    const nodeCoOrdinatesMap: Map<string, TopologyCoordinates> = new Map<string, TopologyCoordinates>();
```
itssharmasandeep, line 8: can these be on the class level instead of the function level? As I assume, there will be only one instance of that per layout class.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+        nodeToLevelMap.set(node, 0);
+        levelToNodesMap.has(0) ? levelToNodesMap.get(0)?.push(node) : levelToNodesMap.set(0, [node])
+      }
+    });
+  }
+
+  private fillNodeAndLevelMaps(levelToNodesMap: Map<number, RenderableTopologyNode[]>, nodeToLevelMap: Map<RenderableTopologyNode, number>) {
+
+    const rootNodes: RenderableTopologyNode[] = levelToNodesMap.get(0) ?? [];
+    const visited = new Set<string>();
```
itssharmasandeep, line 28: we're basing it on the name. Are we sure that name is a unique string every time? I hope so but we can't be sure. We can use nodes for this and as these are the objects stored by reference. So if we do that, I think it should work as expected.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+    let curr = 0;
+
+    while (curr >= 0) {
+      const parent = rootNodes[curr];
+      const level = nodeToLevelMap.get(parent)! + 1;
+
+      parent.outgoing.forEach(edge => {
+        const child = edge.target;
+        if (!visited.has(child.userNode.data?.name!)) {
```
itssharmasandeep, line 37: If we use node here, we don't even need `!`  here. And this basically says the same thing, if data can be undefined then we shouldn't base it on the data properties

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
itssharmasandeep: If we want to layout again we can probably clear the maps and refill them again

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+  }
+
+  private fillNodeAndLevelMaps(levelToNodesMap: Map<number, RenderableTopologyNode[]>, nodeToLevelMap: Map<RenderableTopologyNode, number>) {
+
+    const rootNodes: RenderableTopologyNode[] = levelToNodesMap.get(0) ?? [];
+    const visited = new Set<string>();
+    let curr = 0;
+
+    while (curr >= 0) {
```
itssharmasandeep, line 31: can we not do `curr < rootNodes.length`, instead of breaking it later?

Greedy-Geek: won't it calculate the length once, i.e., in the starting and use the same length for the entire loop?

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
itssharmasandeep: You can also do this, every time a new node is created, generate a unique id for that node and use it. There's a `uniqueId()` in the loadash

Greedy-Geek: Oh yeah right, I converted it that way to use name in cytoscape as it accepts only strings and not objects for nodes and ended up using the names here as well, while here we can very well work with nodes as well. 
Thanks for that!

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Yeah right, updating

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Yeah right, will update that

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
itssharmasandeep: Ahh, you're updating the rootNodes in the loop

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: updated

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: updated

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: updated, it do check the length each time

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+    this.fillNodeAndLevelMaps();
+    this.fillNodeCoOrdinatesMap();
+    this.setNodeCoOrdinates(topology);
+  }
+
+  private findRootNodes(topology: RenderableTopology<TopologyNode, TopologyEdge>) {
+    topology.nodes.forEach(node => {
+      if (node.incoming.length === 0) {
+        this.nodeToLevelMap.set(node, 0);
+        this.levelToNodesMap.has(0) ? this.levelToNodesMap.get(0)?.push(node) : this.levelToNodesMap.set(0, [node])
```
aaron-steinfeld, line 21: nit: Initialize levelToNodesMap before iterating, then simplify the logic below. E.g.
```ts
  this.levelNodesMap.set(0, []);

  ...

  this.levelToNodesMap.get(0)!.push(node); // Prefer the non null assertion to avoid hiding a logic error
```

It also might be helpful to use methods like getRootNodes rather than access things by index (which we do on line 28 too) to indicate the meaning to the reader.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  private levelToNodesMap = new Map<number, RenderableTopologyNode[]>();
+  private nodeToLevelMap = new Map<RenderableTopologyNode, number>();
+  private nodeCoOrdinatesMap: Map<TopologyNode, TopologyCoordinates> = new Map<TopologyNode, TopologyCoordinates>();
+
+  public layout(topology: RenderableTopology<TopologyNode, TopologyEdge>): void {
+
+    this.findRootNodes(topology);
+    this.fillNodeAndLevelMaps();
+    this.fillNodeCoOrdinatesMap();
```
aaron-steinfeld, line 13: please fix capitalization - Coordinates is one word

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
+import { RenderableTopology, RenderableTopologyNode, TopologyCoordinates, TopologyEdge, TopologyNode } from "../../topology";
+
+export class GraphLayout {
+
+  private levelToNodesMap = new Map<number, RenderableTopologyNode[]>();
```
aaron-steinfeld, line 5: nit: a map keyed by integers is also known as an array 😉

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+  private nodeToLevelMap = new Map<RenderableTopologyNode, number>();
+  private nodeCoOrdinatesMap: Map<TopologyNode, TopologyCoordinates> = new Map<TopologyNode, TopologyCoordinates>();
```
aaron-steinfeld, line 7: nit: why does this field have a type declaration but not the others? If the linter allows it, would omit it since it's redundant.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+  private findRootNodes(topology: RenderableTopology<TopologyNode, TopologyEdge>) {
+    topology.nodes.forEach(node => {
+      if (node.incoming.length === 0) {
+        this.nodeToLevelMap.set(node, 0);
+        this.levelToNodesMap.has(0) ? this.levelToNodesMap.get(0)?.push(node) : this.levelToNodesMap.set(0, [node])
+      }
+    });
+  }
+
+  private fillNodeAndLevelMaps() {
```
aaron-steinfeld, line 26: personal pref: doing this with recursion rather than a mutating while loop would probably be more readable.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+      const parent = nodes[0];
+      const level = this.nodeToLevelMap.get(parent)! + 1;
+
+      parent.outgoing.forEach(edge => {
+        const child = edge.target;
+        if (!visited.has(child.userNode)) {
+          visited.add(child.userNode);
+          nodes.push(child);
+          this.nodeToLevelMap.set(child, level);
+          this.levelToNodesMap.has(level) ? this.levelToNodesMap.get(level)?.push(child) : this.levelToNodesMap.set(level, [child]);
```
aaron-steinfeld, line 41: same point as above on root. Alternatively can pull it into a function (E.g. addToLevel)

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+      });
+
+      nodes.shift();
+    }
+  }
+
+  private fillNodeCoOrdinatesMap() {
+
+    console.log(this.levelToNodesMap);
```
aaron-steinfeld, line 51: log statement

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    let curX = 1;
+
+    Array.from(this.levelToNodesMap.keys()).forEach(level => {
```
aaron-steinfeld, line 55: Why are you iterating keys and then just using them to access the values? Can just iterate the values

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+      const nodes = this.levelToNodesMap.get(level)!;
+
+      let curY = 1;
```
aaron-steinfeld, line 58: So we always start rendering at (1,1)? That doesn't seem desirable - but we can revisit. I think ideally we'd support centering (and if not centering, at least padding). If there were two nodes, right now this would render in the top left rather than right in the middle if I'm reading this right?

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+      let maxWidth = 0;
+      nodes.forEach((node) => {
+        this.nodeCoOrdinatesMap.set(node.userNode, { x: curX, y: curY });
```
aaron-steinfeld, line 61: Why are we writing the position to a new map, just to turn around late and write the position back to the node? At a minimum, we could cut down on the maps and have a single node-metadata map, but I'm not sure I see a reason why not to just write the position to node directly here.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Oh yeah, I missed removing it from here

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,76 @@
...
+        curY += (node.renderedData()?.getBoudingBox()?.height ?? 36) + 20;
+        maxWidth = Math.max(maxWidth, node.renderedData()?.getBoudingBox()?.width ?? 400);
+      })
+
+      curX += maxWidth + 240;
```
palbizu, line 66: nit: could we use constants to save the values (to clarify what means each)? It's not necessary but maybe better to understand why there are used.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Oh yeah, that was so dumb of me 😅. It doesn't need to be like that, I'll update it

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Yeah you are right, it will, as of now, start render rendering from the top left, but we can bring each level into the center, as we know the height of that level and we know the height of the container, so we can just translate each of the nodes to the desired position

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Oops this got happened after [this fix](https://github.com/hypertrace/hypertrace-ui/pull/1876#discussion_r1000269330). Previously it was just name of the service there. I'll update here as well

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Agreed, will update it

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: oh yeah will remove that

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: agreed, will update

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: It was just a choice between going Depth First or Breadth First and since all our logic is revolving around the levels I thought it would be better to go with the level order algorithm

aaron-steinfeld: Either algo can be implemented either way.

Greedy-Geek: Oh my bad, I got you wrong 😅. Since the first thing that comes to mind (at least for me) by hearing recursion I thought you were suggesting to go that route. Yeah we can implement it recursively as well.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: here if we want to intialize level to nodes map, we'll need to compute the no. of levels in advance, that I think can be a overkill

aaron-steinfeld: correct - however we can do it where we assign level, rather than checking for each edge (in this approach, be careful because this may not be the first time we visited the level depending on algo used).

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: updated

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: removed nodeCoOrdinatesMap

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Made it recursive

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Oh yeah, that will also improve the time, updated

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: updated

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: updated

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: actually this is becoming a config in [this PR](https://github.com/hypertrace/hypertrace-ui/pull/1880)

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -1,76 +1,75 @@
...
       }
     });
   }
 
   private fillNodeAndLevelMaps() {
 
-    const nodes = [...(this.levelToNodesMap.get(0) ?? [])];
-    const visited = new Set<TopologyNode>([...nodes.map(node => node.userNode)]);
+    const nodesThatAreGoingToBeExplored = [...(this.levelToNodesMap.get(0) ?? [])];
+    const nodesThatHaveBeenOrAreGoingToBeExplored = new Set<TopologyNode>([...nodesThatAreGoingToBeExplored.map(node => node.userNode)]);
```
aaron-steinfeld, line 27: nit - you call them both nodes, but you switch between working with `RenderableTopologyNode` and `TopologyNode`. Doesn't seem to be a clear reason why the distinction - we're just doing more work each time we work with this to map to userNode (lines 27,46,47). Only reason would be if there's some possibility that the same userNode could show up in multiple different renderable nodes, but don't believe that's the case.

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -1,76 +1,75 @@
...
-        }
-      });
+  private levelOrderTraversal(nodes: RenderableTopologyNode[], visited: Set<TopologyNode>): void{
+    if(nodes.length === 0) {
+      return;
+    }
 
-      nodes.shift();
+    const parent = nodes[0];
+    nodes.shift();
```
aaron-steinfeld, line 38:
```suggestion
    const parent = node.shift();
```

(also, given that you're working with a graph not a tree, suggest renaming "parent" to "currentNode")

aaron-steinfeld: actually, given that it's a graph, how do cycles look? Does it draw edges back across the nodes?

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Vertically center aligned the nodes

Greedy-Geek: we can make it configurable in this refactor [PR](https://github.com/hypertrace/hypertrace-ui/pull/1880)

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
Greedy-Geek: Oh that `.map` was supposed to be removed.
Updated

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -91,6 +95,21 @@ export class D3Topology implements Topology {
...
+      case TopologyLayoutType.GraphLayout:
+        return new GraphLayout();
+      case TopologyLayoutType.TreeLayout:
+        return new TreeLayout();
+      case TopologyLayoutType.ForceLayout:
+        return new ForceLayout();
+      default:
+        return new CustomTreeLayout();
+    }
+    // TODO: Make this configurable with Node and edge renderers
```
anandtiwary, line 112: nit: is this comment relevant?

Greedy-Geek: I also don't know about that, since the parenthesis is still empty, I think we can level it there so that we don't forget about it, if we want to make that upgrade

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -0,0 +1,86 @@
...
+  private assignCoordinatesToNodes(): void {
+    let curX = 1;
+
+    Array.from(this.levelToNodesMap.values()).forEach(nodes => {
+      let curY = 1;
+      let maxWidth = 0;
+      nodes.forEach(node => {
+        node.x = curX;
+        node.y = curY;
+        curY += (node.renderedData()?.getBoudingBox()?.height ?? 36) + 20;
```
anandtiwary, line 68: nit: could you please name all the constants and make then a static variable in the class?

Greedy-Geek: they are actually being made configurable in [this PR](https://github.com/hypertrace/hypertrace-ui/pull/1880)

projects/observability/src/shared/dashboard/widgets/topology/topology-widget.model.ts
```diff
@@ -71,6 +73,13 @@ export class TopologyWidgetModel {
   @ModelInject(MODEL_API)
   private readonly api!: ModelApi;
 
+  @ModelInject(FeatureStateResolver)
+  private readonly featureStateResolver!: FeatureStateResolver;
+
+  public isFeatureEnabled(feature: ApplicationFeature): Observable<boolean> {
```
anandtiwary, line 79: This should not be in the model. The model should have a layout property like we discussed. Similar to showBrush

```
  @ModelProperty({
    key: 'showBrush',
    displayName: 'Show Brush',
    type: BOOLEAN_PROPERTY.type,
    required: false
  })
  public showBrush?: boolean = true;
  ```

Comment (anandtiwary) [reply projects/observability/src/shared/dashboard/widgets/topology/topology-widget.model.ts:79]: Then the place where we are sending topology model there we need to put the FF logic.

Comment (Greedy-Geek) [reply projects/observability/src/shared/dashboard/widgets/topology/topology-widget.model.ts:79]: I added it as a config in topology component as that's the one exposed to the outer world. And created a property in TopologyConfiguration

Comment (Greedy-Geek) [reply projects/observability/src/shared/dashboard/widgets/topology/topology-widget.model.ts]: Updated

File: projects/observability/src/pages/apis/topology/application-flow.component.ts [hunk 22]
```diff
@@ -15,203 +19,221 @@ import {
   template: `
     <div class="vertical-flex-layout">
       <ht-page-header></ht-page-header>
-      <ht-navigable-dashboard [navLocation]="this.location" [defaultJson]="this.defaultJson"> </ht-navigable-dashboard>
+      <ht-navigable-dashboard
+        *htLetAsync="this.getDefaultJson() as defaultJson"
+        [navLocation]="this.location"
+        [defaultJson]="this.getDefaultJson()"
```
anandtiwary, line 25: Please do this
```
[defaultJson]="this.getDefaultJson | htMemoize | async
```

projects/common/src/constants/application-constants.ts
```diff
@@ -1,5 +1,6 @@
 export const enum ApplicationFeature {
   PageTimeRange = 'ui.page-time-range',
   FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map',
-  ExplorerOrderBySupport = 'ui.explorer-order-by-support'
+  ExplorerOrderBySupport = 'ui.explorer-order-by-support',
+  ApplicationFlowLayout = 'ui.application-flow-layout'
```
anandtiwary, line 5: please make sure that this flag is set to false in hypertrace feature flag resolver service. it is in source app

Greedy-Geek: oh yeah, updated

projects/observability/src/pages/apis/topology/application-flow.component.ts
Greedy-Geek: can't use htMemoize here as we are using this.isGraphLayoutEnabled() here

Greedy-Geek: should I bring that inside the function? I kept it outside to keep it readable

aaron-steinfeld:
This PR seems to have grown greatly in scope since the initial reviews. Any chance you can split it back down and keep the layout as its own review before moving on to integrating it?

Greedy-Geek:
> This PR seems to have grown greatly in scope since the initial reviews. Any chance you can split it back down and keep the layout as its own review before moving on to integrating it?

It got increased just now, in order to add a FF, had to put the entire defaultLayout inside a method in projects/observability/src/pages/apis/topology/application-flow.component.ts. That's what have increased the size

projects/observability/src/shared/dashboard/widgets/topology/topology-widget.model.ts
```diff
@@ -68,6 +69,14 @@ export class TopologyWidgetModel {
   })
   public enableBoxStyle?: boolean = false;
 
+  @ModelProperty({
+    key: 'layoutType',
+    displayName: 'Layout Type',
+    type: STRING_PROPERTY.type,
+    required: false
+  })
+  public layoutType?: TopologyLayoutType = TopologyLayoutType.CustomTreeLayout;
```
aaron-steinfeld, line 78: What's the purpose of a feature flag if it's also a widget input? The latter seems better - but we certainly don't want both and having to reconcile the two.

projects/observability/src/pages/apis/topology/application-flow.component.ts
```diff
@@ -15,203 +19,217 @@ import {
   template: `
     <div class="vertical-flex-layout">
       <ht-page-header></ht-page-header>
-      <ht-navigable-dashboard [navLocation]="this.location" [defaultJson]="this.defaultJson"> </ht-navigable-dashboard>
+      <ht-navigable-dashboard [navLocation]="this.location" [defaultJson]="this.getDefaultJson() | async">
```
aaron-steinfeld, line 22: Ah I missed this file because the diff was too big. So the flag is meant to change the application flow page's default layout, not the component's?  Probably could use a clarifier in the name (e.g. ApplicationFlow*Page*Layout ) if we really want to go that road. Might be a good time to pull the json out of the page component as we've done on other pages too.

Greedy-Geek: The FF is for very short peroid of time, I believe. It's there just to test out the graphLayout in selected customers' env. Check for the issues or improvements and then open for all. Till then, they would be seeing the layout as it is today. Right @anandtiwary?

projects/observability/src/shared/dashboard/widgets/topology/topology-widget.model.ts
Greedy-Geek: The FF is for this widget input `layoutType ` only. We'll just enable the FF for selected customers to see the graphLayout, rest will see the old CustomTreeLayout. Once we are sure with our layout, we'll enable it for the rest and remove the FF.

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -91,6 +95,21 @@ export class D3Topology implements Topology {
     this.hover = new TopologyHover(this.d3Util, this.domRenderer);
     this.click = new TopologyClick(this.d3Util, this.domRenderer);
     this.zoom = new TopologyZoom();
+    this.layout = this.initializeLayout(config.layoutType);
+  }
+
+  private initializeLayout(layoutType?: TopologyLayoutType): TopologyLayout {
+    switch (layoutType) {
+      case TopologyLayoutType.GraphLayout:
```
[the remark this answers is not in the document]
Greedy-Geek: The last one is the currently used layout, that we see in the app-dev. 

And regarding the names,
Tree layout: for graphs without cycle
Graph layout: for graphs with cycle
Forced layout: don't know it's purpose
Custom Tree layout: It's basically tree layout with a different layout
So, aren't they clear already?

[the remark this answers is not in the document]
aaron-steinfeld: > So, aren't they clear already?

Appears not! They're all for graphs with cycles, they just use different algorithms to visualize the graph (we only work with graphs, so any that aren't handling graphs or cycles are just bugs). Force layout treats nodes as particles (https://github.com/d3/d3-force), we stopped using it because for whatever reason people felt the hierarchy version was prettier.

projects/common/src/constants/application-constants.ts
anandtiwary: As discussed, please remove this flag from here

projects/observability/src/pages/apis/topology/application-flow.component.ts
anandtiwary: As discussed, Please remove this code from here.

anandtiwary (approved):
LGTM. Let's wait for Aaron to review again

aaron-steinfeld:
I believe the last remaining thing we had discussed was renaming the exposed enum for configuration. I think they're very confusing now so to me it seems like it would be a good idea to make them more descriptive (with the actual classes behind them being easier to change later, since they're internal impl) now, before they start getting written into JSON.

Not a deal breaker though, so if you guys disagree feel free to proceed.

PR #1879, opened 2022-10-21
fix: allow writing query params while nav in flight

aaron-steinfeld:
## Description
2 Changes
1. We've had a long standing issue where two different calls to add query params will cause the first to be ignored (because it cancels the nav in flight and starts a new one, so the query params don't get merged. Now, we hold query params until a nav finishes, and if another request arrives, we merge the prior in.
2. If time range refresh is called very early in app bootstrap, it leads to an error as the time range hasn't been initialized yet. Instead, treat it as a no-op.

### DIFF
--- a/projects/common/src/time/time-range.service.ts
+++ b/projects/common/src/time/time-range.service.ts
@@ -61,8 +61,10 @@ export class TimeRangeService {
   }
 
   public refresh(): void {
-    const currentStringTimeRange = this.getCurrentTimeRange().toUrlString();
-    this.applyTimeRangeChange(this.timeRangeFromUrlString(currentStringTimeRange));
+    if (this.isInitialized()) {
+      const currentStringTimeRange = this.getCurrentTimeRange().toUrlString();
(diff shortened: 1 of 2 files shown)


projects/common/src/time/time-range.service.ts
```diff
@@ -61,8 +61,10 @@ export class TimeRangeService {
   }
 
   public refresh(): void {
-    const currentStringTimeRange = this.getCurrentTimeRange().toUrlString();
-    this.applyTimeRangeChange(this.timeRangeFromUrlString(currentStringTimeRange));
+    if (this.isInitialized()) {
+      const currentStringTimeRange = this.getCurrentTimeRange().toUrlString();
```
palbizu, line 65: nit: does it necessary this const here? does not better to pass the variable directly?

aaron-steinfeld: it's pre-existing, I just wrapped it with the if check - but I would say it probably was written that way for readability to avoid nesting 3-4 function calls

adisreyaj (approved):

palbizu (approved):

PR #1882, opened 2022-10-22
feat(gql): query operation name in network tab

adisreyaj:
## Description
Add support for query naming. This will make identifying queries in Network panel much easier.

## Usage
```ts
  private querySomething() {
    return this.gql.query(
      {
        requestType: ''
      },
      {
        cacheability: GraphQlRequestCacheability.Cacheable,
        name: 'FetchEntities' // <-- provide name
      }
    );
  }
```

image

### DIFF
--- a/projects/graphql-client/src/model/graphql-selection.ts
+++ b/projects/graphql-client/src/model/graphql-selection.ts
@@ -2,6 +2,7 @@ import { GraphQlArgument } from './graphql-argument';
 
 export interface GraphQlSelection {
   path: string;
+  name?: string;
   arguments?: GraphQlArgument[];
   children?: GraphQlSelection[];
   alias?: string;
--- a/projects/graphql-client/src/utils/builders/request/graphql-request-builder.ts
+++ b/projects/graphql-client/src/utils/builders/request/graphql-request-builder.ts
@@ -31,7 +31,10 @@ export class GraphQlRequestBuilder {
     return new Map(
       this.originalSelections.map(selection => {
         const key = this.getKeyForSelection(selection);
-        const requestString = `{ ${this.translateRequestFragmentToGql(key, this.mergedRequest[key])} }`;
+        const requestString = `${selection.name ?? selection.path} { ${this.translateRequestFragmentToGql(
+          key,
+          this.mergedRequest[key]
+        )} }`;
 
         return [selection, requestString];
       })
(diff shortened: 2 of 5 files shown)


projects/graphql-client/src/model/graphql-selection.ts
```diff
@@ -2,6 +2,7 @@ import { GraphQlArgument } from './graphql-argument';
 
 export interface GraphQlSelection {
   path: string;
+  name?: string;
```
aaron-steinfeld, line 5: how is this different than alias?

aaron-steinfeld: My guess from the other code is this is used by the gql client to pass metadata from the request options to the root selection, but this type is used for general request building so we don't want to leak an internal impl detail into it.

projects/graphql-client/src/utils/builders/request/graphql-request-builder.ts
```diff
@@ -31,7 +31,10 @@ export class GraphQlRequestBuilder {
     return new Map(
       this.originalSelections.map(selection => {
         const key = this.getKeyForSelection(selection);
-        const requestString = `{ ${this.translateRequestFragmentToGql(key, this.mergedRequest[key])} }`;
+        const requestString = `${selection.name ?? selection.path} { ${this.translateRequestFragmentToGql(
```
aaron-steinfeld, line 34: I would need to go review the exact impl - it's been years - but i would guess this will lead to some bugs when account for request merging. I'm not even sure how it would be expected to behave - which request's operation name is used? And if it prevents merging, is that really desirable - to fire multiple requests just to include debug metadata like operation name (maybe better in dev mode but not prod mode)

aaron-steinfeld (commented):
please make sure to add tests, particular for the cases that deal with multiple requests and merging (e.g. two requests with the same name, different name, one without a name)

projects/graphql-client/src/model/graphql-selection.ts
adisreyaj: Yes...I think I could hide it with a new interface which is not exposed.

aaron-steinfeld: It's very similar to what you're doing - it provides a custom name for a field (so not usually used at root). Most commonly, we see this when the same field is queried multiple times with different arguments.

e.g. an entity requesting
```
  name: attribute(key: "name")
  otherAttr: attribute(key: "otherAttr")
```

PR #1888, opened 2022-10-25
fix: don't count a navigation until history change

aaron-steinfeld:
## Description
Fixed the tracking of first navigation (used to handle in-app back behavior), to ignore if we update the URL without changing history. This was previously causing erroneous back navigation out of the app.

### Testing
Local E2E

### DIFF
--- a/projects/common/src/navigation/navigation.service.ts
+++ b/projects/common/src/navigation/navigation.service.ts
@@ -5,6 +5,7 @@ import {
   ActivatedRoute,
   ActivatedRouteSnapshot,
   Event,
+  Navigation,
   NavigationEnd,
   NavigationExtras,
   ParamMap,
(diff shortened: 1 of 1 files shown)


projects/common/src/navigation/navigation.service.ts
```diff
@@ -144,6 +149,10 @@ export class NavigationService {
     };
   }
 
+  public getActiveNavigation(): Navigation | undefined {
```
Greedy-Geek, line 152: what is the role of undefined here?

Greedy-Geek (approved):

palbizu (approved):

projects/common/src/navigation/navigation.service.ts
aaron-steinfeld: This is based off https://angular.io/api/router/Router#getCurrentNavigation - it returns the in-flight navigation if there is one, otherwise null. I'm switching the null over to undefined here to keep consistency with our codebase.

PR #1889, opened 2022-10-27
feat(panel): add new variant for panel title with action

adisreyaj:
## Description
Add a new variant of panel title.

image


23nobody:
Please add some basic tests

projects/components/src/panel/header/title-with-action/panel-title-with-action.component.ts
```diff
@@ -0,0 +1,38 @@
...
+      </div>
+    </div>
+  `
+})
+export class PanelTitleWithActionComponent {
+  @Input()
+  public expanded!: boolean;
+
+  @Input()
+  public actionIcon: IconType = IconType.Edit;
```
23nobody, line 34: The icon should be hidden with an `*ngIf` when input is not given?

adisreyaj: The panel is made to be used when you need an icon, so the consumer has to provide an icon or the default icon will be used.

23nobody: Why not add this support to `PanelTitleComponent` itself?

23nobody: With icon hidden with an `*ngIf`

adisreyaj: We could, but thought to keep it as it is because if the consumer wants more control they should be using the `PanelTitleComponent` and not this component. Made a separate component as the patterns is becoming more common.

adisreyaj:
> Please add some basic tests

Waiting till this is good to go.

projects/components/src/panel/header/title-with-action/panel-title-with-action.component.ts
itssharmasandeep: 1. Actually, I also think this is a small addition to the panel title component, we shouldn't create a new component for it, we can have an optional property for the action. Because currently this component is also re-using the same code of already existing component
2. What about the multiple actions part? Do we have a use case for this?

adisreyaj: 1. For now its a small addition, just didn't want to keep adding knobs to the component.
2. We don't have that use case now, but could come in the future.

itssharmasandeep: > For now its a small addition, just didn't want to keep adding knobs to the component.

My issue with this approach is this, suppose we want to add another config(suppose an info icon) to the panel title component, in that case, we would need to make changes to both components and it can make things harder in the future. I agree sometimes it becomes hard to make changes to the same component but this may be a more dangerous route going forward.

There are a few components where this works, like a specific component but here it is a generic component, and making two generic components for this will become an issue IMO.

adisreyaj:
Closing in favor of https://github.com/hypertrace/hypertrace-ui/pull/1900

projects/components/src/panel/header/title-with-action/panel-title-with-action.component.ts
adisreyaj: Added new PR updating existing one.

We can add more inputs to configure the actions going forward.

PR #1892, opened 2022-10-28
fix: extracted application flow dashboard, created variable layoutType

Greedy-Geek:


projects/observability/src/pages/apis/topology/application-flow.dashboard.ts
```diff
@@ -0,0 +1,209 @@
...
+  defaultPrimaryEdgeMetricCategories,
+  defaultSecondaryEdgeMetricCategories
+} from '../../../shared/dashboard/widgets/topology/metric/edge-metric-category';
+import {
+  defaultPrimaryNodeMetricCategories,
+  defaultSecondaryNodeMetricCategories
+} from '../../../shared/dashboard/widgets/topology/metric/node-metric-category';
+import { MetricAggregationType } from '../../../shared/graphql/model/metrics/metric-aggregation';
+
+export const applicationFlowDefaultJson: DashboardDefaultConfiguration = {
```
palbizu, line 12: just extracting the const, what's the reason? Are we reusing it elsewhere?

Greedy-Geek: We are using the same dashboard in t-ui as well as we have FF'ed the new layout, and controlling it t-ui.

aaron-steinfeld (approved):

PR #1898, opened 2022-11-02
Exported IconData from tableCellParser in public-api

chahat99:
## Description
Exported IconData from tableCellParser in public-api


projects/components/src/public-api.ts
```diff
@@ -351,6 +351,7 @@ export { TableCellParser } from './table/cells/table-cell-parser';
 export * from './table/cells/data-parsers/table-cell-no-op-parser';
 export * from './table/cells/data-parsers/table-cell-string-parser';
 export * from './table/cells/data-parsers/table-cell-timestamp-parser';
+export { IconData } from './table/cells/data-parsers/table-cell-icon-parser';
```
adisreyaj, line 354: Why not just export `*` like others?

adisreyaj (approved):

projects/components/src/public-api.ts
chahat99: Updated

palbizu (approved):

PR #1899, opened 2022-11-02
refactor: no deselect mode for color picker

Christian862:
## Description
Adds input config option to color picker to disallow de-selecting current element, and only allow change selection to another option

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/color-picker/color-picker.component.ts
+++ b/projects/components/src/color-picker/color-picker.component.ts
@@ -49,8 +49,9 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   @Input()
   public disabled?: boolean = false;
 
+  // Disables de-selecting current option, meaning only allow selection to another option
   @Input()
-  public noDeselectMode?: boolean = false;
+  public disableDeselection?: boolean = false;
 
   @Output()
   private readonly selectedChange: EventEmitter<string> = new EventEmitter<string>();
@@ -81,7 +82,7 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   }
 
   public selectColor(color: string): void {
-    if (this.disabled || (this.noDeselectMode && color === this.selected)) {
+    if (this.disabled || (this.disableDeselection && color === this.selected)) {
       return;
     }
 
--- a/projects/components/src/color-picker/color-picker.component.ts
+++ b/projects/components/src/color-picker/color-picker.component.ts
@@ -51,7 +51,7 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
 
   // Disables de-selecting current option, meaning only allow selection to another option
   @Input()
-  public disableDeselection?: boolean = false;
+  public disableDeselection: boolean = false;
 
   @Output()
   private readonly selectedChange: EventEmitter<string> = new EventEmitter<string>();
--- a/projects/components/src/color-picker/color-picker.component.ts
+++ b/projects/components/src/color-picker/color-picker.component.ts
@@ -49,6 +49,9 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   @Input()
   public disabled?: boolean = false;
 
+  @Input()
+  public noDeselectMode?: boolean = false;
+
   @Output()
   private readonly selectedChange: EventEmitter<string> = new EventEmitter<string>();
 
@@ -78,7 +81,7 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   }
 
   public selectColor(color: string): void {
-    if (this.disabled) {
+    if (this.disabled || (this.noDeselectMode && color === this.selected)) {
       return;
     }
 
--- a/projects/components/src/color-picker/color-picker.component.ts
+++ b/projects/components/src/color-picker/color-picker.component.ts
@@ -2,7 +2,6 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Input, OnChanges, Out
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { IconType } from '@hypertrace/assets-library';
 import { Color, TypedSimpleChanges } from '@hypertrace/common';
-import { isNil } from 'lodash-es';
 import { IconSize } from '../icon/icon-size';
 
 @Component({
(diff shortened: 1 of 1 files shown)


projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -49,6 +49,9 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   @Input()
   public disabled?: boolean = false;
 
+  @Input()
+  public noDeselectMode?: boolean = false;
```
itssharmasandeep, line 53: should we call it `disableColorDeselection` or `disableDeselection`?

itssharmasandeep: And add a comment here if possible

Christian862: yeah better name, change coming

projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -49,6 +49,10 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   @Input()
   public disabled?: boolean = false;
 
+  // Disables de-selecting current option, meaning only allow selection to another option
+  @Input()
+  public disableDeselection?: boolean = false;
```
itssharmasandeep, line 54: we don't need making it optional, if we have assigned it a value

aaron-steinfeld: what a name! how about `required` ?

projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -78,7 +82,7 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
   }
 
   public selectColor(color: string): void {
-    if (this.disabled) {
+    if (this.disabled || (this.disableDeselection && color === this.selected)) {
```
aaron-steinfeld, line 85: nit - the special casing for deselecting happens twice - here and on 89. Would be a bit nicer to resolve the selection, then handle if the result is allowed. Would be even nicer to prevent deselecting in the first place - disabling the interaction on the current color.

projects/components/src/color-picker/color-picker.component.ts
aaron-steinfeld: The reason we often make it optional with an assigned value is to support async assignments. Since angular won't prevent us from writing something like
`[disableDeselection] = "isRequired$ | async"`, that will result in a null value being assigned and overwriting the default. So the optional - even though it's not null, which would be more accurate - makes sure our code doesn't break if that happens.

Christian862: I like `disableDeselection` because this case seems a bit different than the standard required form case, however nbd to me, i can change it if you like.

> The reason we often make it optional with an assigned value is to support async assignments. Since angular won't prevent us from writing something like
[disableDeselection] = "isRequired$ | async", that will result in a null value being assigned and overwriting the default. So the optional - even though it's not null, which would be more accurate - makes sure our code doesn't break if that happens.

Thats a good point, and safety measure, i'll change back.

projects/components/src/color-picker/color-picker.component.ts
anandtiwary: how is it different from `disabled`?

projects/components/src/color-picker/color-picker.component.ts
anandtiwary: If deselection is not allowed, can we disable the color picker after selecting a value?

projects/components/src/color-picker/color-picker.component.ts
Christian862: It still allows user to select between the options, just not de-selecting the currently selected one

projects/components/src/color-picker/color-picker.component.ts
Christian862: No because we still want to allow them to select between the available options

Christian862: yeah i suppose I thought of it differently because of the unset behavior.  With required it implies there is the option to not have it selected, and then manually select it again, however here we remove the option to deselect entirely....requiring the selection to be present...and now i've changed my mind lol. Changing it to required

aaron-steinfeld: FWIW, I would look at whatever we do in select component for consistency. This is just a specific cased version of that.

Christian862: Looked however it appears it doesn't have a required behavior option. its normally handled by the parent with validation, it appears its just this component with the special behavior

aaron-steinfeld: It has `showClearSelected` which is kind of equivalent to what you have - the one side of validation to prevent _reaching_ an empty state, with validator behavior handled in the parent. Do with that what you will!

palbizu (approved):
Looks good to me but if it's possible should be better to wait for @anandtiwary or @aaron-steinfeld review.

projects/components/src/color-picker/color-picker.component.ts
```diff
@@ -83,6 +88,9 @@ export class ColorPickerComponent implements ControlValueAccessor, OnChanges {
     }
 
     const clickedColor = this.selected === color ? undefined : color;
+    if (this.required && isNil(clickedColor)) {
```
anandtiwary, line 91:
```
const isDeselect = this.selected === color;

if (this.required && isDeselect) {
  return;
}

const clickedColor = isDeselect ? undefined : color;
...
```

projects/components/src/color-picker/color-picker.component.ts
aaron-steinfeld: The second part here seems like the important one btw. Not handling this in code but in UX.

projects/components/src/color-picker/color-picker.component.ts
Christian862: changed

anandtiwary (approved):

aaron-steinfeld:
Thoughts on the above comment? https://github.com/hypertrace/hypertrace-ui/pull/1899#discussion_r1012314829

Christian862:
> Thoughts on the above comment? Guessing it was missed [#1899 (comment)](https://github.com/hypertrace/hypertrace-ui/pull/1899#discussion_r1012314829)

Wow my bad, i merged this late-ish last night after discussing with Anand about it and totally missed your comment. Although i should have seen it it was right in front of me lol.

I opened a new PR to address the comment: https://github.com/hypertrace/hypertrace-ui/pull/1903

PR #1904, opened 2022-11-07
fix: large legend shrinks chart in Explorer

### LINKED ISSUE #1905: Bug: Legend in Explorer hides the chart from view
## Use Case

If group_by crosses a certain number of results, the legend's height increases which shrinks the chart to a point it becomes impossible to see.

Screenshot_2022-09-22_at_11_28_46_AM

## Proposal

Either the legend should expand down below, or add a scroll when there are too many results to fit in view. Either way, the legend should never interfere with the chart being displayed.

## Questions to address (if any)
[Slack discussion](https://hypertrace.slack.com/archives/C020N63M0BC/p1666666769175119)

cseas:
## Description

[Slack][1] | Fixes #1905 

Setting a max-height on the Explorer chart legend so it doesn't shrink/hide the chart when there are too many entries in the legend.

This change fixes a bug that's currently rendering Hypertrace unusable for some of our users. We have multiple use-cases where the legend size gets too big  and sometimes even completely hides the chart from view.

### Screenshots

|Before|
|:-----:|
| before |
|After|
| after |

### Testing

CSS changes. Tested manually.

### Checklist:

### Documentation
N/A

[1]: https://hypertrace.slack.com/archives/C020N63M0BC/p1666666769175119


cseas:
Tests are passing. CI is failing at Codecov upload, no idea why.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.scss
```diff
@@ -104,6 +104,10 @@
     }
 
     .legend-entries {
+      /* Prevent large legends from shrinking chart */
+      max-height: 3.5rem;
```
itssharmasandeep, line 108: Two things:
1. we do not use the relative size units anywhere (`em` is used in one place)
2. Since legend text and other things (children of this class)  are fixed (`px`) width and height, here it will not give us the benefit of using the relative size.

Maybe we can use a fixed height here in `px`, you can try and convert this `3.5rem` into `px` by looking at the dom.

projects/observability/src/shared/components/cartesian/cartesian-chart.component.scss
```diff
@@ -104,6 +104,10 @@
...
+      overflow: auto;
```
itssharmasandeep, line 109: I am guessing it is only for vertical scroll, if yes then we can do `overflow-y: auto` instead.

itssharmasandeep:
> Tests are passing. CI is failing at Codecov upload, no idea why.

I re-ran the tests, and it is passing now :)

projects/observability/src/shared/components/cartesian/cartesian-chart.component.scss
cseas: Sure. One question, should the height be assigned to fit 2 rows exactly (screenshot in PR description), or show a little bit of the third row, like this?

scroll

projects/observability/src/shared/components/cartesian/cartesian-chart.component.scss
cseas: Makes sense. Done. ✅

itssharmasandeep (approved):
LGTM

PR #1909, opened 2022-11-08
Expose getContrast color function

Christian862:
## Description
Exposes getContrast color function

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/common/src/color/color-palette.ts
+++ b/projects/common/src/color/color-palette.ts
@@ -20,8 +20,15 @@ export class ColorPalette {
     return this.getColorCombinations(colorSetSize)[Math.abs(hashCode(id)) % colorSetSize];
   }
 
-  private getContrast(rgbColorString: string): string {
-    return getContrastingColor(rgbColorString);
+  public getContrast(rgbColorString: string): string {
+    // Convert to RGB value
+    const rgbColor = rgb(rgbColorString);
+
+    // Get YIQ ratio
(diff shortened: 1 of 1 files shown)


palbizu (commented):
LGTM but I'm not completely sure about the approach.

projects/common/src/color/color-palette.ts
```diff
@@ -57,3 +50,14 @@ export class ColorPalette {
     return [...this.basisColors];
   }
 }
+
+export const getContrastingColor = (rgbColorString: string): string => {
```
aaron-steinfeld, line 54: would suggest making the method public rather than a separate const function - unless there's more to this?

Christian862: Thats fine, it just necessitates there already being a color palette, and in my case it works because i have a color palette to use, however it seems like a special case. Normally i would think the consumer doesn't have a color palette

aaron-steinfeld: Ah - I missed this was color palette, assumed it was color service. I would probably put it there eventually - eventually figuring out how to get it in here (such as by providing a color service on construction).

Christian862: kk, i'll leave here for now then

Christian862: Removed exported, and made function public

aaron-steinfeld (approved):

PR #1919, opened 2022-11-17
fix: conditional file upload notifications

itssharmasandeep:
## Description
This adds a way to show notification conditionally when `showNotification` is true (defaulted to true)

### Testing
Manual testing is done locally.

### Checklist:

### DIFF
--- a/projects/components/src/file-upload/file-upload.service.ts
+++ b/projects/components/src/file-upload/file-upload.service.ts
@@ -74,7 +74,7 @@ export class FileUploadService {
 
           return undefined;
         }),
-        this.showFileUploadNotification(showNotification, `File upload successful`, `File upload failed`),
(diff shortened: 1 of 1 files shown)


projects/components/src/file-upload/file-upload.service.ts
```diff
@@ -72,7 +74,7 @@ export class FileUploadService {
 
           return undefined;
         }),
-        this.notificationService.withNotification(`File upload successful`, `File upload failed`),
+        this.showFileUploadNotification(showNotification, `File upload successful`, `File upload failed`),
```
anandtiwary, line 77: This is a bit weird. Probably rename the method to `maybeShowFileUploadNotification` or just dont call it when boolean is false

arjunlalb (approved):

PR #1920, opened 2022-11-17
Adds opacity adjustment function to color service

Christian862:
## Description
The function takes in a hex or rgb string and returns an rgb string with the adjusted opacity.

Edit: moved to draft

### Testing

### Checklist:

### Documentation


projects/common/src/color/color.service.ts
```diff
@@ -41,6 +41,13 @@ export class ColorService {
     return rgb(colorHex).brighter(basis).formatHex();
   }
 
+  public adjustOpacity(color: string, basis: number): string {
```
itssharmasandeep, line 44: why it is named `basis` here?

projects/common/src/color/color.service.ts
```diff
@@ -41,6 +41,13 @@ export class ColorService {
...
+    const rgbColor = rgb(color);
+    rgbColor.opacity = basis;
+
+    return rgbColor + '';
```
arjunlalb, line 48: instead of this, you can do `return rgb(rgbColor.r, rgbColor.g, rgbColor.b, opacity).toString`, I think.

little cleaner than the implicit string conversion at L48 and setting opacity.

projects/common/src/color/color.service.ts
arjunlalb: +1

projects/common/src/color/color.service.ts
Christian862: fixed

projects/common/src/color/color.service.ts
Christian862: fixed

itssharmasandeep (approved):

PR #1927, opened 2022-11-23
fix: support for start and end time in buildNavParamsWithFilters()

Greedy-Geek:


projects/observability/src/pages/explorer/explorer-service.ts
```diff
@@ -17,7 +18,9 @@ export class ExplorerService {
   ) {}
   public buildNavParamsWithFilters(
     scopeQueryParam: ScopeQueryParam,
-    filters: ExplorerDrilldownFilter[]
+    filters: ExplorerDrilldownFilter[],
+    startTime?: Date,
+    endTime?: Date
```
itssharmasandeep, line 23: why not use `TimeRange` here? It will also solve twice the `null` check at L53.
Any particular reason for making it generic?

projects/observability/src/pages/explorer/explorer-service.ts
```diff
@@ -13,11 +21,13 @@ import { ScopeQueryParam } from './explorer.component';
 export class ExplorerService {
   public constructor(
     private readonly metadataService: MetadataService,
-    private readonly filterBuilderLookupService: FilterBuilderLookupService
+    private readonly filterBuilderLookupService: FilterBuilderLookupService,
+    private readonly timeRangeService: TimeRangeService
   ) {}
   public buildNavParamsWithFilters(
     scopeQueryParam: ScopeQueryParam,
```
anandtiwary, line 28: what is in ScopeQueryParam? And how are we applying filters currently?

projects/observability/src/pages/explorer/explorer-service.ts
```diff
@@ -37,14 +40,21 @@ export class ExplorerService {
...
           scope: scopeQueryParam
-        }
-      }))
+        };
+
+        return {
+          navType: NavigationParamsType.InApp,
+          path: '/explorer',
+          queryParams:
+            !isNil(startTime) && !isNil(endTime)
```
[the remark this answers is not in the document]
Greedy-Geek: updated

projects/observability/src/pages/explorer/explorer-service.ts
Greedy-Geek: oh yeah, updated

projects/observability/src/pages/explorer/explorer-service.ts
```diff
@@ -37,14 +47,20 @@ export class ExplorerService {
...
-        }
-      }))
+        };
+
+        return {
+          navType: NavigationParamsType.InApp,
+          path: '/explorer',
+          queryParams: !isNil(timeRange)
+            ? { ...queryParam, ...this.timeRangeService.toQueryParams(timeRange) }
+            : queryParam
```
itssharmasandeep, line 61: You don't need another block to do it. You can do it like this
```ts
    return forkJoinSafeEmpty(filterStrings$).pipe(
      map(filterStrings => ({
        path: '/explorer',
        queryParams: {
          filter: filterStrings,
          scope: scopeQueryParam,
          ...(!isNil(timeRange) ? this.timeRangeService.toQueryParams(timeRange) : {})
        }
      }))
```

Greedy-Geek: oh yeah, updated

projects/observability/src/pages/explorer/explorer-service.test.ts
```diff
@@ -52,14 +48,25 @@ describe('Explorer service', () => {
...
+  const createService = createServiceFactory({
+    service: ExplorerService,
+    providers: [
+      mockProvider(FilterBuilderLookupService, { lookup: lookupMock }),
+      mockProvider(MetadataService, { getAttribute: getAttributeMock }),
+      mockProvider(TimeRangeService, {
+        toQueryParams: () => ({
+          ['time']: new FixedTimeRange(
+            new Date('2019-09-19T16:40:45.141Z'),
+            new Date('2019-09-19T16:55:45.141Z')
```
itssharmasandeep, line 60: nit: this seems the same below, and can be converted into a const

itssharmasandeep (approved):

projects/observability/src/pages/explorer/explorer-service.test.ts
Greedy-Geek: Oh yeah, but since there's just two instances I left it as it is

PR #1935, opened 2022-12-01
Fix table not re-fetching on column change

Christian862:
## Description

Problem:
1. Table row data is dependent on the data source `getData()` function, this is dependent on the specification backed columns in the request. Any time there is a row change, we detect it, and fetch data (either new or from cache). Any time there is a column configs change(which the request depends on) however, we are not re-fetching. Instead a column config change occurs, we see that there are cached rows, and we use the cached rows - which are outdated because the cached rows did not come from a request that uses the newly updated column configs.

2. While discovering the above problem i came across another: The table depends on the row data returned from the data source `getData()` with a TableDataRequest. <- Here the row data has a tight dependency on the column configs, which it should.  However in the component UI rendering, row data has **no** dependency on columns - meaning that in tables with editable columns, once the columns configs are changed, we immediately change the displayed columns without having the row data to support them.

PR Solutions:
1. Adds check for table column change to fetch new data. 
2. Make datasource responsible for keeping visible columns up to date. This way, only columns that were used with the request (i.e. rows) will be used for display.

### Testing

### Checklist:

### Documentation


projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -284,14 +301,23 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
-    return this.hasCachedRowsForRequest(request) ? this.fetchCachedData(request) : this.fetchNewData(request);
+    return this.hasCacheForRequest(request) ? this.fetchCachedData(request) : this.fetchNewData(request);
+  }
+
+  private haveColumConfigsChanged(request: TableDataRequest): boolean {
+    if (isNil(this.cachedData.request)) {
+      return true;
+    }
+
+    return !isEqual(request.columns, this.cachedData.request.columns);
```
anandtiwary, line 312: if we don't change the object type, wouldn't this condition still do a valid check?

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -42,6 +42,18 @@ export class TableCdkDataSource implements DataSource<TableRow> {
 
   public loadingStateChange$: Observable<TableLoadingState> = this.loadingStateSubject.asObservable();
 
+  private readonly requestColumnConfigsSubject: BehaviorSubject<TableColumnConfigExtended[]> = new BehaviorSubject<
```
anandtiwary, line 45: What do we gain by moving these here?

Christian862:
so in `table-edit-columns-modal.component.ts` we still need to be creating a new object in the array.  Because otherwise, `this.cache.request.columns` and `request.columns` (data source) share the same array and object references, so when the modification is made `column.visible = $event` (header edit modal), they will always pass the check  `isEqual(this.cache.request.columns, request.columns)` in `haveColumnConfigsChanged` , because they are the exact same objects - resulting in never re-fetching.

However when we create the new object in header edit modal:
```
this.editColumns[index] = {
      ...this.editColumns[index],
      visible: checked
};
```
 You might initially think that `this.cache.request.columns` and `request.columns` both get that mutation change(because before this point they are the same object/share reference), however they don’t….one will have object with visible true, the other corresponding object with visible false.
Its almost like modifying an index of the array to have a new obj reference(like above snippet), causes the array to copy itself with a new reference as well...

projects/components/src/table/data/table-cdk-data-source.ts
Christian862: Changes discussed offline made

projects/components/src/table/data/table-cdk-data-source.ts
Christian862: comment below addresses this

anandtiwary (approved):

PR #1938, opened 2022-12-05
fix: changed style of selected card

Greedy-Geek:
Currently, both the blocked and selected card have same background color, which should not be the case. Applied the background color suggested by the UX


Exac (approved):

arjunlalb:
@Greedy-Geek why is this still open?

PR #1944, opened 2022-12-08
fix: added method for external navigation in explorer service

Greedy-Geek:


projects/observability/src/pages/explorer/explorer-service.test.ts
```diff
@@ -113,4 +138,28 @@ describe('Explorer service', () => {
...
+          ],
+          mockTimeRange
+        )
+      ).toBe('(x|)', {
+        x: {
+          navType: NavigationParamsType.External,
+          url:
+            '/explorer?filter=duration_gt_200&filter=status_eq_404&scope=endpoint-traces' +
+            `&time=${mockTimeRange.toUrlString()}`,
+          windowHandling: ExternalNavigationWindowHandling.NewWindow
```
anandtiwary, line 160: We just need the `windowHandling`. No need to set the nav type to External

anandtiwary: We could use `OpenInNewTabModule`

aaron-steinfeld: Oh yeah I forgot we had that. We should update it to just take in nav params or string and have the component handle any externalization necessary

PR #1952, opened 2022-12-13
fix: change explore spec name to the expression key

itssharmasandeep:


aaron-steinfeld (dismissed):

projects/observability/src/shared/graphql/request/builders/specification/explore/explore-specification-builder.ts
```diff
@@ -45,7 +45,7 @@ export class ExploreSpecificationBuilder {
 
     return {
       resultAlias: () => this.buildResultAlias(expression, aggregation),
-      name: expressionString,
+      name: expression.key,
```
Christian862, line 48: shouldn't it be `this.attributeExpressionAsString(expression)`

aaron-steinfeld (approved):

PR #1955, opened 2022-12-15
feat: update ignore global filters handling

adisreyaj:
## Description
The idea of `ignoreGlobalFilters` was to only ignore the Api discovery state. So avoid any confusion around it, the key is now called `ignoreApiDiscoveryStateFilter`.

- Added new `ignoreGlobalEnvironmentFilter` in `GraphQlEntitiesRequest`.
- Add ability to exclude global filters

### DIFF
--- a/projects/observability/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
+++ b/projects/observability/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
@@ -5,12 +5,27 @@ import { GraphQlFilter } from './graphql-filter';
 export class GlobalGraphQlFilterService {
   private readonly filterRules: Map<symbol, FilterRule> = new Map();
 
-  public getGlobalFilters(scope: string): GraphQlFilter[] {
-    return Array.from(this.filterRules.values()).flatMap(rule => rule.filtersForScope(scope));
+  public getGlobalFilters(scope: string, filtersToExclude: symbol[] = []): GraphQlFilter[] {
+    const filtersToExcludeSet = new Set(filtersToExclude);
+
+    return Array.from(this.filterRules.entries())
+      .reduce((acc, curr) => {
+        const [ruleKey, rule] = curr;
+        if (!filtersToExcludeSet.has(ruleKey)) {
+          acc.push(rule);
+        }
+
+        return acc;
+      }, [] as FilterRule[])
+      .flatMap(rule => rule.filtersForScope(scope));
   }
 
-  public mergeGlobalFilters(scope: string, localFilters: GraphQlFilter[] = []): GraphQlFilter[] {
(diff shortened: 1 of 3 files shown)


projects/observability/src/shared/graphql/model/schema/filter/global-graphql-filter.service.ts
```diff
@@ -5,12 +5,27 @@ import { GraphQlFilter } from './graphql-filter';
 export class GlobalGraphQlFilterService {
   private readonly filterRules: Map<symbol, FilterRule> = new Map();
 
-  public getGlobalFilters(scope: string): GraphQlFilter[] {
-    return Array.from(this.filterRules.values()).flatMap(rule => rule.filtersForScope(scope));
+  public getGlobalFilters(scope: string, filtersToExclude: symbol[] = []): GraphQlFilter[] {
```
aaron-steinfeld, line 8: This feels like we're digging the hole even deeper - the idea of global filters is they're opaque - we shouldn't be requesting to exclude them.

projects/observability/src/shared/graphql/request/handlers/entities/query/entity/entity-graphql-query-handler.service.ts
```diff
@@ -61,7 +61,7 @@ export class EntityGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
       // If querying for a single API by ID, then usually want to includeInactive
       includeInactive: request.includeInactive !== false,
       // If querying for a single API by ID, ignore all global filters on entities query
-      ignoreGlobalFilters: true
+      ignoreApiDiscoveryStateFilter: true
```
aaron-steinfeld, line 64: I would change the entity query handler to not ever apply global query handlers, rather than try to pick and choose which should be ignore and which shouldn't.

projects/observability/src/shared/graphql/request/handlers/entities/query/entities-graphql-query-builder.service.ts
```diff
@@ -105,7 +106,8 @@ export interface GraphQlEntitiesRequest {
   filters?: GraphQlFilter[];
   includeTotal?: boolean;
   includeInactive?: boolean;
-  ignoreGlobalFilters?: boolean;
+  ignoreApiDiscoveryStateFilter?: boolean;
+  ignoreGlobalEnvironmentFilter?: boolean;
```
aaron-steinfeld, line 110: Curious what the use case is to ignore the global environment filter - that sounds like a big on the caller side.

arjunlalb: purpose of this is on some of the config pages which allow for environment selection. We want subsequent API/service selection in such workflows to be limited by the environment choice the user made.

PR #1965, opened 2022-12-21
fix: add border and border radius property for label tag

itssharmasandeep:


projects/components/src/label-tag/label-tag.component.ts
```diff
@@ -30,4 +38,14 @@ export class LabelTagComponent {
...
   public prefixIcon?: string;
+
+  @Input()
+  public borderColor?: Color;
+
+  @Input()
+  public borderRadius: number = 0; // In Pixels
+
+  public get border(): string | undefined {
+    return this.borderColor === undefined ? undefined : `1px solid ${this.borderColor}`;
```
adisreyaj, line 49: Why not have the border (transparent) always so that there is no layout change.

itssharmasandeep: I don't think anyone would be setting and un-setting the borderColor for this particular component and didn't want to use border by default (just a preference). If needed, I can make this change

anandtiwary: We should ideally have a  `viewMode` enum property. The caller would pass a view property, and we would apply the style from inside this component. We need to restrict this component honestly. By adding new CSS properties, we are making it too generic.

arjunlalb: +1 on the viewmode enum.

itssharmasandeep: But in that case, how would we have the border-specific configurations?
Then will probably take text color as border color may be?

itssharmasandeep:
Not needed

PR #1966, opened 2022-12-22
fix(span-detai): error handling if tabs array is empty

adisreyaj:
## Description
image

- Component errors out when the `tabs` array is empty. Added extra check to prevent change of tabs if the tabs array is empty.
- Added no data message for span detail if there is no tabs.
- Summary value doesn't render if the value provided is `0`

### DIFF
--- a/projects/components/src/summary-value/summary-value.component.ts
+++ b/projects/components/src/summary-value/summary-value.component.ts
@@ -7,7 +7,7 @@ import { IconSize } from '../icon/icon-size';
   styleUrls: ['./summary-value.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="ht-summary-value" *ngIf="this.value" [htTooltip]="this.tooltipText">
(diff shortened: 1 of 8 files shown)


projects/components/src/summary-value/summary-value.component.ts
```diff
@@ -7,7 +7,7 @@ import { IconSize } from '../icon/icon-size';
   styleUrls: ['./summary-value.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="ht-summary-value" *ngIf="this.value" [htTooltip]="this.tooltipText">
+    <div class="ht-summary-value" *ngIf="!(this.value | htIsNil)" [htTooltip]="this.tooltipText">
```
itssharmasandeep, line 10: Why exactly do we need this pipe here?

itssharmasandeep: We wanna show this section if the string is empty, is it?

adisreyaj: Currently if value=0, it won't render. So the `isNil` checks for `value === undefined || value === null`

arjunlalb (approved):

PR #1967, opened 2022-12-22
fix: improvements to input pill list component

arjunlalb:
Screenshot 2022-12-22 at 3 34 39 PM

## Description
Improvements to pill list component.
1. Styling updates.
2. Support to pasting content.
3. Optional override for comma parsing

### Testing
N/A

### Checklist:

### DIFF
--- a/projects/components/src/form-field/form-field.component.ts
+++ b/projects/components/src/form-field/form-field.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
+import { Color } from '@hypertrace/common';
 import { IconSize } from '../icon/icon-size';
 
 @Component({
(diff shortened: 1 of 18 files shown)


projects/components/src/form-field/form-field.component.ts
```diff
@@ -25,6 +26,7 @@ import { IconSize } from '../icon/icon-size';
           'show-border': this.showBorder,
           'error-border': this.showFormError && this.errorLabel
         }"
+        [ngStyle]="{ 'background-color': this.contentBgColor }"
```
adisreyaj, line 29: Couldn't we work with `background-color: inherit` so that we don't have to rely on inputs to enable us to style items?

arjunlalb: content block is a child inside the form-field . so, inherit doesn't work.

arjunlalb: alternative is ng-deep. I prefer this over that.

adisreyaj (approved):

PR #1969, opened 2022-12-26
feat(stepper): provide custom stepper controls

adisreyaj:
## Description
Add ability to provide custom tab step controls which overrides the default footer controls in the stepper.

## Usage
```html
<ht-stepper class="stepper" [isLinear]="true">
  <ht-stepper-tab label="Save">
    <div>CONTENT</div>
    <ht-stepper-tab-controls #controls>
      <div class="action-buttons">
        <ht-button class="cancel" label="Cancel" role="${ButtonRole.Tertiary}"></ht-button>
        <div class="back-next">
          <ht-button class="back" display="${ButtonStyle.Outlined}" label="Back" (click)="controls.previous()"></ht-button>
          <ht-button class="next" label="Save & Continue" (click)="this.save()"></ht-button>
        </div>
      </div>
    </ht-stepper-tab-controls>
  </ht-stepper-tab>
</ht-stepper>
```

### Checklist:


projects/components/src/stepper/stepper.component.scss
```diff
@@ -47,20 +47,23 @@
...
-    display: flex;
-    justify-content: space-between;
-    z-index: 20;
 
-    .back-next {
+    .action-buttons {
       display: flex;
-      gap: 16px;
+      justify-content: space-between;
+      z-index: 20;
```
23nobody, line 61: Any specific reason for number `20`?

adisreyaj: Not added as part of this PR.

It was added before so that the footer is not covered by the content.

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
```diff
@@ -0,0 +1,23 @@
...
+    super();
+  }
+
+  public next(): void {
+    this.stepper?.next();
+  }
+
+  public previous(): void {
+    this.stepper?.previous();
+  }
```
23nobody, line 22: I believe provided templates should call these methods. Need to find a way to mandate that for a user?

adisreyaj: These are some basic methods that the custom control would want to call, so we expose them. The consumer could do this:
```html
<ht-stepper-tab-controls #controls> <!-- Get access to component instance -->
      <div>
        <div class="back-next">
          <ht-button (click)="controls.stepper.previous()"></ht-button> <!-- This here on click -->
        </div>
      </div>
    </ht-stepper-tab-controls>
```

projects/components/src/stepper/stepper.component.ts
```diff
@@ -151,6 +160,14 @@ export class StepperComponent implements AfterContentInit {
...
   }
 
+  public next(): void {
+    this.stepper?.next();
+  }
+
+  public previous(): void {
+    this.stepper?.previous();
+  }
+
```
23nobody, line 170: Getting confused amongst these and the ones which are inside `StepperTabComponent`. Can you elaborate diff usages?

adisreyaj: https://github.com/hypertrace/hypertrace-ui/pull/1969#discussion_r1057474260

For making it easier for the custom controls to perform do common actions like next and prev, we expose the methods `next` and `previous` in the `ht-stepper-tab-controls` component. So the consumer can get the reference to the component by adding a `#variable`:
```html
<ht-stepper-tab-controls #controls></ht-stepper-tab-controls>
```

They can access methods exposed by the `StepperTabControlsComponent` by doing:
```html
<button (click)="controls.previous()"></button>
```

projects/components/src/stepper/stepper.component.ts
```diff
@@ -48,29 +48,38 @@ import { StepperTabComponent } from './stepper-tab.component';
...
-          ></ht-button>
-          <ht-button
-            class="next"
-            [label]="this.getActionButtonLabel | htMemoize: stepper.selectedIndex:steps"
-            (click)="this.nextOrSubmit(stepper)"
-            [disabled]="this.isNextDisabled(stepper.selectedIndex)"
-          ></ht-button>
-        </div>
-      </div>
+      <footer class="footer">
```
23nobody, line 51: this can be moved inside `ng-container` at L33 ?

adisreyaj: It has styles applied which is needed.

projects/components/src/stepper/tab/stepper-tab.component.ts
```diff
@@ -1,15 +1,19 @@
...
 
 @Component({
   selector: `ht-stepper-tab`,
   template: CONTENT_HOLDER_TEMPLATE,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class StepperTabComponent extends ContentHolder {
+  @ContentChild(StepperTabControlsComponent)
+  public tabControls?: StepperTabControlsComponent;
+
```
23nobody, line 16: Do we also want to define positions of buttons? If that is the case we can also go with just taking in templates as input. Eg. `backTemplate`, `nextTemplate`

adisreyaj: We could leave it for the consumer to decide what they want to show.

projects/components/src/stepper/stepper.component.ts
23nobody: Do we really need the methods which are defined here?

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Component({
+  selector: 'ht-stepper-tab-controls',
+  template: CONTENT_HOLDER_TEMPLATE,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class StepperTabControlsComponent extends ContentHolder {
+  public constructor(public stepper: StepperComponent, public step: StepperTabComponent) {
+    super();
+  }
```
arjunlalb, line 14: do we need the parent elements to be public?

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
23nobody: What happens if we have two tabs defined `tab1`, `tab2` and two set of controls `#controls1`, `#controls2`. 
Now, active tab is `tab1` but user calls `#controls2`?

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
adisreyaj: Not right now. Either we expose APIs like we did below for `next` and `previous` or allow access to stepper which gives more access to the custom controls.

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
23nobody: I believe we should expose `next/back` controls from a single place i.e. from `StepperComponent` instead of tab level i.e. `StepperTabComponent`.

projects/components/src/stepper/stepper.component.ts
adisreyaj: If we make the `stepper` private, we'll need to expose these methods.

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
23nobody: For any handling on the parent we can emit events

adisreyaj: It is all referencing the same `stepper` instance. So nothing could go wrong here, it just for convenience.

23nobody: > What happens if we have two tabs defined `tab1`, `tab2` and two set of controls `#controls1`, `#controls2`. Now, active tab is `tab1` but user calls `#controls2`?

Accessing same instance from different places can get cumbersome? Component should expose only one API for a single type of action.

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
adisreyaj: Marked as `private`

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
```diff
@@ -0,0 +1,23 @@
...
+import { StepperComponent } from '../stepper.component';
+import { StepperTabComponent } from '../tab/stepper-tab.component';
+
+@Component({
+  selector: 'ht-stepper-tab-controls',
+  template: CONTENT_HOLDER_TEMPLATE,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class StepperTabControlsComponent extends ContentHolder {
+  public constructor(private readonly stepper: StepperComponent, public readonly step: StepperTabComponent) {
```
arjunlalb, line 12: why do we need the `step` here?

adisreyaj: If the consumer would want to access any information in the step. Not used now.

projects/components/src/stepper/tab-controls/stepper-tab-controls.component.ts
adisreyaj: Stepper instance inside controls is made private now.

arjunlalb (approved):

PR #1975, opened 2022-12-28
feat: undo icon

itssharmasandeep:


Exac (approved):
It is a pity we don't have something like Storybook to preview these icons.

itssharmasandeep:
> It is a pity we don't have something like Storybook to preview these icons.

Whatever we put in the PR, GitHub actually gives us an preview

Exac:
> > It is a pity we don't have something like Storybook to preview these icons.
> 
> Whatever we put in the PR, GitHub actually gives us an preview

Yeah, I see the preview in GitHub's PR GUI, but it would be nice to have a cataloge of all the re-usable components.

PR #1977, opened 2022-12-29
feat: add load more for the code viewer

itssharmasandeep:


projects/components/src/viewer/code-viewer/code-viewer.component.ts
```diff
@@ -74,6 +77,24 @@ import { SyntaxHighlighterService } from './syntax-highlighter/syntax-highlighte
           <ht-message-display icon="${IconType.NoData}" [title]="this.noDataMessage"></ht-message-display>
         </ng-template>
       </div>
+      <div class="footer">
```
anandtiwary, line 80: Instead of load more, can we explore virtual scroll bar?

anandtiwary: https://material.angular.io/cdk/scrolling/api#CdkVirtualForOf

PR #1986, opened 2023-01-06
fix: restrict ht-filter-bar max height

arjunlalb:
## Description
Specify max height (equal to 5 rows) for advanced filter bar to avoid filter bar taking up all the available space when there are many filters in explorer page.

### Testing
Verified

### Checklist:


projects/components/src/filtering/filter-bar/filter-bar.component.scss
```diff
@@ -6,11 +6,13 @@
   flex-direction: column;
   justify-content: center;
   min-height: 36px;
+  overflow-y: scroll;
   width: 100%;
   padding: 0 1px 0 8px;
   background: white;
   border: 1px solid $color-border;
   border-radius: 6px;
+  max-height: 180px;
```
itssharmasandeep, line 15: why and where this change is needed?

adisreyaj (approved):

projects/components/src/filtering/filter-bar/filter-bar.component.scss
```diff
@@ -6,11 +6,13 @@
   flex-direction: column;
   justify-content: center;
   min-height: 36px;
+  overflow-y: scroll;
```
anandtiwary, line 9: This is debatable. Can we get ux team's opinion?

anandtiwary: Also, this is the advanced filter bar with the same name. It needs refactoring.

arjunlalb: @anandtiwary intention was to do it on the advance filter bar in the explorer page. the bug was filed for that.

projects/components/src/filtering/filter-bar/filter-bar.component.scss
arjunlalb: description :)

PR #1997, opened 2023-01-12
fix: created shared form validators

Greedy-Geek:


projects/common/src/utilities/validators/form-validators.ts
```diff
@@ -0,0 +1,14 @@
...
+import { areEmailAddressesValid } from './email-validator';
+
+export const validateCommaSeparatedEmails = (control: AbstractControl) => {
+  const emails = getStringsFromCommaSeparatedList(control.value);
+
+  // tslint:disable: no-null-keyword
+  return control.value.length === 0
+    ? { emptyEmails: true }
+    : !areEmailAddressesValid(emails)
+    ? { invalidEmails: true }
```
itssharmasandeep, line 12: IMO, having two different error types for the same validator is not a good idea. It is confusing to me. It will anyways give you an invalid state with the same object.

projects/common/src/utilities/validators/form-validators.ts
```diff
@@ -0,0 +1,14 @@
+import { AbstractControl } from '@angular/forms';
+import { getStringsFromCommaSeparatedList } from '../formatters/string/string-formatter';
+import { areEmailAddressesValid } from './email-validator';
+
+export const validateCommaSeparatedEmails = (control: AbstractControl) => {
```
itssharmasandeep, line 5: I think `commaSeparatedEmailsValidator` is a better name for this validator.
IMO, we should namespace all the custom validators (at least the new ones) with something like
`CustomValidators.commaSeparatedEmails`.
Here it will be easier to do since we only have one validator for now. We can move others later to this.

itssharmasandeep:
Can we please write the tests as well, if possible?

projects/common/src/utilities/validators/form-validators.ts
Greedy-Geek: updated

Greedy-Geek:
> Can we please write the tests as well, if possible?

added tests

projects/common/src/utilities/validators/form-validators.ts
```diff
@@ -0,0 +1,12 @@
+import { AbstractControl } from '@angular/forms';
+import { getStringsFromCommaSeparatedList } from '../formatters/string/string-formatter';
+import { areEmailAddressesValid } from './email-validator';
+
+export namespace CustomFormValidators {
```
itssharmasandeep, line 5: You'll need a lint rule probably

Greedy-Geek: updated

projects/common/src/utilities/validators/form-validators.test.ts
```diff
@@ -0,0 +1,21 @@
+import { FormControl } from '@angular/forms';
+import { CustomFormValidators } from '@hypertrace/common';
+
+// tslint:disable: no-null-keyword
+describe('CustomFormValidators', () => {
```
itssharmasandeep, line 5: nit: should we create a single test for a validator, I think this is a good use case to do that, since in the future we may have multiple validators, and keeping 1 for 1, would be easy to read. Use comments to tell what we're going to test.

"commaSeparatedEmails validator should work as expected'" or something like that??

Greedy-Geek: description contains the parent i.e., "CustomFormValidators".
The individual tests are where name of the validators like "commaSeparatedEmails" are used

itssharmasandeep (approved):

projects/common/src/utilities/validators/form-validators.ts
```diff
@@ -0,0 +1,13 @@
+import { AbstractControl } from '@angular/forms';
+import { getStringsFromCommaSeparatedList } from '../formatters/string/string-formatter';
+import { areEmailAddressesValid } from './email-validator';
+
+// tslint:disable-next-line: no-namespace
+export namespace CustomFormValidators {
```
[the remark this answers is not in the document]
itssharmasandeep: In the ideal scenario, we can have something like @hypertrace/forms and use `HtValidators ` or `CustomValidators` to not use "Form" in between

projects/common/src/utilities/validators/form-validators.ts
```diff
@@ -0,0 +1,13 @@
...
+  export const commaSeparatedEmails = (control: AbstractControl) => {
```
adisreyaj, line 7: I think we should just call it `email` and also handle the `single` or `multiple` (comma separated) case in the same validator.

```ts
const emailControl = new FormControl('', [HtValidators.email])
```

Greedy-Geek: Not sure, won't it reduce its readability? It can validate single email too though

Greedy-Geek: Because there can be many ways emails can land up here if we call it just email(s), single email, comma separated emails, array of emails (which it is not ready for yet), etc

projects/common/src/utilities/validators/form-validators.ts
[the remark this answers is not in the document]
Greedy-Geek: It is meant for validators in forms. So that's why I think `CustomFormValidators` is the right name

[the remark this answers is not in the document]
itssharmasandeep: That depends, on how you look at it, validators in angular forms are not named `FormValidators`, they are just `Validators`. That's understandable because we're importing from `@angular/forms`. SO Currently it is fine, because we're importing from `@hypertrace/common`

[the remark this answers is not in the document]
Exac: Agreed, if we are able to package it in `@hypertrace/forms`, we can export it as `HtValidators`.

Speaking of which, is it a lot of work to setup a project like `@hypertrace/forms`, or something preventing us from doing it now (complexity setting it up?). Then we don't have to make breaking changes (moving from `@hypertrace/common` to `@hypertrace/forms` later on).

PR #1998, opened 2023-01-13
fix(Time): added method to return Time from a label

Greedy-Geek:


projects/common/src/time/time.ts
```diff
@@ -32,11 +32,27 @@ export class Time {
     return this._date;
   }
 
+  public static timeFromLabel(label: string, isUTC: boolean = false): Time {
```
arjunlalb, line 35: please rename to `parse` :)

Greedy-Geek: updated

projects/common/src/time/time.ts
```diff
@@ -32,11 +32,27 @@ export class Time {
     return this._date;
   }
 
+  public static parse(label: string, isUTC: boolean = false): Time {
+    const time: string[] = label.split(' ');
```
adisreyaj, line 36:
```ts
const [time, meridian]: string[] = label.split(' ');
const [hours, minutes, seconds, milliseconds] = time.split(':').map(value => +value);
```

If you want strong typing:
```ts
public static parse(label: TimeString, isUTC: boolean = false): Time {}

export type Meridian = 'AM' | 'PM' | 'am' | 'pm'
export type TimeString = `${string}:${string} ${Meridian}` | `${string}:${string}:${string} ${Meridian}` | `${string}:${string}:${string}:${string} ${Meridian}`
```

projects/common/src/time/time.ts
```diff
@@ -32,11 +32,27 @@ export class Time {
...
+    const meridian: string = time.pop()!;
+    const [hours, minutes, seconds, milliseconds] = time[0].split(':').map(value => +value);
```
arjunlalb, line 38: do we always get 4 components here? if not, there could be undefined values here.

projects/common/src/time/time.ts
```diff
@@ -32,11 +32,27 @@ export class Time {
...
+
   public toISOString(): string {
     return this.date.toISOString().substring(11);
   }
 
   public equals(other?: Time): boolean {
     return this.toISOString() === other?.toISOString();
   }
+
+  private static get24HrClockHours(hours: number, meridian: string): number {
```
arjunlalb, line 51: meridian should be limited to `'AM' | 'PM'`. it'll be good to type it.

projects/common/src/time/time.ts
Greedy-Geek: Today `label` only return "hh:mm merdian", but kept seconds and milliseconds here so that we don't need to change it in the future if we decide to show seconds and milliseconds too in the label or something like that. This method can handle that. We anyways assign 0 to minutes, seconds and milliseconds if they are undefined

Greedy-Geek: image

projects/common/src/time/time.ts
```diff
@@ -32,11 +32,27 @@ export class Time {
...
+    if (meridian === 'AM') {
+      return hours === 12 ? 0 : hours;
+    }
+
+    return hours === 12 ? 12 : hours + 12;
+  }
```
arjunlalb, line 57: can be shortened to

```
return meridian === 'AM' ? hour %12 : hour + 12
```

Greedy-Geek: oh yeah, mods

Greedy-Geek: updated

projects/common/src/time/time.ts
Greedy-Geek: updated

projects/common/src/time/time.ts
Greedy-Geek: updated

projects/common/src/time/time.ts
Greedy-Geek: > can be shortened to
> 
> ```
> return meridian === 'AM' ? hour %12 : hour + 12
> ```

This will fail at 12:30 PM, updated to 
`return meridian === 'AM' ? hours % 12 : hours === 12 ? 12 : hours + 12;`

projects/common/src/time/time.ts
```diff
@@ -32,11 +32,29 @@ export class Time {
     return this._date;
   }
 
+  public static parse(label: TimeString, isUTC: boolean = false): Time {
+    const time: [string, Meridian] = label.split(' ') as [string, Meridian];
+    const meridian: Meridian = time[1];
+    const [hours, minutes, seconds, milliseconds] = time[0].split(':').map(value => +value);
```
arjunlalb, line 38: do not do implicit conversion using `+`. Use numberCoercer instead.

projects/common/src/time/time.ts
```diff
@@ -32,11 +32,29 @@ export class Time {
...
+  private static get24HrClockHours(hours: number, meridian: Meridian): number {
+    return meridian === 'AM' ? hours % 12 : hours + 12;
+  }
 }
+
+export type Meridian = 'AM' | 'PM';
+export type TimeString =
+  | `${string}:${string} ${Meridian}`
+  | `${string}:${string}:${string} ${Meridian}`
+  | `${string}:${string}:${string}:${string} ${Meridian}`;
```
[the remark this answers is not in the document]
Greedy-Geek: okay, updated 😅

projects/common/src/time/time.ts
Greedy-Geek: got it, saw it being used somewhere else also, that's why I thought it would be fine.

Greedy-Geek: Updated

projects/common/src/time/time.ts
aaron-steinfeld: I'm really questioning this use case - where are we getting a non ISO formatted date that we'd want to parse? This sounds like we're fixing that should be fixed downstream.

projects/common/src/time/time.ts
```diff
@@ -32,6 +32,19 @@ export class Time {
     return this._date;
   }
 
+  public static parse(time: string, isUTC: boolean = false): Time {
```
aaron-steinfeld, line 35: The time string will have the time zone encoded in it - shouldn't be a separate param. For example in your test above, you gave 4 utc times, and only had it treat 2 of them as utc - the other two give the appear to be local times, but that's incorrect.

I would suggest you produce a UTC `Time` object out of the time string regardless of it's timezone:
```ts
    return new Time(
      scheduledDate.getUTCHours(),
      scheduledDate.getUTCMinutes(),
      scheduledDate.getUTCSeconds(),
      scheduledDate.getUTCMilliseconds(),
      true
    );
```

Greedy-Geek: oh okay

Greedy-Geek: updated

projects/common/src/time/time.test.ts
```diff
@@ -19,4 +19,11 @@ describe('Time', () => {
     expect(mockedTime.equals(new Time(8, 30))).toBeTruthy();
     expect(mockedTime.equals(new Time(9, 30))).toBeFalsy();
   });
+
+  test('can return correct Time from an ISOString', () => {
+    expect(Time.parse('00:00:00.000Z')).toEqual(new Time(0, 0, 0, 0, true));
+    expect(Time.parse('12:30:00.000Z')).toEqual(new Time(12, 30, 0, 0, true));
+    expect(Time.parse('02:00:00.000Z')).toEqual(new Time(2, 0, 0, 0, true));
```
aaron-steinfeld, line 26: nit a test with a non UTC time zone would be a good idea

aaron-steinfeld (dismissed):

Exac (dismissed):
FYI, as promised, the `E164Number` type I was looking for last night is here:
https://github.com/catamphetamine/libphonenumber-js/blob/9b25e3df0f231feee4e1961e41decb2caa8619b1/types.d.ts#L47

The implementation is:
```typescript
type Tagged<A, T> = A & { __tag?: T };
type E164Number = Tagged<string, "E164Number">;
const phone: E164Number = '+11231231234';
```

Greedy-Geek:
> FYI, as promised, the `E164Number` type I was looking for last night is here: https://github.com/catamphetamine/libphonenumber-js/blob/9b25e3df0f231feee4e1961e41decb2caa8619b1/types.d.ts#L47
> 
> The implementation is:
> 
> ```ts
> type Tagged<A, T> = A & { __tag?: T };
> type E164Number = Tagged<string, "E164Number">;
> const phone: E164Number = '+11231231234';
> ```

Thanks for this!

Exac (approved):
No problem. That was created before template literal types were added to TypeScript, so perhaps aren't the most optimal way to go.

PR #2000, opened 2023-01-13
Paginator visibllity

Exac:
- This reverts commit 1aac787bd0f31f0bd5e7096b6bed98583b057b73.

## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/paginator/paginator.component.ts
+++ b/projects/components/src/paginator/paginator.component.ts
@@ -20,7 +20,11 @@ import { PaginationProvider } from './paginator-api';
   styleUrls: ['./paginator.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="paginator" [class.compact]="this.showCompactView">
+    <div
+      class="paginator"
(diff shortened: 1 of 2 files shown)


aaron-steinfeld (approved):

projects/components/src/paginator/paginator.component.ts
```diff
@@ -115,6 +119,8 @@ export class PaginatorComponent implements OnChanges, PaginationProvider {
   // Caused either by a change in the provided page, or user change being emitted
   public readonly pageEvent$: Observable<PageEvent> = merge(this.pageChange, this.pageSizeInputSubject);
 
+  public readonly minItemsBeforeDisplay: number = 10;
```
Exac, line 122: I think this should be configurable (by the parent component), but I don't want to expose it with `@Input()` and then be faced with the prospect of making breaking changes later on. Also, we may be able to come up with a better name if we are to expose this to the parent.

aaron-steinfeld: That's fair. Let's wait for our first use case where we want to modify it, then we can cross that bridge.

PR #2016, opened 2023-01-26
feat: add backdrop to sheet popover

Exac:
## Description

This PR adds a configurable backdrop option to the sheet popover component. When configured, the sheet popover component will render as normal, but the user will not be able to interact with any components they can see "behind" the popover sheet, eg: the environment selector.

The sheet popover  can be configured with a visible, opaque grey backdrop, or an invisible backdrop. By default, the backdrop is not created when opening a sheet popover, and there should be no regression.

### Testing

I added tests to the `popover.service.ts` to ensure the popover is not applied to sheets that do not have popover configuration supplied, and a test to ensure the popover background renders when configured.

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/overlay/overlay.service.ts
+++ b/projects/components/src/overlay/overlay.service.ts
@@ -4,7 +4,7 @@ import { PopoverBackdrop, PopoverFixedPositionLocation, PopoverPositionType } fr
 import { PopoverRef } from '../popover/popover-ref';
 import { PopoverService } from '../popover/popover.service';
 import { DefaultSheetRef } from './sheet/default-sheet-ref';
-import { SHEET_DATA, SheetOverlayConfig, SheetRef } from './sheet/sheet';
+import { SheetOverlayConfig, SheetRef, SHEET_DATA } from './sheet/sheet';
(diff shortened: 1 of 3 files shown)


projects/components/src/overlay/overlay.service.ts
```diff
@@ -32,7 +32,8 @@ export class OverlayService {
         type: PopoverPositionType.Fixed,
         location: config.position ?? PopoverFixedPositionLocation.RightUnderHeader
       },
-      data: metadata
+      data: metadata,
+      backdrop: config.backdrop ?? PopoverBackdrop.None
```
anandtiwary, line 36: How are we handling backdrop styling?

Exac: Users have three options, they can pass in:
- `PopoverBackdrop.None`
- `PopoverBackdrop.Opaque`
- `PopoverBackdrop.Transparent`

The opaque background applies the default backdrop from the Material library. The transparent backdrop is invisible. If we want to enable more styling options for the backdrop, we could refactor this to allow developers to pass in their own CSS class name.

anandtiwary (approved):

PR #2022, opened 2023-02-01
fix(stepper): goto step handling

adisreyaj:
## Description
Updated the logic for `goToStep` method.

### Testing
Tests added to verify the new logic.

### DIFF
--- a/projects/components/src/stepper/stepper.component.test.ts
+++ b/projects/components/src/stepper/stepper.component.test.ts
@@ -173,4 +173,71 @@ describe('Stepper Component', () => {
     expect(submitButton.label).toBe('Click me!');
     expect(submitButton.disabled).toBeTruthy();
   });
+
+  test('should be able to goto particular step', () => {
+    spectator = createHost(
+      `<ht-stepper class="stepper" [isLinear]="isLinear">
+            <ht-stepper-tab label="Hello" [completed]="tabOneStatus"> Hello World!</ht-stepper-tab>
+            <ht-stepper-tab label="Yo" [completed]="tabTwoStatus"> Hey!</ht-stepper-tab>
+            <ht-stepper-tab label="Kem cho" [completed]="tabThreeStatus" actionButtonLabel="Click me!"> Test!</ht-stepper-tab>
+          </ht-stepper>`,
+      {
+        hostProps: {
+          isLinear: true,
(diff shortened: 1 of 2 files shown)


projects/components/src/stepper/stepper.component.test.ts
```diff
@@ -173,4 +173,71 @@ describe('Stepper Component', () => {
...
+          tabOneStatus: true,
+          tabTwoStatus: true,
+          tabThreeStatus: false,
+          tabFourStatus: false
+        }
+      }
+    );
+
+    spectator.component.goToStep(3);
+    expect(spectator.query(MatStepper)?.selectedIndex).toBe(0);
```
Greedy-Geek, line 218: Shouldn't we able to jump to step 3 if step 1 and step 2 are complete?

Greedy-Geek: Oh wait, steps are 0 indexed, right? Ignore the comment

projects/components/src/stepper/stepper.component.ts
```diff
@@ -203,6 +211,16 @@ export class StepperComponent implements AfterContentInit {
   private isLastStep(stepper: MatStepper): boolean {
     return stepper.selectedIndex === stepper.steps.length - 1;
   }
+
+  /**
+   * Check if all the steps in between the selected index and the index are completed.
+   * Mandatory for linear stepper when navigating forward.
+   */
+  private areStepsInBetweenCompleted(selectedIndex: number, index: number): boolean {
```
Greedy-Geek, line 219: `selected index` here is the currently active index, right? I think we should rename it to `active index` and `index` can be renamed to `selected index`

adisreyaj: @23nobody Kept the naming in sync with what material is using I believe.

itssharmasandeep: I think we should move the `isGoingForward` code in this function itself.
and change the condition above
`this.isLinear  ? this.areStepsInBetweenCompleted(this.stepper.selectedIndex, index) : true`

adisreyaj: Updated.

itssharmasandeep (approved):

23nobody (approved):

PR #2043, opened 2023-02-17
feat: introduce bar gauge sizing

itssharmasandeep:


projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -98,6 +98,9 @@ export class BarGaugeComponent implements OnChanges, AfterViewInit {
   @Input()
   public display: BarGaugeStyle = BarGaugeStyle.Regular;
 
+  @Input()
```
Christian862, line 101: nit: Since it only applies to single bar gauges right now we should include a comment here mentioning that

Christian862 (approved):

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
itssharmasandeep: Done

projects/observability/src/shared/components/bar-gauge/bar-gauge.component.ts
```diff
@@ -98,6 +98,9 @@ export class BarGaugeComponent implements OnChanges, AfterViewInit {
...
+  public size: BarGaugeSize = BarGaugeSize.Small; // Only used for single-bar display type as of now
```
anandtiwary, line 102: useful but we should also try to have a consistent ux.

PR #2045, opened 2023-02-17
fix: Fixing Null argument in GraphqlRequestBuilder

palbizu:
## Description
Prevents generating and Argument with an empty object '{}' when the argument is null.

### Testing
Tested locally

### Checklist:

### Documentation


projects/graphql-client/src/utils/builders/request/graphql-request-builder.ts
```diff
@@ -132,7 +132,7 @@ export class GraphQlRequestBuilder {
       return `[${arrayContents}]`;
     }
 
-    if (typeof argumentValue === 'object' && !(argumentValue instanceof Date)) {
+    if (typeof argumentValue === 'object' && !isNil(argumentValue) && !(argumentValue instanceof Date)) {
```
aaron-steinfeld, line 135: See my earlier suggestion offline - we should translate any nil argument as null. this would handle null correctly, but undefined would not be converted to null.

palbizu: done

aaron-steinfeld (approved):

PR #2046, opened 2023-02-17
feat: provide option to use offset timezone format ISO string

jake-bassett:
## Description

### Testing

### Checklist:

### Documentation


aaron-steinfeld (commented):
Could this use our existing formatter (may need to add a new format string) instead of implementing by hand?

projects/common/src/time/time.ts
```diff
@@ -45,10 +45,39 @@ export class Time {
     );
   }
 
-  public toISOString(): string {
+  public toISOString(useTimezoneOffset: boolean = false): string {
```
Exac, line 48: I think we should add a test for this.

projects/common/src/utilities/formatters/date/date-formatter.ts
```diff
@@ -72,11 +76,23 @@ export class DateFormatter {
...
-    return formatDate(coercedValue, this.getFormatString(), 'en_US');
+    return formatDate(coercedValue, this.getFormatString(coercedValue), 'en_US');
   }
 
-  private getFormatString(): string {
+  private getFormatString(date: Date): string {
     switch (this.options.mode) {
+      case DateFormatMode.TimeWithTimeZoneOffset:
+        /*
+         * WARNING: This format is currently unsupported by at least some backend APIs e.g. Reporting
```
aaron-steinfeld, line 86: Not relevant here

projects/common/src/utilities/formatters/date/date-formatter.ts
```diff
@@ -72,11 +76,23 @@ export class DateFormatter {
...
+         */
+        const timezone = date.getTimezoneOffset();
+        const sign = timezone <= 0 ? '+' : '-';
+
+        const pad = (n: number): string => `${n < 10 ? '0' : ''}${n}`;
```
aaron-steinfeld, line 91: This still shouldn't be needed. Angular format definitely supports ISO, right?

projects/common/src/utilities/formatters/date/date-formatter.ts
```diff
@@ -72,11 +76,23 @@ export class DateFormatter {
...
+
+        const offset = `${sign}${pad(Math.floor(timezone / 60))}:${pad(timezone % 60)}`;
+
+        return `hh:mm:ss${offset}`;
```
aaron-steinfeld, line 95: Isn't this the 12 hour format?

projects/common/src/utilities/formatters/date/date-formatter.test.ts
```diff
@@ -28,6 +28,14 @@ describe('Date formatter', () => {
...
-    ).toEqual('19 Aug 2021 11:35 PM GMT+00:00');
+    ).toEqual('19 Aug 2021 1:02 PM GMT+00:00');
+  });
+
+  test('can format a date string with and time with offset time zone', () => {
+    expect(
+      new DateFormatter({
+        mode: DateFormatMode.TimeWithTimeZoneOffset
+      }).format(dateString)
+    ).toEqual('01:02:03+00:00');
```
aaron-steinfeld, line 39: This expectation is wrong, it's showing the bug I described in my other comment. This time should match the original time if they're both utc, but a better test would be to use a different time zone for input.

projects/common/src/utilities/formatters/date/date-formatter.ts
jake-bassett: Ah, good catch. Thanks.

projects/common/src/utilities/formatters/date/date-formatter.ts
jake-bassett: Ah, my mistake. I was looking at the native JS time formatting options. Angular expands what is available. Updated. Thanks!

projects/common/src/utilities/formatters/date/date-formatter.ts
jake-bassett: Eeeeh, I think its useful here too, but its fine. Removed.

projects/common/src/utilities/formatters/date/date-formatter.test.ts
jake-bassett: The tests run in GMT, so the timezone stuff is a little wonky in the tests. Let me update the impl to allow specifying timezones so we can have better tests.

projects/common/src/utilities/formatters/date/date-formatter.ts
```diff
@@ -58,25 +62,27 @@ export class DateFormatter {
     this.options = this.applyOptionDefaults(options);
   }
 
-  public format(value: Date | number | undefined | string): string {
-    return this.convertDateToString(value);
+  public format(value: Date | number | undefined | string, timezone?: string): string {
```
aaron-steinfeld, line 65: I thought a date already encodes a time zone? If I'm mistaken, this makes sense. If it does, we shouldn't accept a different one since that's already expressible and leads to ambiguity

jake-bassett: Yeah, `Date` does have timezone information, the issue though is that the formatter takes a union that may or may not be a `Date` so the extra param would _sometimes_ be necessary.

Maybe I should name it `timezoneOverride`? Not sure; open to suggestions.

aaron-steinfeld: My 2p: the caller has the option of using a date object if they wish to include time zone information, the string and number versions are just conveniences and wouldn't try to make the api support using them with a custom time zone.

jake-bassett: Okay, so played with this a bit more and while you can apply timezone info to `Date`, when that gets formatted it it just adjusts the time to zulu time. You can see this in the matching test. If our `Date` is built with timezone info like `2021-08-19T14:23:56.789-08:30`, once we run it through the formatter without a `timezone` arg, it spits out the zulu time equivalent of `22:53:56Z` and the original timezone is lost.

Keeping the `timezone` argument is actually additional information that tells the formatter to spit out the time for that timezone. So in the same example passing `2021-08-19T14:23:56.789-08:30` with `timezone` set to `PST` would provide us the expected `14:53:56-08:00`.

aaron-steinfeld (approved):

PR #2055, opened 2023-02-21
feat: add panel collapsed body component

itssharmasandeep:
`UseCase` -> We want to have some minimized content when the Panel is collapsed.
`Implementation` -> Added a new `ht-panel-collapsed-body` component

### DIFF
--- a/projects/components/src/panel/panel.component.test.ts
+++ b/projects/components/src/panel/panel.component.test.ts
@@ -41,6 +41,9 @@ describe('Panel component', () => {
           <div class="title">{{title}}</div>
           <div class="summary">{{summary}}</div>
         </ht-panel-header>
+        <ht-panel-collapsed-body>
+          <div>Collapsed Body Content</div>
+        </ht-panel-collapsed-body>
         <ht-panel-body>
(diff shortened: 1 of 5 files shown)


projects/components/src/panel/panel.component.test.ts
```diff
@@ -56,9 +59,11 @@ describe('Panel component', () => {
     expect(spectator.query('.title')).toHaveText('Results');
     expect(spectator.query('.summary')).toHaveText('(508 Records)');
     expect(spectator.query('.body')).not.toExist();
+    expect(spectator.query('.collapsed-body')).toHaveText('Collapsed Body Content');
 
     spectator.click('.title');
     expect(spectator.query('.body')).toHaveText('Body Content');
+    expect(spectator.query('.collapsed-body')).not.toExist();
```
arjunlalb, line 66: also add a test to check absence of collapsed-body when its not supplied. i.e, the default case.

itssharmasandeep: done

arjunlalb (approved):

PR #2056, opened 2023-02-23
fix: separate mat-icon implementation for svg and ligature

arjunlalb:
## Description
mat-icon component does not render correctly when we switch from svg icon to ligature icon in place.
We get a blank element instead of a ligature icon.

Using separate instances of mat-icon inside the ht-icon component to isolate this problem and supply only relevant values to each usage.

After
https://user-images.githubusercontent.com/63222211/220830360-068e39f7-6e22-45df-9ea6-2df07419c49b.mov

Before
https://user-images.githubusercontent.com/63222211/220830483-bbcb5f83-6950-4f4a-af59-ce01dc151252.mov

### Testing
Tested locally. Added UTs.

### Checklist:

### DIFF
--- a/projects/components/src/icon/icon.component.ts
+++ b/projects/components/src/icon/icon.component.ts
@@ -13,7 +13,7 @@ import { IconSize } from './icon-size';
     <ng-container>
       <mat-icon
         *ngIf="this.svgIcon; else matIconTemplate"
-        class="ht-icon"
+        class="ht-icon svg-icon"
         [ngClass]="this.styleClasses"
         [ngStyle]="this.customStyles"
         [attr.aria-label]="this.labelToUse"
@@ -23,7 +23,7 @@ import { IconSize } from './icon-size';
 
       <ng-template #matIconTemplate>
         <mat-icon
-          class="ht-icon"
+          class="ht-icon ligature-icon"
           [ngClass]="this.styleClasses"
           [ngStyle]="this.customStyles"
           [attr.aria-label]="this.labelToUse"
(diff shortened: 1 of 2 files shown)


projects/components/src/icon/icon.component.ts
```diff
@@ -9,21 +10,29 @@ import { IconSize } from './icon-size';
...
-        background: this.borderType === '${IconBorder.InsetBorder}' && this.borderColor ? this.borderColor : '',
-        borderRadius: this.borderRadius ? this.borderRadius : ''
-      }"
-      [attr.aria-label]="this.labelToUse"
-      [htTooltip]="this.showTooltip ? this.labelToUse : undefined"
-      [fontSet]="this.fontSet"
-      [svgIcon]="this.svgIcon"
-      >{{ this.ligatureText }}</mat-icon
-    >
+    <ng-container>
```
itssharmasandeep, line 13: Why do we need an additional ng-container?

projects/components/src/icon/icon.component.ts
```diff
@@ -82,4 +91,25 @@ export class IconComponent implements OnChanges {
...
+
+  public get styleClasses(): string[] {
+    return [this.size, this.borderType ? this.borderType : ''];
+  }
+
+  public get customStyles(): Record<string, string> {
+    return {
+      color: this.color ? this.color : '',
+      borderColor: this.borderType !== IconBorder.InsetBorder && this.borderColor ? this.borderColor : '',
+      background: this.borderType === IconBorder.InsetBorder && this.borderColor ? this.borderColor : '',
```
itssharmasandeep, line 107: repetitive, can make this into another getter.

projects/components/src/icon/icon.component.ts
arjunlalb: removed.

aaron-steinfeld (approved):

PR #2057, opened 2023-02-23
feat: add emit mode control to search box

arjunlalb:
## Description
Add an emit mode control to search box.

If emit mode is onSubmit, the debounceTime is set to 5s. This helps us achieve a behaviour of triggering a search when the user submits or 5s has elapsed since last input.

No changes to incremental mode search.

Adding an FF to control this behaviour across all query powered search usecases in the app.
Using this FF and associated pipe to switch to `onSubmit` search as required.

### Testing
UT added. Manually verified.

### Checklist:


projects/common/src/constants/application-constants.ts
```diff
@@ -1,4 +1,5 @@
 export const enum ApplicationFeature {
   PageTimeRange = 'ui.page-time-range',
-  FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map'
+  FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map',
+  DisableIncrementalSearch = 'ui.disable-incremental-search'
```
aaron-steinfeld, line 4: Make feature flags consistent, it should be enabling only. Add a true default for backwards compat.

projects/components/src/search-box/search-box.component.ts
```diff
@@ -46,6 +46,9 @@ export class SearchBoxComponent implements OnInit, OnChanges {
   @Input()
   public displayMode: SearchBoxDisplayMode = SearchBoxDisplayMode.Border;
 
+  @Input()
```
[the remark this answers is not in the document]
arjunlalb: We already have usecases. We don't want search usecases inside multi-select, select and other client side filtering needs to be trigger based.

projects/components/src/search-box/search-mode-on-submit-if-feature-enabled.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { FeatureState } from '@hypertrace/common';
+import { SearchBoxEmitMode } from './search-box.component';
+
+@Pipe({
+  name: 'htSearchModeOnSubmitIfEnabled',
+  // tslint:disable-next-line: no-pipe-impure delegating to impure async pipe
+  pure: false
```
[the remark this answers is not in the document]
arjunlalb: updated.

projects/common/src/constants/application-constants.ts
arjunlalb: updated.

projects/components/src/search-box/search-box.component.ts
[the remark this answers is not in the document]
arjunlalb: I left the FF logic out of the component so that the consumer has complete flexibility to pick the mode (based on one or more FFs or other parameters) and also to keep a widely used basic component as clean as possible.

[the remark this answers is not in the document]
aaron-steinfeld: I buy the select cases, but client side or not is an impl detail and they should have the same UX if it's the same control. Leaving the input with flag controlled default behavior gives the same functionality as you describe, it's just much easier to use on the typical consumer.

projects/common/src/constants/application-constants.ts
```diff
@@ -1,4 +1,5 @@
 export const enum ApplicationFeature {
   PageTimeRange = 'ui.page-time-range',
-  FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map'
+  FeatureDefaultTimeRangeMap = 'ui.feature-default-time-range-map',
+  EnableTriggerBasedSearch = 'ui.enable-trigger-based-search'
```
aaron-steinfeld, line 4: only question here is whether to remove `enable` from ff name/key for consistency.

projects/components/src/search-box/search-box.component.ts
```diff
@@ -89,9 +106,25 @@ export class SearchBoxComponent implements OnInit, OnChanges {
...
+      map(enabled => {
+        const defaultDebounceTime = this.debounceTime ?? 0;
+        // If incremental search mode is enabled, then use the inputs to compute debounce
+        if (this.searchMode === SearchBoxEmitMode.Incremental) {
+          return defaultDebounceTime;
+        }
+
+        // If on-submit search mode is enabled via the FF and the input, then use the overridden debounce time.
+        // Doing 'and' here to ensure the default behaviour continues until overridden by an FF.
+        return enabled && this.searchMode === SearchBoxEmitMode.OnSubmit ? 5000 : defaultDebounceTime;
```
aaron-steinfeld, line 126: nit - enabled here reads like whether the component is enabled, when it's more specific than that.

aaron-steinfeld (commented):
lgtm other than minor comments.

projects/common/src/constants/application-constants.ts
arjunlalb: updated

projects/components/src/search-box/search-box.component.ts
arjunlalb: updated

aaron-steinfeld (approved):

PR #2061, opened 2023-02-28
fix: added enhancements in TimeDuration class

23nobody:
## Description
Added a few enhancements to the `TimeDuration` class.
- Enabled support to query for `hours`, `minutes`, `seconds`, `days`, `months`, `years`.
- Exposed new logic for `Iso8601DurationString` conversion. The original one converted duration to seconds by default before computing. Exposed new one as `toIso8601DurationStringMilitaryTime()` that persists the units as per military time. For eg, 2 hours 20minutes translates to `PT2H20M`.

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/common/src/time/time-duration.ts
+++ b/projects/common/src/time/time-duration.ts
@@ -1,17 +1,24 @@
 import { parse, toSeconds } from 'iso8601-duration';
+import { toString } from 'lodash-es';
 import { assertUnreachable } from '../utilities/lang/lang-utils';
 import { TimeUnit } from './time-unit.type';
 
 export class TimeDuration {
   private readonly millis: number;
+  private readonly militaryTime: Date;
 
   public constructor(public readonly value: number, public readonly unit: TimeUnit) {
     this.toUnitString(); // Fail if unrecognized TimeUnit
     this.millis = this.normalizeToMillis(isNaN(value) ? 0 : value, unit);
+    this.militaryTime = this.buildMilitaryTime();
   }
 
   public toMillis(): number {
     return this.millis;
+  }
+
+  public toDate(): Date {
+    return this.militaryTime;
   }
 
   public static parse(durationString: string): TimeDuration {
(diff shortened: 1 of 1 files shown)


projects/common/src/time/time-duration.ts
```diff
@@ -111,6 +178,10 @@ export class TimeDuration {
     }
   }
 
+  private buildMilitaryTime(): Date {
+    return new Date(this.millis);
+  }
```
arjunlalb, line 183: what is military time? why is it required in a duration object?

this.millis is a number capturing the duration. it has nothing to do with a specific time. here, its used to create a timestamp. thats the equivalent of saying 3 hours can be converted into a specific timestamp, which is not possible.

projects/common/src/time/time-duration.ts
```diff
@@ -26,6 +29,70 @@ export class TimeDuration {
...
+    return this.militaryTime.getUTCHours();
+  }
+
+  public get minutes(): number {
+    return this.militaryTime.getUTCMinutes();
+  }
+
+  public get seconds(): number {
+    return this.militaryTime.getUTCSeconds();
+  }
```
arjunlalb, line 94: all of these don't belong in a `Duration` object

projects/common/src/time/time-duration.ts
```diff
@@ -26,6 +29,70 @@ export class TimeDuration {
     return `PT${this.toMillis() / 1000}S`;
   }
 
+  public toIso8601DurationStringMilitaryTime(): string {
```
arjunlalb, line 32: if the goal of this is to convert a duration object into an Iso8601 duration string, it should be done using the unit and value in the constructor.

arjunlalb (changes requested):

projects/common/src/time/time-duration.ts
```diff
@@ -1,13 +1,16 @@
 import { parse, toSeconds } from 'iso8601-duration';
+import { toString } from 'lodash-es';
 import { assertUnreachable } from '../utilities/lang/lang-utils';
 import { TimeUnit } from './time-unit.type';
 
 export class TimeDuration {
   private readonly millis: number;
+  private readonly militaryTime: Date;
```
aaron-steinfeld, line 8: please don't use the term "militaryTime" - it's not universally understood. Further, even though I think I know what it means, I'm not sure why we'd have both millis and this.

projects/common/src/time/time-duration.ts
aaron-steinfeld: Agreed. I'd also just update the existing method, that was just implemented with a simple approach due to laziness.

projects/common/src/time/time-duration.ts
```diff
@@ -1,13 +1,16 @@
...
 
   public constructor(public readonly value: number, public readonly unit: TimeUnit) {
     this.toUnitString(); // Fail if unrecognized TimeUnit
     this.millis = this.normalizeToMillis(isNaN(value) ? 0 : value, unit);
+    this.militaryTime = this.buildMilitaryTime();
```
aaron-steinfeld, line 13: Avoid no-arg methods that rely on state. If these two calls had their order flipped, it would break.

projects/common/src/time/time-duration.ts
```diff
@@ -26,6 +29,70 @@ export class TimeDuration {
     return `PT${this.toMillis() / 1000}S`;
   }
 
+  public toIso8601DurationStringMilitaryTime(): string {
+    const timeObject = this.toTimeObject();
```
aaron-steinfeld, line 33: Likewise, avoid stateful calls. I'm not sure why we need both a date object and a time object, they both accomplish the same thing.

projects/common/src/time/time-duration.ts
```diff
@@ -26,6 +29,70 @@ export class TimeDuration {
...
+
+    let timeString = 'P';
```
aaron-steinfeld, line 35: Avoid building time strings by hand like this - use angular's date formatter.

projects/common/src/time/time-duration.ts
```diff
@@ -26,6 +29,70 @@ export class TimeDuration {
...
+      seconds: this.seconds
+    };
+  }
+
+  public get years(): number {
+    return this.militaryTime.getUTCFullYear() - new Date(0).getUTCFullYear();
+  }
+
+  public get months(): number {
+    return this.militaryTime.getUTCMonth() - new Date(0).getUTCMonth();
```
aaron-steinfeld, line 77: All of these things based on the epoch instant are statics, and would be more clearly just defined as such. `new Date(0).getUTCMonth()` is always going to be 0 for example. `new Date(0).getUTCDate()` will always be 1 (because confusingly, days are the one unit that's based)

aaron-steinfeld (changes requested):

PR #2065, opened 2023-03-03
fix: slight update in url filter matching logic

23nobody:
## Description
There was a small glitch in how we match filters from `url` to the `filterAttributes` logic today. 
- To construct parameters for filter in the url, we generally use some key from the filterAttribute eg. `name` or `displayName`. 
- When we parse this filter from the url, we can pass the key names from the attributes that are used to construct the url param for the filter inside `nameFields`  argument in `tryParseStringForAttribute` method call.
-  The `tryParseStringForAttribute` method's job is to check whether the url string key starts with the attribute.
- `tryParseStringForAttribute` looks for `nameFields` one by one and looks for a match. The existing logic for this method was not fully correct.
- Here, 
```ts
const matchingNameField = nameFields.find(nameField =>
    stringContainingFullAttribute.toLowerCase().startsWith(attributeToTest[nameField].toLowerCase())
  );
```
matches some substring:
```ts
stringContainingFullAttribute -> 'domainName'
attributeToTest = {
  displayName: 'Domain',
  name: 'domainName'
}

const matchingNameField = 'displayName'; // 'domainname'.startsWith('domain') -> true
```

Then

---

Then
```ts
if (stringAfterAttributeName.match(/^\w/)) {
    return undefined;
  }
```
incorrectly identifies it as a bad match because of the trailing “name".

## Fix 🔧 : 
Applied a sort before the match logic based upon length of the `nameField` `value`. We check from `longest length of the key value` to the `shortest length of the key value` order to find the best possible match. (Because if the longer ones match, we don't need to check for the shorted ones then. But the opposite is not true as it causes the issue highlighted above cause shorter ones can be a prefix of the longer ones.)

### Before
```ts
// method -> `tryParseStringForAttribute`

/**
   * `attributeToTest` = {displayName: 'Domain', name:'domainName' }
   * `text` = "domainName"
   *  --> `displayName`
   */
  const matchingNameField = nameFields.find(nameField =>
    stringContainingFullAttribute.toLowerCase().startsWith(attributeToTest[nameField].toLowerCase())
  );
  if (!matchingNameField) {
    return undefined;
  }

  /**
   * `domainName`.slice("Domain".length) 
   *  --> `Name`
   */
  const stringAfterAttributeName = text.slice(attributeToTest[matchingNameField].length).trim();

  /**
   * `Name`.match(/^\w/) 
   *  --> true
   *    --> `undefined`
   */
  if (stringAfterAttributeName.match(/^\w/)) {
    return undefined;
  }

```
### After
```ts
// method -> `tryParseStringForAttribute`
  /**
   * `attributeToTest` = {displayName: 'Domain', name:'domainName' }
   * `text` = "domainName"
   *  --> `name`
   */
  const matchingNameField = nameFields
    .sort((key1, key2) => attributeToTest[key2].length - attributeToTest[key1].length)
    // --> ['name', 'displayName']
    .find(nameField =>
      stringContainingFullAttribute.toLowerCase().startsWith(attributeToTest[nameField].toLowerCase())
    );
  if (!matchingNameField) {
    return undefined;
  }

  /**
   * `"domainName".slice("domainName".length)`
   *  --> ``
   */
  const stringAfterAttributeName = text.slice(attributeToTest[matchingNameField].length).trim();

  /**
   * `"".match(/^\w/)`
   *  --> false 
   */
  if (stringAfterAttributeName.match(/^\w/)) {
    return undefined;
  }
```

### Testing

### Checklist:

### Documentation


adisreyaj (commented):
LGTM!

aaron-steinfeld (approved):

PR #2066, opened 2023-03-03
feat: add not contains key filter support

aaron-steinfeld:
## Description
Add support for not contains key operator


palbizu (approved):
LGTM!

PR #2074, opened 2023-03-07
feat: set selected items first - multiselect

palbizu:
## Description
Add selected items' first variable to reorder elements when open

### Testing
Tested locally

### Checklist:

### Documentation


projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -235,11 +239,13 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
...
-        isEmpty(searchText)
-          ? options.toArray()
-          : options.filter(option => option.label.toLowerCase().includes(searchText.toLowerCase()))
+    this.filteredOptions$ = combineLatest([
+      this.allOptions$,
+      this.caseInsensitiveSearchSubject,
+      this.reorderSelectedItemsSubject
+    ]).pipe(
+      map(([options, searchText, reorderSelectedItemsSubject]) =>
+        this.getFilteredOptions(options, searchText, reorderSelectedItemsSubject)
```
arjunlalb, line 248: we don't need to do anything special here. we just have to put the selected items on top of the list always.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -206,6 +206,9 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
   @Input()
   public customControlTemplate?: TemplateRef<unknown>;
 
+  @Input()
+  public setSelectedItemsFirst: boolean = false;
```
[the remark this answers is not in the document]
palbizu: removed, apply by default

projects/components/src/multi-select/multi-select.component.ts
palbizu: done

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -362,6 +375,18 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
     this.propagateControlValueChange?.(value);
     this.propagateControlValueChangeOnTouch?.(value);
   }
+
+  private reorderSelectedItemsFirst(
+    filteredOptions: SelectOptionComponent<V>[],
+    selected?: V[]
+  ): SelectOptionComponent<V>[] {
+    return isNil(selected)
```
arjunlalb, line 383: we should also be checking for empty.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -362,6 +375,18 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
...
+      ? filteredOptions
+      : [
+          ...filteredOptions.filter(filteredOption => selected.includes(filteredOption.value)),
+          ...filteredOptions.filter(filteredOption => !selected.includes(filteredOption.value))
```
arjunlalb, line 387: please use `partition` method from lodash.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -249,6 +257,11 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
     this.setTriggerLabel();
   }
 
+  public onPopoverOpen(): void {
+    this.popoverOpen = true;
+    this.reorderSelectedItemsSubject.next(true);
```
arjunlalb, line 262: we do not need this subject. can be removed. so, by extension, this method also isn't required.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -235,11 +236,18 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
 
   public ngAfterContentInit(): void {
     this.allOptions$ = this.allOptionsList !== undefined ? queryListAndChanges$(this.allOptionsList) : EMPTY;
-    this.filteredOptions$ = combineLatest([this.allOptions$, this.caseInsensitiveSearchSubject]).pipe(
+    this.filteredOptions$ = combineLatest([
+      this.allOptions$,
+      this.caseInsensitiveSearchSubject,
+      this.reorderSelectedItemsSubject
```
arjunlalb, line 242: not used.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -218,6 +218,7 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
 
   public filteredOptions$!: Observable<SelectOptionComponent<V>[]>;
   private readonly caseInsensitiveSearchSubject: BehaviorSubject<string> = new BehaviorSubject('');
+  private readonly reorderSelectedItemsSubject: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
```
arjunlalb, line 221: not required

projects/components/src/multi-select/multi-select.component.ts
palbizu: done

projects/components/src/multi-select/multi-select.component.ts
palbizu: it is used to re-execute the combineLatest to call reordering after click on open. How do you suggest to call this combineLatest without execute a subject?

projects/components/src/multi-select/multi-select.component.ts
palbizu: updated

projects/components/src/multi-select/multi-select.component.ts
palbizu: I couldn't find a way to force the combineLatests without to use this behaviorSubject

projects/components/src/multi-select/multi-select.component.ts
palbizu: I moved it to the Close

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -362,6 +375,19 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
...
+    filteredOptions: SelectOptionComponent<V>[],
+    selected?: V[]
+  ): SelectOptionComponent<V>[] {
+    if (isEmpty(selected)) {
+      return filteredOptions;
+    }
+    const filteredOptionsPartitions = compact(
+      partition(filteredOptions, filteredOption => selected?.includes(filteredOption.value))
+    );
+    return [...filteredOptionsPartitions[0], ...filteredOptionsPartitions[1]];
```
Exac, line 389: Will there always be two items in this array if we use `compact`? Perhaps a unit test would make this more understandable?

projects/components/src/multi-select/multi-select.component.ts
Exac: @arjunlalb Can you clarify?

arjunlalb: @Exac I had discussed this with Pato. We don't need to reorder everytime we open the dropdown. Instead, we should do it on every close. This way, even for large and asynchronous lists, the delay in opening is kept minimal and the reordering happens behind the scenes.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -235,11 +236,18 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
...
+    ]).pipe(
       map(([options, searchText]) =>
         isEmpty(searchText)
-          ? options.toArray()
-          : options.filter(option => option.label.toLowerCase().includes(searchText.toLowerCase()))
+          ? this.reorderSelectedItemsFirst(options.toArray(), this.selected)
+          : this.reorderSelectedItemsFirst(
+              options.filter(option => option.label.toLowerCase().includes(searchText.toLowerCase())),
+              this.selected
+            )
```
arjunlalb, line 250: we don't need to worry about reordering every time. reordering needs to happen on two occasions
1. On init.
2. after the user closes the popover, reorder in the background.

We shouldn't need to worry about reordering every time there's a search.

The way to achieve that would be to not combine reorder subject and search subject. do the reordering before the search.

palbizu: done

projects/components/src/multi-select/multi-select.component.ts
palbizu: it's splitting the array between selected and not selected items

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -362,6 +365,30 @@ export class MultiSelectComponent<V> implements ControlValueAccessor, AfterConte
...
   }
+
+  private reorderSelectedItemsFirst(
+    filteredOptions: SelectOptionComponent<V>[],
+    selected?: V[]
+  ): SelectOptionComponent<V>[] {
+    if (isEmpty(selected)) {
+      return filteredOptions;
+    }
+    const filteredOptionsPartitions = compact(
```
arjunlalb, line 376: compact is not needed here since there won't be any falsey values in the array.

arjunlalb (approved):

projects/components/src/multi-select/multi-select.component.ts
palbizu: done

jake-bassett (approved):

PR #2075, opened 2023-03-10
feat(load-async): add support for custom template for no data and err…

adisreyaj:
Add ability to provide `TemplateRef` for `noData` and `error` states of LoadAsync Directive.


projects/components/src/load-async/wrapper/load-async-wrapper.component.ts
```diff
@@ -66,36 +79,53 @@ export class LoadAsyncWrapperComponent {
...
 
-  private getNoDataIcon(): IconType {
-    return this.config?.noData?.icon ?? IconType.NoData;
+  private getNoDataIcon(data?: NoDataOrErrorStateConfig): IconType {
+    return data?.icon ?? IconType.NoData;
   }
 
-  private getErrorIcon(): IconType {
-    return this.config?.error?.icon ?? IconType.Error;
+  private getErrorIcon(data?: NoDataOrErrorStateConfig): IconType {
```
arjunlalb, line 119: we should perhaps keep the current error and no data icons so that it is consistent when we want to show them.

else for template usecases, you'll have to supply the icons as well. its avoidable if we can default in this wrapper.

anandtiwary: would like to be careful here. We want to have consistency and that is why we did not expose the config early on. 
@adisreyaj could you please share the use case on slack?

adisreyaj: @anandtiwary Can you check the PR once again...made few more changes after the discussions.

adisreyaj: @anandtiwary ping on this ^^

anandtiwary (approved):

anandtiwary (commented):
Please test few screens before merging this.

adisreyaj:
> Please test few screens before merging this.

Tested in couple of the pages (Loading, Error and No Data states). Seems to be working fine.

PR #2083, opened 2023-03-20
feat: adding support for composite specification returning object

itssharmasandeep:
## Description
Initially, for composite specifications, extarctFromServerData always returned `unknown[]` and this was making some of the cell renderers types as `[string, number, ...]` and we could not use the composite specs with a cell renderer which has a cell data typed as the interface (without doing another mapping). This PR introduces that functionality.

### Testing
Local testing is done.

### Checklist:


projects/observability/src/shared/graphql/request/builders/specification/specification-builder.ts
```diff
@@ -50,4 +74,24 @@ export class SpecificationBuilder {
...
+    const extractedData: Dictionary<unknown> = {};
+    const specs = Object.values(specifications);
+
+    return Object.keys(specifications)
+      .map((key, index) => [key, specs[index].extractFromServerData(serverData)] as [string, unknown])
+      .reduce((acc, val) => {
+        acc[val[0]] = val[1];
+
+        return acc;
+      }, extractedData);
```
adisreyaj, line 95:
```suggestion
      .reduce((acc, [key, value]) => {
	      return {
		      ...acc,
		      [key]: value
	      }
      }, extractedData);
```

nit: This makes it a little more readable.

adisreyaj (approved):
LGTM. I would wait for @arjunlalb or @anandtiwary s review as well.

projects/observability/src/shared/graphql/request/builders/specification/specification-builder.ts
```diff
@@ -10,8 +11,31 @@ export class SpecificationBuilder {
...
+
+      return {
+        resultAlias: () => alias,
+        name: alias,
+        asGraphQlSelections: () => this.selectionBuilder.fromSpecifications(specs),
+        extractFromServerData: serverData => this.extractForSpecificationsObject(serverData, specifications),
+        asGraphQlOrderByFragment: () => ({
+          expression: { key: orderByKey }
+        })
+      };
```
arjunlalb, line 35: would be good to define the separate flows (array vs object extraction) in independent methods.

projects/observability/src/shared/graphql/request/builders/specification/specification-builder.ts
```diff
@@ -50,4 +74,26 @@ export class SpecificationBuilder {
...
+
+  private buildSpecName(specs: Specification[]): string {
+    return specs.map(specification => specification.resultAlias()).join('_');
+  }
+
+  private extractForSpecificationsObject(
+    serverData: Dictionary<unknown>,
+    specifications: Dictionary<Specification>
+  ): Dictionary<unknown> {
+    const extractedData: Dictionary<unknown> = {};
```
arjunlalb, line 86: not required since you are doing reduce. can start with `{}` in L96

projects/observability/src/shared/graphql/model/specifications/composite-specification.ts
```diff
@@ -2,5 +2,5 @@ import { Dictionary } from '@hypertrace/common';
 import { Specification } from '../schema/specifier/specification';
 
 export interface CompositeSpecification extends Specification {
-  extractFromServerData(resultContainer: Dictionary<unknown>): unknown[];
+  extractFromServerData(resultContainer: Dictionary<unknown>): unknown[] | Dictionary<unknown>;
```
arjunlalb, line 5: we should consider using generic typing here.

arjunlalb (approved):

projects/observability/src/shared/graphql/request/builders/specification/specification-builder.ts
itssharmasandeep: We thought about the same, but discussed and chose the same path
cc: @anandtiwary

projects/observability/src/shared/graphql/request/builders/specification/specification-builder.ts
itssharmasandeep: Okay, Thanks

projects/observability/src/shared/graphql/model/specifications/composite-specification.ts
itssharmasandeep: Yeah, tried that also, didn't work that well

PR #2088, opened 2023-03-21
fix: specifications should also take object as input in composite specification model

itssharmasandeep:
Follow up of -> https://github.com/hypertrace/hypertrace-ui/pull/2083


projects/observability/src/shared/dashboard/data/graphql/specifiers/composite-specification.model.ts
```diff
@@ -15,7 +15,7 @@ export class CompositeSpecificationModel extends SpecificationModel<CompositeSpe
     displayName: 'Specifications',
     type: ARRAY_PROPERTY.type
```
anandtiwary, line 16: this needs to change as well

itssharmasandeep: I marked it as `UNKNOWN_PROPERTY`

jake-bassett (approved):

PR #2097, opened 2023-03-29
feat(stepper): enhancements

adisreyaj:
## Description
Show `checked` icon for completed and non-editable steps. Gives a visual indication about the progress.

image


projects/components/src/stepper/stepper.component.ts
```diff
@@ -34,17 +35,20 @@ import { StepperTabComponent } from './tab/stepper-tab.component';
...
           <mat-step
-            [completed]="step?.stepControl ? false : step.completed"
+            [completed]="step?.stepControl?.valid ?? step.completed"
             [stepControl]="step?.stepControl"
             [optional]="step.optional"
             [editable]="step.editable"
           >
             <ng-template matStepLabel>
-              <ht-label class="header-label" [label]="step.label"></ht-label>
+              <ht-label class="header-label" [attr.data-editable]="step.editable" [label]="step.label"></ht-label>
```
anandtiwary, line 44: why do we have to use attributes for this? Can we do it by setting a css class?

adisreyaj: We can.

23nobody:
This solution was not working the last time we tried it. Wondering what have changed since that makes it work. 😅

23nobody:
Please look for all the scenarios back and forth before merging.

adisreyaj:
> e scenarios back and forth before merging.

Adding necessary icons did the job.

PR #2105, opened 2023-04-04
Upgrade to v14

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation


arjunlalb (approved):

package.json
```diff
@@ -11,38 +11,35 @@
...
-    "build:components": "ng build components",
-    "build:dashboards": "ng build dashboards",
     "build:ci": "node --max_old_space_size=3584 node_modules/@angular/cli/bin/ng build --configuration production --no-progress",
-    "test": "ng test hypertrace-ui --cache --maxWorkers=2",
+    "test": "ng test hypertrace-ui --cache --max-workers=2",
     "lint": "ng lint hypertrace-ui",
     "lint:fix": "ng lint --fix hypertrace-ui",
     "prettier:check": "prettier --check '**'",
-    "test:ci": "ng test hypertrace-ui --maxWorkers=2 --ci --coverage",
+    "test:ci": "ng test hypertrace-ui --ci --coverage",
```
aaron-steinfeld, line 19: intentionally removing max workers here?

projects/observability/src/shared/components/cartesian/d3/cartesian-no-data-message.ts
```diff
@@ -8,7 +8,7 @@ export class CartesianNoDataMessage {
 
   public constructor(
     private readonly hostElement: Element,
-    private readonly series: Series<{}>[],
+    private readonly series: Series<unknown>[],
```
aaron-steinfeld, line 11: why this change?

projects/observability/src/shared/components/utils/d3/zoom/d3-zoom.ts
```diff
@@ -92,7 +88,7 @@ export abstract class D3Zoom<TContainer extends Element = Element, TTarget exten
     this.translateToRect(requestedRect);
   }
 
-  public panToRect(viewRect: ClientRect): void {
+  public panToRect(viewRect: Omit<DOMRect, 'x' | 'y' | 'toJSON'>): void {
```
aaron-steinfeld, line 91: You replaced clientrect (was that a native type that's been expanded) but narrow it back in several places. Maybe add an alias for the dom rect without these fields?

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/entity-topology-graphql-query-handler.service.ts
```diff
@@ -123,8 +123,10 @@ export class EntityTopologyGraphQlQueryHandlerService
     }
   }
 
-  public convertResponse(response: TopologyServerResponse, request: GraphQlEntityTopologyRequest): EntityNode[] {
-    return this.convertToEntityNodes(request, response.results);
+  public convertResponse(response: unknown, request: GraphQlEntityTopologyRequest): EntityNode[] {
```
aaron-steinfeld, line 126: why did this lose typing?

tsconfig.json
```diff
@@ -1,13 +1,14 @@
...
+    "rootDir": ".",
     "baseUrl": "./",
     "importHelpers": true,
     "outDir": "./dist/out-tsc",
     "strict": true,
     "sourceMap": true,
     "declaration": false,
     "module": "es2020",
-    "target": "es2015",
+    "target": "es2020",
```
aaron-steinfeld, line 11: Is changing the target required? Just curious why we're doing it now

package.json
anandtiwary: It is not needed for ci environment because we only have 1 or 2 CPU machines there. I can bring it back if you prefer.

projects/observability/src/pages/explorer/explorer.component.test.ts
```diff
@@ -164,19 +164,17 @@ describe('Explorer component', () => {
         limit: 1000,
         interval: new TimeDuration(15, TimeUnit.Second)
       }),
-      expect.objectContaining({})
+      undefined
```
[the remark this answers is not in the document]
anandtiwary: Yeah, i am not sure if the tests were working as intended before. I will double check this page to see if there is any issue.

projects/observability/src/shared/components/cartesian/d3/cartesian-no-data-message.ts
anandtiwary: reverted

package.json
anandtiwary: reverted

projects/observability/src/shared/components/utils/d3/zoom/d3-zoom.ts
anandtiwary: Yeah, ClientRect is deprecated now. DomRect is the preferred way. Updated with alias

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/entity-topology-graphql-query-handler.service.ts
anandtiwary: We have used this pattern at a few places. We tend to avoid exporting "server" types from the handler but in this case, we must export them if we want the tests to be able to call this method. 

Here is an example:

```
  public convertResponse(response: unknown, request: GraphQlEntitiesQueryRequest): Observable<EntitiesResponse> {
    return this.entitiesGraphQlQueryBuilderService.buildResponse(response, request);
  }
```

projects/observability/src/shared/components/cartesian/d3/cartesian-no-data-message.ts
anandtiwary: scratch that. This is the error we get from the new ts

```
Argument of type 'Series<TData>[]' is not assignable to parameter of type 'Series<{}>[]'.
  Type 'Series<TData>' is not assignable to type 'Series<{}>'.
    Type 'TData' is not assignable to type '{}'.ts(2345)
    
```

package.json
aaron-steinfeld: It was actually originally introduced for CI - it's possible this isn't the case any more, but historically it was getting the CPU count from the underlying hardware, while the actual container running the job had less CPUs available, causing a big slowdown. In GHA, since we're running on VMs rather than in docker it may not be relevant, so you could try opening it up and see how times compare but wouldn't do it in this PR.

anandtiwary:
@aaron-steinfeld updated. Also, please help with the test failure in ci? Is it a memory issue?

package.json
anandtiwary: Reverted it and it also fixed the weird test failure. :)

tsconfig.json
aaron-steinfeld: Would have to go through compatibility - it doesn't really matter to us since TS handles the down-compilation, so no harm (other than missed optimizations) in targeting the older version with more compatibility

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/entity-topology-graphql-query-handler.service.ts
aaron-steinfeld: > we must export them if we want the tests to be able to call this method.

I thought it should work as long as we don't need a direct type reference?

tsconfig.json
Exac: In that case, what about `.browserlistrc`? Just by virtue of updating Angular, we are limiting browser support to the most recent browsers at the time of Angular 14's release (Q2 2022). (See [browserlist](https://github.com/browserslist/browserslist/blob/main/README.md)).

IMO we should update this to ES2021 or ES2020, and we should check for compatibility after updating Angular anyways, regardless of any internal SLA that might be different than Angular's (which is "[the most recent browsers](https://angular.io/guide/browser-support)", presumably when Angular 14 was released last year).

Exac (commented):
I'm not sure how thorough we want to be with this PR in the interest of time. Since we are bumping TypeScript by four minor versions, I took a look at the changelog blogposts:

- https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/

The first section of the 4.4 update mentions that TS can now detect if a constant value contains a type guard, so we could improve confirmation-modal.component.ts:
```typescript
    this.isContentString = isString(config.content);
    if (isString(config.content)) {
    // calling isString twice is now redundant, we can replace the second call with:
    if (this.isContentString) {
```
There are probably more. If our goal is just to bump the version up to catch up, then that's okay too.

anandtiwary:
> I'm not sure how thorough we want to be with this PR in the interest of time. Since we are bumping TypeScript by five minor versions, I took a look at the changelog blogposts:
> 
>     * https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/
> 
>     * https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/
> 
>     * https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/
> 
>     * https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/
> 
>     * https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/
> 
> 
> The first section of the 4.4 update mentions that TS can now detect if a constant value contains a type guard, so we could improve confirmation-modal.component.ts:
> 
> ```ts
>     this.isContentString = isString(config.content);
>     if (isString(config.content)) {
>     // calling isString twice is now redundant, we can replace the second call with:
>     if (this.isContentString) {
> ```
> 
> There are probably more. If our goal is just to bump the version up to catch up, then that's okay too.

Yeah, these are some things we should do with follow up PRs. We will also have to do similar upgrades for angular related code.

tsconfig.json
anandtiwary: I think we can decouple this one with the current PR. I will do some research and take your concern into account and put a new PR.

package.json
```diff
@@ -94,31 +96,42 @@
...
-    "codelyzer": "^6.0.2",
+    "@typescript-eslint/eslint-plugin": "5.43.0",
+    "@typescript-eslint/parser": "5.43.0",
     "commitizen": "^4.3.0",
     "cz-conventional-changelog": "^3.3.0",
+    "eslint": "^8.28.0",
+    "eslint-config-prettier": "^8.5.0",
+    "eslint-plugin-import": "2.27.5",
+    "eslint-plugin-jsdoc": "40.0.0",
+    "eslint-plugin-no-null": "1.0.2",
```
Exac, line 113: Why do we install this, but then remove all the comments from tslint that enable this behavior?
```typescript
// tslint:disable-next-line: no-null-keyword
```
I think we don't actually need to install this plugin if we removed all the comments.

Also, FWIW, I think allowing the `null` keyword is a great thing.

tsconfig.json
aaron-steinfeld: Agree, decouple. There's a difference in angular (and our) published SLA and best effort. Like there's no way angular _actually_ only supports the last 2 versions of chrome (i.e. the march and april releases, but not feb), or other evergreens. That's just what they're putting in writing. We've got to be careful with older support given the nature of larger enterprise users and slower upgrade cadences.

projects/observability/src/shared/components/cartesian/d3/cartesian-no-data-message.ts
aaron-steinfeld: adding a bound on the definition, `Series<TInterval>` would resolve that right?

projects/observability/src/pages/explorer/explorer.component.test.ts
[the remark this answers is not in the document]
aaron-steinfeld: this seems like the only outstanding issue. If they were returning an object before and now returning undefined, something's been functionally changed. It may be on the test side, may be on the component side but seems like we should understand this issue before merging.

aaron-steinfeld:
> There are probably more. If our goal is just to bump the version up to catch up, then that's okay too.

We should bump the version in its own PR (obviously, this one's huge already), then we can go through and see what features we can take advantage of. Would be good to spend some time on weekly call going over the new features available.

> Why do we install this, but then remove all the comments from tslint that enable this behavior?

I thought we had just replaced the tslint disable overrides with eslint ones? It's still on by default right

> Also, FWIW, I think allowing the null keyword is a great thing.

Separate discussion, we should try to migrate as-is then we can consider rules changes. I disagree, but certainly open to debate!

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/entity-topology-graphql-query-handler.service.ts
anandtiwary: It complains. Probably because we didn't pass all properties. I didn't want to debug and change this in the code as part of this PR. Do you mind if we address all similar occurrences in a new PR?

projects/observability/src/shared/components/cartesian/d3/cartesian-no-data-message.ts
anandtiwary: umm.. may be, but unknown makes sense to me here. No data doesn't need to know the generic type of series data. I can define TInterval, but i think i will also have to default to unknown or always pass a type from the caller side.

aaron-steinfeld: In the context of the larger review, didn't look where we were. Agree, no data doesn't care (i.e. doesn't make any assumptions that it's an object) and as long as it works with unknown it's fine.

Exac:
> > There are probably more. If our goal is just to bump the version up to catch up, then that's okay too.
>
> We should bump the version in its own PR (obviously, this one's huge already), then we can go through and see what features we can take advantage of. Would be good to spend some time on weekly call going over the new features available.

I remember updating Angular in the past and running into issues when changing the version of typescript that the auto update tool suggested.

I think I would personally rather have the stronger typing from the newer version of TypeScript sooner, and then go back and look at improvements that we missed later. We could create some tasks to look for improvements from each minor version of TypeScript. @anandtiwary has already taken care of the breaking changes that throw compile-time errors, which is the difficult part.

<hr/>
> > Also, FWIW, I think allowing the null keyword is a great thing.
>
> Separate discussion, we should try to migrate as-is then we can consider rules changes. I disagree, but certainly open to debate!

It would be great if we could have a linter rule smart enough to see when we are providing `null` to a library function outside of our source code (which is 100% of the places where we currently invoke the `no-null` rule. We can discuss when we have time.

projects/observability/src/pages/explorer/explorer.component.test.ts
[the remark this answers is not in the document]
anandtiwary: Just tested this and things seem to work as expected. Not seeing any issues at all. Imo, we can go ahead with this.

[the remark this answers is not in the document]
aaron-steinfeld: Please follow up on it - something's clearly broken, sounds like it's on the test side.

aaron-steinfeld (approved):

PR #2106, opened 2023-04-05
feat: setting default TR from current url when ff is disabled

anandtiwary:
## Description
feat: setting default TR from current url when ff is disabled

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/common/src/time/page-time-range-preference.service.ts
+++ b/projects/common/src/time/page-time-range-preference.service.ts
@@ -37,7 +37,7 @@ export class PageTimeRangePreferenceService {
       this.featureStateResolver.getFeatureFlagValue<Dictionary<string>>(ApplicationFeature.FeatureDefaultTimeRangeMap)
     ]).pipe(
       map(([pageTimeRangeStringDictionary, featureState, featureDefaultTimeRangeMap]) => {
-        if (featureState === FeatureState.Enabled) {
+        if (featureState !== FeatureState.Enabled) {
           if (!isNil(pageTimeRangeStringDictionary[rootLevelPath])) {
             return () => this.timeRangeService.timeRangeFromUrlString(pageTimeRangeStringDictionary[rootLevelPath]);
           }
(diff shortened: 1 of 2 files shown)


arjunlalb (dismissed):

projects/common/src/time/page-time-range-preference.service.ts
```diff
@@ -93,7 +93,11 @@ export class PageTimeRangePreferenceService {
   }
 
   public getGlobalDefaultTimeRange(): TimeRange {
-    return new RelativeTimeRange(new TimeDuration(1, TimeUnit.Day));
+    return this.maybeGetCurrentTimeRange() ?? new RelativeTimeRange(new TimeDuration(1, TimeUnit.Day));
```
aaron-steinfeld, line 96: This seems off to me - it would make the default _always_ the current after startup. Maybe instead, have the `initializeTimeRange` subscription save the requested time range and expose that?

anandtiwary: I only want to call this when FF is disabled. I missed that it is used elsewhere. I can just update this method to have the new behavior. What do you think?

```
// When FF is disabled
return () => this.getGlobalDefaultTimeRange();
```

anandtiwary: Updated. @aaron-steinfeld  please check

projects/common/src/time/page-time-range-preference.service.ts
```diff
@@ -49,7 +49,7 @@ export class PageTimeRangePreferenceService {
         }
 
         // When FF is disabled
-        return () => this.getGlobalDefaultTimeRange();
+        return () => this.maybeGetCurrentTimeRange() ?? this.getGlobalDefaultTimeRange();
```
aaron-steinfeld, line 52: nit - private method isn't really doing anything any more, can remove
```suggestion
        return () => this.timeRangeService.maybeGetCurrentTimeRange() ?? this.getGlobalDefaultTimeRange();
```

aaron-steinfeld (approved):

projects/common/src/time/page-time-range-preference.service.ts
anandtiwary: I can address this in a follow up.

PR #2109, opened 2023-04-10
feat(refresh): add new refresh button component

adisreyaj:
## Description
Currently the refresh button is coupled to the ht-time-range. This PR makes the **Refresh Button** decoupled from the **Time Range** so it can be used as standalone in the **Page Header**.

Introduced Display mode for **Page Header Component**:
```ts
export const enum PageHeaderDisplayMode {
  Default = 'default',
  WithTimeRange = 'with-time-range',
  WithRefreshButton = 'with-refresh'
}
```


projects/components/src/header/page/page-header.component.ts
```diff
@@ -99,6 +104,9 @@ export class PageHeaderComponent implements OnInit {
   @Input()
   public hidePageTimeRange?: boolean = false;
 
+  @Input()
+  public showRefreshButton?: boolean = false;
```
aaron-steinfeld, line 108: this feels off - both because other boolean inputs are hideXyz (so this is inverted), but also because not all pairings of the two booleans make sense. This is only checked if we're hiding the time range.

projects/components/src/refresh-button/refresh-button.service.ts
```diff
@@ -0,0 +1,12 @@
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
```
aaron-steinfeld, line 4: Not sure this makes sense as a global service - more so as a button output, since multiple things might want refreshing independently.

projects/components/src/time-range/time-range.component.ts
```diff
@@ -90,7 +90,14 @@ export class TimeRangeComponent {
...
   public readonly timeRangeSelected: EventEmitter<TimeRange> = new EventEmitter();
 
-  public constructor(private readonly timeRangeService: TimeRangeService, private readonly ngZone: NgZone) {}
+  public constructor(
+    private readonly timeRangeService: TimeRangeService,
+    private readonly ngZone: NgZone,
+    private readonly refreshButtonService: RefreshButtonService,
+    private readonly subs: SubscriptionLifecycle
+  ) {
+    this.subs.add(this.refreshButtonService.refresh$.subscribe(() => this.timeRangeService.refresh()));
```
aaron-steinfeld, line 99: so what benefit do we get from the service vs component output approach?

adisreyaj: Do you think I should just work with output then? Add an extra output in the **PageHeader** and let the consumer listen to it?

My initial thought was to have the service inside the timerange itself, but that won't be a good idea I think.

projects/components/src/refresh-button/refresh-button.service.ts
adisreyaj: Right.

projects/components/src/header/page/page-header.component.ts
adisreyaj: Updated.

projects/components/src/refresh-button/refresh-button.service.ts
adisreyaj: Removed the service.

projects/components/src/time-range/time-range.component.ts
adisreyaj: Using event emitter instead of service.

projects/components/src/header/page/page-header.component.ts
aaron-steinfeld: Addressed the inversion, but not the fact that 2 booleans introduces 4 states when only 3 are actually meaningful. Maybe combine these two booleans into a mode enum (would need to handle compatibility, but that's straightforward by deprecating boolean but using to set mode default if that's unset).

adisreyaj: Makes sense. Will update.

projects/components/src/time-range/time-range.component.ts
```diff
@@ -57,19 +58,18 @@ import { PopoverRef } from '../popover/popover-ref';
...
         [role]="refreshButton.role"
-        (click)="refreshButton.onClick()"
+        (click)="this.onRefresh()"
       >
-      </ht-button>
+      </ht-refresh-button>
     </div>
-  `
+  `,
+  providers: [SubscriptionLifecycle]
```
aaron-steinfeld, line 72: No longer needed

projects/components/src/header/page/page-header.component.ts
```diff
@@ -159,3 +185,12 @@ export const enum PageHeaderContentAlignment {
   Column = 'column-alignment',
   Row = 'row-alignment'
 }
+
+/**
+ * @param Default - Default mode with no time range or refresh button
+ */
+export const enum PageHeaderDisplayMode {
+  Default = 'default',
+  WithTimeRange = 'with-time-range',
```
aaron-steinfeld, line 194: Nit might rename these to clarify. Time range may or may not have refresh, and most commonly we show it so default feels off too. May make sense to just be verbose

projects/components/src/time-range/time-range.component.ts
adisreyaj: Removed

aaron-steinfeld (approved):

projects/components/src/header/page/page-header.component.ts
adisreyaj: I'll revisit this part.

PR #2144, opened 2023-05-04
Fix button aria role

jake-bassett:


projects/components/src/button/button.component.ts
```diff
@@ -59,7 +59,7 @@ export class ButtonComponent {
   public type: ButtonType = ButtonType.Button;
 
   @Input()
-  public role: ButtonRole = ButtonRole.Secondary;
+  public variant: ButtonVariant = ButtonVariant.Secondary;
```
anandtiwary, line 63: In addition to this, I think we should set an aria role attribute on the root element

jake-bassett: It's there on line 15. I searched for all button elements in the code. I think there were only like 4 or 5 because we use our own wrappers.

anandtiwary: Sorry, i think i commented before the latest commit.

anandtiwary (approved):

projects/components/src/button/button.component.ts
jake-bassett: Oh, hah. Yeah, I tried to break up the commits for easier reviewing.

Exac (approved):

PR #2147, opened 2023-05-05
feat(accessibility): aria label support for number input and toggle s…

adisreyaj:
Add aria label support for:
- Number Input
- Toggle Switch

### DIFF
--- a/projects/components/src/navigation/navigation-list.component.scss
+++ b/projects/components/src/navigation/navigation-list.component.scss
@@ -105,7 +105,7 @@
       .link {
         margin-right: auto;
         @include link;
-        @include footnote($gray-5);
+        @include footnote($gray-2);
       }
     }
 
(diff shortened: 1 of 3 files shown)


projects/components/src/navigation/navigation-list.component.scss
```diff
@@ -105,7 +105,7 @@
       .link {
         margin-right: auto;
         @include link;
-        @include footnote($gray-5);
+        @include footnote($gray-2);
```
itssharmasandeep, line 108: any reason for this change?

projects/components/src/toggle-switch/toggle-switch.component.ts
```diff
@@ -12,6 +12,7 @@ import { ToggleSwitchSize } from './toggle-switch-size';
       <mat-slide-toggle
         color="primary"
         [checked]="this.isChecked"
+        [aria-label]="this.ariaLabel ?? this.label"
```
itssharmasandeep, line 15: should we default it?

adisreyaj: Ideally we should provide one of these.

itssharmasandeep (approved):

PR #2152, opened 2023-05-05
feat: choose the first value for the breadcrumb when an entity contains a list of values

Exac:
## Description
When provided multiple values for the breadcrumb, show the first one (for now).

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
+++ b/projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
@@ -99,10 +99,14 @@ export class ApiDetailBreadcrumbResolver<T extends EntityBreadcrumb> extends Ent
     if (!parentTypeMetadata) {
       return {};
     }
+    const parentName: string | string[] = entity[this.getParentNameAttribute(parentTypeMetadata)] as string;
 
     return {
       parentId: entity[this.getParentIdAttribute(parentTypeMetadata)] as string,
-      parentName: entity[this.getParentNameAttribute(parentTypeMetadata)] as string
+      // TODO: (ENG-30212) How do we know which domain name to display when
+      //  there is an array? There is no guarantee that the first domain chosen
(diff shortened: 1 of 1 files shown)


projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
```diff
@@ -99,10 +99,14 @@ export class ApiDetailBreadcrumbResolver<T extends EntityBreadcrumb> extends Ent
...
     }
+    const parentName: string | string[] = entity[this.getParentNameAttribute(parentTypeMetadata)] as string;
 
     return {
       parentId: entity[this.getParentIdAttribute(parentTypeMetadata)] as string,
-      parentName: entity[this.getParentNameAttribute(parentTypeMetadata)] as string
+      // TODO: (ENG-30212) How do we know which domain name to display when
+      //  there is an array? There is no guarantee that the first domain chosen
+      //  will match the domainId for the domain.
+      parentName: Array.isArray(parentName) ? parentName[0] : parentName
```
jake-bassett, line 109: Is there always guaranteed to be at least one element if there is an array?

In other words, do we need any null checking here?

jake-bassett (approved):
Commented, but approved if we don't need any null checks.

projects/observability/src/pages/apis/api-detail/api-detail-breadcrumb.resolver.ts
Exac: Hmmn, that is a good point. We should simply return `{}` (like on line 100) if that were the case.

Exac: Added a check in 81e0269fed865201347ffba4bc5f44f8252d2c1b.

Exac:
Going to close this for now, it shouldn't be necessary.

PR #2158, opened 2023-05-08
feat: updates to notification component and service

arjunlalb:
## Description
Update notification component style.
Add support for taking custom content template.

Screenshot 2023-05-08 at 3 12 08 PM

### Testing
Updated UTs. Tested locally.

### Checklist:

### DIFF
--- a/projects/components/src/notification/notification.component.scss
+++ b/projects/components/src/notification/notification.component.scss
@@ -2,12 +2,12 @@
 
 .notification-container {
   display: flex;
-  background-color: $gray-7;
+  background-color: $blue-4;
   color: white;
   align-items: center;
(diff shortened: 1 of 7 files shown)


projects/components/src/notification/notification.component.scss
```diff
@@ -35,5 +35,5 @@
 }
 
 .info {
-  color: $blue-3;
+  color: white;
```
adisreyaj, line 38: We need to invert this as well.

arjunlalb: @adisreyaj this was intentional. We don't have any usages of info toast anywhere.

adisreyaj (approved):

PR #2174, opened 2023-05-09
fix: Update blue color and add textarea aria-label

palbizu:
## Description
Updating blue color to support accessibility and adding ariaLabel into textArea

### Testing
Tested Locally

### Checklist:

### Documentation

### DIFF
--- a/projects/assets-library/assets/styles/_color-palette.scss
+++ b/projects/assets-library/assets/styles/_color-palette.scss
@@ -45,7 +45,7 @@ $green-9: #03110a;
 $blue-1: #f0f6ff;
 $blue-2: #b8d3ff;
 $blue-3: #70a7ff;
-$blue-4: #2478ff;
+$blue-4: #0666ff;
 $blue-5: #0053d7;
 $blue-6: #0043ad;
 $blue-7: #003385;
--- a/projects/components/src/textarea/textarea.component.ts
+++ b/projects/components/src/textarea/textarea.component.ts
@@ -29,6 +29,7 @@ import { LoggerService } from '@hypertrace/common';
         [rows]="this.rows"
         [disabled]="this.disabled"
         [placeholder]="this.placeholder"
+        [attr.aria-label]="this.ariaLabel || 'textarea'"
         [ngModel]="this.value"
         (ngModelChange)="this.onValueChange($event)"
       >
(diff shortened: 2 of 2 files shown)


projects/assets-library/assets/styles/_color-palette.scss
```diff
@@ -45,7 +45,7 @@ $green-9: #03110a;
 $blue-1: #f0f6ff;
 $blue-2: #b8d3ff;
 $blue-3: #70a7ff;
-$blue-4: #2478ff;
+$blue-4: #0666ff;
```
jake-bassett, line 48: Do we want to change blue-4 everywhere or should we just use a different blue like blue-3 or blue-5 for whatever instance you are trying to fix?

Exac: If we change this we will presumably have to look at all the pages we have already changed again, because the color contrast between the new `blue-4` and other colors changes now.

Exac: If we change this we will presumably have to look at all the pages we have already changed again, because the color contrast between the new blue-4 and other colors changes now.

palbizu: I'm increasing the color value to just 7%. For the user will be really similar and we can keep the logic of hover, already applied in links but this passes the accessibility check.

image

jake-bassett: Okay, I'm cool with changing the color, but @Exac is right that we will need to recheck areas where this might cause regressions.

@palbizu please ensure no regressions before merging. @Exac if you know any specific places Pato should check, please let him know.

I'll approve, but please check before merging.

jake-bassett (approved):

projects/assets-library/assets/styles/_color-palette.scss
Exac: I don't know anywhere in particular that we should check. The point being that the tool we use to check for accessibility on all the other pages did so when `blue-4` was `2478ff`. Now that we are changing one of our palette colors, we will have to go back and run the accessibility tool for every page that we have already merged, in order to be sure.

It shouldn't take too long, we can just look at the PRs that have been merged and look at the accessibility tool for each page to confirm we didn't change anything. It can even be done as a final pass once we have completed all these accessibility stories.

I support this change, just want to make sure that we don't get to the end of all these accessibility changes and realize that we inadvertently introduced different contrast problems.

Exac (approved):

projects/assets-library/assets/styles/_color-palette.scss
palbizu: Yes I understand that could be an issue, I took a look at several components and they are working fine. There are many uses for links with this color this change helps a lot to accessibility

PR #2182, opened 2023-05-12
fix: table detail content rendering

itssharmasandeep:
## Description
This fixes the re-rendering for the detailed content if any other row is expanded.

`Current Behaviour`
Currently, if we have a table of type `detail` and we expand a row, the first one works fine but if we expand another row then it renders both the rows. That means it will always render all the previously expanded rows if we expand any other rows.

`FIx`
There are checks needed to fix this.
1. Selections change should not be triggered if current and previous both selections are empty.
2. The rowChange emission should only happen when the table of `tree` type, since it causes the whole table to re-render

### Testing
Manual testing is done.

### Checklist:

### DIFF
--- a/projects/components/src/table/table.component.ts
+++ b/projects/components/src/table/table.component.ts
@@ -494,7 +494,11 @@ export class TableComponent
       this.initializeData();
     }
 
-    if (changes.selections) {
+    // Current and previous selections both should not be empty.
+    if (
+      changes.selections &&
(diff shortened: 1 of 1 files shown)


jake-bassett:
Be sure to test trace waterfall.

palbizu (approved):

itssharmasandeep:
> Be sure to test trace waterfall.

Don't see any issues :)

PR #2197, opened 2023-05-19
fix: table column resizing

itssharmasandeep:
## Description
`Current Behavior`
Currently, we have a table that supports horizontal scrolling but the table column resize only works for adjacent columns which makes it harder to understand for a user and doesn't work in the expected way. It column width update doesn't make the table scrollable automatically.
2nd problem is, we do column resize as soon as we start moving the resize handler, which makes is very less performant due to a lot of calculations and DOM updates in a short time.

`New Behavior`
1. Now, we'll have a resizer for each non-state column which changes the width of that particular column and scrolls the table if it goes beyond the container width
2. We'll change the column width once resize action is done. Users will have a way to move the resize handler and once they finish the action, the table column will resize to that width.

### Testing
Manual testing is done.

### Checklist:


projects/components/src/table/table.component.ts
```diff
@@ -626,18 +673,30 @@ export class TableComponent
...
-    if (this.isStateColumn(columns[index - 1])) {
+    if (this.isStateColumn(columns[index])) {
       return false;
     }
 
     return true;
   };
 
+  public showColumnResizeHandler = (columns: TableColumnConfig[], index: number): boolean => {
+    if (!(this.resizable ?? true)) {
```
jake-bassett, line 689: This is some fun boolean logic. Seems a roundabout way of doing these checks. Is this because `resizable` can be undefined? 

I feel like a lot of the logic in `showColumnDivider` and  `showColumnResizeHandler` could be cleaned up a bit, but maybe not.

projects/components/src/table/table.component.scss
```diff
@@ -88,34 +88,43 @@ $data-row-border: 1px solid $gray-2;
 }
 
 .header-cell {
-  display: flex;
```
anandtiwary, line 91: Is it possible to keep these styles? Just trying to have minimal changes to the code.

projects/components/src/table/table.component.ts
itssharmasandeep: I added functions so that we don't have to make some changes in the template and now two different methods are needed for showing the divider and resize handler as I decoupled them with each other.
It is because of resizable as well the fact that resizing can be done on the last column but divider won't be there for that and due to this reason we need one more extra check

projects/components/src/table/table.component.scss
itssharmasandeep: Visual Styles are almost the same, nothing major change. It's just it is moved around because of the decoupling of the column divider and resize handler in the template.

projects/components/src/table/table.component.ts
```diff
@@ -561,14 +579,52 @@ export class TableComponent
...
+    }
+  }
+
+  public onResizeMouseUp(event: MouseEvent): void {
+    this.checkAndResizeColumn(event);
+    this.changeDetector.detectChanges();
+  }
+
+  private addEventListeners(): void {
+    this.document.addEventListener('mousemove', event => this.onResizeMouseMove(event));
```
anandtiwary, line 594: We could attach it on the root element (inject element). We could also use rxjs but not a requirement

projects/components/src/table/table.component.ts
```diff
@@ -561,14 +579,52 @@ export class TableComponent
...
+
+  private removeEventListeners(): void {
+    this.document.removeEventListener('mousemove', event => this.onResizeMouseMove(event));
+    this.document.removeEventListener('mouseup', event => this.onResizeMouseMove(event));
   }
 
-  @HostListener('mouseup')
-  public onResizeMouseUp(): void {
-    this.resizeColumns = undefined;
+  private checkAndResizeColumn(event: MouseEvent): void {
```
anandtiwary, line 603: can we do auto fit content for all columns?

itssharmasandeep: discussed offline

anandtiwary (approved):

PR #2227, opened 2023-05-29
feat: run change detection after updating the visible columns

Exac:
- If we do not run change detection after updating the visible columns, then we run the risk of columns having a width of  (in such a case, the table renders visually as a tall 1px column)

## Description
- This is a bug fix
- if we do not run change detection after updating the visible columns, then we run the risk of columns having a width of  (in such a case, the table renders visually as a tall 1px column)

### Testing
Added a unit test, and tested in my local env.

### Checklist:

### Documentation


palbizu (approved):
LGTM. Seems to be a valid solution

PR #2234, opened 2023-06-01
feat(table): add options to allow customization of cell renderers

adisreyaj:
## Description

- Add extra `options` prop to the `TableColumnConfig`
- Use it to pass extra data to the renderers which can then be used for customizing the renderers.

### DIFF
--- a/projects/components/src/table/cells/table-cell-injection.ts
+++ b/projects/components/src/table/cells/table-cell-injection.ts
@@ -9,6 +9,7 @@ export const TABLE_DATA_PARSER: InjectionToken<TableCellParserBase<unknown, unkn
 );
 export const TABLE_CELL_DATA: InjectionToken<unknown> = new InjectionToken('TABLE_CELL_DATA');
 export const TABLE_ROW_DATA: InjectionToken<unknown> = new InjectionToken('TABLE_ROW_DATA');
+export const TABLE_CELL_OPTIONS: InjectionToken<unknown> = new InjectionToken('TABLE_CELL_OPTIONS');
 
 export const createTableCellInjector = (
   columnConfig: TableColumnConfig,
@@ -35,6 +36,10 @@ export const createTableCellInjector = (
       {
         provide: TABLE_CELL_DATA,
         useValue: cellData
+      },
+      {
+        provide: TABLE_CELL_OPTIONS,
+        useValue: columnConfig?.options
       },
       {
         provide: TABLE_ROW_DATA,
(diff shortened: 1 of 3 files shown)


projects/components/src/table/table-api.ts
```diff
@@ -4,20 +4,35 @@ import { FieldFilter, FilterValue } from '../filtering/filter/filter';
 import { FilterOperator } from '../filtering/filter/filter-operators';
 import { TableCellAlignmentType } from './cells/types/table-cell-alignment-type';
 
-export interface TableColumnConfig {
-  id: string; // This is the unique ID for the column (often same as 'name' except for composite fields)
-  name?: string; // Attribute name (for composite columns use the attribute that should be filtered/sorted)
+export interface TableColumnConfig<TableColumnOptions = unknown> {
```
arjunlalb, line 7: I think its safe to assume `Dictionary<unknown>` as type.

arjunlalb (dismissed):

projects/components/src/table/cells/table-cell-injection.ts
```diff
@@ -36,6 +37,10 @@ export const createTableCellInjector = (
         provide: TABLE_CELL_DATA,
         useValue: cellData
       },
+      {
+        provide: TABLE_CELL_OPTIONS,
```
anandtiwary, line 41: Why do we need to inject it? Column config is already injected, can we pull from there?

adisreyaj: Yeah. I think will make the change then.

projects/components/src/table/table-api.ts
adisreyaj: Started off with that, but had type conflicts downstream.

anandtiwary: Could we do `TableColumnOptions = Dictionary<boolean>` so that we only allow a bunch of booleans?

adisreyaj: There can be some other cases where passing enum would also make sense for options.

arjunlalb: agree with possible need for enums. but why wouldn't `TableColumnOptions extends Dictionary<unknown>` type work?

arjunlalb (approved):

projects/components/src/table/table-api.ts
adisreyaj: I can come around to it in some time.

PR #2238, opened 2023-06-05
style: add line clamp mixin

adisreyaj:
Add mixin for line-clamps

### DIFF
--- a/projects/assets-library/assets/styles/_font.scss
+++ b/projects/assets-library/assets/styles/_font.scss
@@ -341,3 +341,10 @@ $font-family: 'Work Sans', sans-serif;
 $mat-typography: mat.define-typography-config(
   $font-family: $font-family
 );
+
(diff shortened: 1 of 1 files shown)


projects/assets-library/assets/styles/_font.scss
```diff
@@ -341,3 +341,10 @@ $font-family: 'Work Sans', sans-serif;
 $mat-typography: mat.define-typography-config(
   $font-family: $font-family
 );
+
+@mixin line-clamp($maxLines: 1) {
```
arjunlalb, line 345: What's the most common usecase? Is it 1?

arjunlalb (approved):

projects/assets-library/assets/styles/_font.scss
adisreyaj: Generally we truncate on single lines now.

PR #2246, opened 2023-06-06
feat: support only allowlisted domains for external navigation

arjunlalb:
## Description
support only allowlisted domains for external navigation

### Testing
UTs added


projects/common/src/constants/external-urls-allowlist.ts
```diff
@@ -0,0 +1,4 @@
+import { InjectionToken } from '@angular/core';
+
+export const externalUrlDomainAllowList: string[] = ['hypertrace.org'];
```
Exac, line 3: When our on-prem customers use the app, what URL do they visit for it?

adisreyaj (approved):

projects/common/src/external/external-url-navigator.test.ts
```diff
@@ -6,42 +6,67 @@ import {
...
 describe('External URL navigator', () => {
   let spectator: SpectatorService<ExternalUrlNavigator>;
   const buildNavigator = createServiceFactory({
     service: ExternalUrlNavigator,
-    providers: [mockProvider(NavigationService)]
+    providers: [
+      mockProvider(NavigationService),
+      {
+        provide: EXTERNAL_URL_DOMAIN_ALLOWLIST,
+        useValue: ['traceable.ai']
```
anandtiwary, line 19: can we avoid this domain in ht repo?

PR #2247, opened 2023-06-06
fix: topology graph layout

itssharmasandeep:


projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
```diff
@@ -1,18 +1,22 @@
...
   public layout(topology: RenderableTopology<TopologyNode, TopologyEdge>): void {
+    this.initialize();
     this.findRootNodes(topology);
     this.fillNodeAndLevelMaps();
     this.assignCoordinatesToNodes();
     this.verticallyCenterAlignNodes();
   }
 
+  private initialize(): void {
+    this.levelToNodesMap = new Map([[0, []]]);
```
Greedy-Geek, line 16: if we are initializing `levelToNodesMap` here then do we need the initialization at L4?

Greedy-Geek (approved):

projects/observability/src/shared/components/topology/d3/layouts/graph-layout.ts
itssharmasandeep: I don't think so.

anandtiwary (approved):

PR #2249, opened 2023-06-07
fix: topology edge rendering with update pattern

itssharmasandeep:


anandtiwary:
@itssharmasandeep can we close this PR?

PR #2251, opened 2023-06-08
fix: table resizing issues

itssharmasandeep:


projects/components/src/table/table.component.ts
```diff
@@ -504,6 +508,18 @@ export class TableComponent
       .subscribe(sort => this.updateSort(sort));
   }
 
+  public onLayoutChange(): void {
+    const latestTableWidth = this.table.nativeElement.offsetWidth;
+
+    if (this.tableWidth === 0) {
```
anandtiwary, line 514: so, we are only overriding for the first time?

itssharmasandeep: Initially, we just assign the width the first time (when the page loads). After that when we change the nav then we distribute the width

anandtiwary (approved):

PR #2293, opened 2023-07-06
feat: add external urls allow list handling

itssharmasandeep:
We're introducing `urlAllowList` and `domainAllowList` in the case of external URL redirection.
An external will only be redirected if
1. It is a full URL and inside the urlAllowList
2. It is a full URL and the domain is inside domainAllowList
3. It is a partial URL (for in-app navigation with external nav params)


arjunlalb:
@itssharmasandeep could you please update the PR title and description here?

itssharmasandeep:
Not required for now

PR #2295, opened 2023-07-07
feat: new isFeatureEnabled pipe

arjunlalb:
## Description
A pipe to check if the feature state is enabled.


projects/components/src/feature-check/is-feature-state-enabled.pipe.ts
adisreyaj, line 1: Test 🙂

adisreyaj (approved):

PR #2303, opened 2023-07-11
fix: external url navigation on reload

itssharmasandeep:


projects/common/src/navigation/navigation.service.ts
```diff
@@ -42,6 +43,11 @@ export class NavigationService {
   ) {
     this.event$(RoutesRecognized)
       .pipe(
+        tap(event => {
+          if (isEmpty(this.firstNavigatedUrl)) {
+            this.firstNavigatedUrl = decodeURIComponent(event.url);
+          }
+        }),
```
arjunlalb, line 50: please leave a comment here on why we are storing this.

itssharmasandeep: Added

arjunlalb (approved):

PR #2304, opened 2023-07-12
feat: registered icons

itssharmasandeep:


projects/assets-library/src/icons/icon-registry.service.ts
```diff
@@ -1,16 +1,23 @@
...
 import { DomSanitizer } from '@angular/platform-browser';
-import { isEmpty } from 'lodash-es';
+import { isEmpty, orderBy } from 'lodash-es';
 import { Observable } from 'rxjs';
 
 @Injectable({ providedIn: 'root' })
 export class IconRegistryService {
   private static readonly SVG_ICON_PREFIX: string = 'svg:';
 
+  public registeredIcons: string[] = [];
```
arjunlalb, line 11: why public? if you need to access the registry, it should be via a getter method.

itssharmasandeep: Right, Used getter.

projects/assets-library/src/icons/icon-registry.service.ts
```diff
@@ -1,16 +1,23 @@
...
 @Injectable({ providedIn: 'root' })
 export class IconRegistryService {
   private static readonly SVG_ICON_PREFIX: string = 'svg:';
 
+  private registeredIcons: string[] = [];
+
   public constructor(private readonly matIconRegistry: MatIconRegistry, private readonly sanitizer: DomSanitizer) {}
 
   public registerSvgIcon(iconInfo: SvgIconRegistrationInfo): void {
+    if (!this.registeredIcons.includes(iconInfo.key)) {
```
anandtiwary, line 16: Qq: can we pull this from matIconRegistry?

projects/assets-library/src/icons/icon-registry.service.ts
```diff
@@ -1,16 +1,23 @@
...
+      this.registeredIcons.push(iconInfo.key);
+      this.registeredIcons = orderBy(this.registeredIcons, key => key, 'asc');
```
arjunlalb, line 18: do we need to store them in order? if the order is important only in the read path, we could do it there instead of ordering for every insert.

itssharmasandeep: RIght :), removed  as we do not need it here

arjunlalb (approved):

projects/assets-library/src/icons/icon-registry.service.ts
anandtiwary: @itssharmasandeep Also, this is taking O(n2). Why not use a map?

itssharmasandeep: Right, used set instead, here is the follow up PR https://github.com/hypertrace/hypertrace-ui/pull/2304#discussion_r1261592929

PR #2323, opened 2023-07-25
fix: external url copy

itssharmasandeep:
## 😵‍💫 ISSUE

Recently we had a fix related to external navigation where we banned the external navigation as the first navigation. Due to this copy and paste the external navigation is broken and does not work.

```ts
// Suppose we have an external URL as 'http://localhost/external;url=https://google.com'
// So if we click on this URL from the app, it would work as this is not the first navigation.
// But if we copy the URL and paste it into the new window and try to navigate it would through an error.
```

## 🔨  FIX

To fix this, we have used a separate anchor tag inside `ht-link` component for the external URL which has `href` exactly equal to the external URL itself.

### DIFF
--- a/projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
+++ b/projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
@@ -8,9 +8,11 @@ import { IconSize } from '../icon/icon-size';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div *ngIf="this.paramsOrUrl" class="open-in-new-tab" htTooltip="Open in a new tab">
-      <ht-link [paramsOrUrl]="this.paramsOrUrl" ariaLabel="Open in a new tab">
-        <ht-icon icon="${IconType.OpenInNewTab}" [size]="this.iconSize"></ht-icon>
-      </ht-link>
+      <ht-event-blocker event="click">
+        <ht-link [paramsOrUrl]="this.paramsOrUrl" ariaLabel="Open in a new tab">
+          <ht-icon icon="${IconType.OpenInNewTab}" [size]="this.iconSize"></ht-icon>
+        </ht-link>
+      </ht-event-blocker>
     </div>
   `
 })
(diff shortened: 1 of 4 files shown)


projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
```diff
@@ -8,11 +8,9 @@ import { IconSize } from '../icon/icon-size';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div *ngIf="this.paramsOrUrl" class="open-in-new-tab" htTooltip="Open in a new tab">
-      <ht-event-blocker event="click">
-        <ht-link [paramsOrUrl]="this.paramsOrUrl" ariaLabel="Open in a new tab">
-          <ht-icon icon="${IconType.OpenInNewTab}" [size]="this.iconSize"></ht-icon>
-        </ht-link>
-      </ht-event-blocker>
+      <ht-link [paramsOrUrl]="this.paramsOrUrl" ariaLabel="Open in a new tab">
```
arjunlalb, line 11: we need to add the event blocker back in here since this component could be used within a cell of a table or card.

arjunlalb (approved):

projects/components/src/open-in-new-tab/open-in-new-tab.component.ts
itssharmasandeep: It is not working with the event blocker. When I removed it, it started working. My guess would be since we're using anchor tag directly with href. In the internal one, we're using `routerLink` instead, so angular was taking care of the navigation. Should we also use a click method for this which can do the navigation?

itssharmasandeep: Now I think about it that means, every link with event blocker wouldn't work for external navigation

itssharmasandeep: Updated the PR to fix this.

arjunlalb (approved):

PR #2324, opened 2023-07-26
fix: Table related issues

itssharmasandeep:
This contains the following changes:

1. On the initial load, we're NOT assigning the pixel widths now. CSS will take care of the widths accordingly. This is done to prevent unnecessary tasks on load and scrolling issue was coming because of this.
2. Once a user resizes the column, we will assign the pixel width to the column, so that the resizing works seamlessly.
3. Resize handler's size has been increased from `3px` to `5px`
4. Type for the column width has been updated for better type-safety
    - Now we support flex value as a number pixel value and percent value as a string.
    - Code will throw an error if the width is not compatible, this is an additional check in case of dashboards with table widgets.
5. Template changes only update a few of the properties which fixes the CSS


arjunlalb:
Could you please summarize the changes in the PR in comments or description?

projects/components/src/table/table.component.ts
```diff
@@ -93,9 +95,10 @@ import { CdkDragDrop, moveItemInArray } from '@angular/cdk/drag-drop';
             <cdk-header-cell
               [attr.data-column-index]="index"
               *cdkHeaderCellDef
-              [style.flex-basis]="columnDef.width"
-              [style.max-width]="columnDef.width"
-              [style.min-width]="columnDef.minWidth ?? columnDef.width ?? this.minColumnWidth"
+              [style.flex-basis]="this.getFlexBasis | htMemoize: columnDef.width"
```
adisreyaj, line 98: Why not have single method which return a json which can be passed as input to `ngStyle`? We'll get rid of these multiple memoize directives.

arjunlalb: +1

projects/components/src/table/table.component.ts
```diff
@@ -158,9 +161,10 @@ import { CdkDragDrop, moveItemInArray } from '@angular/cdk/drag-drop';
             </cdk-header-cell>
             <cdk-cell
               *cdkCellDef="let row"
-              [style.flex-basis]="columnDef.width"
-              [style.min-width]="columnDef.minWidth ?? columnDef.width ?? this.minColumnWidth"
-              [style.max-width]="columnDef.width"
+              [style.flex-basis]="this.getFlexBasis | htMemoize: columnDef.width"
```
arjunlalb, line 164: if we use a style object, we can use the source here as well.

projects/components/src/table/table.component.ts
```diff
@@ -1093,12 +1143,32 @@ export class TableComponent
...
+
+    const widthInPx = this.getColWidthInPx(width) - (hasScrollbar ? TableComponent.SCROLLBAR_WIDTH_IN_PX : 0);
+
+    const nonStateResizableVisibleColumnConfigs = this.visibleColumnConfigs
+      .filter(column => !this.isStateColumn(column))
+      .filter(column => {
+        const colWidth = this.initialColumnConfigIdWidthMap.get(column.id) ?? -1;
+        const isPxWidthColumn =
+          typeof colWidth === 'string' ? colWidth.substring(colWidth.length - 2) === 'px' : colWidth > -1;
+
```
arjunlalb, line 1164: good candidate for the util

projects/components/src/table/util/column-width.util.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    if (!Number.isNaN(value) && unit === '%') {
+      return true;
+    }
+
+    return false;
+  }
+
+  public static getWidth(value?: TableColumnWidth): string | undefined {
+    if (value === undefined || typeof value === 'number') {
```
arjunlalb, line 29: you can define a shared predicate/method for this check

projects/components/src/table/table-api.ts
```diff
@@ -25,8 +25,8 @@ export interface TableColumnConfig<TableColumnOptions = unknown> {
   sortable?: boolean;
   filterable?: boolean;
   alignment?: TableCellAlignmentType;
-  width?: number | string;
-  minWidth?: number | string;
+  width?: TableColumnWidth;
```
Exac, line 28: Nitpick: I think it would be a good idea to add some type doc explaining that passing a number without a unit is a flex width, and that `px` and `%` are also supported. Either here (and on the `maxWidth`, or on the `TableColumnWidth` type.

arjunlalb: +1

arjunlalb (approved):

PR #2325, opened 2023-07-27
fix: making table column width type safe

itssharmasandeep:


itssharmasandeep:
Closing this. will fix this with https://github.com/hypertrace/hypertrace-ui/pull/2324

PR #2335, opened 2023-08-02
fix: add support for showing metadata in list-view and use it in span detail

arjunlalb:
Adding support for `metadata` in list view.
Metadata corresponding to each list key will get added below the corresponding row.

Using this to support showing cookie metadata in span detail.


projects/components/src/list-view/list-view.component.ts
```diff
@@ -23,24 +24,34 @@ import { ListViewValueRendererDirective } from './list-view-value-renderer.direc
...
       </div>
-      <div class="data-row" [class]="this.display" *ngFor="let record of this.records">
-        <div class="key">
-          <ng-container
-            *ngTemplateOutlet="
-              this.keyRenderer ? this.keyRenderer!.getTemplateRef() : defaultKeyRenderer;
-              context: { $implicit: record }
-            "
-          ></ng-container>
+      <ng-container *ngFor="let record of this.records">
```
arjunlalb, line 27: data-row is inside a container now. no other changes here

projects/components/src/list-view/list-view.component.ts
```diff
@@ -23,24 +24,34 @@ import { ListViewValueRendererDirective } from './list-view-value-renderer.direc
...
-        <div class="value">
-          <ng-container
-            *ngTemplateOutlet="
-              this.valueRenderer ? this.valueRenderer!.getTemplateRef() : defaultValueRenderer;
-              context: { $implicit: record }
-            "
-          ></ng-container>
-        </div>
-      </div>
+        <ng-container *ngIf="!(this.getMetadataForKey | htMemoize: record.key | htIsEmpty)">
```
arjunlalb, line 46: added segment

projects/components/src/list-view/list-view.component.scss
```diff
@@ -57,6 +57,30 @@ $horizontal-offset: 12px;
...
+
+    .marker {
+      width: 4px;
+      height: 4px;
+      border-radius: 50%;
+      background: $gray-5;
+    }
+    .key {
+      color: $gray-5;
+      font-weight: 400;
```
itssharmasandeep, line 75: nit: can we use an existing mixin (if any)?

projects/observability/src/shared/components/span-detail/span-data.ts
```diff
@@ -11,6 +11,7 @@ export interface SpanData {
   requestBody: string;
   responseHeaders: Dictionary<unknown>;
   responseCookies: Dictionary<unknown>;
+  responseCookiesMetadata: Dictionary<Dictionary<unknown>>;
```
itssharmasandeep, line 14: should this be optional?

arjunlalb: since other related fields aren't optional, making this also of the same category.

projects/components/src/list-view/list-view.component.scss
arjunlalb: we usually override single properties like this. there is no mixin for font-weights

itssharmasandeep (approved):

adisreyaj (approved):

PR #2338, opened 2023-08-04
fix: support for entity context options in explore query

itssharmasandeep:


projects/observability/src/shared/graphql/request/handlers/explore/explore-graphql-query-builder.service.ts
```diff
@@ -108,4 +111,14 @@ export class ExploreGraphqlQueryBuilderService {
       this.specBuilder.exploreSpecificationForAttributeExpression(expression)
     );
   }
+
+  private buildEntityContextOptionsArgument(entityContextOptions?: EntityContextOptions): GraphQlArgument[] {
+    if (!isNil(entityContextOptions) && !isNil(entityContextOptions.includeNonLiveEntities)) {
```
adisreyaj, line 116:
```suggestion
    if (!isNil(entityContextOptions?.includeNonLiveEntities)) {
```

itssharmasandeep: Can't do this because the compiler will say this 
Screenshot 2023-08-04 at 6 10 53 PM

itssharmasandeep: We would have to use assertion

arjunlalb (approved):

PR #2358, opened 2023-08-20
fix: support async rendered menu items

aaron-steinfeld:
## Description
Previously, we were using the length of the query list to determine whether the menu is active. This means if it changes (either from or to zero items, for example when using an ngIf with an async pipe), it won't be updated with on push change detection. Switched to a more robust approach here.

### DIFF
--- a/projects/components/src/menu-dropdown/menu-dropdown.component.ts
+++ b/projects/components/src/menu-dropdown/menu-dropdown.component.ts
@@ -1,7 +1,10 @@
-import { ChangeDetectionStrategy, Component, ContentChildren, Input, QueryList } from '@angular/core';
+import { AfterContentInit, ChangeDetectionStrategy, Component, ContentChildren, Input, QueryList } from '@angular/core';
 import { IconType } from '@hypertrace/assets-library';
+import { queryListAndChanges$ } from '@hypertrace/common';
+import { Observable, of } from 'rxjs';
 import { IconSize } from '../icon/icon-size';
 import { MenuItemComponent } from './menu-item/menu-item.component';
+import { map } from 'rxjs/operators';
 
 @Component({
   selector: 'ht-menu-dropdown',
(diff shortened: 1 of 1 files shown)


projects/components/src/menu-dropdown/menu-dropdown.component.ts
```diff
@@ -37,4 +40,12 @@ export class MenuDropdownComponent {
 
   @ContentChildren(MenuItemComponent)
   public items?: QueryList<MenuItemComponent>;
+
+  public hasItems$: Observable<boolean> = of(false);
```
anandtiwary, line 44: we can mark it protected.

anandtiwary (approved):

projects/components/src/menu-dropdown/menu-dropdown.component.ts
aaron-steinfeld: ah I didn't realize that was supported now. will do that going forwarded.

aaron-steinfeld, line 44:
```suggestion
  protected hasItems$: Observable<boolean> = of(false);
```

projects/components/src/menu-dropdown/menu-dropdown.component.ts
```diff
@@ -37,4 +40,12 @@ export class MenuDropdownComponent {
 
   @ContentChildren(MenuItemComponent)
   public items?: QueryList<MenuItemComponent>;
```
anandtiwary, line 42: nit: this can be private now

anandtiwary (approved):

projects/components/src/menu-dropdown/menu-dropdown.component.ts
aaron-steinfeld: It's assigned by angular - unless something changed anything managed by angular has always been public.

PR #2366, opened 2023-08-22
Attribute sort specification

jake-bassett:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
@@ -22,10 +22,7 @@ export class AttributeSortSpecificationModel extends SpecificationModel<Specific
     key: 'direction',
     type: {
       key: ENUM_TYPE.type,
-      values: [
-        TableSortDirection.Ascending,
-        TableSortDirection.Descending
-      ]
+      values: [TableSortDirection.Ascending, TableSortDirection.Descending]
     } as EnumPropertyTypeInstance
   })
   public direction!: TableSortDirection;
--- a/projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
@@ -1,7 +1,6 @@
 import { Dictionary } from '@hypertrace/common';
-import { TableSortDirection } from '@hypertrace/components';
-import { ENUM_TYPE, EnumPropertyTypeInstance } from '@hypertrace/dashboards';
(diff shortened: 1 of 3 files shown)


Exac (approved):
Quick question: I noticed that in some of our code we have three sort directions, `'ASC'`, `'DESC'`, and `undefined`. Since we are asserting that the `direction` is always present, I suppose we don't have to worry about that in this particular model?

Otherwise LGTM.

projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
```diff
@@ -0,0 +1,33 @@
...
+  @ModelProperty({
+    key: 'key',
+    type: STRING_PROPERTY.type,
+    required: true
+  })
+  public key!: string;
+
+  @ModelProperty({
+    key: 'direction',
+    type: STRING_PROPERTY.type,
```
itssharmasandeep, line 22: we need to mark it as required if it is

projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
```diff
@@ -0,0 +1,33 @@
...
+import { GraphQlSortDirection } from '../../../../graphql/model/schema/sort/graphql-sort-direction';
+import { Specification } from '../../../../graphql/model/schema/specifier/specification';
+import { SpecificationBuilder } from '../../../../graphql/request/builders/specification/specification-builder';
+import { SpecificationModel } from './specification.model';
+
+@Model({
+  type: 'attribute-sort-specification',
+  displayName: 'Attribute Sort'
+})
+export class AttributeSortSpecificationModel extends SpecificationModel<Specification> {
```
itssharmasandeep, line 12: Confused about why we need this to extend the specification model?

projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
palbizu: direction could be undefined IMO (when is not applied). Do you think that should be mandatory anyways?

projects/observability/src/shared/dashboard/data/graphql/specifiers/attribute-sort-specification.model.ts
jake-bassett: You are the man. Thanks for calling this out. Its unnecessary for this model.

jake-bassett:
> Quick question: I noticed that in some of our code we have three sort directions, `'ASC'`, `'DESC'`, and `undefined`. Since we are asserting that the `direction` is always present, I suppose we don't have to worry about that in this particular model?
> 
> Otherwise LGTM.

Yeah, in this case just the omission of the model itself is the third state. If no sorting was required, this model would never be invoked.

PR #2388, opened 2023-09-11
fix: topology edge filter support

itssharmasandeep:


projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
```diff
@@ -22,6 +23,7 @@ import {
 } from '../../../../graphql/request/handlers/entities/query/topology/entity-topology-graphql-query-handler.service';
 import { GraphQlDataSourceModel } from '../graphql-data-source.model';
 import { TopologyMetricsData, TopologyMetricsModel } from './metrics/topology-metrics.model';
+import { AttributeExpression, GraphQlFieldFilter } from '../../../../../public-api';
```
anandtiwary, line 26: please fix the imports

projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
```diff
@@ -90,20 +100,25 @@ export class TopologyDataSourceModel extends GraphQlDataSourceModel<TopologyData
       metricSpecifications: this.getAllMetricSpecifications(this.edgeMetricsModel)
     };
 
-    return this.query<EntityTopologyGraphQlQueryHandlerService>(
-      filters => ({
+    return this.query<EntityTopologyGraphQlQueryHandlerService>(filters => {
+      const topologyFilters = this.getTopologyFilters(filters as GraphQlFieldFilter[]);
```
anandtiwary, line 104: we should do a check before this cast.

projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
```diff
@@ -160,3 +214,13 @@ export interface TopologyData {
...
   edgeMetrics: TopologyMetricsData;
 }
+
+export interface TopologyEdgeFilterConfig {
+  entityType: string;
+  fields: string[];
+}
+
+interface TopologyFilters {
+  forNode: GraphQlFieldFilter[];
```
anandtiwary, line 224: could we just call it `nodes` and `edges`? The <for> in the variable name gives it an impression that we are writing a for loop

itssharmasandeep: Done

projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
itssharmasandeep: What type of check? Checking any fields in the data? The  `GraphQlFIlter` interface only has one function.  and we do not know how to extract data from there.

anandtiwary: As in, before casting filters to `GraphQlFieldFilter`, we should check if it has properties of GraphQlFieldFilter or not. There can be Id Filters in the array that may break this method at run time.

itssharmasandeep: Okay, SO I updated the logic, Now we're checking the field filters and if it is not then we're assigning that to node filters for now.

anandtiwary (approved):

palbizu (approved):

PR #2402, opened 2023-09-15
fix: topology query caching

itssharmasandeep:
### 😵‍💫 ISSUE 
Whenever we apply an edge filter for a particular entity, we need to omit the edges for all the other entities and only query for the particular entity edge. Once we do that and remove the filter, the query fired still doesn't contain the omitted edges and only contains the edge filter once, which leads to the wrong results

### FIX 🔨 
This is happening because of the `id` field in the entities query. This field is used by Apollo to cache the query structure. Once we change this and use an alias `entityId` this starts working fine. We did a similar thing earlier for another issue with entities query. https://github.com/hypertrace/hypertrace-ui/pull/2334

_NOTE: `EntityTopologyGraphQlQueryHandlerService` has its own implementation of `entities` query_


projects/observability/src/shared/graphql/request/handlers/entities/query/topology/entity-topology-graphql-query-handler.service.ts
```diff
@@ -154,8 +154,8 @@ export class EntityTopologyGraphQlQueryHandlerService
     nodeMap: Map<string, EntityNode>,
     serverEntity: TopologyEntity
   ): EntityNode {
-    if (nodeMap.has(serverEntity.id)) {
-      return nodeMap.get(serverEntity.id)!;
+    if (nodeMap.has(serverEntity.entityId)) {
```
anandtiwary, line 157: @aaron-steinfeld Please check this change. @itssharmasandeep Please explain the change.

itssharmasandeep: Updated @anandtiwary @aaron-steinfeld

aaron-steinfeld: So by not using the special value 'id' we just tell apollo not to cache by ref, which is fine. But given this keeps coming up, we really should make the correct change and fix the apollo config to disable this feature - we don't want to ever use it in our current design.

aaron-steinfeld: If I understood the initial issue correctly though, are you modifying the object returned from the query directly? (If so, removing the caching by id won't resolve that)

itssharmasandeep: I tried this https://www.apollographql.com/docs/react/caching/cache-configuration#disabling-normalization earlier, do not remember if I was making a mistake somewhere but it did not work at that time

aaron-steinfeld: That's the right place to look. Feel free to go ahead with this one in the meantime, but if you have a chance and can put together a PR with the client approach, we can try to dig into what was going wrong.

aaron-steinfeld (approved):

projects/observability/src/shared/graphql/request/handlers/entities/query/topology/entity-topology-graphql-query-handler.service.ts
itssharmasandeep: Sure, I will try to do that. Thanks :)

PR #2404, opened 2023-09-15
feat: support for path while setting a cookie

arjunlalb:
## Description
Support for setting a path in the cookie.

### Testing
Verified manually.


anandtiwary (approved):

aaron-steinfeld:
Very curious on use case here

PR #2412, opened 2023-09-18
feat: Adding filters for supported subpath operators

palbizu:
## Description
Adding Filters for Supported Subpath Operators

### Testing
Locally Tested

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
+++ b/projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
@@ -1,7 +1,6 @@
 import { ChangeDetectionStrategy, Component, Inject, InjectionToken } from '@angular/core';
 import { TargetAttribute } from '@hypertrace/observability';
 import { FilterValue } from '../filter/filter';
-import { FilterAttribute } from '../filter/filter-attribute';
 
 export const FILTER_BUTTON_WRAPPER = new InjectionToken<FilterButtonWrapperData>('INTERVAL_DATA');
 
@@ -11,7 +10,6 @@ export const FILTER_BUTTON_WRAPPER = new InjectionToken<FilterButtonWrapperData>
   template: `
     <ht-filter-button
       *ngIf="this.filterButtonWrapperData"
-      [metadata]="this.filterButtonWrapperData.allAttributes"
       [attribute]="this.filterButtonWrapperData.targetAttribute.attribute"
       [value]="this.filterButtonWrapperData.value"
       [subpath]="this.filterButtonWrapperData.targetAttribute.subpath"
@@ -23,7 +21,6 @@ export class FilterButtonWrapperComponent {
 }
 
 export interface FilterButtonWrapperData {
-  allAttributes: FilterAttribute[];
   targetAttribute: TargetAttribute;
   value: FilterValue;
 }
--- a/projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
+++ b/projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
@@ -13,6 +13,7 @@ export const FILTER_BUTTON_WRAPPER = new InjectionToken<FilterButtonWrapperData>
       [metadata]="this.filterButtonWrapperData.metadata"
       [attribute]="this.filterButtonWrapperData.attribute"
       [value]="this.filterButtonWrapperData.value"
+      [subpath]="this.filterButtonWrapperData.subpath"
     ></ht-filter-button>
   `
 })
@@ -24,4 +25,5 @@ export interface FilterButtonWrapperData {
   metadata?: FilterAttribute[];
   attribute?: FilterAttribute;
   value?: FilterValue;
+  subpath?: string;
 }
--- a/projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
+++ b/projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
@@ -1,4 +1,5 @@
 import { ChangeDetectionStrategy, Component, Inject, InjectionToken } from '@angular/core';
+import { TargetAttribute } from '@hypertrace/observability';
 import { FilterValue } from '../filter/filter';
 import { FilterAttribute } from '../filter/filter-attribute';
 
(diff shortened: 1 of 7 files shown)


projects/observability/src/shared/components/cartesian/chart.ts
```diff
@@ -66,6 +66,7 @@ export interface Band<TInterval> {
 export interface GroupByFilter {
```
aaron-steinfeld, line 66: What's a "groupByFillter"? Was this added recently? Given there's no operator or value here, guessing this is just misnamed and describes a group by, not a filter.

projects/components/src/filtering/filter/builder/types/abstract-filter-builder.ts
```diff
@@ -19,6 +19,14 @@ export abstract class AbstractFilterBuilder<TValue extends FilterValue> {
     return [...this.supportedTopLevelOperators(), ...this.supportedSubpathOperators()];
   }
 
+  public buildFiltersForSupportedSubpathOperators(
+    attribute: FilterAttribute,
+    value: TValue,
+    subpath: string
```
aaron-steinfeld, line 25: I suspect you'll end up with much cleaner code if you fix FilterAttribute to reflect that an attribute may include a subpath rather than trying to move them around in tandem. That said, haven't looked at the filter button stuff in depth, so it's possible it needs a more thorough refactor to fix that and it's out of scope - I'd at least suggest giving it a try though.

projects/components/src/filtering/filter-button/filter-button.component.ts
```diff
@@ -60,13 +63,21 @@ export class FilterButtonComponent implements OnChanges {
...
+        ? this.buildAvailableFilters(this.attribute, this.value, this.subpath)
         : [];
   }
 
   public onFilterClick(filter: Filter): void {
     this.filterUrlService.addUrlFilter(this.metadata || [], filter);
   }
+
+  private buildAvailableFilters(attribute: FilterAttribute, value: FilterValue, subpath?: string): Filter[] {
+    if (subpath !== undefined) {
```
aaron-steinfeld, line 75: Push this down - filter builder should handle the optional subpath.

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -145,7 +145,12 @@ export abstract class ExploreCartesianDataSourceModel extends GraphQlDataSourceM
                 attribute: request.attributes?.find(
                   attribute => attribute.name === request.groupBy?.keyExpressions?.[0].key
                 ),
-                metadata: request.attributes
+                metadata: request.attributes,
```
aaron-steinfeld, line 148: Pre-existing, but if we're picking out the appropriate attribute why do we need to send the whole list down too? Also attribute and metadata are basically synonyms in our system - as can be seen here since they have the same datatype. We should be using consistent names, especially in the same type (e.g. targetAttribute and allAttributes).

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
```diff
@@ -145,7 +145,12 @@ export abstract class ExploreCartesianDataSourceModel extends GraphQlDataSourceM
...
+                ...(request.groupBy?.keyExpressions?.[0].subpath
+                  ? {
```
aaron-steinfeld, line 150: Just use optional access here, no need for the ternary.

projects/observability/src/shared/components/cartesian/chart.ts
palbizu: renamed to groupBy

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
palbizu: done

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
palbizu: allAttributes necessary to redirect on filter click (from filter-url.service), anyway renaming them

aaron-steinfeld: Setting the url (going from object to string) shouldn't require all attributes; reading it does (to go from string to object)

projects/components/src/filtering/filter-button/filter-button.component.ts
palbizu: done

projects/observability/src/shared/dashboard/data/graphql/explore/explore-cartesian-data-source.model.ts
palbizu: yes, you're right. let me remove it

projects/observability/src/shared/components/cartesian/chart.ts
```diff
@@ -63,9 +63,13 @@ export interface Band<TInterval> {
   hide?: boolean;
 }
 
-export interface GroupByFilter {
-  attribute?: FilterAttribute;
-  metadata: FilterAttribute[];
+export interface GroupBy {
```
aaron-steinfeld, line 66: I would suggest not defining these two in chart.ts, since the concepts are not limited (or even used directly by) charts. Further, do we need new types or could we just update FilterAttribute to have an optional subpath for example (not sure on the scope of that change). Another relevant type would be `AttributeExpression` which models an attribute key and optional subpath. Basically I'm just trying to avoid us introducing a new type for each new usage when the data is the same.

palbizu: Agreed, I prefer to use AttributeExpression. Do you agree? I need to extract the reference but straightforward.

aaron-steinfeld: I agree that's all that's needed, so if that can be used without a big refactor that seems to make sense. If you needed to wrap it (say, updating FilterAttribute's reference string to be an AttributeExpression) that could work too.

palbizu: done

aaron-steinfeld (approved):

projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
```diff
@@ -21,7 +21,6 @@ export class FilterButtonWrapperComponent {
 }
 
 export interface FilterButtonWrapperData {
-  metadata?: FilterAttribute[];
-  attribute?: FilterAttribute;
-  value?: FilterValue;
+  targetAttribute: FilterAttributeExpression;
```
anandtiwary, line 24: @aaron-steinfeld  @palbizu   `targetAttribute` can be null, and we need to handle it. UI code failed at run time, crashing the app and causing prod issues. 

https://traceableai.atlassian.net/browse/ENG-35622
https://traceableai.atlassian.net/browse/ENG-35603

aaron-steinfeld: Commented on the fix and can continue discussion there https://github.com/hypertrace/hypertrace-ui/pull/2421#discussion_r1334677536

PR #2413, opened 2023-09-19
feat: handle empty string in displaystring pipe

adisreyaj:


Greedy-Geek (approved):

projects/common/src/utilities/formatters/string/string-formatter.ts
```diff
@@ -13,7 +15,7 @@ export const displayString = (provided?: unknown, defaultValueOnEmpty: string =
     case 'function':
       return 'Function';
     case 'string':
-      return provided;
+      return !isEmpty(provided.trim()) ? provided : defaultValueOnEmpty;
```
aaron-steinfeld, line 18: Do we know every place we use this pipe we want to treat empty string as undefined? Most places this is probably safe, but this seems like a risky change especially for something that feels like a caller bug.

23nobody (approved):

PR #2417, opened 2023-09-20
feat: add column config local storage support to table

arjunlalb:
Remove column config local storage support from table widget renderer.
Adapt and add it to table component directly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "@ngbracket/ngx-layout": "^15.0.0",
         "@types/d3-hierarchy": "^2.0.2",
         "@types/d3-transition": "1.1.5",
-        "apollo-angular": "^5.0.0",
+        "apollo-angular": "^5.0.1",
         "core-js": "^3.32.2",
         "d3-array": "^2.12.0",
         "d3-axis": "^2.1.0",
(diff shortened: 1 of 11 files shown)


itssharmasandeep (commented):
I haven't looked at the full code yet but I have some doubts.
1. Will we have a way to not set preferences?
2. Will the preference setting be a default behavior?

The problem that I see here is with the column resizing. We set pixel widths after resizing and if we save everything to the preferences then we will never have a way to revert to the original state in terms of the column sizing. and it will create a problem if someone goes from one screen to another.
Should we have a way to reset that as well? I really think we should.
We can add an action in the table menu to reset columns the columns.

jake-bassett (approved):
Code looks good; no red flags here that I could see. However, table is a beast, so please be sure to test with all forms - flat, expandable, and tree versions. In particular, tree/waterfall.

arjunlalb:
> I haven't looked at the full code yet but I have some doubts.
> 
> 1. Will we have a way to not set preferences?
> 2. Will the preference setting be a default behavior?
> 
> The problem that I see here is with the column resizing. We set pixel widths after resizing and if we save everything to the preferences then we will never have a way to revert to the original state in terms of the column sizing. and it will create a problem if someone goes from one screen to another. Should we have a way to reset that as well? I really think we should. We can add an action in the table menu to reset columns the columns.

We are currently saving only two attributes from the column for user preferences. `id` and the boolean flag `visible`. Other column properties will continue to work as intended.

arjunlalb:
> Code looks good; no red flags here that I could see. However, table is a beast, so please be sure to test with all forms - flat, expandable, and tree versions. In particular, tree/waterfall. Also explorer table, despite being expandable, works a little different and can be thought of as a fourth version worth testing.
> 
> Approved, but please work with @itssharmasandeep on the column resizing concerns he has above as he recently improved that logic and knows it best.

I've tested all the variants you mentioned. LGTM.

jake-bassett (approved):

PR #2419, opened 2023-09-21
fix: add gql request name support

itssharmasandeep:
#### ⚠️ Issue
Currently, we fire a lot of queries of the same type, for example: we may fire up to 10 different entity queries for different use cases. Since we have a number of queries of the same type, debugging becomes difficult for the teams as they have to look for the schema to realize which query is failing and what is the use case.

#### 💡 Solution
Added an optional attribute `name` in the Graphql options which can be used to fire a names query depending upon the use case

`Before`

```ts
query {...}
```

`After`

```ts
query abcd {...}
```

### DIFF
--- a/projects/graphql-client/src/graphql-request.service.ts
+++ b/projects/graphql-client/src/graphql-request.service.ts
@@ -145,7 +145,7 @@ export class GraphQlRequestService {
     options: GraphQlRequestOptions
   ): Observable<TResponse> {
     return type === GraphQlHandlerType.Mutation
-      ? this.executeMutation(requestString)
+      ? this.executeMutation(requestString, options)
       : this.executeQuery(requestString, options);
   }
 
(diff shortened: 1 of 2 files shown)


projects/graphql-client/src/graphql-request.service.ts
```diff
@@ -155,7 +155,7 @@ export class GraphQlRequestService {
   ): Observable<TResponse> {
     return this.apollo
       .query<TResponse>({
-        query: gql(requestString),
+        query: gql(`query ${options?.name ?? ''} ${requestString}`),
```
aaron-steinfeld, line 158: Will this work with current batching behavior? I would think it would either break or prevent batching - neither good. Work to push the name further down so multiple named queries can be grouped (you might want to use alias rather than a query name, for example).

PR #2420, opened 2023-09-22
feat: inbuilt csv download support for tables

arjunlalb:
## Description
Adding generic csv download support for tables.

Strategy:

- Introduce the concept of csv generators for cells. Like parsers/renderers, custom ones can be provided.
- Introduce a table csv downloader service that the consumers can use to communicate with table. 
- If the table gets a request to download csv for its ID via the service above, it'll requery the data source with a high limit and use column configs to generate csv for each cell and then download is triggered using file download service. All cells with no csv generator attached will be skipped.
- CSV Generators are picked up based on the cell renderer used.

Improvements in the pipeline
- Make limits for csv customizable.
- Take column visibility into account.
- Support batching to fetch required data in chunks instead of one go.

### Testing
Testing in progress.


projects/components/src/table/table.component.ts
```diff
@@ -648,6 +657,53 @@ export class TableComponent
...
+              })
+            : of({ data: [], totalCount: 0 })
+          ).pipe(
+            take(1),
+            map((response: TableDataResponse<TableRow>) => response.data),
+            map(rows => {
+              const csvGeneratorMap = new Map(
+                columnConfigs
+                  .filter(
+                    column => !isNil(column.csvGenerator) && !(column.csvGenerator instanceof TableCellSkipCsvGenerator)
```
aaron-steinfeld, line 678: two ways of modeling the same thing, so could just skip the skip generator and omit only.

projects/components/src/table/cells/table-cell-renderer-base.ts
```diff
@@ -54,4 +60,8 @@ export abstract class TableCellRendererBase<TCellData, TValue = TCellData, TColu
   public get tooltip(): string {
     return this._tooltip;
   }
+
+  public getValueForCsv(): string | undefined {
```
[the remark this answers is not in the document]
aaron-steinfeld: Later it looks like we're not - we're calling the csv generator in table component against all data, so why is the renderer involved I guess would be the next question

projects/components/src/table/table.component.ts
```diff
@@ -648,6 +657,53 @@ export class TableComponent
     this.changeDetector.detectChanges();
   }
 
+  private downloadCsv(): void {
```
aaron-steinfeld, line 660: This file is a crime against humanity already, would refactor as much of this complexity outside as possible

projects/components/src/table/cells/table-cell-csv-generator-base.ts
```diff
@@ -0,0 +1,10 @@
+import { Injector } from '@angular/core';
+import { CoreTableCellCsvGeneratorType } from './types/core-table-cell-csv-generator-type';
+
+export abstract class TableCellCsvGeneratorBase<TCellData, TRowData = unknown> {
```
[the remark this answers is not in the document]
aaron-steinfeld: That's fair, although I think I mentioned in another comment you can probably shift the generators to services so they can support directly injecting like any other service.

projects/components/src/table/table.component.ts
aaron-steinfeld: I had mainly meant to reduce file size, but now that you mention it would be at least worth considering whether table controls would be a better home to this (not sure haven't looked in there in a while and not sure what it has access to, but this _is_ a control).

projects/components/src/table/table.component.ts
arjunlalb: fixed

projects/components/src/table/table.component.ts
arjunlalb: Table controls do not have access to data. Only thing it can do is to be a proxy between the table and the messaging service we added. Since we have a messaging service, I think we should avoid another intermediary.

projects/components/src/table/cells/table-cell-csv-generator-lookup.service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/*
+ * @TableCellCsvGenerator decorator is used to configure type for cell CsvGenerators
+ */
+
+export function TableCellCsvGenerator(
+  tableCellCsvGeneratorMetadata: TableCellCsvGeneratorMetadata
+): TableCellCsvGeneratorDecorator {
+  return (constructor: TableCellCsvGeneratorConstructor<unknown, unknown>): void => {
+    constructor.type = tableCellCsvGeneratorMetadata.type;
```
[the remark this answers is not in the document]
arjunlalb: do you mean something like a single service that contains all the column to csv mapper thats injected via a token, and if its available, let the table directly consume it?

projects/components/src/table/table.component.ts
aaron-steinfeld: Messaging service? can you point me to that? I looked at the controls and agree they don't have access to data, but I always see them being wired directly to table (which agree, isn't a great model - should have accept the table as content and wired directly to it).

projects/components/src/table/cells/table-cell-csv-generator-lookup.service.ts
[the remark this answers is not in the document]
aaron-steinfeld: Reading my comment further, the second point I was making was actually that we can make the generators themselves injectable (similar to how our gql handlers are) and forgo the decorators at all. This has several benefits
- reduce registration complexity
- simplifies and optimizes consumption. A single root injectable instance rather than new'ing an instance of the generator for every usage.
- allows generators to inject like any other service.

projects/components/src/table/table.component.ts
arjunlalb: @aaron-steinfeld I have added a `TableCsvDownloaderService` which is just a way to let the table know that it should download CSV.

projects/components/src/table/cells/table-cell-csv-generator-lookup.service.ts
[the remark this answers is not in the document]
arjunlalb: Updated this to reduce registration complexity.

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
```diff
@@ -0,0 +1,13 @@
+import { TableCellCsvGeneratorBase } from '../table-cell-csv-generator-base';
+import { Injectable } from '@angular/core';
+import { isUndefined } from 'lodash-es';
+
+@Injectable({ providedIn: 'root' })
+export class TableCellBooleanCsvGenerator implements TableCellCsvGeneratorBase<boolean> {
+  public readonly type = 'CSV_GENERATOR';
```
aaron-steinfeld, line 7: What's the purpose of this value?

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
```diff
@@ -0,0 +1,13 @@
+import { TableCellCsvGeneratorBase } from '../table-cell-csv-generator-base';
+import { Injectable } from '@angular/core';
+import { isUndefined } from 'lodash-es';
+
+@Injectable({ providedIn: 'root' })
+export class TableCellBooleanCsvGenerator implements TableCellCsvGeneratorBase<boolean> {
```
aaron-steinfeld, line 6: naming nit - TableCellCsvGeneratorBase -> TableCellCsvGenerator (try to avoid things like "abstract" or "base" - especially for interfaces)

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
```diff
@@ -0,0 +1,13 @@
...
+  public readonly type = 'CSV_GENERATOR';
+  public readonly cellType: string = 'boolean';
```
aaron-steinfeld, line 8: Does this need to match `CoreTableCellCsvGeneratorType.Boolean` ? if so, can we use that value

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  public generateSafeCsv(cellData?: boolean): string | undefined {
```
aaron-steinfeld, line 10: how confident are we that every value passed to this will be boolean|undefined? do we enforce it? Stray nulls etc. often cause issues with these type of contracts.

projects/components/src/table/cells/table-cell-csv-generator-base.ts
```diff
@@ -0,0 +1,13 @@
+import { Dictionary } from '@hypertrace/common';
+import { InjectionToken } from '@angular/core';
+
+export const TABLE_CELL_CSV_GENERATORS = new InjectionToken<unknown[][]>('TABLE_CELL_CSV_GENERATORS');
+
+export interface TableCellCsvGeneratorBase<TCellData, TRowData = unknown> {
+  readonly type: CSV_GENERATOR_TYPE;
+  readonly cellType: string;
+
+  generateSafeCsv(cellData?: TCellData, rowData?: TRowData): string | Dictionary<string> | undefined;
```
aaron-steinfeld, line 10: When would we produce a dictionary?

projects/components/src/table/cells/table-cell-injection.ts
```diff
@@ -1,12 +1,16 @@
...
 import { TableColumnConfig, TableRow } from '../table-api';
 import { TableCellParserBase } from './table-cell-parser-base';
+import { CoreTableCellCsvGeneratorType } from './types/core-table-cell-csv-generator-type';
 
 export const TABLE_COLUMN_CONFIG: InjectionToken<TableColumnConfig> = new InjectionToken('TABLE_COLUMN_CONFIG');
 export const TABLE_COLUMN_INDEX: InjectionToken<TableColumnConfig> = new InjectionToken('TABLE_COLUMN_INDEX');
 export const TABLE_DATA_PARSER: InjectionToken<TableCellParserBase<unknown, unknown, unknown>> = new InjectionToken(
   'TABLE_DATA_PARSER'
 );
+export const TABLE_CELL_CSV_GENERATOR: InjectionToken<
```
aaron-steinfeld, line 11: are the changes in table cell renderers still needed?

projects/components/src/table/cells/table-cell-csv-generator-management.service.ts
```diff
@@ -0,0 +1,50 @@
...
+})
+export class TableCellCsvGeneratorManagementService {
+  private readonly csvGenerators: TableCellCsvGeneratorBase<unknown>[] = [];
+
+  public register(csvGenerators: TableCellCsvGeneratorBase<unknown> | TableCellCsvGeneratorBase<unknown>[]): void {
+    isArray(csvGenerators) ? csvGenerators.forEach(item => this.addGenerator(item)) : this.addGenerator(csvGenerators);
+  }
+
+  private addGenerator(csvGenerator: TableCellCsvGeneratorBase<unknown>): void {
+    if (!includes(this.csvGenerators, csvGenerator)) {
```
aaron-steinfeld, line 17: nit - use your findMatchingGenerator function so we don't register two with the same type

projects/components/src/table/cells/table-cell-csv-generator-management.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  public findMatchingGenerator(type?: string): TableCellCsvGeneratorBase<unknown> | undefined {
+    return this.csvGenerators.find(generator => generator.cellType === type);
+  }
+}
+
+/*
+ * TableCellRendererConstructor is used by lookup service to dynamically instantiate cell CsvGenerators
+ */
+export interface TableCellCsvGeneratorConstructor<TCellData, TRowValue>
```
aaron-steinfeld, line 30: do we still use decorators/constructors?

projects/components/src/table/table.component.ts
```diff
@@ -512,8 +515,15 @@ export class TableComponent
     private readonly domElementMeasurerService: DomElementMeasurerService,
     private readonly tableService: TableService,
     private readonly modalService: ModalService,
+    private readonly fileDownloadService: FileDownloadService,
+    private readonly tableCsvDownloaderService: TableCsvDownloaderService,
+    private readonly notificationService: NotificationService,
     private readonly preferenceService: PreferenceService
   ) {
+    this.tableCsvDownloaderService.csvDownloadRequest$.pipe(filter(tableId => tableId === this.id)).subscribe(() => {
```
aaron-steinfeld, line 523: subscribing to a long lived observable here, need to clean it up.

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
arjunlalb: We are getting rid of csv generator type and directly linking them to cell renderer types.

projects/components/src/table/cells/table-cell-csv-generator-base.ts
arjunlalb: if the column is an object like `{ key1: "value", key2: "value" }`, returning a dictionary allows us to create separate entries in the CSV for these.

projects/components/src/table/cells/table-cell-csv-generator-management.service.ts
arjunlalb: updated

projects/components/src/table/cells/table-cell-csv-generator-management.service.ts
arjunlalb: nope. removed.

projects/components/src/table/cells/table-cell-injection.ts
arjunlalb: not required. removed.

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
arjunlalb: we should make sure the generate method accounts for null/undefined. updated all existing generators to account fo rthis.

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
arjunlalb: so that the injection token can correctly pick up all the generators. Without an identifying property like this, it does not work.

projects/components/src/table/table.component.ts
arjunlalb: fixed

projects/components/src/table/cells/csv-generators/table-cell-boolean-csv-generator.ts
arjunlalb: actually, this may not be needed anymore. let me check

arjunlalb: removed

projects/components/src/table/table.component.ts
```diff
@@ -648,6 +658,67 @@ export class TableComponent
...
+            ? this.data.getData({
+                columns: columnConfigs,
+                position: { limit: 1000, startIndex: 0 },
+                filters: filters ?? []
+              })
+            : of({ data: [], totalCount: 0 })
+          ).pipe(
+            take(1),
+            map((response: TableDataResponse<TableRow>) => response.data),
+            map(rows => {
```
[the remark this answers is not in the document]
arjunlalb: moved out.

projects/components/src/table/cells/table-cell-csv-generator-base.ts
aaron-steinfeld: interesting... but what happens if one row to the next doesn't have consistent keys?

arjunlalb: usually that does not happen. all objects corresponding to a column have the same structure. in the off chance there are optional keys, the generator should account for it.

aaron-steinfeld: I don't think we have this guarantee, and we can't have inconsistent numbers of values being output for rows. can come back to this one separately, but avoid using it until we resolve this.

aaron-steinfeld (approved):

projects/components/src/table/table-csv-downloader.service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  public constructor(
+    private readonly fileDownloadService: FileDownloadService,
+    private readonly notificationService: NotificationService
+  ) {}
+  public triggerDownload(tableId: string): void {
+    this.csvDownloadRequestSubject.next(tableId);
+  }
+
+  public executeDownload(
```
anandtiwary, line 24: @arjunlalb We show a confirmation modal before downloading csv. In a follow-up pr, do you think we should add that logic inside this method?

anandtiwary (approved):

projects/components/src/table/table-csv-downloader.service.ts
arjunlalb: @anandtiwary we could.

PR #2421, opened 2023-09-22
fix: handle null values for filter button

itssharmasandeep:


adisreyaj (approved):

projects/components/src/filtering/filter-button-wrapper/filter-button-wrapper.component.ts
```diff
@@ -10,9 +10,9 @@ export const FILTER_BUTTON_WRAPPER = new InjectionToken<FilterButtonWrapperData>
   template: `
     <ht-filter-button
       *ngIf="this.filterButtonWrapperData"
-      [attribute]="this.filterButtonWrapperData.targetAttribute.attribute"
+      [attribute]="this.filterButtonWrapperData.targetAttribute?.attribute"
```
aaron-steinfeld, line 13: Please revert and fix the caller - the types here declare it can't be null, so if someone is passing a null fix that rather than ignoring the types. Or if we think it's not fixable, then fix the types. Either way, please treat the types as the source of truth and fix the violation, not the consequences.

palbizu: Yes,  the validation should be done directly on cartesian-legend.ts 
```
if(!isNil(legendEntry.node()) && !isNil(this.activeSeries[0].groupBy)){
      this.drawFilterControl(legendEntry.node(), this.activeSeries[0].groupBy).location.nativeElement.classList.add('filter');
    }
```

aaron-steinfeld: I dug into this a little bit and the high level issue precedes the recent change. Fundamentally, this filter button is an explorer feature that was pushed into chart without appropriate abstraction.

Currently, we always draw this filter button, so Pato's suggestion above is one part of an improvement. Other parts include
- Use typing when creating the `FILTER_BUTTON_WRAPPER` injector. It's untyped, which is why the compiler didn't catch the missing presence check.
- Changing the naming. I had asked for it to be renamed from groupByFilter to groupBy without enough context. After digging into the code a bit here, group by is the specific explorer use case, not the generic feature being built in chart. This is just a filter, group by is irrelevant.
- More fundamentally, when the button is clicked it modifies the application url to add a filter. That's the filter parameter associated with the explorer page. Many pages use charts without filters, or could choose a different query parameter name for their filters. Rather than making this a "smart component", this seems like it would best be set up as a callback (or observer) on series that can be notified when a legend value is clicked. Then, it can make context-specific decisions on what to do with that value. To further abstract it, let that series config describe things like what icon to use etc. Basically, a generic control in the legend rather than a very use-case specific filter button. You could even go further and accept a component, template etc.
```

PR #2423, opened 2023-09-25
feat(radio): add tooltip support

adisreyaj:
Add tooltip support for radio group.


projects/components/src/time-picker/time-picker.component.scss
```diff
@@ -8,6 +8,7 @@
     @include body-1-medium($gray-7);
     border-radius: 4px;
     height: 36px;
+    width: 100%;
```
adisreyaj, line 11: Let time picker take available space of parent.

projects/components/src/radio/radio-group.component.ts
```diff
@@ -40,6 +40,7 @@ import { RadioOption } from './radio-option';
         [ngClass]="[this.optionsDirection, this.disabled || option.disabled ? 'disabled' : '']"
         [value]="option.value"
         [disabled]="option.disabled"
+        [htTooltip]="option.tooltip"
```
palbizu, line 43: is this an input parameter or a directive? if is an input name should be tooltip instead of htTooltip.

adisreyaj: It's the ht tooltip directive.

palbizu (approved):

anandtiwary (approved):

PR #2438, opened 2023-09-27
feat: adding attribute columns to traces

palbizu:
## Description
Adding Attribute's columns to Traces

### Testing
Locally Tested

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
+++ b/projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
@@ -112,7 +112,7 @@ export class ExplorerDashboardBuilder {
               resultsQuery.properties
             )
           ]),
-          map(columnsMetadata => this.removeColumnsDuplicated(columnsMetadata)),
+          map(columnsMetadata => this.removeDuplicatedColumns(columnsMetadata)),
           map(columnsMetadata => this.buildColumnModelJson(request.context, columnsMetadata)),
           map(json => ({
             json: json,
@@ -180,7 +180,7 @@ export class ExplorerDashboardBuilder {
     }
   }
 
-  protected getDefaultTableColumns(context: ExplorerGeneratedDashboardContext): ModelJson[] {
+  protected getDefaultTableColumns(context: ExplorerGeneratedDashboardContext): ModelSpecificationJson[] {
     switch (context) {
       case ObservabilityTraceType.Api:
         return [
@@ -570,7 +570,7 @@ export class ExplorerDashboardBuilder {
     }
   }
 
-  private removeColumnsDuplicated(columns: ModelSpecificationJson[]): ModelJson[] {
+  private removeDuplicatedColumns(columns: ModelSpecificationJson[]): ModelJson[] {
     return uniqBy(columns, column =>
       column.value?.type === ValueSpecificationType.Attribute ? column.value.attribute : column
     );
@@ -581,21 +581,25 @@ export class ExplorerDashboardBuilder {
     context: ExplorerGeneratedDashboardContext,
     selectedProperties: Specification[]
   ): ModelSpecificationJson[] {
-    return attributes.map(attribute => ({
-      type: 'table-widget-column',
-      title: attribute.displayName,
-      width: 1,
-      display: this.getRendererForType(attribute.type),
-      filterable: this.filterBuilderLookupService.isBuildableType(toFilterAttributeType(attribute.type)),
-      value: {
-        type: ValueSpecificationType.Attribute,
-        attribute: attribute.name
-      },
-      visible: selectedProperties.find(selectedProperty => selectedProperty.name === attribute.name) ? true : false,
-      'click-handler': {
-        type: context === SPAN_SCOPE ? 'span-trace-navigation-handler' : 'api-trace-navigation-handler'
-      }
-    }));
+    const attributesToExclude = this.getAttributesToExcludeFromUserDisplay(context);
+
+    return attributes
+      .filter(attribute => !attributesToExclude.has(attribute.name))
+      .map(attribute => ({
+        type: 'table-widget-column',
+        title: attribute.displayName,
+        width: 1,
+        display: this.getRendererForType(attribute.type),
+        filterable: this.filterBuilderLookupService.isBuildableType(toFilterAttributeType(attribute.type)),
+        value: {
+          type: ValueSpecificationType.Attribute,
+          attribute: attribute.name
+        },
+        visible: selectedProperties.find(selectedProperty => selectedProperty.name === attribute.name) ? true : false,
+        'click-handler': {
+          type: context === SPAN_SCOPE ? 'span-trace-navigation-handler' : 'api-trace-navigation-handler'
+        }
+      }));
   }
 }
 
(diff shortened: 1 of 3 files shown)


projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -565,9 +570,14 @@ export class ExplorerDashboardBuilder {
     }
   }
 
-  private getUserRequestedNonDefaultColumns(
+  private removeColumnsDuplicated(columns: ModelJson[]): ModelJson[] {
+    return uniqBy(columns, 'title'); // Todo: this should be improved
```
aaron-steinfeld, line 574: can we do this by attribute rather than title?

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -583,6 +593,7 @@ export class ExplorerDashboardBuilder {
           type: 'attribute-specification',
           attribute: attribute.name
         },
+        visible: selectedProperties.find(selectedProperty => selectedProperty.name === attribute.name) ? true : false,
```
aaron-steinfeld, line 596: visible defaults to true, how was this working in the past?

If a user selects to display something, does that get preserved?

palbizu: Yes, I could ignore the field in the case that 'true' because is true by default. Answering the second question, they are shown after selecting the each metric.

aaron-steinfeld: > Yes, I could ignore the field in the case that 'true' because is true by default. 

Ah - in the past this code was only being used for the added column which we wanted to display. The default ones were the ones that had hidden values. Nothing to do here - explicit as you have here is better, I was just confused on the mapping since the purpose of this code changed even though the diff shows it existed.

> Answering the second question, they are shown after selecting each metric.

Still a bit confused here - every time the query changes, wouldn't we be rebuilding the columns and tossing the user's selections (the columns they've checked off in the edit cols modal)? Or are they being preserved somewhere?

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: updated.

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -565,9 +570,14 @@ export class ExplorerDashboardBuilder {
     }
   }
 
-  private getUserRequestedNonDefaultColumns(
+  private removeColumnsDuplicated(columns: ModelJson[]): ModelJson[] {
+    return uniqBy(columns, column => (column?.value as AttributeValue).attribute ?? '');
```
aaron-steinfeld, line 574: Rather than casting here, change the declared return types on your two methods (getGeneratedTableColumns, getDefaultTableColumns) to more descriptively describe their results. That's the whole benefit of using types - if your data doesn't conform to that, then this isn't a safe transform to do.

palbizu: I improved it a bit, not possible to change return type of getGeneratedTableColumns because is protected

aaron-steinfeld: Why does it being protected stop us? It just means any child classes would need to be similarly updated. Not ideal, but if it's the best solution that's never stopped us before.

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -570,34 +570,32 @@ export class ExplorerDashboardBuilder {
     }
   }
 
-  private removeColumnsDuplicated(columns: ModelJson[]): ModelJson[] {
-    return uniqBy(columns, column => (column?.value as AttributeValue).attribute ?? '');
+  private removeColumnsDuplicated(columns: ModelSpecificationJson[]): ModelJson[] {
```
aaron-steinfeld, line 573:
```suggestion
  private removeDuplicatedColumns(columns: ModelSpecificationJson[]): ModelJson[] {
```

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -570,34 +570,32 @@ export class ExplorerDashboardBuilder {
...
+      column.value?.type === ValueSpecificationType.Attribute ? column.value.attribute : column
+    );
   }
 
   private getGeneratedTableColumns(
     attributes: AttributeMetadata[],
     context: ExplorerGeneratedDashboardContext,
     selectedProperties: Specification[]
-  ): ModelJson[] {
-    const attributesToExclude = this.getAttributesToExcludeFromUserDisplay(context);
```
aaron-steinfeld, line 572: We need to be careful about removing this. Functionally uniq check should take its place - but please go through the attributes and make sure there's nothing in there that wouldn't be covered by uniq. I believe this is also protected so subclasses can override, so we need to consider that as well and check any subclasses.

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -606,9 +604,25 @@ export interface ExplorerGeneratedDashboard {
   onReady(dashboard: Dashboard): void;
 }
 
-interface AttributeValue {
-  type: string;
-  attribute: string;
+interface ModelSpecificationJson extends ModelJson {
```
aaron-steinfeld, line 607: looks good here. Is value optional though? do we have columns with no value?

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: renamed

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: needs to be mandatory

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: Calls and Status Code are present after changing this and they should not be there. So I will keep the code for now.

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: Fixed

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: updated

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -604,8 +608,8 @@ export interface ExplorerGeneratedDashboard {
   onReady(dashboard: Dashboard): void;
 }
 
-interface ModelSpecificationJson extends ModelJson {
-  value?:
+export interface ModelSpecificationJson extends ModelJson {
```
aaron-steinfeld, line 611: If you're exporting this, need a clearer name (e.g. SpecificationBackedColumnModelJson - I believe that's we call this form of column)

aaron-steinfeld (commented):
otherwise lgtm

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: reanmed

PR #2440, opened 2023-09-29
Remove page time range

anandtiwary:
## Description
Remove page time range

### Testing

### Checklist:

### Documentation


src/app/shared/navigation/navigation.component.ts
```diff
@@ -85,37 +85,29 @@ export class NavigationComponent {
...
-    const navItems = this.navItemDefinitions.map(definition =>
+    this.navItems = this.navItemDefinitions.map(definition =>
       this.navigationListService.decorateNavItem(definition, this.activatedRoute)
     );
-    // Decorate the nav items with the corresponding time ranges, depending on the FF state.
-    // The time ranges in nav items are streams that get the most recent value from page time range preference service
-    this.navItems$ = this.navListComponentService.resolveNavItemConfigTimeRanges(navItems);
 
     this.isCollapsed$ = this.preferenceService.get(NavigationComponent.COLLAPSED_PREFERENCE, false);
+    this.updateDefaultTimeRangeIfUnset();
```
aaron-steinfeld, line 96: Are we sure this isn't too early? Looks like before it was only being initialized once the active item was set  which was due to the requirement to check what the item was. We don't care what the item is any more, but we do need to make sure this code isn't called before nav service's regular init has completed and applied any deep linked time range.

anandtiwary: @aaron-steinfeld I think we don't need to call this method from here. The TR is already getting initialized. I should probably move the default time range code to the TR service. It is possible that we are exposing a public method `updateDefaultTimeRangeIfUnset()` so that we can set different default tr from outside. 

```
  public constructor(
    private readonly navigationService: NavigationService,
    private readonly timeDurationService: TimeDurationService
  ) {
    this.initializeTimeRange();
    this.navigationService.registerGlobalQueryParamKey(TimeRangeService.TIME_RANGE_QUERY_PARAM);
  }
 ```

Comment (anandtiwary) [reply src/app/shared/navigation/navigation.component.ts:96]: Please also check the TR service changes in the latest commit.

File: projects/components/src/header/page/page-header.component.ts [hunk 2]
```diff
@@ -24,19 +23,15 @@ import { NavigableTab } from '../../tabs/navigable/navigable-tab';
       class="page-header"
       [class.bottom-border]="!this.tabs?.length"
     >
-      <!--      Code Path 1-->
-      <div
-        *htIfFeature="'${ApplicationFeature.PageTimeRange}' | htFeature; else noTimeRangeHeaderLayoutTemplate"
-        class="column-alignment"
-      >
+      <div class="column-alignment">
```
arjunlalb, line 26: nit: this div seems unnecessary now?

src/app/shared/navigation/navigation.component.ts
arjunlalb: Do we even need this method and call anymore since there is only one way to set a timerange? The navigation component should not care about what time range is set?

aaron-steinfeld: +1 - please move the default time range assignment into TR service. Otherwise the changes to TR service look good.

projects/common/src/time/time-range.service.ts
```diff
@@ -88,46 +87,10 @@ export class TimeRangeService {
...
-      })
-    );
-  }
-
   private initializeTimeRange(): void {
     this.getInitialTimeRange().subscribe(timeRangeFromInitialUrl => {
       this.applyTimeRangeChange(timeRangeFromInitialUrl);
     });
-
-    this.getPageTimeRangeChanges().subscribe(timeRange => {
```
anandtiwary, line 110: @aaron-steinfeld @arjunlalb I think we need to keep the logic in `getPageTimeRangeChanges()` so that the app can respond to time range changes through URL. We are currently setting new time range from app via this method

```
  private setTimeRangeInUrl(timeRange: TimeRange): this {
    this.navigationService.addQueryParametersToUrl({
      [TimeRangeService.TIME_RANGE_QUERY_PARAM]: timeRange.toUrlString()
    });
    return this;
  }
```

anandtiwary: Or we modify `this.getInitialTimeRange()` to respond to later tr changes. I am trying to understand why we are using `activeRoute.snapshot` in `getPageTimeRangeChanges`

aaron-steinfeld: Yeah, probably just poorly named. Can be simplified though now. Can probably just look back in history and revert

anandtiwary: Good point. No need to reinvent the wheel :)

anandtiwary: Reverted. `Refresh` query param is also gone.

aaron-steinfeld (dismissed):

projects/common/src/time/time-range.service.ts
```diff
@@ -70,7 +70,18 @@ export class TimeRangeService {
         map(activeRoute => activeRoute.snapshot.queryParamMap),
         distinctUntilChanged((prevParamMap, currParamMap) => this.shouldNotUpdateTimeRange(prevParamMap, currParamMap)),
         map(paramMap => paramMap.get(TimeRangeService.TIME_RANGE_QUERY_PARAM)), // Extract the time range value from it
-        filter((timeRangeString): timeRangeString is string => !isEmpty(timeRangeString)), // Only valid time ranges
+        filter((timeRangeString): timeRangeString is string => {
+          if (isEmpty(timeRangeString)) {
+            /**
+             * This will add the a time param to the url and re-trigger navigation$
```
aaron-steinfeld, line 76: If this wasn't needed before, why is it now? This seems like it would risk refreshing the page

anandtiwary: I think it was broken before. When you hit the URL with no TR, the default TR is set, but the URL is not updated. This breaks the nav history. 

The logic around page time range fixed this issue because we set a TR explicitly.

anandtiwary: I tested the changes, and they seem to work well. Not leading to duplicate emits.

aaron-steinfeld (approved):

PR #2450, opened 2023-10-05
fix: handle null values in time ago cell renderer

itssharmasandeep:


projects/components/src/table/cells/data-renderers/time-ago/time-ago-table-cell-renderer.component.ts
```diff
@@ -20,17 +20,22 @@ import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
   styleUrls: ['./time-ago-table-cell-renderer.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div class="cell-value" [htTooltip]="this.value | htDisplayDate: this.dateFormat">
+    <div
+      *ngIf="!(this.value | htIsNil); else nilValueTemplate"
```
adisreyaj, line 24: I see that the `htDisplayTimeAgo` already handles `null` value:

```ts
public transform(value?: DateOrMillis | null, suffix: string = 'ago', dateCoercer?: DateCoercer): string {
    if (value === null || value === undefined || value === 0) {
      return '-';
    }

    const durationAgo = new TimeDuration(this.calcSecondsAgo(value, dateCoercer ?? this.dateCoercer), TimeUnit.Second);
    if (durationAgo.getAmountForUnit(TimeUnit.Minute) < 1) {
      return 'Just now';
    }

    return `${durationAgo.getMostSignificantUnitOnly().toLongString()} ${suffix}`;
  }
```

itssharmasandeep: Ahh okay, then it was only erroring for parser, removed the null handling from here

adisreyaj (approved):

PR #2451, opened 2023-10-05
feat: Update Filters appending

palbizu:
## Description

### Testing

### Checklist:

### Documentation


projects/observability/src/shared/components/span-detail/call/headers/span-detail-call-headers.component.ts
```diff
@@ -15,15 +13,23 @@ import { ScopeQueryParam } from '../../../../../pages/explorer/explorer.componen
...
             <div class="record-value" *htListViewValueRenderer="let record">
               <div class="value">{{ record.value }}</div>
-              <ht-explore-filter-link
-                class="filter-link"
-                [paramsOrUrl]="this.buildExploreNavigationParams | htMemoize: record | async"
+              <ht-filter-button
+                *ngIf="this.attribute && this.metadata"
+                class="filter-button"
+                [attribute]="this.attribute"
+                [metadata]="this.metadata"
```
aaron-steinfeld, line 28: Consider whether we really need to have every caller of this pass metadata in. Should it (or the filter url service) just query it internally. Seems cleaner.

palbizu: done

projects/observability/src/shared/components/span-detail/tags/span-tags-detail.component.ts
```diff
@@ -16,12 +14,15 @@ import { ScopeQueryParam } from '../../../../pages/explorer/explorer.component';
...
           <div class="tag-value" *htListViewValueRenderer="let record">
             <div class="value">{{ record.value }}</div>
-            <ht-explore-filter-link
-              class="filter-link"
-              [paramsOrUrl]="this.getExploreNavigationParams | htMemoize: record | async"
+            <ht-filter-button
+              *htLoadAsync="this.metadata$ as metadata"
+              class="filter-button"
+              [attribute]="this.getFilterAttribute | htMemoize: metadata"
+              [metadata]="metadata"
```
aaron-steinfeld, line 21: Could we keep pushing this down? Can the filter button fetch its own metadata and just accept a scope. Probably just have the filter url service accept a scope and let it fetch metadata.

palbizu: Finalize, MetadataService was used in both places

palbizu: No, not possible because they are placed in different routes filterButton in components and MetadataService under observability and move FilterButton means to move more components related. A way to fix this would be to add the method to MetadataService returning FilterAttribute[] or create a new service just with getAllAttributes method.

anandtiwary: `getFilterAttributes` we already have this method. Can we use it?

anandtiwary (approved):

PR #2470, opened 2023-10-17
Replacing group by with interval as none chart

palbizu:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
+++ b/projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
@@ -107,7 +107,7 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
       .withSeries(...this.series)
       .withTooltip(
         this.chartTooltipBuilderService.constructTooltip<TData, Series<TData>>(data =>
-          this.convertToDefaultTooltipRenderData(Array.isArray(data?.[0].dataPoint) ? [data[3]] : data)
+          this.convertToDefaultTooltipRenderData(data)
         )
       )
       .withEventListener(ChartEvent.Hover, data => {
@@ -210,7 +210,7 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
     }
 
     return {
-      title: this.resolveTooltipTitle(data[0]),
+      title: Array.isArray(data?.[0].dataPoint) ? '' : this.resolveTooltipTitle(data[0]),
       labeledValues: data.map(singleValue => ({
         label: singleValue.context.name,
         value: defaultYDataAccessor<number | string>(singleValue.dataPoint),
--- a/projects/observability/src/shared/components/utils/chart-tooltip/default/default-chart-tooltip.component.html
+++ b/projects/observability/src/shared/components/utils/chart-tooltip/default/default-chart-tooltip.component.html
@@ -24,7 +24,7 @@
       </td>
     </tr>
   </table>
-  <div class="tooltip-timestamp">
+  <div class="tooltip-timestamp" *ngIf="data.title">
     {{ data.title }}
   </div>
 </div>
--- a/projects/observability/src/shared/components/utils/chart-tooltip/default/default-chart-tooltip.component.ts
+++ b/projects/observability/src/shared/components/utils/chart-tooltip/default/default-chart-tooltip.component.ts
@@ -16,6 +16,6 @@ export class DefaultChartTooltipComponent {
 }
 
 export interface DefaultChartTooltipRenderData {
-  title: string;
+  title?: string;
   labeledValues: { label: string; value: number | string; color: string; units?: string }[];
 }


projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -102,13 +102,12 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
...
       return;
     }
-
     this.chart = this.chartBuilderService
       .build<TData>(this.strategy, this.container.nativeElement, this.renderer, this.groupId)
       .withSeries(...this.series)
       .withTooltip(
         this.chartTooltipBuilderService.constructTooltip<TData, Series<TData>>(data =>
-          this.convertToDefaultTooltipRenderData(data)
+          this.convertToDefaultTooltipRenderData(Array.isArray(data?.[0].dataPoint) ? [data[3]] : data)
```
aaron-steinfeld, line 110: We'll need to get these assumptions cleaned up. Having a magic format here is no good

projects/observability/src/shared/components/cartesian/d3/data/series/cartesian-column.ts
```diff
@@ -86,20 +86,28 @@ export class CartesianColumn<TData> extends CartesianSeries<TData> {
...
+      : this.xScale.getGroupedColumnSeriesLength(this.series);
     const groupedColumnAdjustment = (this.xScale.getBandwidth() - columnWidth * totalGroupedSeries) / 2;
 
     return totalGroupedSeries === 1
       ? groupedColumnAdjustment
       : groupedColumnAdjustment + this.xScale.getGroupedColumnSeriesPosition(this.series) * columnWidth;
   }
 
+  private hasArrayData(series: Series<TData>): boolean {
+    return Array.isArray(series.data?.[0]);
```
aaron-steinfeld, line 108: What's the expected data format here?

palbizu: Checks data structure to distinguish the correct column size

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
palbizu: moved to tooltip

projects/observability/src/shared/components/cartesian/d3/data/series/cartesian-column.ts
aaron-steinfeld: Fundamentally though, this whole implementation embeds into chart the assumption that we should change behavior arbitrarily if series data contains an array. This seems to me like a magic assumption. The use case - for a simple bar chart - should be handled more elegantly. @anandtiwary what do you think?

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
```diff
@@ -211,7 +210,7 @@ export class CartesianChartComponent<TData> implements OnChanges, OnDestroy {
     }
 
     return {
-      title: this.resolveTooltipTitle(data[0]),
+      title: Array.isArray(data?.[0].dataPoint) ? '' : this.resolveTooltipTitle(data[0]),
```
anandtiwary, line 213: Why do we need this conditional?

projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis-crosshair.ts
```diff
@@ -28,7 +28,8 @@ export class CartesianAxisCrosshair {
     this.updatePosition(
       crosshairSelection,
       position,
-      locationData.map(data => data.location)
+      locationData.map(data => data.location),
+      Array.isArray(locationData?.[0].dataPoint)
```
anandtiwary, line 32: We should isolate this logic in a separate method. No need to repeat it everywhere. 

Imo, this should be controlled via an input property and set from outside the generic cartesian chart. But

projects/observability/src/shared/components/cartesian/d3/data/series/cartesian-column.ts
anandtiwary: Yeah, I agree. @palbizu let's sync up on this

projects/observability/src/shared/components/cartesian/cartesian-chart.component.ts
palbizu: just used to remove duplicated label

projects/observability/src/shared/components/cartesian/d3/axis/cartesian-axis-crosshair.ts
palbizu: Problem is that I'm checking the data here. I should follow a better logic

projects/observability/src/shared/components/cartesian/d3/data/series/cartesian-column.ts
anandtiwary: @aaron-steinfeld We discussed this, and we are trying the following approach:
1. Add an optional `color` property to each data point. If it is present, then the cartesian chart will override the series color. The responsibility of setting the right color of data points lies with the parent component/service. 
2. We already support `Series Name: x-value` type display on Legends when` Interval and group by are present`. We will extend the same behavior to when `interval is absent but group by is present`. Datapoint color will override the series color.

Please let us know what you think of this approach.

aaron-steinfeld: > Add an optional color property to each data point. If it is present, then the cartesian chart will use it else. It will fallback to the series color. The responsibility of setting the right color of data points lies with the parent component/service.

I think that makes sense at a high level if I'm following - I think I had asked for having color at the data point level instead of series too.

> We already support the Series Name: x-value type display on Legends when Interval and group by are present. We will extend the same behavior to when interval is absent but group by is present. Datapoint color will override the series color.

Here, we have to be a bit careful. This doesn't seem quite right. The existing cartesian shouldn't (doesn't, I hope?) care about group by directly. If I recall, it groups legends based on some series key that shows which series are associated. That should already be independent of time at the chart level, however explorer may be changing how its building series such that we only group these on a time series.

We want to make sure that a legend with a single series renders in the optimal way for that series - I don't think that would be grouping since the grouping of legend values is to avoid repetition in UX only. This is really just the simple cartesian bar chart case.

> Datapoint color will override the series color.

What we're really doing here is deriving the legend entries from the data rather than the series. If I provide data like
```ts
[ 
  {
    value: 10,
    color: "red",
    label: "Red Data"
  },
  {
    value: 15,
    color: "blue",
    label: "Blue Data"
  }
]
```
then there's only one series defined (which doesn't directly appear in the legend), but there should be two legend entries like 
``` 
🔴  Red Data
🔵  Blue Data
```

palbizu: Not really clear about following your comment. Currently, for this type of data we are generating/receiving an object <name,value>[] in the same series. For this reason the representation uses just one color. What you suggest is to modify the data that we are sending to the chart? instead of updating the color representation?

aaron-steinfeld: I'm suggesting updating the legend so that it generates an entry for each value in the series rather than one entry per series (which is why earlier attempts had split into multiple series, I assume). Whether legend itself generates the color or allows the value in the series to specify it I think isn't critical. Ideally it uses a provided value and only falls back to generated if not provided, but as long as it can work with a provided one I think we're fine for now.

PR #2471, opened 2023-10-17
fix: support to get states of multiple feature flags

Greedy-Geek:


projects/common/src/feature/state/feature-state.resolver.ts
```diff
@@ -13,6 +13,16 @@ export abstract class FeatureStateResolver {
     );
   }
 
+  public getFeatureStates(features: string[]): Observable<Map<string, FeatureState>> {
```
aaron-steinfeld, line 16: how does this differ from the existing function right below this getCombinedFeatureState ? If you want multiple values and are working with them independently, make multiple calls. If you're resolving them together, use the existing merger

Greedy-Geek: we want multiple values, i.e., separate states of each FF, but want to use them in the same place like
```
FF1 === Enabled ? returnSomething1() : returnNothing1();
FF2 === Enabled ? returnSomething2() : returnNothing2();
```

aaron-steinfeld: Why not get each individually then?

Greedy-Geek: So I thought having these methods in each service, let's common them out in this service

aaron-steinfeld: If we really want this, it's not a huge deal - just not sure we're providing a lot of value. the difference between (pseudo)
`combineLatest(resolver.getState(first), resolver.getState(second))` and `resolver.getStates(first, second)` just doesn't seem worth adding a dedicated API for. In either case, you're likely going to be writing extraction code to reshape it right after.

PR #2473, opened 2023-10-19
feat: improvements to table csv download logic

arjunlalb:
## Description
Improvements to table CSV download flow.

1. Add optional input to table for CSV download configs.
2. Use limit from the above for download data query. If limit is unavailable, default to 1000.
3. Use the filename from config. If unavailable, use fallback logic using table ID.
4. For column keys, preference order is column title -> column name -> column id
5. Update timestamp csv generator.

### DIFF
--- a/projects/components/src/table/table.component.ts
+++ b/projects/components/src/table/table.component.ts
@@ -1287,6 +1287,6 @@ interface TableLocalPreferences {
 type PersistedTableColumnConfig = Pick<TableColumnConfig, 'id' | 'visible'>;
 
 export interface TableCsvDownloadConfig {
-  limit: number;
-  filename: CsvFileName;
+  limit?: number;
+  filename?: CsvFileName;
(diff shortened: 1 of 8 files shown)


projects/components/src/table/table.component.ts
```diff
@@ -676,7 +680,10 @@ export class TableComponent
       )
     );
 
-    this.tableCsvDownloaderService.executeDownload(downloadConfig$, this.id);
+    this.tableCsvDownloaderService.executeDownload(
```
itssharmasandeep, line 683: Should we add the confirmation logic as well?

itssharmasandeep (approved):

projects/components/src/table/table.component.ts
arjunlalb: we should. will add in a separate follow up.

PR #2474, opened 2023-10-19
feat: Adding Simplified Filters (null)

palbizu:
## Description
Adding a null filter by default

### Testing
Locally Tested

Screenshot 2023-10-19 at 18 33 13

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/filtering/filter-button/filter-button.component.ts
+++ b/projects/components/src/filtering/filter-button/filter-button.component.ts
@@ -55,7 +55,7 @@ export class FilterButtonComponent implements OnChanges {
   @Output()
   public readonly popoverOpen: EventEmitter<boolean> = new EventEmitter();
 
-  private static readonly DEFAULT_VALUE: string = 'null';
+  private static readonly NULL_VALUE: string = 'null';
   public availableFilters: Filter[] = [];
 
   public constructor(
@@ -67,7 +67,7 @@ export class FilterButtonComponent implements OnChanges {
     if (isNil(this.attribute)) {
       this.availableFilters = [];
     } else if (isNil(this.value)) {
-      this.availableFilters = this.buildSimplifiedFilters(this.attribute);
+      this.availableFilters = this.buildFiltersForNullValue(this.attribute);
     } else {
       this.availableFilters = this.buildAvailableFilters(this.attribute, this.value, this.subpath);
     }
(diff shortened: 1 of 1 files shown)


projects/components/src/filtering/filter-button/filter-button.component.ts
```diff
@@ -62,10 +64,13 @@ export class FilterButtonComponent implements OnChanges {
...
 
   public ngOnChanges(): void {
-    this.availableFilters =
-      !isNil(this.attribute) && !isNil(this.value)
-        ? this.buildAvailableFilters(this.attribute, this.value, this.subpath)
-        : [];
+    if (isNil(this.attribute)) {
+      this.availableFilters = [];
+    } else if (isNil(this.value)) {
+      this.availableFilters = this.buildSimplifiedFilters(this.attribute);
```
anandtiwary, line 70: if value is undefined or null, then we want to show only eq and not_eq filters?

palbizu: In this case, is receiving undefined as a value. I think that is the same if we receive null or undefined. Instead of showing the blank popover with this change, we show equals/not equals to null.

palbizu: @aaron-steinfeld requested this change. Do you think that it's fine. I'm not finding a case where this logic doesn't work well.

palbizu: Two filters added after to discuss with Aaron offline

aaron-steinfeld: Basically equality filters (=, !=, in, not_in) always make sense for any value, null included. Type specific filters (numeric comparisons, regex, contains etc.) should not be shown for nulls.

projects/components/src/filtering/filter-button/filter-button.component.ts
```diff
@@ -77,4 +82,21 @@ export class FilterButtonComponent implements OnChanges {
       .lookup(attribute.type)
       .buildFiltersForSupportedOperators(attribute, value, subpath);
   }
+
+  private buildSimplifiedFilters(attribute: FilterAttribute, value?: FilterValue): Filter[] {
```
aaron-steinfeld, line 86: This function can be renamed and remove the value arg. Renamed because it's "simplified" in that it's hardcoded to null (so maybe buildFiltersForNullValue), and remove the arg because we've already checked it outside before calling it. The behavior here is dependent on it being null, so just use the value literal (which will also help readability). If value were undefined for example, we still want to use null in the filter, not undefined (since server only works in null, not undefined).

palbizu: Done, I kept the read-only const but updated the rest

aaron-steinfeld (approved):

PR #2476, opened 2023-10-20
feat: csv generator for string enum type

arjunlalb:
## Description
csv generator for string enum type

### DIFF
--- a/projects/components/src/table/cells/csv-generators/table-cell-string-enum-csv-generator.ts
+++ b/projects/components/src/table/cells/csv-generators/table-cell-string-enum-csv-generator.ts
@@ -1,0 +1,16 @@
+import { TableCellCsvGenerator } from '../table-cell-csv-generator';
+import { Injectable } from '@angular/core';
+import { CoreTableCellRendererType } from '../types/core-table-cell-renderer-type';
+import { DisplayStringEnumPipe } from '@hypertrace/common';
+import { isNil } from 'lodash-es';
+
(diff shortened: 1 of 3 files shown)


projects/components/src/table/cells/csv-generators/table-cell-string-enum-csv-generator.ts
```diff
@@ -0,0 +1,16 @@
...
+import { isNil } from 'lodash-es';
+
+@Injectable({ providedIn: 'root' })
+export class TableCellStringEnumCsvGenerator implements TableCellCsvGenerator<string> {
+  public cellType: string = CoreTableCellRendererType.StringEnum;
+
+  private readonly stringEnumPipe = new DisplayStringEnumPipe();
+
+  public generateSafeCsv(cellData: string | null | undefined): string | undefined {
+    return isNil(cellData) ? undefined : this.stringEnumPipe.transform(cellData);
```
adisreyaj, line 14: I think we can just directly use the `displayStringEnum` method used by the pipe instead of using the pipe itself.

adisreyaj (approved):

projects/components/src/table/cells/csv-generators/table-cell-string-enum-csv-generator.ts
arjunlalb: We could. My intention was to keep it as close to the cell renderer as possible.

PR #2477, opened 2023-10-20
feat: topology group node

itssharmasandeep:
https://github.com/hypertrace/hypertrace-ui/assets/78212931/7f6dd5d1-1904-496f-836b-5eea053d46bd

Screenshot 2023-10-19 at 11 21 23 PM


projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -159,8 +164,10 @@ export class D3Topology implements Topology {
   }
 
   private updateLayout(): void {
+    this.runAndDrainCallbacks(this.dataClearCallbacks);
+    this.collapseGroupNodesIfPresent();
```
anandtiwary, line 168: can we redraw only if group node is present? Otherwise, we can continue to update the positions of already rendered nodes.

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -420,6 +429,97 @@ export class D3Topology implements Topology {
...
+    this.runAndDrainCallbacks(this.dataClearCallbacks);
+    userNode.expanded = !userNode.expanded;
+
+    if (!userNode.expanded) {
+      this.userNodes = this.userNodes.filter(n => !childNodes.includes(n));
+    } else {
+      this.userNodes = [this.userNodes, childNodes].flat();
+    }
+
+    this.topologyData = this.topologyConverter.convertTopology(
```
anandtiwary, line 449: This re-init should not be part of this method. Or we should make it more dynamic else there will be perf hits.

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -420,6 +429,97 @@ export class D3Topology implements Topology {
...
+
+    this.updateLayoutForGroupNode(this.topologyData, userNode);
+    this.drawData(this.topologyData, this.nodeRenderer, this.edgeRenderer);
+  }
+
+  private updateLayoutForGroupNode(
+    topology: RenderableTopology<TopologyNode, TopologyEdge>,
+    groupNode: TopologyGroupNode
+  ): void {
+    const renderableGroupNode = topology.nodes.find(n => n.userNode === groupNode);
```
anandtiwary, line 465: What is this one doing?

anandtiwary: as in, how is it updating the layout?

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -410,6 +417,8 @@ export class D3Topology implements Topology {
       this.neighborhoodFinder.singleNodeNeighborhood(node.userNode)
     );
 
+    this.checkAndHandleGroupNodeClick(node);
```
anandtiwary, line 420: Can you separate the entire new code with an input property? This way, we can protect the existing app flow and avoid bugs.

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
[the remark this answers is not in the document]
itssharmasandeep: Moved the group node related code to a util class, but still it is going over around 20 lines 😑

projects/observability/src/shared/components/topology/d3/d3-topology.ts
itssharmasandeep: Done

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -420,6 +429,97 @@ export class D3Topology implements Topology {
...
 
+  private checkAndHandleGroupNodeClick(node: RenderableTopologyNode): void {
+    const userNode = node.userNode;
+
+    if (!isTopologyGroupNode(userNode)) {
+      return;
+    }
+
+    const childNodes = userNode.children;
+    this.runAndDrainCallbacks(this.dataClearCallbacks);
```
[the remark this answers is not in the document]
itssharmasandeep: We need this because we're rendering newer nodes

projects/observability/src/shared/components/topology/d3/d3-topology.ts
itssharmasandeep: Done

projects/observability/src/shared/components/topology/d3/d3-topology.ts
```diff
@@ -95,7 +98,8 @@ export class D3Topology implements Topology {
     this.hover = new TopologyHover(this.d3Util, this.domRenderer);
     this.click = new TopologyClick(this.d3Util, this.domRenderer);
     this.zoom = new TopologyZoom();
-    this.layout = this.initializeLayout(config.layoutType);
+    this.layout = this.config.customLayout ?? this.initializeLayout(this.config.layoutType);
+    this.supportGroupNode = this.config.supportGroupNode ?? false;
```
anandtiwary, line 102: Can we just infer this from this.layout == GroupedLayout?

projects/observability/src/shared/components/topology/d3/interactions/drag/topology-node-drag.ts
```diff
@@ -22,7 +24,16 @@ export class TopologyNodeDrag extends TopologyEventBehavior {
     if (topologyData.nodes.length === 0) {
       return EMPTY;
     }
-    const nodeLookup = this.buildLookupMap(topologyData.nodes, node => nodeRenderer.getElementForNode(node));
+
+    const childUserNodesOfGroup = topologyData.nodes
```
anandtiwary, line 28: can we use the groupNode property here as well?

projects/observability/src/shared/components/topology/utils/topology-converter.ts
```diff
@@ -90,12 +91,26 @@ export class TopologyConverter {
     domRenderer: Renderer2,
     stateManager: TopologyStateManager
   ): RenderableTopologyEdge[] {
-    return edges.map(edge => {
-      const sourceNode = nodeMap.get(edge.fromNode)!;
-      const targetNode = nodeMap.get(edge.toNode)!;
+    return edges
```
anandtiwary, line 94: can we isolate new code with the new property?

itssharmasandeep: Done

projects/observability/src/shared/components/topology/d3/interactions/drag/topology-node-drag.ts
itssharmasandeep: Done

projects/observability/src/shared/components/topology/d3/d3-topology.ts
itssharmasandeep: No really, layout and group nodes are two different things

projects/observability/src/shared/components/topology/d3/d3-topology.ts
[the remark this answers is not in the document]
anandtiwary: Need to break it up further. Can do it in a follow up.

projects/observability/src/shared/components/topology/topology.component.ts
```diff
@@ -58,9 +59,18 @@ export class TopologyComponent implements OnChanges, OnDestroy {
   @Input()
   public shouldAutoZoomToFit?: boolean = false;
 
+  @Input()
+  public draggableNodes?: boolean = true;
```
anandtiwary, line 63: Nit: this sounds like it is storing the nodes which are draggable. Could we rename it to allowNodeDrag?

projects/observability/src/shared/components/topology/topology.component.ts
```diff
@@ -58,9 +59,18 @@ export class TopologyComponent implements OnChanges, OnDestroy {
...
+
   @Input()
   public layoutType?: TopologyLayoutType;
 
+  @Input()
+  public customLayout?: TopologyLayout; // This will override `layoutType` property
+
+  @Input()
+  public supportGroupNode?: boolean;
```
anandtiwary, line 72: Do we need this as an input or can  we extract this info from our custom group node layout property?

anandtiwary (commented):
Looks fine otherwisr

projects/observability/src/shared/components/topology/topology.component.ts
itssharmasandeep: Hmm, this was the property already present in the topology config, so I used the same name. do we change at both places or just input?

projects/observability/src/shared/components/topology/topology.component.ts
itssharmasandeep: The layout and group nodes are different IMO. The current flow looks like this.

1. We create group nodes like we create entity nodes using the data that we get from the backend.
2. We pass this data to the topology.
3. Now topology handles group nodes based on the property.
4. custom layout (with some business logic) is passed from the parent component and used to position the nodes on some layout logic. The layout may or may not care about the group nodes.

The one thing that we can do here is we can extract this property by checking if the nodes contain a group node or not. If this works I can make the changes

projects/observability/src/shared/components/topology/topology.component.ts
anandtiwary: It is fine if it is an existing term

anandtiwary (approved):

PR #2478, opened 2023-10-21
fix: use keys from across all rows to construct csv object

arjunlalb:
## Description
CSV download logic should look at all keys across the row list, not just first row.


projects/components/src/download-file/service/file-download.service.test.ts
```diff
@@ -49,8 +49,10 @@ describe('File Download Service', () => {
 
   test('should download as csv correctly', () => {
     const spectator = createService();
-    const csvData$ = of([{ name: 'traceable', headCount: 123 }]);
-
+    const csvData$ = of([
+      { name: 'traceable', headCount: 123 },
```
aaron-steinfeld, line 53:
```suggestion
      { name: 'example', headCount: 123 },
```

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -44,6 +47,13 @@ export class FileDownloadService {
 
     // Convert values into strings
     const values$ = config.dataSource.pipe(
+      // All rows are not guaranteed to have all keys, so add all keys to all rows
+      map(data => {
+        const allKeys = getAllDataKeys(data);
+        const allKeysDict = allKeys.reduce((acc, key) => ({ ...acc, [key]: undefined }), {});
+
+        return data.map(row => ({ ...allKeysDict, ...row }));
```
aaron-steinfeld, line 55: Because we're using object.values to iterate this below, we're still relying here on insertion order being consistent, which it won't be if we're spreading like this (also IIRC there's some nuances to the iteration order of object key/values, so best not to rely on it). suggest we iterate on keys to access values rather than using them as-is to overwrite empty values like this.

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -31,11 +31,14 @@ export class FileDownloadService {
    * @param config Csv download config
    */
   public downloadAsCsv(config: CsvDownloadFileConfig): Observable<FileDownloadEvent> {
+    const getAllDataKeys = (data: Dictionary<unknown>[]): string[] => uniq(data.map(row => Object.keys(row)).flat());
     // If given header is empty, then create header from the data keys
     const header$ = isEmpty(config.header)
       ? config.dataSource.pipe(
-          map(data => Object.keys(data[0] ?? [])),
-          map(keys => keys.map(startCase))
+          map(
```
aaron-steinfeld, line 38: give there's a decent amount of logic here, can we also write a real test?

projects/components/src/download-file/service/file-download.service.ts
adisreyaj: Are you suggesting we do this?:
```ts
Object.fromEntries(allKeys.map((key) => [key, row?.[key] ?? undefined]));
```

adisreyaj: Are you suggesting we do this instead?:
```ts
data.map((row) => {
  const newRow = Object.fromEntries(allKeys.map((key) => [key, row?.[key] ?? undefined]));
  return newRow;
})
```

aaron-steinfeld: there's multiple approaches here, but I probably would avoid mapping it into an object and back out again. Not only is it extra allocations, but more important - you're relying on semantics (both of .fromEntries and of .values). Instead, map allKeys to the stringify calls.

adisreyaj: Updated.

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -31,20 +31,28 @@ export class FileDownloadService {
...
     // Value replacer for null and undefined values
     const replacer = (_: string, value: string) => value ?? '-';
 
     // Convert values into strings
-    const values$ = config.dataSource.pipe(
-      map(data => data.map(datum => Object.values(datum).map(value => JSON.stringify(value, replacer))))
+    const values$: Observable<string[][]> = config.dataSource.pipe(
+      // All rows are not guaranteed to have all keys, so add all keys to all rows
+      map(data => {
+        const allKeys = getAllDataKeys(data);
```
aaron-steinfeld, line 52: nit: now that I read this a little closer, we're being pretty inefficient on doing this. We're iterating all the values to calculate the keys twice, iterating to calculate the values once and then iterating again to merge them. If you think that makes it more readable, by all means go for it. Personally, I think it does the opposite - I'd try to pipe off the data source a single time.

More importantly though, I see the header can be specified via config. If that's the case, we still have an issue. If I give a config of "header1, header2" and data containing extra keys (or keys in a different order) both the prior code and our new code would produce mismatched rows. However we calculate the header, we need to use that same result to extract the data.

projects/components/src/download-file/service/file-download.service.ts
aaron-steinfeld: bump here.

projects/components/src/download-file/service/file-download.service.ts
adisreyaj: Updated: [`1a62ad4` (#2478)](https://github.com/hypertrace/hypertrace-ui/pull/2478/commits/1a62ad49e92d0f63a75450ee3b546bfd7af2052e)

projects/components/src/download-file/service/file-download.service.ts
itssharmasandeep: +1 one the test

adisreyaj: Added: [`6b4d3d1` (#2478)](https://github.com/hypertrace/hypertrace-ui/pull/2478/commits/6b4d3d18f8937d8d8deebb2422e1a4610bbfc6d5)

itssharmasandeep (approved):

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -31,25 +31,17 @@ export class FileDownloadService {
...
   public downloadAsCsv(config: CsvDownloadFileConfig): Observable<FileDownloadEvent> {
-    // If given header is empty, then create header from the data keys
-    const header$ = isEmpty(config.header)
-      ? config.dataSource.pipe(
-          map(data => Object.keys(data[0] ?? [])),
-          map(keys => keys.map(startCase))
-        )
-      : of(config.header!);
-
+    const getAllDataKeys = (data: Dictionary<unknown>[]): string[] => uniq(data.map(row => Object.keys(row)).flat());
```
aaron-steinfeld, line 34: nit - now we only do this once, don't need to declare it as a separate function.

projects/components/src/download-file/service/file-download.service.ts
```diff
@@ -31,25 +31,17 @@ export class FileDownloadService {
...
     // Value replacer for null and undefined values
     const replacer = (_: string, value: string) => value ?? '-';
+    const csvData$ = config.dataSource.pipe(
+      map(data => {
+        const allKeys = isEmpty(config.header) ? getAllDataKeys(data) : config.header!;
```
aaron-steinfeld, line 39: One major concern here is there's a big assumption that the header provided is keys, and not already cased/renamed. Do we know that to be true for all use cases? If it is, I wouldn't even call it the header any more since that sounds like it should not impact the extraction. Would call it like csvColumn with a key and optional title or something.

PR #2485, opened 2023-10-24
fix: export method from time range service

23nobody:
exported `timeRangeFromUrlString` method

### DIFF
--- a/projects/common/src/time/time-range.service.ts
+++ b/projects/common/src/time/time-range.service.ts
@@ -99,7 +99,7 @@ export class TimeRangeService {
     );
   }
 
-  private timeRangeFromUrlString(timeRangeFromUrl: string): TimeRange {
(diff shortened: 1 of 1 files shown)


adisreyaj (dismissed):

projects/observability/src/pages/explorer/explorer.component.ts
```diff
@@ -339,5 +339,6 @@ export const enum ExplorerQueryParam {
   GroupLimit = 'limit',
   Series = 'series',
   Order = 'order',
-  Filters = 'filter'
+  Filters = 'filter',
+  Time = 'time'
```
aaron-steinfeld, line 343: I'm not sure I'd call this an explorer param. Consider treating time separately and always going through TR service.

23nobody: Makes sense, reverted

aaron-steinfeld (approved):

PR #2506, opened 2023-11-09
feat: supporting aggregation queries in visualization

palbizu:
## Description
Supporting Aggregation queries in Visualization

### Testing
Locally Tested.

### Checklist:

### Documentation

### DIFF
--- a/projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
+++ b/projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
@@ -71,23 +71,7 @@ export class ExplorerDashboardBuilder {
           type: 'cartesian-explorer-selection-handler',
           'show-context-menu': false
         },
-        ...(!isEmpty(request.groupBy) && isEmpty(request.interval)
-          ? {
-              'x-axis': {
-                type: 'cartesian-axis',
-                'scale-type': ScaleType.Band
-              }
-            }
-          : {}),
-        ...(isEmpty(request.groupBy) && isEmpty(request.interval)
-          ? {
-              'x-axis': {
-                type: 'cartesian-axis',
-                'scale-type': ScaleType.Band,
-                'show-tick-labels': false
-              }
-            }
-          : {})
+        ...this.buildXAxis(request)
       },
       onReady: dashboard => {
         dashboard.createAndSetRootDataFromModelClass(ExplorerVisualizationCartesianDataSourceModel);
@@ -175,6 +159,29 @@ export class ExplorerDashboardBuilder {
         trace: context
       }
     };
+  }
+
+  private buildXAxis(request: ExploreVisualizationRequest): object {
+    if (isEmpty(request.interval)) {
+      if (isEmpty(request.groupBy)) {
+        return {
+          'x-axis': {
+            type: 'cartesian-axis',
+            'scale-type': ScaleType.Band,
+            'show-tick-labels': false
+          }
+        };
+      } else {
+        return {
+          'x-axis': {
+            type: 'cartesian-axis',
+            'scale-type': ScaleType.Band
+          }
+        };
+      }
+    }
+
+    return {};
   }
 
   private getRendererForType(type: AttributeMetadataType): string {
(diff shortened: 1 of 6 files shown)


projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
```diff
@@ -78,6 +78,15 @@ export class ExplorerDashboardBuilder {
                 'scale-type': ScaleType.Band
               }
             }
+          : {}),
+        ...(isEmpty(request.groupBy) && isEmpty(request.interval)
+          ? {
```
aaron-steinfeld, line 83: the conditions make it so only one (or none) of these will ever be added, but it's a bit complex to inline. Suggest moving out a function like `buildXAxis(request)`

projects/observability/src/shared/dashboard/data/graphql/explore/explore-result.ts
```diff
@@ -64,10 +68,24 @@ export class ExploreResult {
     return this.resultsContainingSpec(spec).map(result => this.resultToGroupData(result, groupBySpecs, spec));
   }
 
+  private extractMetricAggregationForSpec(spec: ExploreSpecification): MetricAggregationData {
+    return this.resultToMetricAggregationDataData(this.resultsContainingSpec(spec)?.[0]!, spec);
```
aaron-steinfeld, line 72: We're asserting here a result will always be returned - what happens if we don't get any data back? Not sure if that can happen or not (i.e. does the server default if no data in range for aggregation), but seems like we can handle it regardless to be safe.

palbizu: So in the case that this doesn't return data we can return an empty array?

projects/observability/src/pages/explorer/explorer-dashboard-builder.ts
palbizu: sure, let me update it

projects/observability/src/shared/dashboard/data/graphql/explore/explore-result.ts
aaron-steinfeld: Yeah I think you'd ultimately want to render "No data" - so that probably translates to no result series.

aaron-steinfeld: Alternatively a series with an empty data array could make sense too

palbizu: what if I return the value as 0 in this case? I believe that is an unreachable scenario but we could handle this scenario returning 0 as value.

aaron-steinfeld: I'd aim for the empty version (either no series or series of empty data) - a zero and a null (no result) are sometimes equivalent but not always, it depends on the query. For example if I ask "how many calls occurred in the last 1h" showing either 0 or no data makes sense. If I ask "What is the average latency for calls in the last hour" then 0 is not correct - we'd want no data. 

Hope that makes sense

palbizu: done

projects/observability/src/shared/dashboard/data/graphql/explore/explore-result.ts
```diff
@@ -69,20 +69,24 @@ export class ExploreResult {
...
-    result: GraphQlExploreResult,
+    result: GraphQlExploreResult | undefined,
     spec: ExploreSpecification
   ): MetricAggregationData {
     return {
       key: spec.name,
-      value: result[spec.resultAlias()].value as number
+      ...(result
+        ? {
+            value: result?.[spec.resultAlias()].value as number
```
aaron-steinfeld, line 87: if we sent value as undefined, what would happen? It'd read a lot cleaner if we can remove the ternary and typically there wouldn't be much difference between `{value: undefined}` and `{}`

aaron-steinfeld (approved):

PR #2519, opened 2023-11-15
fix: fix placeholder alignment in input component

arjunlalb:
## Description
Fix placeholder alignment in input component


projects/components/src/input/input.component.scss
```diff
@@ -14,7 +14,6 @@
           &.mat-form-field-empty {
             // Placeholder
             color: $gray-4;
-            top: 0.8em;
```
23nobody, line 17: Oh, so this was the issue. Good catch

23nobody (approved):

PR #2520, opened 2023-11-15
feat(file-upload): support custom file requirements

adisreyaj:
## Description
image

### DIFF
--- a/projects/components/src/file-upload/file-upload.component.ts
+++ b/projects/components/src/file-upload/file-upload.component.ts
@@ -46,6 +46,13 @@ import { FileTypeUtil, SupportedFileType } from './file-types';
 
     <ng-template #defaultSubTpl>
       <ul class="file-requirements-list">
+        <li class="item" *ngFor="let req of (this.config?.customRequirementsInfo ?? [])">
+          <ht-icon class="icon" icon="${IconType.CheckCircleFill}" size="${IconSize.ExtraSmall}"></ht-icon>
+          <div class="text">
+            <strong>{{ req.key }}</strong
+            >: {{ req.value | htDisplayString }}
+          </div>
+        </li>
         <li class="item">
           <ht-icon class="icon" icon="${IconType.CheckCircleFill}" size="${IconSize.ExtraSmall}"></ht-icon>
           <div class="text"><strong>File Types</strong>: {{ this.config.supportedFileTypes | htFilePipe }}</div>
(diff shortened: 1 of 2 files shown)


projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -46,6 +46,13 @@ import { FileTypeUtil, SupportedFileType } from './file-types';
 
     <ng-template #defaultSubTpl>
       <ul class="file-requirements-list">
+        <li class="item" *ngFor="let req of (this.config?.customRequirementsInfo ?? [])">
+          <ht-icon class="icon" icon="${IconType.CheckCircleFill}" size="${IconSize.ExtraSmall}"></ht-icon>
+          <div class="text">
+            <strong>{{ req.key }}</strong
```
23nobody, line 52: need `htDisplayString` for `key` also?

adisreyaj: Not necessary I think. But not harm in adding.

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -46,6 +46,13 @@ import { FileTypeUtil, SupportedFileType } from './file-types';
 
     <ng-template #defaultSubTpl>
       <ul class="file-requirements-list">
+        <li class="item" *ngFor="let req of (this.config?.customRequirementsInfo ?? [])">
```
23nobody, line 49: what's the expected list size? just to double check that it doesn't break anything when the list in long.

adisreyaj: Ideally, 1 or 2.

projects/components/src/file-upload/file-upload.component.ts
Greedy-Geek: +1

adisreyaj: Added.

projects/components/src/file-upload/file-upload.component.ts
```diff
@@ -198,4 +205,10 @@ export interface UploaderConfig {
   supportedFileTypes: SupportedFileType[];
   maxFileSizeInBytes: number;
   maxNumberOfFiles: number;
+  customRequirementsInfo?: FileUploadCustomRequirementInfo[];
```
Greedy-Geek, line 208: Do we want this to be optional? Can't it just be an empty array instead of undefined?

adisreyaj: We don't want the consumer to always provide the field if they really don't want to use it.

23nobody (approved):

projects/components/src/file-upload/file-upload.component.ts
Greedy-Geek: Optional fields are easy to miss in case one don't know about it. So that's why

Greedy-Geek: No problem

Greedy-Geek (approved):

PR #2521, opened 2023-11-15
fix: table should not cache if filter or sort changes

aaron-steinfeld:
## Description

Caching on table was preventing requests from firing when the request changed

### DIFF
--- a/projects/components/src/table/data/table-cdk-data-source.ts
+++ b/projects/components/src/table/data/table-cdk-data-source.ts
@@ -297,27 +297,24 @@ export class TableCdkDataSource implements DataSource<TableRow> {
 
   private hasCacheForRequest(request: TableDataRequest): boolean {
     if (
-      this.cachedData.rows.length !== 0 &&
-      this.cachedData.rows.length === this.cachedData.total &&
-      request.position.limit <= this.cachedData.rows.length &&
-      !this.haveColumConfigsChanged(request)
+      !isEqual(this.cachedData.request?.sort, request.sort) ||
+      !isEqual(this.cachedData.request?.filters, request.filters) ||
+      this.haveColumConfigsChanged(request)
     ) {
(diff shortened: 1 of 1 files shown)


palbizu (approved):

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -297,27 +297,24 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
+      !isEqual(this.cachedData.request?.filters, request.filters) ||
+      this.haveColumConfigsChanged(request)
     ) {
-      // Check if we already have all available results cached
-      return true;
+      // If the columns, sort or filters have changed, we need to make a new request
+      return false;
     }
 
-    if (this.cachedData.rows.length < request.position.limit || !isEqual(this.cachedData.request?.sort, request.sort)) {
```
aaron-steinfeld, line 309: The first clause here is redundant with the final calculation I believe. The second one moved to the top.

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -297,27 +297,24 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
+      return true;
     }
 
     const offsetWithinCachedRows = this.calcOffsetWithinCachedRows(request);
 
     // Check if requested startOffset + limit is within the cached data
     return (
-      offsetWithinCachedRows >= 0 &&
-      offsetWithinCachedRows + request.position.limit <= this.cachedData.rows.length &&
-      !this.haveColumConfigsChanged(request)
```
aaron-steinfeld, line 320: pulled this clause to the top rather than in each individual check.

jake-bassett (approved):
Logic looks right in isolation. Please test with all forms of table before merging.

PR #2522, opened 2023-11-16
Fix: Request should be cloned when caching table data

aneeshsharma:
## Description
When storing the request object in the cache for the table, only the reference was being assigned to the cache for the column config. So, the new request object and the cached request object would always have the same column config even after it is updated leading to new data not being fetch when the column config changes.

Now, made the change so a deep copy of the request is created when caching.

This fixes the issue where new data was not being fetched when sort state changes in the column config.

### DIFF
--- a/projects/components/src/table/data/table-cdk-data-source.ts
+++ b/projects/components/src/table/data/table-cdk-data-source.ts
@@ -1,6 +1,6 @@
 import { DataSource } from '@angular/cdk/collections';
 import { Dictionary, forkJoinSafeEmpty, isEqualIgnoreFunctions, RequireBy, sortUnknown } from '@hypertrace/common';
-import { isEqual, isNil } from 'lodash-es';
+import { cloneDeep, isEqual, isNil } from 'lodash-es';
 import { BehaviorSubject, combineLatest, NEVER, Observable, of, Subject, Subscription, throwError } from 'rxjs';
 import { catchError, debounceTime, map, mergeMap, startWith, switchMap, tap } from 'rxjs/operators';
 import { PageEvent } from '../../paginator/page.event';
@@ -92,7 +92,7 @@ export class TableCdkDataSource implements DataSource<TableRow> {
 
   private cacheNewData(total: number, rows: StatefulTableRow[], request?: TableDataRequest): void {
     this.cachedData = {
-      request: request,
+      request: cloneDeep(request),
       rows: rows.map(TableCdkRowUtil.cloneRow),
       total: total,
     };


projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -92,7 +92,7 @@ export class TableCdkDataSource implements DataSource<TableRow> {
 
   private cacheNewData(total: number, rows: StatefulTableRow[], request?: TableDataRequest): void {
     this.cachedData = {
-      request: request,
+      request: cloneDeep(request),
```
aaron-steinfeld, line 95: The idea is solid, but we need to be careful here because columns can hold state that should not be accounted for in caching. Perhaps we can add a custom function that pulls out the important bits of the request rather than cloning the whole thing.

aneeshsharma: Would it be a good idea to create a copy, but in the comparison logic inside `haveColumnConfigsChanged`, we only compare the important parts.

I don't think anything else should break with creating a copy, since we don't really expect a situation where change in the request should automatically update the cache in anyway other than through the `cacheNewData` call.

@aaron-steinfeld

aaron-steinfeld: >  but in the comparison logic inside haveColumnConfigsChanged, we only compare the important parts.

we compare the whole column right?
```
  private haveColumConfigsChanged(request: TableDataRequest): boolean {
    if (isNil(this.cachedData.request)) {
      return true;
    }

    return !isEqual(request.columns, this.cachedData.request.columns);
  }
```

> I don't think anything else should break with creating a copy, since we don't really expect a situation where change in the request should automatically update the cache in anyway other than through the cacheNewData call.

Not sure I'm following the situation you describe - the problem is the inverse, that we're going to start getting erroneous cache misses due to cloning, where no change in the request isn't going to the cache. At least that's my read - because cloneDeep wouldn't copy functions for example, which at least some columns have. Then, isEqual will compare the clone and original value and return false (column configs _have_ changed) every single time, no matter what.

aneeshsharma: So, the functions will be the only objects that will not be compared correctly.
If we copy everything else apart from functions, this should work. Does that sound good?

The other thing is if there are fields that shouldn't be considered for cache misses, the comparison is still unchanged and will return false in case something changes in the request.

For example if the alignment or width of the column changes, then we don't really need to refresh the cache. In the current comparison since the entire column is compared anyways, even a change in alignment will trigger a cache miss.

So, along those lines I was suggesting that we might need to change the comparison logic as well.

aaron-steinfeld: > If we copy everything else apart from functions, this should work. Does that sound good?

> The other thing is if there are fields that shouldn't be considered for cache misses, the comparison is still unchanged and will return false in case something changes in the request.

We don't know what consumers will put in their column as you said, so this still seems off.

I think it's unlikely that the columns will change the actual data they represent without changing references (and in the spirit of the on push change detection we use across the codebase anyway), so IMO just comparing the refs would be just fine. Yes, it means the caching would be broken if someone mutated their column to change how it aligns to a data fetch, but that's how on push works everywhere. The real use case of this comparison is (I'm guessing) - were any columns added or removed?

PR #2523, opened 2023-11-16
Fix: Separate collapsable and search history behaviour

aneeshsharma:
## Description
The collapsable behaviour of the search box was tied to whether search history was enabled or not.

Now added a separate input `collapsable` which can be set to `true` to make the search box collapsable and vice versa.

Search history is now independent of this.

### DIFF
--- a/projects/components/src/search-box/search-box.component.scss
+++ b/projects/components/src/search-box/search-box.component.scss
@@ -31,7 +31,7 @@
     }
   }
 
-  &.with-search-history {
+  &.collapsable {
     width: 48px;
 
     .input {
--- a/projects/components/src/search-box/search-box.component.test.ts
+++ b/projects/components/src/search-box/search-box.component.test.ts
@@ -155,26 +155,45 @@ describe('Search box Component', () => {
     flush();
   }));
 
-  test('collapsable enabled should add collapsable class', fakeAsync(() => {
+  test('collapsable enabled should add collapsable class', () => {
     spectator = createHost(
-      `<ht-search-box [placeholder]="placeholder" [debounceTime]="debounceTime" [searchMode]="searchMode" [enableSearchHistory]="enableSearchHistory" [collapsable]="true"></ht-search-box>`,
+      `<ht-search-box [placeholder]="placeholder" [debounceTime]="debounceTime" [searchMode]="searchMode" [enableSearchHistory]="enableSearchHistory" [collapsable]="collapsable"></ht-search-box>`,
       {
         hostProps: {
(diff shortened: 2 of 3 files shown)


projects/components/src/search-box/search-box.component.scss
```diff
@@ -31,7 +31,7 @@
     }
   }
 
-  &.with-search-history {
+  &.collapsable {
```
arjunlalb, line 34: if we rename this, does the enable search history route lose any of its styling? if not, should we just remove L41 in the component file?

palbizu: +1 we should have a separate style for this particular class or remove the other line if is useless right now.

aneeshsharma: `.with-search-history` doesn't have any styling rules specific to search history and the search history box is not affected by this change.

All the styling rules in `.with-search-history` is only related to the collapsing logic.

aneeshsharma: Also the class is not being used anywhere so we can remove it I think

projects/components/src/search-box/search-box.component.test.ts
```diff
@@ -154,4 +154,38 @@ describe('Search box Component', () => {
 
     flush();
   }));
+
+  test('collapsable enabled should add collapsable class', fakeAsync(() => {
```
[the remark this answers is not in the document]
aneeshsharma: Removed fakeAsync since not needed

projects/components/src/search-box/search-box.component.test.ts
```diff
@@ -154,4 +154,38 @@ describe('Search box Component', () => {
...
+    spectator = createHost(
+      `<ht-search-box [placeholder]="placeholder" [debounceTime]="debounceTime" [searchMode]="searchMode" [enableSearchHistory]="enableSearchHistory" [collapsable]="true"></ht-search-box>`,
```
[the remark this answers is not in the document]
aneeshsharma: Added a test for default value

projects/components/src/search-box/search-box.component.scss
aneeshsharma: So should we remove the `.with-search-history` class altogether?

anandtiwary (approved):

projects/components/src/search-box/search-box.component.scss
arjunlalb: if its not used, lets remove it :)

PR #2533, opened 2023-11-21
Feat: Provide external loading triggers

aneeshsharma:
## Description
Any changes that were external to the table that would trigger a new data fetch, couldn't be accounted for when displaying the loading state in the table. Also, since the data source is external and a stream, it is not possible for the table to know when a new data fetch might have been triggered.

So, added `loadingTrigger: Observable` to the `TableDataSource` as an optional field, which can be provided by the user which should emit whenever a change happens that will trigger a new data fetch.

This will require the user to provide the trigger but I couldn't find any other way for the table to know by itself if a new data fetch is going to be triggered.

### DIFF
--- a/projects/components/src/table/table.component.ts
+++ b/projects/components/src/table/table.component.ts
@@ -38,7 +38,7 @@ import {
   TypedSimpleChanges,
 } from '@hypertrace/common';
 import { isNil, without } from 'lodash-es';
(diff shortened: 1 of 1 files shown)


aneeshsharma:
Instead of passing the `loadingTrigger$` as metadata to the data source, we can now pass the trigger as input to the table component itself.

PR #2539, opened 2023-11-27
Fix traces table replacing last

rahul-p-rajesh:
## Description
Replaced "Last"  in the paginator with the totalCount.

1. on first page image
2. on any other last page 
image

### Checklist:
- [✅] My changes generate no new warnings
- [✅] I have added tests that prove my fix is effective or that my feature works
- [✅] Any dependent changes have been merged and published in downstream modules


projects/components/src/paginator/paginator.component.ts
```diff
@@ -217,7 +217,9 @@ export class PaginatorComponent implements OnChanges, PaginationProvider {
     }
 
     if (this.totalItems === PaginatorTotalCode.Last) {
-      return 'last';
+      const approxTotal = (this.pageIndex + 1) * this.pageSize;
```
aaron-steinfeld, line 220: We should never guess, but this case shouldn't exist - we would want to remove this enum value entirely  and update all usages (as you did in traces table data source model).

rahul-p-rajesh: Removed the last enum value and swapped with exact total being passed from "projects/observability/src/shared/dashboard/data/graphql/table/traces/traces-table-data-source.model.ts" line no 64

aaron-steinfeld (approved):

projects/components/src/paginator/paginator.component.ts
```diff
@@ -232,7 +228,7 @@ export class PaginatorComponent implements OnChanges, PaginationProvider {
       return this.pageSize;
     }
 
-    if (this.totalItems === PaginatorTotalCode.Unknown || this.totalItems === PaginatorTotalCode.Last) {
+    if (this.totalItems === PaginatorTotalCode.Unknown) {
```
aaron-steinfeld, line 231: Sorry I missed this before approving PaginatorTotalCode.Last should be removed entirely.

aaron-steinfeld (changes requested):

projects/components/src/paginator/paginator.component.ts
rahul-p-rajesh: PaginatorTotalCode.Last has been completely removed from the file

aaron-steinfeld (approved):

palbizu (approved):

PR #2546, opened 2023-11-28
fix: table caching on limit change

itssharmasandeep:


projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -349,7 +349,7 @@ export class TableCdkDataSource implements DataSource<TableRow> {
 
   private fetchCachedData(request: TableDataRequest): Observable<StatefulTableRow[]> {
     if (this.cachedData.rows.length < request.position.limit) {
-      return of([]);
+      return of(this.cachedData.rows);
```
aaron-steinfeld, line 352: This ignores the actual request. It would work for the first page only. This if statement is completely unnecessary since we don't call this fetchCached method until we've made a determination that we have the data in the cache already.

aaron-steinfeld: Discussed offline, it _is_ unnecessary but it shouldn't break because we've already decided the cache has the data. Leaning on that we know we must be on the first page for this condition to hold (if the page size changed, we'd go back to 1; if we already had less rows than the last limit, we couldn't get to page 2). So a complicated short circuit. Suggest combining the fetch cache and check cache methods together so they don't have to make assumptions.

PR #2551, opened 2023-11-29
feat: add a new ValueOrUndefined common type

arjunlalb:
## Description
Adding a new generic type.

### DIFF
--- a/projects/common/src/utilities/types/types.ts
+++ b/projects/common/src/utilities/types/types.ts
@@ -35,3 +35,4 @@ export interface Json {
 export type JsonValue = Json | Json[] | string | number | boolean | null;
 export type PrimitiveValue = string | number | boolean;
 export type ValueOrNull<T> = T | null;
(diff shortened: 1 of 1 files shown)


itssharmasandeep (approved):

projects/common/src/utilities/types/types.ts
```diff
@@ -35,3 +35,4 @@ export interface Json {
 export type JsonValue = Json | Json[] | string | number | boolean | null;
 export type PrimitiveValue = string | number | boolean;
 export type ValueOrNull<T> = T | null;
+export type ValueOrUndefined<T> = T | undefined;
```
aaron-steinfeld, line 38: How's this differ from `?: T`?

arjunlalb: its the same. I feel this is just a bit more cleaner when we have to return something from a function or use this as a type instead of doing ` | undefined`

PR #2554, opened 2023-11-30
fix: adding edge entity filters support

itssharmasandeep:
Adds support for entity scoped edge filters to topology data source and query handler

### DIFF
--- a/projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
+++ b/projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
@@ -127,9 +127,8 @@ export class TopologyDataSourceModel extends GraphQlDataSourceModel<TopologyData
 
     return this.query<EntityTopologyGraphQlQueryHandlerService>(filters => {
       const topologyFilters = this.getTopologyFilters(filters);
-      const edgeFilterEntityType = this.edgeFilterConfig?.entityType;
-      const requiredEdgeEntityTypes =
-        topologyFilters.edges.length > 0 && edgeFilterEntityType !== undefined ? [edgeFilterEntityType] : undefined;
+      const edgeFilterEntityType = this.getRequiredEdgeEntityTypeFromFilters(filters);
+      const requiredEdgeEntityTypes = edgeFilterEntityType !== undefined ? [edgeFilterEntityType] : undefined;
 
       return {
         requestType: ENTITY_TOPOLOGY_GQL_REQUEST,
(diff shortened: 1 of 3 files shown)


projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
```diff
@@ -163,20 +162,19 @@ export class TopologyDataSourceModel extends GraphQlDataSourceModel<TopologyData
...
-      'keyOrExpression' in gqlFilter && 'operator' in gqlFilter && 'value' in gqlFilter;
-
     filters.forEach(gqlFilter => {
       // Edge filter only supported for `GraphQlFieldFilter` for now
-      if (
-        isFieldFilter(gqlFilter) &&
-        edgeFilterFields.includes(this.getFieldFromExpression(gqlFilter.keyOrExpression))
-      ) {
-        edgeFilters.push(gqlFilter);
+      const field = this.isFieldFilter(gqlFilter) ? this.getFieldFromExpression(gqlFilter.keyOrExpression) : '';
```
arjunlalb, line 171: instead of just skipping past by assuming `''` as default value, can we do something better?

Do we expect non `GraphqlFieldFilters` here? If yes, is it by design or is it an unexpected scenario? If its the former, this is fine. If its latter, we should throw/log an appropriate error.

arjunlalb (dismissed):

projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
itssharmasandeep: We have an interface called `GraphQlFilter` that gets implemented and `GraphQlFieldFilter` is one of them. Now the data source and the query handler accept `GraphQlFilter`.  This logic only handles the edge filters and it only considers the field-filters for now. All the other filters go directly inside the node filters instead so I think we're good on that front.

anandtiwary (approved):

projects/observability/src/shared/dashboard/data/graphql/topology/topology-data-source.model.ts
anandtiwary: replace it with undefined?

itssharmasandeep: Done

anandtiwary (approved):
please address in follow up

PR #2566, opened 2023-12-07
Fix: Enable search history by default for Search Box

aneeshsharma:
## Description
- Set default value for `enableSearchHistory` to `true` in `ht-search-box`
- Added debounce time to the search history values being emitted. This is done to prevent search history storing all iterations of a word when typing if the debounce time is set to `0`.
- Fixed duplicate values in search history


projects/components/src/search-box/search-box.component.ts
```diff
@@ -111,11 +111,14 @@ export class SearchBoxComponent implements OnInit, OnChanges {
   public searchMode: SearchBoxEmitMode = SearchBoxEmitMode.Incremental;
 
   @Input()
-  public enableSearchHistory: boolean = false; // Experimental
+  public enableSearchHistory: boolean = true; // Experimental
```
anandtiwary, line 114: nit: we should get rid of the comment

projects/components/src/search-box/search-box.component.ts
```diff
@@ -111,11 +111,14 @@ export class SearchBoxComponent implements OnInit, OnChanges {
...
 
   @Input()
   public collapsable: boolean = false;
 
+  @Input()
+  public searchHistoryDebounceTime: number | undefined;
```
anandtiwary, line 120: We should use the existing debounce time.

projects/components/src/search-box/search-box.component.ts
```diff
@@ -261,11 +264,14 @@ export class SearchBoxComponent implements OnInit, OnChanges {
...
 
     this.subscriptionLifecycle.add(
-      this.valueChange.asObservable().subscribe(emittedValue => {
-        if (!isEmpty(emittedValue)) {
-          this.lastEmittedValues = [emittedValue, ...this.lastEmittedValues];
-        }
-      }),
+      this.valueChange
+        .asObservable()
+        .pipe(debounceTime(this.searchHistoryDebounceTime ?? 400))
```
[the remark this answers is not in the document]
anandtiwary: I agree with Arjun. We don't need this as a user-configurable input. We can control it from inside this component.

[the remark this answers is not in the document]
itssharmasandeep: Yes, +1 on that.
We should use the debounce of the value emission and if that is zero/undefined then we should default it to `400` and that can be an internal private property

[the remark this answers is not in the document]
aneeshsharma: If the debounce time is set, the search text emitted will already be debounced. So we don't need a separate debounce time for search history in that case.

If its not set we can introduce debounce time in search history emitted values.

I've made the change accordingly, setting the default value of debounce time to `400` if debounce time is not set.

projects/components/src/search-box/search-box.component.ts
anandtiwary: So, we should remove this as an Input.

projects/components/src/search-box/search-box.component.ts
```diff
@@ -261,11 +264,22 @@ export class SearchBoxComponent implements OnInit, OnChanges {
     );
 
     this.subscriptionLifecycle.add(
-      this.valueChange.asObservable().subscribe(emittedValue => {
-        if (!isEmpty(emittedValue)) {
-          this.lastEmittedValues = [emittedValue, ...this.lastEmittedValues];
-        }
-      }),
+      this.debounceTime && this.debounceTime > 0 ?
+      this.valueChange
```
anandtiwary, line 268: Agree with the thought process here, but maybe we should just keep a large 1s or 2s debounce on valueChange. Should be fine. When the user is done with a meaningful search, they likely wait for a few seconds before searching again.

aneeshsharma: I've set a constant debounce time. Which will be used by default if debounce time is not set. If debounce time is set, the minimum of the debounce time and the default debounce time is taken.

projects/components/src/search-box/search-box.component.ts
aneeshsharma: Removed this input

anandtiwary (approved):

PR #2575, opened 2023-12-12
External loading anand

anandtiwary:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/projects/components/src/table/data/table-cdk-data-source.ts
+++ b/projects/components/src/table/data/table-cdk-data-source.ts
@@ -2,7 +2,7 @@ import { DataSource } from '@angular/cdk/collections';
 import { Dictionary, forkJoinSafeEmpty, isEqualIgnoreFunctions, RequireBy, sortUnknown } from '@hypertrace/common';
 import { isEqual, isNil } from 'lodash-es';
 import { BehaviorSubject, combineLatest, NEVER, Observable, of, Subject, Subscription, throwError } from 'rxjs';
-import { catchError, debounceTime, map, mergeMap, startWith, switchMap, tap } from 'rxjs/operators';
+import { catchError, debounceTime, map, mergeMap, startWith, switchMap, take, tap } from 'rxjs/operators';
 import { PageEvent } from '../../paginator/page.event';
 import { PaginationProvider } from '../../paginator/paginator-api';
 import { RowStateChange, StatefulTableRow, StatefulTreeTableRow, TableFilter, TableRow } from '../table-api';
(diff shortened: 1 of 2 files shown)


projects/components/src/table/data/table-data-source.ts
```diff
@@ -1,8 +1,12 @@
 import { Observable } from 'rxjs';
 import { TableColumnConfig, TableFilter, TableSortDirection } from '../table-api';
 
-export interface TableDataSource<TResult, TCol extends TableColumnConfig = TableColumnConfig> {
-  getData(request: TableDataRequest<TCol>): Observable<TableDataResponse<TResult>>;
+export interface TableDataSource<TResult, TCol extends TableColumnConfig = TableColumnConfig, TDataTriggers = unknown> {
+  secondaryDataTriggers$?: Observable<TDataTriggers>; // TBD
```
arjunlalb, line 5: IMO, this is fine how it is used here. Adding it to watched observables in the data source is sound logic. However, does it give us the intended effect. Do we show a loader while the new data is being resolved?

projects/components/src/table/data/table-data-source.ts
```diff
@@ -1,8 +1,12 @@
...
+  getData(
+    request: TableDataRequest<TCol>,
+    secondaryDataTriggers?: TDataTriggers,
```
arjunlalb, line 8: we should be able to avoid this by making new request happen on a trigger change within the table. Is that something we can do? Getting the triggers as part of data source and then the consumer having to implement some logic dependent on them seems wierd.

We should not have to emit filters via these triggers again. These triggers can play the simple role of being signals while the getData method's `request` argument having the desired state always.

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -174,7 +184,12 @@ export class TableCdkDataSource implements DataSource<TableRow> {
       this.filtersProvider.queryProperties$,
       this.columnStateChangeProvider.columnState$,
       this.rowStateChangeProvider.rowState$,
-    ]).pipe(map(values => this.detectRowStateChanges(...values)));
+    ]).pipe(
+      switchMap(values =>
+        combineLatest([of(values), this.tableDataSourceProvider.data?.secondaryDataTriggers$ ?? of(undefined)]),
```
aneeshsharma, line 189: Since we are using `combineLatest`, in terms of emitting the loading state value, it will have affect the same way as 2 subscriptions, one to the existing change observable and one to `secondaryDataTriggers$`.

Correct me if I am wrong.

projects/components/src/table/data/table-data-source.ts
anandtiwary: Yes.

projects/components/src/table/data/table-data-source.ts
anandtiwary: > we should be able to avoid this by making new request happen on a trigger change within the table

This is exactly what we are doing. For requests triggered externally, we do need some `state` (e.g show inactive boolean). The purpose of this argument is to pass on that state to the getData method so that it can be used to build the correct gql query.

PR #2577, opened 2023-12-14
Fix: Export Predefined time duration service

aneeshsharma:
## Description
- Exported predefined time duration service
- ~~Made `getAvailableIntervals` public in `IntervalDurationService`~~


projects/common/src/time/interval-duration.service.ts
```diff
@@ -69,7 +69,7 @@ export class IntervalDurationService {
     return durations[0];
   }
 
-  private getAvailableIntervals(timeRange: TimeRange, minDataPoints: number, maxDataPoints: number): TimeDuration[] {
+  public getAvailableIntervals(timeRange: TimeRange, minDataPoints: number, maxDataPoints: number): TimeDuration[] {
```
itssharmasandeep, line 72: As you mentioned, let's use `getAutoDuration()` instead for our use case. We can remove this change for now

aneeshsharma: yep removed this change

itssharmasandeep (approved):

PR #2587, opened 2023-12-21
feat: enhance string array cell renderer to support view modes

arjunlalb:
## Description
With the new addition of options in column config, now we can customize cell renderers using inputs.
Using this to enhance the rendering options available in string array cell renderer.

### Testing
Verified manually. Added UTs.

### Checklist:


projects/components/src/table/cells/data-renderers/string-array/string-array-table-cell-renderer.component.ts
```diff
@@ -43,20 +48,35 @@ import { TableCellAlignmentType } from '../../types/table-cell-alignment-type';
...
     @Inject(TABLE_COLUMN_INDEX) index: number,
     @Inject(TABLE_DATA_PARSER) parser: TableCellParserBase<string[], string[], boolean>,
     @Inject(TABLE_CELL_DATA) cellData: string[],
     @Inject(TABLE_ROW_DATA) rowData: TableRow,
   ) {
     super(columnConfig, index, parser, cellData, rowData);
+
+    if (this.columnConfigOptions?.viewMode) {
+      this.viewMode = this.columnConfigOptions.viewMode;
+    }
```
itssharmasandeep, line 67: We should make `this.viewMode` readonly and
```ts
 this.viewMode = this.columnConfigOptions?.viewMode ?? StringArrayTableCellDisplayMode.List
```

arjunlalb: don't want to reassign and trigger cd if its not changing. made it readonly.

itssharmasandeep: I did not get you, it is not reassigning. It is the same as if condition.

itssharmasandeep: Since we're assigning in the constructor anyways

arjunlalb: if viewMode is empty, the variable will be reassigned with the same value ('list') again. This could trigger change detection since a reassignment happened. We can avoid it if we do the assignment only when value is supplied.

itssharmasandeep (approved):

PR #2589, opened 2023-12-21
fix(time-duration): Time Duration Most Significant adjusted to not have approax value

rahul-p-rajesh:
## Description
Problem: On Providing an input like 44H, the output is 1 day.
Fix: Added an allowApproxUnit flag which on false keeps on checking the for time unit which is not approximate.

### Checklist:
- ✅ My changes generate no new warnings
- ✅ I have added tests that prove my fix is effective or that my feature works
- ✅  Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/projects/common/src/time/time-duration.test.ts
+++ b/projects/common/src/time/time-duration.test.ts
@@ -80,6 +80,24 @@ describe('Time duration', () => {
     expect(new TimeDuration(0, TimeUnit.Millisecond).getMostSignificantUnitOnly()).toEqual(
       new TimeDuration(0, TimeUnit.Millisecond),
     );
+
+    expect(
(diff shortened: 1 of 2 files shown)


aaron-steinfeld:
> Problem: On Providing an input like 44H, the output is 1 day.
Fix: Added an allowApproxUnit flag which on false keeps on checking the for time unit which is not approximate

That's the whole point of the function though. In that example, 'day' is the most significant unit. The behavior you're describing is something else, so we shouldn't overload the existing function with a boolean flag (code smell!). We'd want a whole new function. That said, given we already can represent this as a 1d24h and as 1d, do we really need a third representation? The UX consistency here is concerning.

PR #2601, opened 2024-01-03
Fix: Added scroll strategy for overlay

aneeshsharma:
## Description
Added `scrollStrategy` to overlay to reposition the overlay while scrolling. This makes the overlay stick to the position on the page and scrolls along with the rest of the page.

This would only work if the `cdk-scrollable` directive is used on the container that is being scrolled through.

So, its not a universal solution, but `cdk-scrollable` needs to be added to any containers that have overlays that are facing this issue.


aneeshsharma:
For reference, this is where I found the solution.

https://github.com/angular/components/issues/6157#issuecomment-318972884

https://github.com/angular/components/issues/6157

arjunlalb (approved):

PR #2603, opened 2024-01-04
feat: copy curl command support in span detail component

Greedy-Geek:


projects/observability/src/shared/components/span-detail/span-detail.module.ts
```diff
@@ -20,6 +21,7 @@ import { SpanRequestDetailModule } from './request/span-request-detail.module';
 import { SpanResponseDetailModule } from './response/span-response-detail.module';
 import { SpanDetailComponent } from './span-detail.component';
 import { SpanTagsDetailModule } from './tags/span-tags-detail.module';
+import { MemoizeModule } from '../../../../../common/src/utilities/memoize/memoize.module';
```
arjunlalb, line 24: bad import?

arjunlalb (approved):

projects/observability/src/shared/components/span-detail/span-detail.module.ts
Greedy-Geek: updated

arjunlalb (approved):

PR #2604, opened 2024-01-05
fix: add `clickOnRowToSelect` option to table

arjunlalb:
## Description
With various usecases of table, it is not always required to select a row when its clicked on. There are usecases that just needs some action to be done on a row click instead of toggling its selection status.

To enable this, this PR introduces a boolean control in the table component.

### Testing
Tested manually. 

### Checklist:

### DIFF
--- a/projects/components/src/table/table.component.ts
+++ b/projects/components/src/table/table.component.ts
@@ -343,6 +343,9 @@ export class TableComponent
   public selectionMode?: TableSelectionMode = TableSelectionMode.Single;
 
   @Input()
+  public clickOnRowToSelect?: boolean = true;
+
+  @Input()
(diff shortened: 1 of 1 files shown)


projects/components/src/table/table.component.ts
```diff
@@ -342,6 +342,9 @@ export class TableComponent
   @Input()
   public selectionMode?: TableSelectionMode = TableSelectionMode.Single;
 
+  @Input()
+  public clickOnRowToSelect?: boolean = true;
```
itssharmasandeep, line 346: We have `TableSelectionMode.None` but we're not using that anywhere in the table component.
can we use that instead of creating a new component?

